PR #12, opened 2024-03-20
Remove argparse from get_client function signature

ryantwolf:
Main items:
- [x] Refactor `get_client` function to not use argparse
- [x] Add `parse_client_args` function to bridge argparse to new `get_client`
- [x] Update all `get_client` invocations
- [x] Allow `get_client` to be imported from root module like `from nemo_curator import get_client`
- [x] Update documentation
- [x] Test all scripts

Other changes:
- Modify how `find_exact_duplicates.py` determines whether to use CPU or GPU.
- Fix broken links in user guide.


Maghoumi:
@ryantwolf I think we should have this functionality. I was trying to use the Curator through a notebook and had to do some gymnastics to create a client using `get_client()`. 

If this is too much work for now, maybe a simple workaround would be to have an equivalent of `add_distributed_args()`, but return a dictionary of the required arguments. This way the user can pass that `dict` instance to `get_client()` and still be able to use the code in a notebook environment.

ayushdg (approved):
Thanks @ryantwolf looks good to me.

PR #13, opened 2024-03-21
Add workflow for running cpu pytests

ayushdg:


ayushdg:
@ryantwolf Looks like Nemo-toolkit install fails in a python 3.9 environment (can reproduce locally). I can skip 3.9 tests here for the time being, but not sure if it's _supposed_ to work with 3.9 or not.

ryantwolf:
@ayushdg ~~hm, why does it fail with 3.9? I can ask around if it's supposed to work.~~ I see the logs now. We can skip for now though.

After talking with folks, looks like NeMo only supports Python 3.10 and above now. So we should bump our minimum version too.

ayushdg:
> After talking with folks, looks like NeMo only supports Python 3.10 and above now. So we should bump our minimum version too.

Thanks, I'll limit the scope of the PR to 3.10 testing. We can bump the minimum version in a followup.

ryantwolf (approved):
Looks good. Thanks!

PR #18, opened 2024-03-27
Add batched decorator

ryantwolf:
Refactors batched to be in a decorator. This frees the user from having to be knowledgable about how the underlying `DocumentFilter` or `DocumentModifier` when initializing `ScoreFilter` or `Modify` respectively.

Unit tests pass. The following scripts were manually tested and work properly.

```
examples/classifier_filtering.py
examples/find_pii_and_deidentify.py
nemo_curator/scripts/find_pii_and_deidentify.py
tutorials/tinystories/main.py
```
The examples and other scripts have been manually tested to work.


ayushdg (changes requested):
Thanks for the changes! Just a couple of places with the stray `batched` argument but generally lgtm.
I like the idea of having `batched` be an attribute of the filters/modifiers. Prevents the caller from having knowledge of which variant should be used.

ayushdg (approved):

PR #32, opened 2024-04-15
Add dataset blending tool

ryantwolf:
Add functionality for large scale dataset blending and shuffling.


ryantwolf:
@ayushdg could you please take a look at why the exact deduplication tests are failing here? I didn't modify anything related to dedup, but evidently one of my changes triggered this.

ayushdg:
Haven't been able to reproduce locally. But on first glance it looks like the dask config options when we get to the assert_eq is no longer "tasks" for the shuffle method. Overriding it in the test would allow things to past but I'm curious to see why it isn't reproducing locally.

nemo_curator/modules/dataset_ops.py
```diff
@@ -0,0 +1,150 @@
...
+
+    def __call__(self, dataset: DocumentDataset) -> DocumentDataset:
+        new_npartitions = (
+            dataset.df.npartitions if self.npartitions is None else self.npartitions
+        )
+
+        rand_df = dataset.df.map_partitions(self._add_rand_col, new_npartitions)
+
+        shuffled_df = rand_df.shuffle(
+            self.rand_col, npartitions=new_npartitions, ignore_index=True
```
ayushdg, line 47: The shuffle as mentioned makes the result non deterministic. While slower, I wonder if `set_index` or `sort_values` on the `rand_col` was running into a lot of issues at scale? 
If it's slower but works, it might be a "slow" alternative is reproducibility is important for folks.

ryantwolf: I never tested `set_index`. I definitely was aware of it, but I can't quite remember why I didn't try it. I think I was worried about destroying the user's existing index if that was useful, but I doubt that many users would do something like that Will it always be deterministic in its ordering?

ayushdg: `set_index`/ `sort_values` will also sort by the `rand_col` therefore ensuring a more or less consistent distribution for the same seed (not sure if it's a stable sort which might slightly impact results if the same rand col values are repeated). 
It is however a lot more expensive and slower than a hash based shuffle

ryantwolf: Makes sense. From my testing it definitely is slower, but not by much. Only 26% slower on a dataset of 116 GB. I added another path for it though.

nemo_curator/modules/dataset_ops.py
```diff
@@ -0,0 +1,182 @@
...
+        self,
+        seed: Optional[int] = None,
+        npartitions: Optional[int] = None,
+        partition_to_filename: Callable[[int], str] = default_filename,
+    ) -> None:
+        """
+        Randomly permutes the dataset. This will make the original "filename" column invalid, so if the column is present it will be overwritten.
+        Args:
+            seed: The random seed that will be used to determine which partition (file) each datapoint goes to.
+                Even with the seed set, the shuffle is not guaranteed to be deterministic if done in parallel.
```
ayushdg, line 25: This can be updated, since it's deterministic now. But we might want to document that perf may take a hit if users want determinism, and the non deterministic shuffle is much faster

docs/user-guide/DocumentDataset.rst
```diff
@@ -137,3 +137,88 @@ In these cases, we recommend processing the input dataset in batches using a sim
...
+  shuffle = nc.Shuffle(seed=42)
+
+  shuffled_books = shuffle(books)
+
+  shuffled_books.to_json("shuffled_books/")
+
+* ``shuffle = nc.Shuffle(seed=42)`` This creates a shuffle operation that can be chained with
+  the various other modules in NeMo Curator. In this example, we fix the seed to be 42.
+  Even with a fixed seed, shuffling is only guaranteed to be deterministic if done with a single-threaded
+  and single process client. Dask allows us to perform the shuffling in parallel for speed gains, so using a
```
ayushdg, line 222: Same for docs here

nemo_curator/modules/dataset_ops.py
```diff
@@ -0,0 +1,182 @@
...
+            return self.shuffle_deterministic(dataset)
+
+    def shuffle_deterministic(self, dataset: DocumentDataset) -> DocumentDataset:
+        new_npartitions = (
+            dataset.df.npartitions if self.npartitions is None else self.npartitions
+        )
+
+        dataset.df[self.rand_col] = dataset.df.map_partitions(self._add_rand_col)
+
+        shuffled_df = dataset.df.set_index(self.rand_col, npartitions=new_npartitions)
```
ayushdg, line 52: Does it make sense to reset the index (rand_col) after shuffling here?

ayushdg (commented):
Some nits around documentation but generally looks good

nemo_curator/modules/dataset_ops.py
ryantwolf: Yeah I think it does. I could see issues later on keep this as the index if other steps in the pipeline want to use the index. I've updated it.

ayushdg (approved):

Issue #31, opened 2024-04-12
[BUG] Serialization Issue with SLURM


ryantwolf:
**Describe the bug**

When running the `filter_documents` utility on SLURM with the FastTextLangId, the following error occurs in the Dask worker log:
```
2024-04-11 16:46:41,873 - distributed.protocol.pickle - INFO - Failed to deserialize b'<long binary string omitted>.'
Traceback (most recent call last):
  File "/usr/local/lib/python3.10/dist-packages/distributed/protocol/pickle.py", line 94, in loads
    return pickle.loads(x, buffers=buffers)
  File "/usr/local/lib/python3.10/dist-packages/nemo_curator/filters/__init__.py", line 16, in <module>
    from .code import (
  File "/usr/local/lib/python3.10/dist-packages/nemo_curator/filters/code.py", line 21, in <module>
    from nemo.collections.common.tokenizers import SentencePieceTokenizer
  File "/usr/local/lib/python3.10/dist-packages/nemo/collections/common/__init__.py", line 16, in <module>
    from nemo.collections.common import data, losses, parts, tokenizers
  File "/usr/local/lib/python3.10/dist-packages/nemo/collections/common/losses/__init__.py", line 15, in <module>
    from nemo.collections.common.losses.aggregator import AggregatorLoss
  File "/usr/local/lib/python3.10/dist-packages/nemo/collections/common/losses/aggregator.py", line 19, in <module>
    from nemo.core.classes import Loss, typecheck
  File "/usr/local/lib/python3.10/dist-packages/nemo/core/__init__.py", line 16, in <module>
    from nemo.core.classes import *
  File "/usr/local/lib/python3.10/dist-packages/nemo/core/classes/__init__.py", line 33, in <module>
    from nemo.core.classes.modelPT import ModelPT
  File "/usr/local/lib/python3.10/dist-packages/nemo/core/classes/modelPT.py", line 32, in <module>
    from nemo.core import optim
  File "/usr/local/lib/python3.10/dist-packages/nemo/core/optim/__init__.py", line 32, in <module>
    from nemo.core.optim.optimizer_with_main_params import MainParamsOptimizerWrapper
  File "/usr/local/lib/python3.10/dist-packages/nemo/core/optim/optimizer_with_main_params.py", line 33, in <module>
    from megatron.core.tensor_parallel import copy_tensor_model_parallel_attributes
  File "/opt/megatron-lm/megatron/__init__.py", line 5, in <module>
[...]

PR #34, opened 2024-04-17
Disable PyTorch Compile Multiprocessing

ryantwolf:
Addresses #31 by disabling multiprocessing for PyTorch according to advice gathered from [this issue](https://github.com/pytorch/pytorch/issues/123329). Essentially, we need to set `os.environ["TORCHINDUCTOR_COMPILE_THREADS"] = "1"` before PyTorch is imported and [this variable](https://github.com/pytorch/pytorch/blob/main/torch/_inductor/config.py#L397) is initialized. I also moved the nemo import statement to inside of the filter so that any performance impact is minimized (though I observed no speed degradation in my tests).

### DIFF
--- a/nemo_curator/modules/__init__.py
+++ b/nemo_curator/modules/__init__.py
@@ -11,6 +11,9 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+import os
+
+os.environ["TORCHINDUCTOR_COMPILE_THREADS"] = "1"
 
 from .add_id import AddId
 from .exact_dedup import ExactDuplicates
--- a/nemo_curator/modules/__init__.py
+++ b/nemo_curator/modules/__init__.py
@@ -13,6 +13,10 @@
 # limitations under the License.
 import os
 
+# Disables multiprocessing in torch.compile calls.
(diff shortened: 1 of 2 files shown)


nemo_curator/modules/__init__.py
```diff
@@ -11,6 +11,9 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+import os
+
```
VibhuJawa, line 15: Minor nitpick of adding a comment here linking the issue a minor description

VibhuJawa (approved):
Looks good to me.  

Double checked we are not compiling any pytorch code using `torch.compile` , the speed of compilation is controlled by this variable.  So we should be fine. 

We should use `torch.compile`  for our model inferences long term though but i doubt compilation speed will matter then, if it does we can look at `worker_start_method` which is  said to fork to spawn (again long term, nothing to do right now). 

```
# The multiprocessing start method to use for inductor workers in the codecache.
# TODO: fork is not safe in a multithreaded environment, we should evaluate changing
# the default to spawn.
```

https://github.com/pytorch/pytorch/blob/c2f8bfae9cf8d3e3729ad8137d72a2c088c6a39c/torch/_inductor/config.py#L356-L359

PR #35, opened 2024-04-18
Fix metadata inference with pandas and dask

ryantwolf:
Prevents Dask from passing `pd.NA` to the filters for type inference on the scoring and filtering functions. Also fixes some issues with task decontamination working with pandas 2.0 strings and exploding.

With task decontamination, we convert the document text column (dtype=string) to a list of split documents (dtype=object). When calling explode on this column of split documents, the column maintains its object datatype even though now it's only strings. We need to recast the column for newer versions of pandas/dask where string and object are different datatypes.

### DIFF
--- a/nemo_curator/modules/filter.py
+++ b/nemo_curator/modules/filter.py
@@ -11,11 +11,18 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
+import pandas as pd
+from dask import dataframe as dd
 from dask.typing import no_default
 
 from nemo_curator.datasets import DocumentDataset
 from nemo_curator.utils.module_utils import is_batched
+
+
+# Override so that pd.NA is not passed during the metadata inference
+@dd.extensions.make_array_nonempty.register(pd.StringDtype)
+def _(dtype):
+    return pd.array(["a", "b"], dtype=dtype)
 
 
 class Score:
--- a/nemo_curator/modules/filter.py
+++ b/nemo_curator/modules/filter.py
@@ -12,17 +12,17 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 import pandas as pd
-from dask import dataframe as dd
+from dask.dataframe.extensions import make_array_nonempty
 from dask.typing import no_default
 
 from nemo_curator.datasets import DocumentDataset
 from nemo_curator.utils.module_utils import is_batched
 
-
 # Override so that pd.NA is not passed during the metadata inference
-@dd.extensions.make_array_nonempty.register(pd.StringDtype)
-def _(dtype):
-    return pd.array(["a", "b"], dtype=dtype)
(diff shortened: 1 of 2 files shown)


ayushdg:
cc: @rjzamora if you want to take a look

nemo_curator/modules/task.py
```diff
@@ -318,7 +320,15 @@ def _remove_ngrams_partition(self, partition, task_ngrams, ngrams_freq_sorted):
...
-        return filtered_partition.explode(self.text_field, ignore_index=True)
+        exploded_partition = filtered_partition.explode(
+            self.text_field, ignore_index=True
+        )
+        # After exploding, the string datatype can become an "object" type
+        exploded_partition[self.text_field] = exploded_partition[
+            self.text_field
+        ].astype(text_type)
+
+        return exploded_partition
```
ayushdg, line 331: Is this a pandas bug that we somehow exposed or maybe expected behavior?

ryantwolf: I can't find anything on the pandas side of things that would suggest it's a bug. I doubt it is, but my understanding could be incorrect. In my head, the resulting column from an explode operation should probably be an `object` type since lists can contain anything. Though, not sure if pandas says somewhere that it should automatically convert the column datatype if all elements of the list are all strings/ints or whatever.

ryantwolf: It's only in our case that we can guarantee all the list elements will be strings. And since we make the promise to Dask in the `from_delayed` call on line 278 that the resulting dataframe will still have the text field as a string, we need to do this conversion.

nemo_curator/modules/filter.py
```diff
@@ -11,13 +11,20 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
+import pandas as pd
+from dask import dataframe as dd
```
rjzamora, line 15: If you are only using `make_array_nonempty`, you can also make the import a bit more targeted:

```python
from dask.dataframe.extensions import make_array_nonempty
```

nemo_curator/modules/filter.py
```diff
@@ -11,13 +11,20 @@
...
 from dask.typing import no_default
 
 from nemo_curator.datasets import DocumentDataset
 from nemo_curator.utils.module_utils import is_batched
 
 
+# Override so that pd.NA is not passed during the metadata inference
+@dd.extensions.make_array_nonempty.register(pd.StringDtype)
+def _(dtype):
+    return pd.array(["a", "b"], dtype=dtype)
```
rjzamora, line 25: LGTM as is. Just to share the non-decorator alternative:

```python
make_array_nonempty.register(
    pd.StringDtype,
    lambda x: pd.array(["a", "b"], dtype=x),
)
```

rjzamora (commented):
`make_array_nonempty`-related changes look correct to me - Thanks! Left two minor comments.

ayushdg (approved):

PR #36, opened 2024-04-19
Improve speed of AddId module

ryantwolf:
Improves the speed of the add id module by reducing the size of the task graph. However, this new method uses a different id structure. If users need the old format that is guaranteed to be ordered, they can still access by passing in a definitive `start_id`.


ayushdg (approved):
Thanks for the changes. Can confirm this doesn't impact the GPU dedup results from small scale testing.

Issue #33, opened 2024-04-17
[BUG]  FastTextLangId doesnt return a list truely


zahramahani:
**Describe the bug**

    lambda score: score[1]
IndexError: string index out of range

**Steps/Code to reproduce bug**

i used examples files to download common crawl with download_common_crawl.py after that i decided to seprate Persian language so i run identify_languages_and_fix_unicode.py  but when i run the python script the above error occurs

here are varibles that i changed in first file

    start_snapshot = "2023-50"
    end_snapshot = "2024-10"
    output_directory = "../../../datasets/common_crawl/"

second file changes

    multilingual_data_path = "../../../datasets/common_crawl"
    language_separated_output_path = "../../../datasets/common_crawl_lang_separated"
    cleaned_data_output_path = "../../../datasets/common_crawl_cleaned"

    # Download a fastText language identification model
    # and see a list of supported languages here:
    # https://fasttext.cc/docs/en/language-identification.html
    model_path = "../../../lid.176.bin"
    target_language = "FA"
    language_field = "language"

**Expected behavior**

run with no error

**Environment overview **

 - Environment location: local server
 - Method of NeMo-Curator install: pip install --extra-index-url https://pypi.nvidia.com .

**Environment details**

- OS version : Debian GNU/Linux 11 (bullseye)
- Dask version : 2024.1.1
- Python version: 3.10.14

ryantwolf:
Thanks for raising this issue! I'm always happy to see people contributing to the project. I think I've discovered the root cause. It appears that when Dask performs type inference on the function, it passes in a string instead of a list. I'll make a PR that explicitly annotates the meta so no type inference is needed.

This bug has seemingly revealed another though, as the mentioned fix isn't actually enough. If you only implement that, you may be left with another error `Exception: 'TypeError("\'>=\' not supported between instances of \'str\' and \'float\'")'`. This comes from [this issue](https://github.com/dask/dask/issues/10631) where Dask seems to improperly convert object types to strings. Therefore, in the `keep_document` call of `FastTextLangId`, `score[0]` is not `0.5435` (or whatever the lang id score is), but instead it's `'['` because `score` was converted to a string. We'll likely need to disable the string conversion for the filter with `dask.config.set({"dataframe.convert-string": False})`.

Thanks again for raising the issue! I should have a PR up shortly for this.

zahramahani:
i added this line in the start of main function the next error occured

```
kwargs:    {}
[...]

PR #37, opened 2024-04-19
Fix lang id example

ryantwolf:
Addresses #33


ryantwolf:
For clarity, I verified that the unit tests I added for emulating the language identification filter do fail without this fix.

tests/test_filters.py
```diff
@@ -756,3 +758,71 @@ def test_per_extension_filter(self):
...
+    """
+
+    def __init__(self, min_langid_score=0.3):
+        super().__init__()
+        self._cutoff = min_langid_score
+
+        # Dask will automatically convert the list score type
+        # to a string without this option.
+        # See https://github.com/NVIDIA/NeMo-Curator/issues/33
+        dask.config.set({"dataframe.convert-string": False})
```
ayushdg, line 794: It might make sense to add another test where this is set to true and mark it as `xfail`. That way we'll know if the behavior changes later on, or can debug what changes to be made if we want to fix this behavior with this set to True

ayushdg (approved):
Generally lgtm! Small suggestion around tests but non blocking

PR #39, opened 2024-04-22
Adds Nemo Curator K8s example

terrykong:


docs/user-guide/KubernetesCurator.rst
```diff
@@ -0,0 +1,382 @@
...
+    kubectl get pods -l "dask.org/cluster-name=$DASK_CLUSTER_NAME"
+
+    # NAME                                                     READY   STATUS    RESTARTS      AGE
+    # rapids-dask-default-worker-587238cf2c-7d685f4d75-k6rnq   1/1     Running   0             57m
+    # rapids-dask-default-worker-f8ff963886-5577fff76b-qmvcd   1/1     Running   3 (52m ago)   57m
+    # rapids-dask-scheduler-654799869d-9bw4z                   1/1     Running   0             57m
+
+.. tab-set::
+
+    .. tab-item:: Running Existing Module
```
ryantwolf, line 191: These tab items don't really stand out on GitHub. They might on the NeMo user guide, but I'd prefer if we could get these docs to look good on both the user guide and GitHub. Can you change this so `Running Existing Module` is more noticeable (like a subheading or something)?

docs/user-guide/KubernetesCurator.rst
```diff
@@ -0,0 +1,382 @@
...
+            #                  adlr_id                           _hashes
+            # 0  cc-2023-14-0397113620  91b77eae49c10a65d485ac8ca18d6c43
+            # 1  cc-2023-14-0397113621  a266f0794cc8ffbd431823e6930e4f80
+            # 2  cc-2023-14-0397113622  baee533e2eddae764de2cd6faaa1286c
+            # 3  cc-2023-14-0397113623  87dd52a468448b99078f97e76f528eab
+            # 4  cc-2023-14-0397113624  a17664daf4f24be58e0e3a3dcf81124a
+            # Finished!
+
+
+    .. tab-item:: Running Custom Module
```
ryantwolf, line 253: Same comment as Running Existing Module

docs/user-guide/KubernetesCurator.rst
```diff
@@ -0,0 +1,382 @@
...
+
+Create a Dask Cluster
+---------------------
+
+Use the ``create_dask_cluster.py`` to create a CPU or GPU dask cluster.
+
+.. note::
+    If you are creating another Dask cluster with the same ``--name <name>``, first delete it via::
+
+        kubectl delete daskcluster <name>
```
ryantwolf, line 144: Nit: Could you wrap this command in a code block or an inline code formatting?

docs/user-guide/KubernetesCurator.rst
```diff
@@ -0,0 +1,382 @@
...
+
+To copy into the ``nemo-workspace`` PVC, we will do so with ``kubectl exec``. You may also
+use ``kubectl cp``, but ``exec`` has fewer surprises regarding compressed files:
+
+.. code-block:: bash
+  
+    # Replace <...> with a path on your local machine 
+    LOCAL_WORKSPACE=<...>
+
+    # This copies $LOCAL_WORKSPACE/big_english to /big_english within the PVC.
```
ryantwolf, line 128: Nit: Could you change all `big_english` names to something that more clearly designates a dataset? Perhaps something as simple as `my_dataset` would be good.

examples/k8s/create_dask_cluster.py
```diff
@@ -0,0 +1,103 @@
...
+        name_to_path = {}
+        for pvc in specs.split(","):
+            # Can be empty
+            if not pvc:
+                continue
+            name, _, path = pvc.partition(":")
+            name_to_path[name] = path
+        return name_to_path
+
+    parser = argparse.ArgumentParser()
```
ryantwolf, line 82: Could you please add a help section to all of the args?

docs/user-guide/KubernetesCurator.rst
```diff
@@ -0,0 +1,382 @@
...
+secret key on the k8s cluster with (replace ``<NGC KEY HERE>`` with your NGC secret
+key. Note that if you have any special characters in your key you might need to wrap
+the key in single quotes (``'``) so it can be parsed correctly by k8s)::
+
+    kubectl create secret docker-registry ngc-registry --docker-server=nvcr.io --docker-username=\$oauthtoken --docker-password=<NGC KEY HERE>
+
+Setup Python Environment
+------------------------
+
+Setup a virtual environment:
```
ryantwolf, line 108: Could you expand on what this environment is used for? Users might get confused and wonder if the need to install NeMo Curator in the environment or just dask kubernetes.

ryantwolf (changes requested):
Looks good overall. Thanks for doing this! Just left a few minor comments on expanding/reformatting some of the documentation. Also I noticed the DCO and precommit checks are failing. You also might not have signed your commits with a GPG key. Make sure you use `git commit-sS` and check the output of precommit to see what's wrong.

ayushdg:
cc: @jacobtomlinson if you're interested in taking a look

docs/user-guide/KubernetesCurator.rst
```diff
@@ -0,0 +1,386 @@
...
+
+.. note::
+    If you are creating another Dask cluster with the same ``--name <name>``, first delete it via::
+
+        kubectl delete daskcluster <name>
+
+.. code-block:: bash
+
+    # Creates a CPU Dask cluster with 1 worker
+    python create_dask_cluster.py \
```
jacobtomlinson, line 151: This command appears to have a lot of overlap with the built-in `dask kubernetes gen cluster` command. I wonder if things can be made more flexible there so it can be used directly instead?

terrykong: Cool, I wasn't aware of that. It looks like the flags are similar to what `make_cluster_spec` accepts
```sh
dask kubernetes gen cluster --help
Usage: dask kubernetes gen cluster [OPTIONS]

  Generate a DaskCluster spec, see ``dask_kubernetes.operator.make_cluster_spec``.

Options:
  --name TEXT                    Name of the DaskCluster  [required]
  --image TEXT                   Container image to use (default
                                 'ghcr.io/dask/dask:latest')
  --n-workers INTEGER            Number of workers
  --resources TEXT               Resource constraints
  -e, --env TEXT                 Environment variables
  --worker-command TEXT          Worker command (default 'dask-worker')
  --scheduler-service-type TEXT  Service type for scheduler (default 'ClusterIP')
  --jupyter                      Start Jupyter on the scheduler (default 
```

So someone still needs to edit the spec to, say, add PVC mounts or remove `resources` from the scheduler (e.g., GPUs or IB devices), they'd need a script like this to do that. In light of this, do you still see a path of using `dask kubernetes gen cluster` by adding some of these configuration options, or do you think this opinionated script still has value?

jacobtomlinson: I was thinking we might be able to satisfy most of what you need by adding a couple more flags. 

If all you need is a flag to set pvc mounts and a flag to set scheduler resources independently then I could see us adding that quite easily.

PR #41, opened 2024-04-23
Fix failing GPU tests with latest pandas bump

ayushdg:
The latest version bumps also resulted in pandas 2.x being pulled in which brought in subtle differences in handling around object/list types. 

This PR fixes failing GPU tests after the bump


ryantwolf (approved):
LGTM. Tested on my end and they passed.

Issue #43, opened 2024-04-29
[BUG] Empty files created when invoking reshard_jsonl method at nemo_curator.utils.file_utils.py


miguelusque:
**Describe the bug**
Sometimes, when invoking `reshard_jsonl` method at `nemo_curator.utils.file_utils.py`, some files are created with size zero, which makes other scripts fail.

This is produced by the method `.to_textfiles` when, because of any reason, one of the partitions in the dask bag is empty (that can happen because of multiple reasons. For instance, when you have a jsonl file consisting on a single line of around 1GB, and 5 lines with only a few megabytes each).

**Steps/Code to reproduce bug**

Create a jsonl file as detailed above, and it will produce those results.

I have some files like that in one of our clusters. Just ping me if you want to have access to them.

**Expected behavior**

No empty files created.

P.S.: I will submit a MR today to fix this issue

ayushdg:
Completed via #57

PR #44, opened 2024-04-29
Fix issue #43 (empty files creation) and improve reading/writing speed

miguelusque:
This commit fixes issue #43 (empty files created when invoking reshard_jsonl method at nemo_curator.utils.file_utils.py) by double-checking the files size after being generated, and deleting them with size zero.

In addition to that, I have noticed there is no need to parse to JSON object the content of the different lines, which should be already in json format. By removing that extra-parsing, there is a significant speed up in the execution of this method.

### DIFF
--- a/nemo_curator/utils/file_utils.py
+++ b/nemo_curator/utils/file_utils.py
@@ -181,7 +181,7 @@ def parse_str_of_num_bytes(s, return_str=False):
 def _save_jsonl(documents, output_path, start_index=0, max_index=10000, prefix=None):
     """Worker function to write out the data to jsonl files"""
 
-    def _output_json(document):
+    def _encode_text(document):
         return document.strip().encode("utf-8")
 
     def _name(start_index, npad, prefix, i):
@@ -194,7 +194,7 @@ def _save_jsonl(documents, output_path, start_index=0, max_index=10000, prefix=N
 
     output_glob_string = os.path.join(output_path, "*.jsonl")
 
-    output_files = documents.map(_output_json).to_textfiles(
+    output_files = documents.map(_encode_text).to_textfiles(
         output_glob_string,
         name_function=name,
     )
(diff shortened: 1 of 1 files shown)


nemo_curator/utils/file_utils.py
```diff
@@ -195,11 +194,22 @@ def _name(start_index, npad, prefix, i):
...
+                os.remove(output_file)
+        except Exception as exception:
+            print(
+                f"An exception occurred when trying to delete {output_file}.\n{exception}",
+                flush=True,
+            )
+
 
 def reshard_jsonl(
     input_dir, output_dir, output_file_size="100M", start_index=0, file_prefix=""
```
ryantwolf, line 215: If you could change the docstring for `start_index` on line 225 to add something like "Note: The indices may not be continuous if the sharding process would output an empty file in its place" (I would add this comment on line 225 itself, but GitHub isn't letting me right now for some reason)

ryantwolf (commented):
Thanks a ton for this @miguelusque! I love the kind of change where I can just run it on a large chunk of data, and use `diff` to see that everything is ok. Good catch with the speedup. I only have one minor suggestion, but we should be good.

nemo_curator/utils/file_utils.py
miguelusque: Hi @ryantwolf , done!

Thanks!

ryantwolf (approved):
Thanks again!

nemo_curator/utils/file_utils.py
```diff
@@ -182,8 +182,7 @@ def _save_jsonl(documents, output_path, start_index=0, max_index=10000, prefix=N
     """Worker function to write out the data to jsonl files"""
 
     def _output_json(document):
-        myjson = json.dumps(document, ensure_ascii=False)
-        return myjson.encode("utf-8")
+        return document.strip().encode("utf-8")
```
ayushdg, line 185: I would prefer renaming this function since it doesn't output a json string anymore, it assumes that the input is already a json. Maybe something like `_encode_text` or something similar to explain what it does.

miguelusque: Sure! It makes sense! I will change it now.

miguelusque: Done!

ayushdg (approved):
Thanks @miguelusque

PR #45, opened 2024-04-30
[Tutorials] Add a tutorial for PEFT data curation

Maghoumi:
This PR adds a new tutorial to demonstrate data curation for PEFT use-cases.


Maghoumi:
@ryantwolf marking as ready for review, but we're still facing the null bug.

ryantwolf:
Null bug should be fixed in #55

Maghoumi:
I rebased onto main (after #55 was merged) and I still see `null` entries. Do I need to modify something on my end to get this to work?

ryantwolf:
@Maghoumi did you reinstall after rebase? I'm running on your branch right now and not seeing any nulls.

Maghoumi:
Oops my bad, old habits!
It works now

arhamm1:
[like]  Arham Mehta reacted to your message:
________________________________

tutorials/peft-curation/docbuilder.py
```diff
@@ -0,0 +1,113 @@
...
+import requests
+
+from nemo_curator.download.doc_builder import (
+    DocumentDownloader,
+    DocumentExtractor,
+    DocumentIterator,
+)
+
+
+class EmailsDownloader(DocumentDownloader):
```
ryantwolf, line 28: These look super clean, do you think you could plug them into the `download_and_extract` [function](https://github.com/NVIDIA/NeMo-Curator/blob/main/nemo_curator/download/doc_builder.py#L151) that we have? It would be great if you could show that off too. It's technically a bit overkill for this, but still gives users a good perspective.

tutorials/peft-curation/docbuilder.py
```diff
@@ -0,0 +1,113 @@
...
+
+class EmailsExtractor(DocumentExtractor):
+    def __init__(self):
+        super().__init__()
+        # The regular expression pattern to extract subject/body/label into groups.
+        self._pattern = re.compile(
+            r"Subject:: (.*?)\nBody:: (.*?)\n.*\[/INST\] (.*?) <s>", re.DOTALL
+        )
+
+    def extract(self, content: str) -> Tuple[Set, str]:
```
ryantwolf, line 101: The return type annotation seems to be incorrect.

tutorials/peft-curation/filters.py
```diff
@@ -0,0 +1,47 @@
...
+    """
+    If the email is too long, discard.
+    """
+
+    def __init__(self, max_length: int = 5000):
+        super().__init__()
+        self.max_length = max_length
+
+    def score_document(self, text: str) -> bool:
+        return isinstance(text, str) and len(text) <= self.max_length
```
ryantwolf, line 28: Given the order you apply the filters, I don't think you need the `isinstance` call here.

tutorials/peft-curation/filters.py
```diff
@@ -0,0 +1,47 @@
...
+
+    def keep_document(self, score) -> bool:
+        return score
+
+
+class FilterEmptyEmails(DocumentFilter):
+    """
+    Discard all empty emails (either empty body, or labeled as empty).
```
ryantwolf, line 36: I feel like saying "Discard all empty emails" might be a bit misleading in this documentation given that the filter (without inverting it) actually just keeps the empty emails.

tutorials/peft-curation/main.py
```diff
@@ -0,0 +1,165 @@
...
+    client.close()
+    return os.path.join(out_path, os.path.basename(jsonl_fp))
+
+
+def main():
+    parser = argparse.ArgumentParser()
+    parser = add_distributed_args(parser)
+    args = parser.parse_args()
+    # Limit the total number of workers to ensure we don't run out of memory.
+    args.n_workers = min(args.n_workers, 1)
```
ryantwolf, line 154: Have you been commonly running out of memory with the default? If so, that might be evidence for us to change the default.

ryantwolf (commented):
Thanks for making this! I left a couple of ideas that I'd be interested in hearing your thoughts on.

tutorials/peft-curation/docbuilder.py
Maghoumi: Unfortunately not possible. There's an implicit assumption about the `text` field [here](https://github.com/NVIDIA/NeMo-Curator/blob/main/nemo_curator/download/doc_builder.py#L133), which makes it unsuitable for this example.

tutorials/peft-curation/docbuilder.py
Maghoumi: Fixed.

tutorials/peft-curation/filters.py
Maghoumi: Removed.

tutorials/peft-curation/filters.py
Maghoumi: Fixed.

Maghoumi:
@ryantwolf I addressed the comments that I could and pushed all the changes. Let me know if there's anything else needed before merging.

tutorials/peft-curation/main.py
ryantwolf: Hm ok we'll need to debug this some more later then. I would prefer you remove this and update the documentation of the tutorial if you think it's necessary.

tutorials/peft-curation/docbuilder.py
ryantwolf: Very well. We'll need to update the logic then to make it more flexible in the future. For now, we'll just go with this.

tutorials/peft-curation/main.py
Maghoumi: I can set this to 8 or something, but more than that will definitely cause OOM. Let me know if you'd prefer 8, which is also what we are using for the TinyStories one.

ryantwolf: yeah let's stick with 8 and work on diagnosing this later.

Maghoumi: Okay, modified to 8 and updated README.md.

ryantwolf (approved):
Thanks again for this!

ayushdg (approved):

PR #54, opened 2024-05-06
Fix #53 - Add batched files reading support to separate_by_metadata script

### LINKED ISSUE #53: [FEA] Add batched files reading to separate_by_metadata.py
**separate_by_metadata.py** script reads all the files at once, and distributes them through the different Dask workers. That could lead to OOMs.

To read the files in batches, to reduce the chances of an OOM.

miguelusque:
Added batched files reading support to **separate_by_metadata.py**, in order to avoid OOMs.

In the current implementation, all the files are read at once, and distributed to the workers.

With this approach, only a batch of them are sent to the workers at once.


miguelusque:
Please, do not review this PR yet. There is something wrong when writing metadata to this.

I will fix it asap. Thanks!

PR #55, opened 2024-05-06
Fix indexing in PII Modifier

ryantwolf:
Fixes an issue reported by @Maghoumi. When the PII Modifier takes in a partition that has been partially filtered, it does not respect the subset of the partition that was given to it. It overrides the index, and the modifications get applied to elements they were not supposed to.


singhva (commented):
LGTM!

singhva (approved):

PR #60, opened 2024-05-10
Align `extract_partitioning_index` logic with upstream shuffling

rjzamora:
Dask modified how `partitioning_index` is used for shuffling in https://github.com/dask/dask/pull/10705 (included in `dask>=2023.12.1`). This PR modifies `extract_partitioning_index` to use the same logic.

TODO:
- [ ] Testing


rjzamora:
cc @VibhuJawa

nemo_curator/utils/fuzzy_dedup_utils/merge_utils.py
```diff
@@ -15,15 +15,20 @@
...
 from dask.dataframe.core import new_dd_object
 from dask.dataframe.shuffle import partitioning_index
 from dask.highlevelgraph import HighLevelGraph
 from dask.utils import M
+from packaging.version import Version
 
 from nemo_curator.utils.fuzzy_dedup_utils.shuffle_utils import rearange_by_column_direct
 
+DASK_GT_2023_12_0 = Version(dask.__version__) > Version("2023.12.0")
+
```
ayushdg, line 31: nit: For consistency can we move this logic to https://github.com/NVIDIA/NeMo-Curator/blob/main/nemo_curator/_compat.py

ayushdg (changes requested):
Initial review since this is a wip. 
- I can test the fix out on larger datasets if you feel it's ready for testing.
- You might need to amend/update existing commits to include a signoff (commit -s), to pass the DCO check

nemo_curator/utils/fuzzy_dedup_utils/merge_utils.py
rjzamora: Thanks, I assumed this kind of "compat" code was living somewhere, but didn't look very hard :)

rjzamora:
@ayushdg - thanks for the review. Still don't have an evironment to test this myself. I will try to do that later today, but if it's easy for you to test it is very welcome on my end :)

ayushdg:
Can confirm I'm seeing expected results on the larger scale dataset with this fix. I'll run a few more tests on my end but it's generally looking good. 
@rjzamora Do you have a small example that checks consistency of the behavior of these two shuffle approaches, that could be added as a unit test with the PR?

rjzamora:
>Do you have a small example that checks consistency of the behavior of these two shuffle approaches, that could be added as a unit test with the PR?

@VibhuJawa - Still trying to fix my environment so I can confirm locally, but won't the `FuzzyDuplicates` tests fail without this fix in place?

ayushdg:
> > Do you have a small example that checks consistency of the behavior of these two shuffle approaches, that could be added as a unit test with the PR?
> 
> @VibhuJawa - Still trying to fix my environment so I can confirm locally, but won't the `FuzzyDuplicates` tests fail without this fix in place?

In theory the shuffle gives incorrect results, but the dataset/num_partitions here is small enough that it doesn't impact the correctness of final results (duplicate documents detected)

VibhuJawa (approved):
Approving, lets merge it in when we add the requested unit test.

tests/test_fuzzy_dedup.py
```diff
@@ -367,3 +370,67 @@ def test_from_yaml(self, tmpdir):
         config = FuzzyDuplicatesConfig.from_yaml(tmpdir / "config.yaml")
         for param in yaml_params:
             assert getattr(config, param) == yaml_params[param]
+
+
+# TODO: This test should also work on CPU. However,
+# `shuffle_utils.py` will still try to import cudf
```
rjzamora, line 376: Do we care that this test also works with pandas-backed data? If so, I can also adda `"backend"` parameter to the test.

tests/test_fuzzy_dedup.py
```diff
@@ -367,3 +370,67 @@ def test_from_yaml(self, tmpdir):
...
+
+    # Use `extract_partitioning_index` to calculate
+    # the partitioning index and assign it as a new
+    # "_partitions" column
+    result, _ = extract_partitioning_index(
+        unshuffled,
+        "part_id",
+        bk_mapping,
+        npartitions_right,
+        npartitions_right,
```
rjzamora, line 419: I'm having trouble wrapping my head around the "correct" way to test the batched case. However, my sense is that this test already covers the critical requirement that we are extracting a partitioning index that is consistent with `ddf.shuffle`.

nemo_curator/utils/fuzzy_dedup_utils/merge_utils.py
```diff
@@ -195,6 +212,10 @@ def merge_left_to_shuffled_right(
     subset_bucket_df,
     merge_on,
 ):
+    from nemo_curator.utils.fuzzy_dedup_utils.shuffle_utils import (
+        rearange_by_column_direct,
+    )
```
rjzamora, line 217: Note that I moved this import into `merge_left_to_shuffled_right`, because `shuffle_utils` currently requires `cudf`/`dask_cuda`, while other logic in this module does not.

tests/test_fuzzy_dedup.py
ayushdg: Yeah. I don't think this specific bug impacts the batched case any differently than the reproducer here so I think it should be good to go.

ayushdg (approved):
Thanks a lot @rjzamora

Issue #59, opened 2024-05-09
[BUG] OOM errors while running Deduplication


vipulraheja:
**Describe the bug**
While trying to run the demo scripts using Common Crawl (CC-MAIN-2020-50) through the deduplication steps, the `gpu_compute_minhashes` and `minhash_buckets` steps are crashing non-deterministically where the workers fail to start. I have confirmed that the GPUs were idle before launching these steps. Specifically, I observe that there are multiple processes that get spawned on one GPU, which eventually goes OOM, while others stay idle.

Here is the error log (for the `gpu_compute_minhashes` stage. The `minhash_buckets` fails similarly.):
```
(venv) vipul.raheja@vr-lm-1:~/NeMo-Curator$ gpu_compute_minhashes   --input-data-dirs /data/commoncrawl/CC-MAIN-2020-50/langs/ENGLISH_cleaned_id   --output-minhash-dir /data/commoncrawl/CC-MAIN-2020-50/langs/ENGLISH_cleaned_minhash   --input-json-text-field text   --input-json-id-field adlr_id   --minhash-length 256   --char-ngram 5   --hash-bytes 4   --seed 42   --log-dir ./log/

2024-05-08 18:47:16,603 - distributed.worker - ERROR - std::bad_alloc: out_of_memory: CUDA error at: /__w/rmm/rmm/include/rmm/mr/device/cuda_async_view_memory_resource.hpp:112: cudaErrorMemoryAllocation out of memory
Traceback (most recent call last):
  File "/home/vipul.raheja/NeMo-Curator/venv/lib/python3.10/site-packages/distributed/utils.py", line 832, in wrapper
    return await func(*args, **kwargs)
  File "/home/vipul.raheja/NeMo-Curator/venv/lib/python3.10/site-packages/distributed/worker.py", line 1873, in plugin_add
    result = plugin.setup(worker=self)
  File "/home/vipul.raheja/NeMo-Curator/venv/lib/python3.10/site-packages/dask_cuda/plugins.py", line 64, in setup
    mr = rmm.mr.CudaAsyncMemoryResource(
  File "memory_resource.pyx", line 338, in rmm._lib.memory_resource.CudaAsyncMemoryResource.__cinit__
MemoryError: std::bad_alloc: out_of_memory: CUDA error at: /__w/rmm/rmm/include/rmm/mr/device/cuda_async_view_memory_resource.hpp:112: cudaErrorMemoryAllocation out of memory
[...]

PR #61, opened 2024-05-10
Only import PII constants during Curator import

ayushdg:
One approach to fixing #59. 
The root cause seems to come from the fact that importing space (which imports and calls a cupy cuda function) somehow impacts the state of the system and prevent a cluster starting up using all available GPUs on the machine. 

Currently the only imports in Curator that lead to this situation is importing `DEFAULT_LANGUAGE` from the pii modules which transitively ends up importing presidio->spacy.

This pr moves these constants to a separate file so that we don't end up importing all other dependencies during Curator import.


nemo_curator/modifiers/pii_modifier.py
```diff
@@ -17,7 +17,7 @@
 import pandas as pd
 
 from nemo_curator.modifiers import DocumentModifier
-from nemo_curator.pii.algorithm import DEFAULT_LANGUAGE
+from nemo_curator.pii.constants import DEFAULT_LANGUAGE
```
VibhuJawa, line 20: n.p. 

Why not import it DEFAULT_MAX_DOC_SIZE here, if its just a constant? 

```suggestion
from nemo_curator.pii.constants import DEFAULT_LANGUAGE, DEFAULT_MAX_DOC_SIZE
```

nemo_curator/pii/algorithm.py
```diff
@@ -30,36 +30,16 @@
 from presidio_anonymizer import AnonymizerEngine, BatchAnonymizerEngine
 from presidio_anonymizer.entities import OperatorConfig
 
+from nemo_curator.pii.constants import DEFAULT_LANGUAGE, SUPPORTED_ENTITIES
 from nemo_curator.pii.custom_batch_analyzer_engine import CustomBatchAnalyzerEngine
 from nemo_curator.pii.custom_nlp_engine import CustomNlpEngine
 from nemo_curator.pii.recognizers.address_recognizer import AddressRecognizer
```
VibhuJawa, line 37: Will suggest adding a comment here (and creating an issue) that importing this module leads to cuda context problems when used with dask_cuda.

VibhuJawa (approved):
Mostly looks good to me, nitpicks around  comments and code structure. 

Thanks a ton for fixing this for us.

nemo_curator/pii/algorithm.py
ayushdg: Opened #64 and added comment

VibhuJawa:
Tested locally, works, please go ahead and merge.

PR #74, opened 2024-05-21
Fuzzy Dedup:  Use text_field instead of hardcoded text column

ayushdg:
One of the functions in jaccard computation for Fuzzy dedup assume the text field of the dataset to be called `text` and doesn't use the `text_field` information provided. 

Additionally, explicitly set query planning to false.


ayushdg:
> Good catch. Thanks!

Whoops just realized I was setting query-planning to True in the PR. just updated.

PR #75, opened 2024-05-22
Fix #63. Add --input-meta parameter to explicitly specify the jsonl field dtypes

### LINKED ISSUE #63: [FEA] Add --meta parameter to explicitly specify the jsonl field dtypes
When reading jsonl files with Dask, the dataframe datatypes are inferred unless explicitly specified.

Inferring the data types can lead to several issues, such as incorrect type inference, degradation of performance and increased memory usage among others.

I think we could mitigate those issues if we would add a `--meta` parameter, which would receive a dictionary of datatypes.

That parameter would be optional, and be similar to the `--meta` parameter available here: [https://docs.dask.org/en/latest/generated/dask.dataframe.read_json.html](https://docs.dask.org/en/latest/generated/dask.dataframe.read_json.html).

miguelusque:
When reading jsonl files with Dask, the dataframe datatypes are inferred unless explicitly specified.

Inferring the data types can lead to several issues, such as incorrect type inference, degradation of performance and increased memory usage among others.

I think we could mitigate those issues if we would add a `--input-meta` parameter, which would receive a dictionary of datatypes.

That parameter would be optional, and be similar to the `--meta` parameter available here: https://docs.dask.org/en/latest/generated/dask.dataframe.read_json.html.


nemo_curator/download/doc_builder.py
```diff
@@ -190,8 +193,9 @@ def download_and_extract(
         extractor=extractor,
         output_type=output_type,
         keep_raw_download=keep_raw_download,
-        force_download=force_download,
+        force_download=force_dssownload,
```
ryantwolf, line 196: Nit, fix typo

examples/blend_and_shuffle.py
```diff
@@ -46,6 +49,14 @@ def attach_args(
         formatter_class=argparse.ArgumentDefaultsHelpFormatter
     ),
 ):
+    parser.add_argument(
```
ryantwolf, line 52: I would not add this argument to the scripts in `examples/`. The examples are meant to be as simple as possible. The only reason why we have the argparse in the first place here is so that the examples can be easily run on SLURM without modifications.

nemo_curator/scripts/download_and_extract.py
```diff
@@ -116,6 +117,13 @@ def attach_args(
         required=False,
         help="Path to input data directory",
     )
+    parser.add_argument(
+        "--input-meta",
```
ryantwolf, line 121: You should add this argument to every script in this folder that deals with a single dataset as input.

nemo_curator/utils/distributed_utils.py
```diff
@@ -186,6 +187,8 @@ def read_single_partition(
         files: The path to the jsonl files to read.
         backend: The backend to use for reading the data. Either "cudf" or "pandas".
         add_filename: Whether to add a "filename" column to the DataFrame.
+        input-meta: A dictionary with the json object field names and data types.
```
ryantwolf, line 190: It would be good if this parameter could be a `Union[str, dict]` so that users could directly pass in a dictionary if they are using the Python API, or use the `str` if they are on the command line. Please update the docstring when you do this too to mention that the string should be "a string representation of a dictionary" or something along those lines.

tutorials/peft-curation/main.py
```diff
@@ -163,7 +165,15 @@ def run_curation_pipeline(args: Any, jsonl_fp: str) -> str:
 def main():
     parser = argparse.ArgumentParser()
     parser = add_distributed_args(parser)
+    parser.add_argument(
+        "--input-meta",
```
ryantwolf, line 169: Since we haven't run into any issues in the tutorial datasets, and for the clarity of the tutorials, I think you should omit the argument from them.

ryantwolf (commented):
I like this idea! I left a few changes I'd like to see made, but nothing major. Also it looks like there are some merge conflicts so you should probably resolve those.

ryantwolf:
After sleeping on it, I do have one more nit. I think the argument for all the `read_*` functions should be `meta`. The argparse arguments can still be `--input-meta`. I want to try and keep the arguments and behavior as close to Dask as possible.

miguelusque:
> After sleeping on it, I do have one more nit. I think the argument for all the `read_*` functions should be `meta`. The argparse arguments can still be `--input-meta`. I want to try and keep the arguments and behavior as close to Dask as possible.

Hi @ryantwolf , thank you for all your feedback.

I am addressing all the comments you have made. My first thought about the name of the parameter was to name it `meta`, but then I thought that, in the near future, we may want to also specify the `output-meta` parameter, where we may want to specify specific formats for the output files.

Even if I think it is very likely that `meta` might work for both use-cases, I thought it would be clearer to differentiate them.

Looking forward to your thoughts. Thanks!!!

nemo_curator/download/doc_builder.py
miguelusque: Done!

nemo_curator/scripts/download_and_extract.py
miguelusque: I tried to do it to any script that invoked (directly or indirectly) any of the methods where `input_meta` is used. Please, let me know if I have missed anything. Thanks!

nemo_curator/utils/distributed_utils.py
miguelusque: Done!

nemo_curator/utils/distributed_utils.py
```diff
@@ -218,6 +226,11 @@ def read_single_partition(
     else:
         raise RuntimeError("Could not read data, please check file type")
 
+    if input_meta:
```
ayushdg, line 229: Same recommendation as above

ayushdg, line 229: Since this argument is only valid for `jsonl` formats, would it make sense to move it to the section above where `filetype=="jsonl"`? We can raise a warning if the user specifies this for other filetypes saying it'll not do anything.

nemo_curator/scripts/verify_classification_results.py
```diff
@@ -136,9 +148,16 @@ def verify_results(
         expected_results_file_path: The path of the expected_result file.
         results_pred_column: The prediction column name for the input files.
         expected_pred_column: The prediction column name for the expected_result file.
+        input_meta: A dictionary or a string formatted as a dictionary, which outlines
+            the field names and their respective data types within the JSONL input file.
 
     """
-    expected_df = pd.read_json(expected_results_file_path, lines=True)
+    if input_meta:
+        read_kwargs["dtype"] = (
```
ayushdg, line 156: This file doesn't use `read_kwargs` perhaps you meant to update `input_meta` here?

nemo_curator/utils/fuzzy_dedup_utils/io_utils.py
```diff
@@ -26,31 +27,41 @@
...
+def _read_json_func(
+    files, engine="cudf", include_path_column=False, columns=None, input_meta=None
+):
     """
     Reads multiple Json Lines files into a cuDF
     dataframe with an additional `path` column denoting the path
     of the input file.
     """
+    if input_meta:
+        read_kwargs["dtype"] = (
```
ayushdg, line 39: This file also doesn't use `read_kwargs`, probably want to update `input_meta`

ayushdg (changes requested):
Thanks for the changes! Left a few suggestions, but generally looks good. 

In addition to the changes, is it also possible to add a couple of tests, that attempt to use `DocumentDataset.read_json` with the `input_meta` param set both as a string and a dict, and verify that the result dtypes are what we'd expect.

nemo_curator/utils/fuzzy_dedup_utils/io_utils.py
miguelusque: Great catch! Thank you!

I have actually changed a bit the condition. Because the interface expects either a dictionary, a string or None, I only the string needs to be converted to dictionary, I have changed the logic a bit as follows:

>     if type(input_meta) == str:
>         input_meta = ast.literal_eval(input_meta)

nemo_curator/scripts/verify_classification_results.py
miguelusque: Great catch! Thank you!

I have actually changed a bit the condition. Because the interface expects either a dictionary, a string or None, I only the string needs to be converted to dictionary, I have changed the logic a bit as follows:

> if type(input_meta) == str:
>     input_meta = ast.literal_eval(input_meta)

nemo_curator/utils/distributed_utils.py
miguelusque: Great comment. Will do it right now.

nemo_curator/scripts/verify_classification_results.py
```diff
@@ -136,9 +148,16 @@ def verify_results(
...
+            ast.literal_eval(input_meta) if type(input_meta) == str else input_meta
```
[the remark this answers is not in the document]
miguelusque: Hi Ayush, you are right. The rationale behind accepting strings is to allow passing them from the CLI.

nemo_curator/scripts/verify_classification_results.py
```diff
@@ -136,9 +148,16 @@ def verify_results(
         expected_results_file_path: The path of the expected_result file.
         results_pred_column: The prediction column name for the input files.
         expected_pred_column: The prediction column name for the expected_result file.
+        input_meta: A dictionary or a string formatted as a dictionary, which outlines
+            the field names and their respective data types within the JSONL input file.
 
     """
-    expected_df = pd.read_json(expected_results_file_path, lines=True)
+    if input_meta:
```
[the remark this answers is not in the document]
miguelusque: Done!

nemo_curator/utils/distributed_utils.py
miguelusque: Done!

miguelusque:
> Thanks for the changes! Left a few suggestions, but generally looks good.
> 
> In addition to the changes, is it also possible to add a couple of tests, that attempt to use `DocumentDataset.read_json` with the `input_meta` param set both as a string and a dict, and verify that the result dtypes are what we'd expect.

Done! See `test_io.py` file, where I think we could start adding other IO-related tests.

ayushdg (approved):
Thanks a lot for ddressing all my comments.

Issue #50, opened 2024-05-03
Better mimic DocumentDataset's `read_*` functions to Dask's `read_*` functions


sarahyurick:
Right now, `DocumentDataset` has a couple of `read_*` functions:
(1)
```
def read_json(
    cls,
    input_files,
    backend="pandas",
    files_per_partition=1,
    add_filename=False,
)
```
(2)
```
def read_parquet(
    cls,
    input_files,
    backend="pandas",
    files_per_partition=1,
    add_filename=False,
)
```
(3)
```
def read_pickle(
    cls,
    input_files,
    backend="pandas",
    files_per_partition=1,
    add_filename=False,
)
```
It would be good if these functions could support Dask's [read_json](https://docs.dask.org/en/latest/generated/dask.dataframe.read_json.html) and [read_parquet](https://docs.dask.org/en/stable/generated/dask.dataframe.read_parquet.html) parameters (there is no `read_pickle` function in Dask but we can perhaps look to [Pandas](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.read_pickle.html) for this).

In addition to this, we can restructure our `to_*` functions as well.

ayushdg:
I believe the reason we have a custom read_json implementation is the ability to specify `files_per_partition` and combine multiple files into a single read_json call from cudf which isn't supported in dask dataframe. Since `parquet` and a few others have support for many params ootb, it makes sense to mimic dask in the parquet case.

sarahyurick:
From #46:
"
https://github.com/NVIDIA/NeMo-Curator/blob/main/examples/fuzzy_deduplication.py#L53-L60

Is there a reason you didn't use `DocumentDataset.read_parquet`? I would prefer to use that or expand its flexibility such that you can do what you need to do.

Yeah, the `DocumentDataset.read_parquet` functionality is a bit lacking in column support and a few other missing config options. I'd prefer the `DocumentDataset.read_parquet` method to mimic Dask's `read_parquet` for the time being.

I would be interested in that discussion as well. My intuition is that we should mimic the behavior of Dask as much as possible, but there might be good reasons to deviate.

Yeah, I agree that the goal should be to mimic Dask's `read_*` functions as best as possible, probably with kwargs.
"

sarahyurick:
From #130:
"
Couple of things here:
[...]

PR #77, opened 2024-05-22
Update Fuzzy dedup params for long strings support.

ayushdg:
Fuzzy deduplication is currently accelerated via cuDF which until release 24.04 had a limit that a string column could not exceed int32 number of characters. Consequently some defaults and core logic in the deduplication pipeline aims to mitigate errors for cases where we may exceed this value.

Starting 24.06, cuDF has experimental support for longer strings (int64 number of chars), and this PR attempts to change defaults and simplify logic around handling long strings.

### DIFF
--- a/nemo_curator/modules/fuzzy_dedup.py
+++ b/nemo_curator/modules/fuzzy_dedup.py
@@ -521,9 +521,9 @@ class FuzzyDuplicates:
                 documents_df=dataset.df,
                 bucket_w_anchors_path=mapped_buckets_w_anchors_path,
                 output_shuffled_docs_path=shuffled_docs_path,
-                bucket_mapping_df_blocksize=256,
-                parts_per_worker=1,
-                bucket_parts_per_worker=8,
+                bucket_mapping_df_blocksize=self.config.bucket_mapping_blocksize,
+                parts_per_worker=self.config.parts_per_worker,
+                bucket_parts_per_worker=self.config.bucket_parts_per_worker,
             )
             print(f"Stage{stage_num} (False Postive Check): Shuffle docs complete!")
             stage_num += 1
@@ -755,7 +755,7 @@ class _MapBuckets:
     ):
         # String bytes limit for cuDF
         # https://github.com/rapidsai/cudf/issues/13733
-        max_text_bytes_per_part = int(np.iinfo(np.int32).max // 1.2)
+        max_text_bytes_per_part = int(np.iinfo(np.int32).max * 1.2)
 
         self._logger.info(f"max_text_bytes_per_part = {max_text_bytes_per_part}")
         # Increasing in an attempt to prevent hitting
--- a/nemo_curator/modules/fuzzy_dedup.py
+++ b/nemo_curator/modules/fuzzy_dedup.py
@@ -755,7 +755,7 @@ class _MapBuckets:
     ):
         # String bytes limit for cuDF
         # https://github.com/rapidsai/cudf/issues/13733
-        max_text_bytes_per_part = int(np.iinfo(np.int32).max * 1.2)
+        max_text_bytes_per_part = int(np.iinfo(np.int32).max * 3)
 
         self._logger.info(f"max_text_bytes_per_part = {max_text_bytes_per_part}")
         # Increasing in an attempt to prevent hitting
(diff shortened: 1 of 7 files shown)


nemo_curator/modules/fuzzy_dedup.py
```diff
@@ -602,7 +602,7 @@ def _get_output_map_from_text_bytes_per_bucket(
     ):
         # String bytes limit for cuDF
         # https://github.com/rapidsai/cudf/issues/13733
-        max_text_bytes_per_part = int(np.iinfo(np.int32).max // 1.2)
+        max_text_bytes_per_part = int(np.iinfo(np.int32).max * 3)
```
VibhuJawa, line 605: Should we also make this configurable ?

nemo_curator/utils/fuzzy_dedup_utils/shuffle_utils.py
```diff
@@ -83,7 +83,7 @@ def get_shuffle_part_ids_df(
     num_workers=0,
 ):
     sizes = agg_df[size_col].values
-    max_text_bytes_per_part = int(np.iinfo(np.int32).max // 1.2)
+    max_text_bytes_per_part = int(np.iinfo(np.int32).max * 3)
```
VibhuJawa, line 86: Same suggestion here.

nemo_curator/modules/fuzzy_dedup.py
VibhuJawa, line 605: We should make this configurable too.

nemo_curator/utils/fuzzy_dedup_utils/shuffle_utils.py
VibhuJawa, line 86: Same suggestion here.

VibhuJawa (changes requested):
Mostly looks good to me,  just a minor nit around ` max_text_bytes_per_part`

nemo_curator/utils/fuzzy_dedup_utils/shuffle_utils.py
ayushdg: We can make this configurable by making it an argument to the function, and adding the option to set it when creating the `_Shuffle` class. I'm a bit hesitant to add it as a config option to `FuzzyDedupConfig` overall since there are a lot of knobs already and these are more advanced options that complicate the high level API. What do you think?

VibhuJawa: Like from a user perspective, I would rather have them use the more advanced option than modify the code, like we do here to deal with OOMs. Which are more probable, now that  we have long strings. 

Of course, We can do that later on, this is not a blocking review for the PR and we should land this in the code freeze today. 

Can you file an issue around this please.

ayushdg: Done: https://github.com/NVIDIA/NeMo-Curator/issues/233

nemo_curator/utils/distributed_utils.py
```diff
@@ -237,6 +237,8 @@ def read_single_partition(
         read_kwargs = {"lines": filetype == "jsonl"}
         if backend == "cudf":
             read_f = cudf.read_json
+            if input_meta is not None:
+                read_kwargs["prune_columns"] = True
```
sarahyurick, line 241: Do you think we should move away from `input_meta` in favor of a keyword like `dtype` (like Pandas' and cuDF's `read_json`) and having the user configure `prune_columns` themselves?

I think this would align with https://github.com/NVIDIA/NeMo-Curator/issues/50 too.

sarahyurick (approved):
LGTM, added a general question.

ryantwolf (approved):
LGTM

VibhuJawa (approved):
LGTM

nemo_curator/utils/distributed_utils.py
ayushdg: I'm generally in favor of overhauling the IO helpers in the current setup for something better. When we tackle #50. I'll share more thoughts there, but moving to encouraging users using the read_xyz api's is easier. 
We can then have a common helper that based on the filetype directs to the relevant `read_xyz` api rather than the other way around where `read_json` goes to a common read method that handles different formats.

Regarding: `prune_columns` specifically: This change is important in newer versions of rapids because many public datasets like rpv1 do not have consistent metadata across all their files. If we do not prune columns to just ID & Text, cuDF will now fail with inconsistent metadata errors.

PR #90, opened 2024-05-30
Add Resiliparse option for text extraction

sarahyurick:
Right now, we only support Common Crawl text extraction with jusText. Resiliparse is known to be a faster text extraction algorithm which may also produce better tokens.

This PR adds optional support for the Resiliparse algorithm while still keeping jusText as the default.


nemo_curator/download/commoncrawl.py
```diff
@@ -288,7 +362,7 @@ def download_common_crawl(
     url_limit=None,
 ) -> DocumentDataset:
```
sarahyurick, line 363: From my understanding, the `download_common_crawl` function is what the user calls in their Python script. However, it doesn't look like they can set any of the `extract_text` parameters themselves (like `length_low`, `length_high`, etc.). Should we update this so that the user has more control over these parameters?

nemo_curator/download/commoncrawl.py
```diff
@@ -280,6 +353,7 @@ def download_common_crawl(
     start_snapshot: str,
     end_snapshot: str,
     output_type: str = "jsonl",
+    algorithm: str = "justext",
```
sarahyurick, line 356: We will keep jusText as the default for this PR. If our research finds it beneficial to switch to Resiliparse, we can update this line to set Resiliparse as the default.

sarahyurick (commented):
Added a couple of comments for discussion. This PR should hopefully be relatively straightforward.

sarahyurick:
Hi @ryantwolf @jojennin this should be ready for review. I will fix the GitHub signed commits issue and branch conflicts shortly.

ryantwolf:
Perfect, I'll give this a review today or Monday. Thanks!

sarahyurick:
Reopened at https://github.com/NVIDIA/NeMo-Curator/pull/128.

PR #94, opened 2024-05-31
[WIP][Experiment] Use multi-threading for multi-file json reads

rjzamora:
Experimental change to improve IO performance when multiple json files are mapped to each dask-dataframe partition.

**Context**: I was originally exploring a similar optimization to improve remote-storage performance, and found a significant perf bump for local storage as well.


rjzamora:
cc @ayushdg

PR #97, opened 2024-06-03
Make sure query-planning is disabled for now

rjzamora:
As far as I can tell, the existing logic used to disable query-planning isn't really doing anything. As soon as `dask.dataframe` is imported, changing the `"dataframe.query-planning"` configuration will no longer have an effect.

### DIFF
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -11,3 +11,16 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+
+import dask
+
+# Disable query planning before any tests are loaded
+# https://github.com/NVIDIA/NeMo-Curator/issues/73
+if dask.config.get("dataframe.query-planning") is True:
+    raise NotImplementedError(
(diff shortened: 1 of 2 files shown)


tests/__init__.py
```diff
@@ -11,3 +11,16 @@
...
+# Disable query planning before any tests are loaded
+# https://github.com/NVIDIA/NeMo-Curator/issues/73
+if dask.config.get("dataframe.query-planning") is True:
+    raise NotImplementedError(
+        "NeMo Curator does not support query planning yet. "
+        "Please disable query planning before importing "
+        "`nemo_curator`, `dask.dataframe` or `dask_cudf`."
+    )
+else:
+    dask.config.set({"dataframe.query-planning": False})
```
rjzamora, line 26: I suppose we could also just import `nemo_curator` here as well.

ayushdg (approved):
Thanks!
I've run into similar issues when running tests. 
@rjzamora Is this good to merge, or do you want to update?

rjzamora:
>Is this good to merge, or do you want to update?

Current version of the PR should be fine, but happy to make revisions if requested :)

PR #99, opened 2024-06-05
Allow multiple filenames per partition when separating by metadata

### LINKED ISSUE #89: ValueError: More than one filename found in partition
**Describe the bug**

Ran into [Issue 49](https://github.com/NVIDIA/NeMo-Curator/issues/49) and tried increasing the number of partitions while reading the data using the `read_json` function from `nemo_curator.datasets import DocumentDataset`

```
File /usr/local/lib/python3.10/dist-packages/nemo_curator/utils/distributed_utils.py:405, in single_partition_write_with_filename()
    403 num_filenames = len(df.filename.unique())
    404 if num_filenames > 1:
--> 405     raise ValueError(
    406         f"More than one filename found in partition: {num_filenames}"
    407     )
    408 filename = Path(filename).stem
    409 output_file_path = os.path.join(output_file_dir, filename)

ValueError: More than one filename found in partition: 20
```

ayushdg:
## Description
Closes #89 

Adds an option to handle scenarios where there are multiple filenames in a single partition when writing with filename. 
This is typically true when files are read in with `files_per_partition` > 1.

## Usage

```python
dataset = DocumentDataset.read_json(path, files_per_partition=5, include_filename=True)
write_to_disk(dataset.df, output_file_dir=path, write_to_filename=True)
```
## Checklist

### DIFF
--- a/nemo_curator/utils/distributed_utils.py
+++ b/nemo_curator/utils/distributed_utils.py
@@ -409,10 +409,10 @@ def single_partition_write_with_filename(df, output_file_dir, output_type="jsonl
     assert "filename" in df.columns
 
     if len(df) > 0:
-        empty_partition = True
+        empty_partition = False
     else:
         warnings.warn(f"Empty partition found")
-        empty_partition = False
+        empty_partition = True
 
     if is_cudf_type(df):
         import cudf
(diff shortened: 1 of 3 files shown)


nemo_curator/utils/distributed_utils.py
```diff
@@ -421,34 +421,50 @@ def single_partition_write_with_filename(df, output_file_dir, output_type="jsonl
...
 
-    if empty_partition:
-        filename = df.filename.iloc[0]
-        num_filenames = len(df.filename.unique())
-        if num_filenames > 1:
+    if not empty_partition:
+        filenames = df.filename.unique()
+        filenames = list(filenames.values_host) if is_cudf_type(df) else list(filenames)
+        num_files = len(filenames)
+        if num_files > 1 and output_type != "jsonl":
```
VibhuJawa, line 428: We can also do it for `parquet` too , any reason we are skipping that ?

nemo_curator/utils/distributed_utils.py
```diff
@@ -421,34 +421,50 @@ def single_partition_write_with_filename(df, output_file_dir, output_type="jsonl
...
+                    out_df.to_json(
+                        output_file_path,
+                        orient="records",
+                        lines=True,
+                        force_ascii=False,
+                        mode="a",
+                    )
+            elif output_type == "parquet":
+                output_file_path = output_file_path + ".parquet"
+                out_df.to_parquet(output_file_path)
```
VibhuJawa, line 465: We can expand to do this `to_parquet` with `partition_on` and that should probably just work .

VibhuJawa (changes requested):
Thanks for working on this @ayushdg , should be helpful across the various modules . Requested some changes around supporting parquet

nemo_curator/utils/distributed_utils.py
ayushdg: The main reason for skipping parquet was the lack of `append` mode in df.to_parquet.
The original functionality of this method is to write out data from a single dataframe to a specific filename/filenames. 
Using `partition_on` will create multiple folders but deviates from the original behavior of writing out files with the same filename as the input.

VibhuJawa: Thanks for clarification. I think we need to support 2 use cases: 

1. Over-write existing file in the directory
2. Append to file in the the directory

Previously we were by default over-writing the file, a behavior i think is a good default because I anticipate folks trying to over-write files instead of wanting to append to existing files and getting confused as a result.  

Suggested Behavior : 


| File Format | Write Mode | Write to filename | Supported |
|-------------|------------|-------------------|-----------|
| jsonl       | Append     | True              |      True     |
| jsonl       | Append     | False             |       True    |
| jsonl       | Overwrite  | True              |      True     |
| jsonl       | Overwrite  | False             |    True       |
| json        | Append     | True              |     True      |
| json        | Append     | False             |      True     |
| json        | Overwrite  | True              |     True      |
| json        | Overwrite  | False             |      True     |
| parquet     | Append     | True              |     False      |
| parquet     | Append     | False             |     False      |
| parquet     | Overwrite  | True              |     True      |
| parquet     | Overwrite  | False             |      True     |

VibhuJawa (approved):
LGTM to merge, thanks for working on this @ayushdg

PR #102, opened 2024-06-10
Fix 69 - Refactor how arguments are added to scripts

miguelusque:
## Description
I have noticed that the code used to parse the scripts is repeated in multiple places.

I have refactored the different `parser.add_argument` calls, grouping them in a single file that is consumed by the different scripts.


examples/domain_classifier_example.py
```diff
@@ -89,52 +89,16 @@ def attach_args(
...
+    argumentHelper = ArgumentHelper(parser)
+
+    argumentHelper.add_device()
+    argumentHelper.add_enable_spilling()
+    argumentHelper.add_nvlink_only()
+    argumentHelper.add_protocol()
+    argumentHelper.add_rmm_pool_size()
+    argumentHelper.add_scheduler_address()
+    argumentHelper.add_scheduler_file()
+    argumentHelper.add_set_torch_to_use_rmm()
```
ayushdg, line 101: I think this should just be an overlap of the `add_distributed_args` functionality.

examples/k8s/create_dask_cluster.py
```diff
@@ -81,51 +83,7 @@ def parse_pvcs(specs: str) -> dict[str, str]:
...
-        help="If --image is from a private registry, specify the appropriate pull secret you created to allow these to be pulled.",
-    )
-    parser.add_argument(
-        "-p",
-        "--pvcs",
-        type=parse_pvcs,
-        default="",
-        help="Comma sep PVC specificiation of $pvc_name_1:$mount_path_1,$pvc_name_2:$mount_path_2. Example: foo:/foo,bar:/bar mounts pvcs named foo and bar to /foo and /bar respectively.",
-    )
+    ArgumentHelper(parser).add_create_dask_cluster_args()
```
ayushdg, line 86: I'm not sure k8's specific configurations should reside in the same `ArgumentHelper` class that also includes a lot of local cluster/ other configurations, and the function should be renamed to something like `add_create_k8s_dask_cluster_args` to avoid confusion. 

My personal preference is to leave k8's configuration in this file since it's not used elsewhere. Not sure if @ryantwolf has opinions.

examples/quality_classifier_example.py
```diff
@@ -60,52 +60,16 @@ def attach_args(
...
+    argumentHelper = ArgumentHelper(parser)
+
+    argumentHelper.add_device()
+    argumentHelper.add_enable_spilling()
+    argumentHelper.add_nvlink_only()
+    argumentHelper.add_protocol()
+    argumentHelper.add_rmm_pool_size()
+    argumentHelper.add_scheduler_address()
+    argumentHelper.add_scheduler_file()
+    argumentHelper.add_set_torch_to_use_rmm()
```
ayushdg, line 72: I suspect this has a high overlap with add_distributed_args

nemo_curator/scripts/add_id.py
```diff
@@ -73,77 +69,25 @@ def attach_args(
...
-        default="jsonl",
-        help="File type of the dataset to be read in. Supported file formats"
-        " include 'jsonl' (default), 'pickle', or 'parquet'.",
+    argumentHelper = ArgumentHelper(parser)
+
+    argumentHelper.add_id_field_name()
+    argumentHelper.add_id_prefix()
+    argumentHelper.add_input_data_dir()
+    argumentHelper.add_input_file_type()
+    argumentHelper.add_output_data_dir(
```
ayushdg, line 78: Since this is so specific to `add_id` not sure if needs to be in the helper class. And if it does, I'd prefer grouping them all in a single `add_id_args` method rather than individual methods.

nemo_curator/scripts/filter_documents.py
```diff
@@ -181,138 +177,24 @@ def attach_args(
...
+    argumentHelper.add_input_data_dir()
+    argumentHelper.add_input_file_type()
+    argumentHelper.add_input_local_data_dir()
+    argumentHelper.add_log_dir()
+    argumentHelper.add_log_frequency()
+    argumentHelper.add_log_scores()
+    argumentHelper.add_output_document_score_dir()
+    argumentHelper.add_output_file_type()
+    argumentHelper.add_output_removed_document_dir()
+    argumentHelper.add_output_retained_document_dir()
```
ayushdg, line 195: Most of these might be filter specific.

ayushdg (changes requested):
Thanks for the changes. Most of my comments are around the same topic:
- There are some parts of the codebase that repeat the same arguments across multiple files. It makes sense to group some of those into the common file for reuse. For other arguments, that are very specific to a use case or notebook, do you have a strong opinion on why they should reside in a common class rather than added directly in the example itself?

examples/domain_classifier_example.py
miguelusque: Done!

examples/k8s/create_dask_cluster.py
miguelusque: Hi @ayushdg , I have renamed the function as follows, `add_args_create_k8s_dask_cluster`, but left it in `ArgumentHelper` yet.

examples/quality_classifier_example.py
miguelusque: Hi @ayushdg , you are right. There is a high overlap with add_distributed_args in this file, `quality_classifier_example`, and also in `domain_classifier_example.py`.

If you do not mind, because the overlap is not 100% and it requires a deeper analysis if I can replace those calls with `add_distributed_args`, I would appreciate if we could pospone this change to a near future pull request. 

I hope you are happy with the above. Thanks!

nemo_curator/scripts/add_id.py
miguelusque: Done!

nemo_curator/scripts/filter_documents.py
miguelusque: Grouped into a single method.

examples/domain_classifier_example.py
miguelusque: Hi @ayushdg , you are right. There is a high overlap with add_distributed_args in this file, `domain_classifier_example.py`, and also in `quality_classifier_example.py`.

If you do not mind, because the overlap is not 100% and it requires a deeper analysis if I can replace those calls with `add_distributed_args`, I would appreciate if we could pospone this change to a near future pull request. 

I hope you are happy with the above. Thanks!

ryantwolf: > If you do not mind, because the overlap is not 100% and it requires a deeper analysis if I can replace those calls with add_distributed_args, I would appreciate if we could pospone this change to a near future pull request

I agree, we can postpone those to later.

nemo_curator/utils/script_utils.py
```diff
@@ -13,335 +13,1276 @@
...
-        parser,
-        "set-torch-to-use-rmm",
-        default=False,
-        help_str="Whether to set torch to use RMM or not",
-    )
-
-    return parser
-
-
-def chunk_list(lst, nchnks):
```
ryantwolf, line 268: Were these functions deleted by mistake?

nemo_curator/utils/script_utils.py
```diff
@@ -13,335 +13,1276 @@
...
+            type=int,
+            default=default,
+            help=help,
+        )
+
+    def add_arg_set_torch_to_use_rmm(self):
+        self.parser.add_argument("--set-torch-to-use-rmm", action="store_true")
+
+    def add_arg_shuffle(self, help: str):
+        ArgumentHelper.__attach_bool_arg(
```
ryantwolf, line 255: I believe this function call is missing an argparse object. Though, fixing it might be unnecessary with my other suggestions.

ryantwolf (changes requested):
In general I like this idea going forward, but I think I want to see a couple of changes. Relating to Ayush's review summary, I think that common arguments (like those found in `add_distributed_args`) should be grouped together. I think this class also makes more sense as a home for `add_distributed_args` and similar functions rather than them being entirely independent.

However, I think if there isn't any reuse of a list of arguments in a script, we shouldn't make a method in `ArgumentHelper` for it. Arguments that are hyper-specific to each script should remain only in that script and not be abstracted away. I'm fine abstracting away collections of arguments that are used in 2+ scripts though (like `input-data-dir`, `input-type`, etc.). So I think I'd like to see `ArgumentHelper` shrink a bit for now and return a lot of the niche arguments to their original scripts.

nemo_curator/utils/script_utils.py
miguelusque: Hi @ryantwolf , I searched for their usage, and there were not being used anywhere.

nemo_curator/utils/script_utils.py
miguelusque: Done! Thank you!

miguelusque:
> In general I like this idea going forward, but I think I want to see a couple of changes. Relating to Ayush's review summary, I think that common arguments (like those found in `add_distributed_args`) should be grouped together. I think this class also makes more sense as a home for `add_distributed_args` and similar functions rather than them being entirely independent.
> 
> However, I think if there isn't any reuse of a list of arguments in a script, we shouldn't make a method in `ArgumentHelper` for it. Arguments that are hyper-specific to each script should remain only in that script and not be abstracted away. I'm fine abstracting away collections of arguments that are used in 2+ scripts though (like `input-data-dir`, `input-type`, etc.). So I think I'd like to see `ArgumentHelper` shrink a bit for now and return a lot of the niche arguments to their original scripts.

Hi Ryan,

Thank you! Just done!

nemo_curator/scripts/prepare_fasttext_training_data.py
```diff
@@ -67,20 +67,19 @@ def attach_args(
...
-    parser.add_argument(
-        "--input-data-dir",
-        type=str,
-        default=None,
-        help="Input directory consisting of .jsonl files that are accessible "
-        "to all nodes. Use this for a distributed file system",
+    argumentHelper = ArgumentHelper(parser)
+
+    argumentHelper.add_arg_input_data_dir()
+    argumentHelper.add_arg_input_local_data_dir()
```
ryantwolf, line 73: We should delete this, it's not actually used in the script and is a relic of an older version.

nemo_curator/scripts/text_cleaning.py
```diff
@@ -71,49 +71,20 @@ def attach_args(
...
+    argumentHelper.add_arg_input_file_type()
+    argumentHelper.add_arg_input_text_field()
+    argumentHelper.add_arg_output_file_type()
+    argumentHelper.add_distributed_args()
     parser.add_argument(
         "--output-clean-dir",
         type=str,
-        default=None,
         required=True,
         help="The output directory to where the cleaned " "jsonl files will be written",
```
ryantwolf, line 86:
```suggestion
        help="The output directory to where the cleaned jsonl files will be written",
```

nemo_curator/scripts/train_fasttext.py
```diff
@@ -135,17 +144,30 @@ def attach_args(
...
     parser.add_argument(
-        "--output-train-file",
+        "--output-model",
         type=str,
-        default="./fasttext_samples.train",
-        help="The concatenated, shuffled samples used "
-        "to train the skip-gram classifier",
+        default=None,
+        required=True,
+        help="The output trained skip-gram classifier written " "as a FastText model",
```
ryantwolf, line 163:
```suggestion
        help="The output trained skip-gram classifier written as a FastText model",
```

nemo_curator/utils/fuzzy_dedup_utils/io_utils.py
```diff
@@ -60,8 +60,9 @@ def _read_json_func(
 
 
 def get_text_ddf_from_json_path_with_blocksize(
-    input_data_paths, num_files, blocksize, id_column, text_column, input_meta
+    input_data_paths, num_files, blocksize, id_column, text_column, input_meta=None
 ):
+    print("Hello world!")
```
ryantwolf, line 65: Looks like this was left by mistake.

nemo_curator/utils/fuzzy_dedup_utils/io_utils.py
```diff
@@ -60,8 +60,9 @@ def _read_json_func(
 
 
 def get_text_ddf_from_json_path_with_blocksize(
-    input_data_paths, num_files, blocksize, id_column, text_column, input_meta
+    input_data_paths, num_files, blocksize, id_column, text_column, input_meta=None
```
ryantwolf, line 63: Thanks for doing this too :)

nemo_curator/scripts/make_data_shards.py
```diff
@@ -61,26 +65,17 @@ def attach_args(
         help="Output directory to where the sharded " ".jsonl files will be written",
```
ryantwolf, line 65:
```suggestion
        help="Output directory to where the sharded .jsonl files will be written",
```

ryantwolf (changes requested):
This version is great and much more in line with what I think is the most useful. I found a couple of outdated arguments (that were a lot easier to spot thanks to you) that would be good to fix before merging this in. Once all my changes are addressed and I verify via the CI that everything is ok, we should be good.

nemo_curator/utils/fuzzy_dedup_utils/io_utils.py
miguelusque: Oops! 😅 Just removed it. Thank you for noticing it!

nemo_curator/scripts/find_pii_and_deidentify.py
```diff
@@ -92,89 +90,60 @@ def attach_args(
...
-
-    parser.add_argument(
-        "--supported-entities",
-        type=str,
-        default=None,
-        required=False,
-        help="Comma separated list of PII entity types. None implies all supported types",
+    argumentHelper.add_arg_input_data_dir(help="Directory containing the input files.")
+    argumentHelper.add_arg_input_file_type(
+        choices=["jsonl", "csv", "text"],
```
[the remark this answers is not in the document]
miguelusque: Hi Ryan,

I am not sure to understand it. arguentHelper.add_arg_input_file_type() method default value is "jsonl", which I think is the one that was originally in the code. Am I missing something? I also removed the `required=True` parameter because we had a default value, so no need to specify both.

Am I missing something? Thanks!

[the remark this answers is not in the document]
ryantwolf: Sorry, let me clarify. You did correctly copy over the original arguments from the file. I am pointing out that the original arguments were actually incorrect haha. Both `choices` and `help` should remain with the default values you have given them. Therefore, this function call should look like `argumentHelper.add_arg_input_file_type()` and the output file type should also be `argumentHelper.add_arg_output_file_type()`.

This is because in the script where `args.input_file_type` is used, the `read_data` function is called which actually supports the usual file types (`jsonl`, `parquet`, etc.) The `help` text is also incorrect for this function call since other file types are supported.

This mistake was easy to spot for me because of your changes, so it looks like this is proof your changes are going to be a big help. :)

examples/k8s/create_dask_cluster.py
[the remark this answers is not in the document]
ryantwolf: This still needs to be addressed.

ryantwolf (changes requested):
Just one comment of mine is left to be addressed. Other than that, should be good! I'll run the CI now just to double check

ryantwolf (approved):
Great, thanks again for this!

ayushdg (approved):
Confirmed it works with the fuzzy dedup cli pipeline

PR #107, opened 2024-06-12
Stricter check for query planning.

ayushdg:
## Description
Importing newer versions of dask (tested with 2024.5) set `dataframe.query-planning` to None instead of `True` or `False`, but the `None` case defaults to using query planning. This PR extends the check to `None` to raise relevant errors w/ newer versions of dask.

## Usage
N/A
## Checklist

### DIFF
--- a/nemo_curator/__init__.py
+++ b/nemo_curator/__init__.py
@@ -12,12 +12,13 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import sys
+
 import dask
 
 # Disable query planning if possible
 # https://github.com/NVIDIA/NeMo-Curator/issues/73
-QUERY_PLANNING = dask.config.get("dataframe.query-planning")
-if QUERY_PLANNING is True or QUERY_PLANNING is None:
+if dask.config.get("dataframe.query-planning") is True or "dask-expr" in sys.modules:
     raise NotImplementedError(
         """
         NeMo Curator does not support query planning yet.
--- a/nemo_curator/__init__.py
+++ b/nemo_curator/__init__.py
@@ -18,7 +18,7 @@ import dask
 
 # Disable query planning if possible
 # https://github.com/NVIDIA/NeMo-Curator/issues/73
-if dask.config.get("dataframe.query-planning") is True or "dask-expr" in sys.modules:
+if dask.config.get("dataframe.query-planning") is True or "dask_expr" in sys.modules:
     raise NotImplementedError(
         """
         NeMo Curator does not support query planning yet.
--- a/nemo_curator/__init__.py
+++ b/nemo_curator/__init__.py
@@ -16,11 +16,17 @@ import dask
 
 # Disable query planning if possible
 # https://github.com/NVIDIA/NeMo-Curator/issues/73
-if dask.config.get("dataframe.query-planning") is True:
+QUERY_PLANNING = dask.config.get("dataframe.query-planning")
+if QUERY_PLANNING is True or QUERY_PLANNING is None:
     raise NotImplementedError(
-        "NeMo Curator does not support query planning yet. "
-        "Please disable query planning before importing "
-        "`nemo_curator`, `dask.dataframe` or `dask_cudf`."
+        """
+        NeMo Curator does not support query planning yet.
+        Please disable query planning before importing
+        `dask.dataframe` or `dask_cudf`. This can be done via:
+        `export DASK_DATAFRAME__QUERY_PLANNING=False`, or
+        importing `dask.dataframe/dask_cudf` after importing
+        `nemo_curator`.
+        """
     )
 else:
     dask.config.set({"dataframe.query-planning": False})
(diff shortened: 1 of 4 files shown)


ayushdg:
cc: @rjzamora If you could take a look

nemo_curator/__init__.py
```diff
@@ -16,11 +16,17 @@
...
+        NeMo Curator does not support query planning yet.
+        Please disable query planning before importing
+        `dask.dataframe` or `dask_cudf`. This can be done via:
+        `export DASK_DATAFRAME__QUERY_PLANNING=False`, or
+        importing `dask.dataframe/dask_cudf` after importing
+        `nemo_curator`.
+        """
     )
 else:
     dask.config.set({"dataframe.query-planning": False})
```
rjzamora, line 32: It doesn't seem like there is any point of this else statement, because the config would already need to be `False`.

rjzamora (commented):
Is the goal here to raise an error in any case that the user does not **explicitly** opt out of query-planning? This means all curator users will see this error unless they explicitly set ther config to false ahead of time.

ayushdg:
> Is the goal here to raise an error in any case that the user does not explicitly opt out of query-planning? This means all curator users will see this error unless they explicitly set ther config to false ahead of time.

The intended goal is to raise this error whenever `Query-Planning` is enabled and dask is going down that route. In newer versions of dask, I've noticed that this value is being set to `None` on importing dask.dataframe, but still going down the query planning route. 
I now realize that it could be `None` in the case where `dask.dataframe` is not imported at all. I'll think about this a bit more, but curious if you see better ways of checking whether `query planning` is truly enabled.

rjzamora:
>I now realize that it could be None in the case where dask.dataframe is not imported at all. I'll think about this a bit more, but curious if you see better ways of checking whether query planning is truly enabled.

Right - I believe you have stumbled upon the primary reason we rapids-24.04 was pinned to such an "old" version of dask: There is currently no reliable way (that I know of) to check if the user has already imported `dask.dataframe` before `dask_cudf` is imported :/

That said, it **may** be possible for us to add a canonical "switch" in `rapids-dask-dependency`.

nemo_curator/__init__.py
```diff
@@ -16,11 +16,17 @@
 
 # Disable query planning if possible
 # https://github.com/NVIDIA/NeMo-Curator/issues/73
-if dask.config.get("dataframe.query-planning") is True:
+QUERY_PLANNING = dask.config.get("dataframe.query-planning")
+if QUERY_PLANNING is True or QUERY_PLANNING is None:
```
ayushdg, line 20: Thinking a bit more about this, it might make sense to replace the `None` check for the config with something like `"dask_expr" in sys.modules`. 
We can then followup with a better approach based on the resolution of upstream dask discussions.

nemo_curator/__init__.py
```diff
@@ -12,15 +12,22 @@
...
 # limitations under the License.
 
+import sys
+
 import dask
 
 # Disable query planning if possible
 # https://github.com/NVIDIA/NeMo-Curator/issues/73
-if dask.config.get("dataframe.query-planning") is True:
+if dask.config.get("dataframe.query-planning") is True or "dask-expr" in sys.modules:
```
rjzamora, line 21: How about:

```python
query_planning_config = dask.config.get("dataframe.query-planning")
if query_planning_config is True or (
    query_planning_config is None and
    "dask.dataframe" in sys.modules and
    "dask_expr" in sys.modules
):
    # Cannot change query-planning config anymore, must raise
    raise NotImplementedError("...")
else:
    dask.config.set({"dataframe.query-planning": False})
```

This seems like the only way to check if you can still modify the query-planning config (pre https://github.com/dask/dask/pull/11183).

ayushdg: These are somewhat contrived examples, but if the following happens:
```python
import dask
from dask import dataframe as dd

dask.config.set({"dataframe.query-planning": False})
import nemo_curator
```

```python
import dask
import dask_expr

dask.config.set({"dataframe.query-planning": False})
import nemo_curator
```

Both cases would not raise an error on the Curator side even though query planning is indeed enabled. Shouldn't the only criteria to check whether query planning might be used is checking the presence of `dask_expr` in sys.modules even if query_planning_config can be false?

rjzamora: Sorry for the ridiculously reply!

>Shouldn't the only criteria to check whether query planning might be used is checking the presence of dask_expr in sys.modules

Okay - I thought about this again and agree with your preference to be as strict as possible. Although it is possible to use `dask_expr` separately from `dask.dataframe`, it seems very unlikely that a user will be doing this. Therefore, I agree that your original logic make sense.

With that said, it should become easier to check if query-planning is enabled after https://github.com/dask/dask/pull/11183 (curator can import `dask.dataframe` after setting the config, and raise an error if `dd.DASK_EXPR_ENABLED` is `True`).

rjzamora:
```suggestion
if dask.config.get("dataframe.query-planning") is True or "dask_expr" in sys.modules:
```

rjzamora (approved):

VibhuJawa (approved):
LGTM

PR #110, opened 2024-06-12
Shuffle CC result on group before writing out

ayushdg:
## Description

This PR shuffles the `connected_components` output by group to ensure that all documents belonging to the same duplicate group/component end up in the same partition. The motivation for doing this ensure the example showcasing document removal work as intended: https://github.com/NVIDIA/NeMo-Curator/blob/main/examples/fuzzy_deduplication.py#L87.

## Usage
N/A
## Checklist

### DIFF
--- a/nemo_curator/modules/fuzzy_dedup.py
+++ b/nemo_curator/modules/fuzzy_dedup.py
@@ -1296,6 +1296,8 @@ class ConnectedComponents:
 
         assert num_nodes == len(labels_df)
         print(f"assert num_nodes:{num_nodes}==labels_df:{len(labels_df)} passed")
+        # Ensure all docs in the same group are in the same partition
+        labels_df = labels_df.shuffle(on=["group"], ignore_index=True)
         labels_df.to_parquet(output_path, write_index=False)
         Comms.destroy()
 


nemo_curator/modules/fuzzy_dedup.py
```diff
@@ -1296,6 +1296,8 @@ def _run_connected_components(
 
         assert num_nodes == len(labels_df)
         print(f"assert num_nodes:{num_nodes}==labels_df:{len(labels_df)} passed")
+        # Ensure all docs in the same group are in the same partition
+        labels_df = labels_df.shuffle(on=["group"], ignore_index=True)
         labels_df.to_parquet(output_path, write_index=False)
```
VibhuJawa, line 1301: Looks good to me.  

Is the reason we dont run into this on MNMG setups that we do jaccard computations ?

VibhuJawa (approved):
LGTM, just one general clarification

nemo_curator/modules/fuzzy_dedup.py
ayushdg: MNMG setups typically go via the CLI scripts that just right out CC.parquet as is. We were using local scripts to read that back in, shuffle on group and then compute duplicate docs. The example however https://github.com/NVIDIA/NeMo-Curator/blob/main/examples/fuzzy_deduplication.py#L87 was an oversight on my part where I forgot to add the shuffle call, but on further thought I think it makes sense to include within the CC results directly rather than expecting users to shuffle before computing duplicates.

VibhuJawa: I agree, this makes more sense .  Please feel free to merge this in.

PR #122, opened 2024-06-18
Hardcode labels for domain and quality classifiers

### LINKED ISSUE #71: [IMP] Hard code labels for the quality and domain classifier

We should prevent feeding in labels to `DomainClassifier `, `Quality Classifier` at every initialization of these classifiers.  

@ryantwolf , mentioned that  the labels are fixed for each classifier right, as in they can't be reordered or altered on-demand and suggested that we should probably refactor this so that the labels are just hardcoded into the DomainClassifier and QualityClassifier so the user doesn't have to manually enter in the labels each time. It would then be good to also add in the docstring of the constructor for each class what the possible options for the filter_by parameter would be.

See link: https://github.com/NVIDIA/NeMo-Curator/pull/58#discussion_r1605337411

sarahyurick:
Closes #71 

Copy of https://github.com/NVIDIA/NeMo-Curator/pull/95 which I had to close due to GitHub issues.

cc @VibhuJawa this should be ready for review.


sarahyurick:
Closing in favor of the work for https://github.com/NVIDIA/NeMo-Curator/issues/72.

We should be able to get labels from HF directly (https://huggingface.co/nvidia/domain-classifier) as they are present in `config.id2label`.

PR #128, opened 2024-06-24
[REVIEW] Add Resiliparse option for text extraction

sarahyurick:
Duplicate of https://github.com/NVIDIA/NeMo-Curator/pull/90 with successful DCO check.

Right now, we only support Common Crawl text extraction with jusText. Resiliparse is known to be a faster text extraction algorithm which may also produce better tokens.

This PR adds optional support for the Resiliparse algorithm while still keeping jusText as the default.


nemo_curator/download/commoncrawl.py
```diff
@@ -280,6 +354,7 @@ def download_common_crawl(
     start_snapshot: str,
     end_snapshot: str,
     output_type: str = "jsonl",
+    algorithm: str = "justext",
```
sarahyurick, line 357: We will keep jusText as the default for this PR. If our research finds it beneficial to switch to Resiliparse, we can update this line to set Resiliparse as the default.

nemo_curator/download/commoncrawl.py
```diff
@@ -288,7 +363,7 @@ def download_common_crawl(
     url_limit=None,
 ) -> DocumentDataset:
```
sarahyurick, line 364: From my understanding, the `download_common_crawl` function is what the user calls in their Python script. However, it doesn't look like they can set any of the `extract_text` parameters themselves (like `length_low`, `length_high`, etc.). Should we update this so that the user has more control over these parameters?

sarahyurick (commented):
Hi @ryantwolf @jojennin sorry for the confusion. I'm reopening https://github.com/NVIDIA/NeMo-Curator/pull/90 to fix the commit signoff issues I was dealing with; the PRs are identical otherwise. Thanks!

nemo_curator/download/commoncrawl.py
```diff
@@ -71,55 +73,126 @@ def lang_detect(decoded_html):
...
-        stop_words,
-        length_low,
-        length_high,
-        stopwords_low,
-        stopwords_high,
-        max_link_density,
-        no_headings,
-    )
+    if algorithm.lower() == "justext":
+        if length_low is None:
```
ryantwolf, line 87: I don't want to over engineer this simple task, but given how many arguments we want to expose I think it's worth refactoring this into some classes even if they're a bit heavier.

I imagine something like this could work

```python
class HTMLExtractionAlgorithm(ABC):
  @abstractmethod
  def extract_text(html, stop_words):
    pass

class JusTextExtraction(HTMLExtractionAlgorithm):
  def __init__(self, length_low=70, ...):
     self.length_low = length_low
     # Etc...

  def extract_text(html, stop_words):
     # Segment the HTML into paragraphs
    try:
        # Form the DOM tree
        dom = justext.core.html_to_dom(html)
    # Etc...
    except:
      # Etc...

class Resiliparse(HTMLExtractionAlgorithm):
  # Etc...
```

Then, when the user calls `download_common_crawl` they could supply something like this:
```
my_algorithm = Resiliparse(...)
download_common_crawl('/path/to/output', '2020-01', '2020-03', algorithm=my_algorithm)
```
Keeping the default as the default `JusTextExtraction()`

My main reasons for thinking this is preferred are:
1. This gets rid of the weird `None` but still default arguments.
    1. The defaults will be easily visible in each class' constructor.
    1. No warnings are needed.
1. It makes clear which arguments are necessary for which algorithm.
    1. It also decouples the defaults. I see that `stopwords_high` is used for both parsers, but it's not immediately obvious that both algorithms should share the same default to me.
1. It keeps each algorithm's logic isolated from each other.

I'd like to hear your opinion on this too if you think there's a better way to handle this.

nemo_curator/download/commoncrawl.py
```diff
@@ -71,55 +73,126 @@ def lang_detect(decoded_html):
...
+                "jusText-specific parameter 'max_heading_distance' is set but will be ignored when using Resiliparse.",
+                UserWarning,
+            )
+        if no_headings is not None:
+            warnings.warn(
+                "jusText-specific parameter 'no_headings' is set but will be ignored when using Resiliparse.",
+                UserWarning,
+            )
+
+        text = extract_plain_text(html, main_content=True, alt_texts=False)
```
ryantwolf, line 172: Probably best to pass these arguments and the remaining ones in the function on to the user.

setup.py
```diff
@@ -46,6 +46,7 @@
         "fasttext==0.9.2",
         "pycld2==0.41",
         "justext==3.0.0",
+        "resiliparse",
```
ryantwolf, line 49: Do you have any concerns with not pinning the version?

ryantwolf (commented):
Overall looks great! I just have some design comments on how I think we can make it easier to swap between algorithms and customize them.

nemo_curator/download/commoncrawl.py
```diff
@@ -71,55 +73,126 @@ def lang_detect(decoded_html):
...
+
+        paragraphs = list(filter(None, text.split("\n")))
```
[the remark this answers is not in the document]
sarahyurick: After Resiliparse does its initial text extraction, it's just a raw string but we need a list of the paragraphs in the text to iterate through (see the for loop below this). Resiliparse's output has a bunch of new lines and spacing stuff, so this line cleans that up and then creates a list of non-empty paragraphs.

setup.py
sarahyurick: Not really. I can pin the most recent version that exists right now, but I don't anticipate any issues since we are only using a very basic Resiliparse function `extract_plain_text` one time. Let me know.

ryantwolf: Sounds good. I'm fine leaving it like this. It's probably worth adding a few small unit tests to ensure that parsing a basic HTML page works properly with our system though just so we can catch if a version change breaks things.

nemo_curator/download/commoncrawl.py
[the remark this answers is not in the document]
ryantwolf: Ah I forgot about the empty paragraphs, that makes sense. Good catch!

nemo_curator/download/commoncrawl.py
sarahyurick: I think the classes are a good idea. My only comment is that I think for whichever one we end up setting as the default, the user shouldn't have to touch any of those parameters (including having to set the algorithm itself). I can still create the class logic and then allow `algorithm` to default to `JusTextExtraction()`.

So while
```
my_algorithm = ResiliparseExtraction(...)
download_common_crawl('/path/to/output', '2020-01', '2020-03', algorithm=my_algorithm)
```
is an option, we should have the default as:
```
download_common_crawl('/path/to/output', '2020-01', '2020-03')  # defaults with JusTextExtraction()
```

ryantwolf: Totally agree

sarahyurick:
Hi @ryantwolf this is ready for another review.

The only question I have is what you think the best way to go about adding the unit tests is? Locally I'm testing it with [download_common_crawl.py](https://github.com/NVIDIA/NeMo-Curator/blob/main/examples/download_common_crawl.py) but that seems a bit heavy for CI? Even doing a single snapshot with `url_limit = 10` takes at least a couple minutes each to run.

Edit: Perhaps it may be sufficient to add examples of `JusTextExtraction()` and `ResiliparseExtraction()` to [download_common_crawl.py](https://github.com/NVIDIA/NeMo-Curator/blob/main/examples/download_common_crawl.py)?

ryantwolf:
> The only question I have is what you think the best way to go about adding the unit tests is?

Good question. I wouldn't add a unit test for the `download_and_extract`. I would just target each algorithm's `extract_text ` method. You could draft like one or two simple html pages and make a unit test using each algorithm. Just something to make sure the behavior stays consistent.

> Perhaps it may be sufficient to add examples of JusTextExtraction() and ResiliparseExtraction() to [download_common_crawl.py](https://github.com/NVIDIA/NeMo-Curator/blob/main/examples/download_common_crawl.py)?

It isn't a bad idea to showcase how users can change the algorithm. Do you mind updating the `download.rst` docs instead? I'm thinking right after we explain what `output_type="jsonl"` does you could add a snippet about the algorithm parameter. Or, you could add another small code block like

```python
from nemo_curator.download import (
    download_common_crawl,
    ResiliparseExtraction,
)

# Change the extraction algorithm
extraction_algorithm = ResiliparseExtraction()
common_crawl = download_common_crawl(
    "/extracted/output/folder", 
    "2020-50",
    "2021-04",
    output_type="jsonl",
    algorithm=extraction_algorithm,
)
```

sarahyurick:
Thanks @ryantwolf ! Should be ready now.

ryantwolf (approved):
Great, glad to have this in. Thanks again!

PR #154, opened 2024-07-18
Prevent plugging an allocator twice

VibhuJawa:
## Description
This PR prevent plugging an allocator twice

### DIFF
--- a/nemo_curator/utils/script_utils.py
+++ b/nemo_curator/utils/script_utils.py
@@ -281,6 +281,22 @@ class ArgumentHelper:
             help="The block size for chunking jsonl files for text ddf in mb",
         )
 
+    def add_arg_model_path(self, help="The path to the model file"):
+        self.parser.add_argument(
+            "--model-path",
+            type=str,
+            help=help,
+            required=True,
+        )
+
+    def add_arg_autocaset(self, help="Whether to use autocast or not"):
+        ArgumentHelper.attach_bool_arg(
+            parser=self.parser,
+            flag_name="autocast",
+            default=True,
+            help=help,
+        )
+
     def add_distributed_args(self) -> argparse.ArgumentParser:
         """
         Adds default set of arguments that are needed for Dask cluster setup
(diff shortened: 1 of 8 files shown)


nemo_curator/utils/script_utils.py
```diff
@@ -472,6 +451,7 @@ def parse_gpu_dedup_args(description: str) -> argparse.ArgumentParser:
 
         # Set default device to GPU for dedup
         argumentHelper.parser.set_defaults(device="gpu")
+        argumentHelper.parser.set_defaults(set_torch_to_use_rmm=False)
```
VibhuJawa, line 454: @ayushdg ,  This is what you wanted me switch off right ?

nemo_curator/utils/script_utils.py
```diff
@@ -392,67 +408,30 @@ def parse_distributed_classifier_args(
...
+        argumentHelper.add_distributed_args()
+        argumentHelper.add_arg_input_data_dir(required=True)
+        argumentHelper.add_arg_output_data_dir(help="The path of the output files")
+        argumentHelper.add_arg_input_file_type()
+        argumentHelper.add_arg_input_file_extension()
+        argumentHelper.add_arg_output_file_type()
+        argumentHelper.add_arg_input_text_field()
+        argumentHelper.add_arg_enable_spilling()
+        argumentHelper.add_arg_set_torch_to_use_rmm()
+        argumentHelper.add_arg_batch_size(
```
VibhuJawa, line 421: Decided to clean this up too because i was in the same fine, @ryantwolf , please take a look.

ryantwolf (commented):
All the changes you made look good on my end. There are a couple of dedup scripts that manually set
```python
args.set_torch_to_use_rmm = False
```

Could you go through and remove that line since you change the default?

VibhuJawa:
> Could you go through and remove that line since you change the default?

Done (Also made it false) in the function call to be safe.

nemo_curator/utils/script_utils.py
ayushdg: That's right!

ayushdg (approved):
Thanks for the changes!

PR #163, opened 2024-07-23
[Tutorials] Reduce the cluster size for the SGD + PEFT notebook

Maghoumi:
## Description
Reduce the ratio of synthetic samples to lower API calls.
Reduce the cluster size for Sem Dedupe.

### DIFF
--- a/tutorials/peft-curation-with-sdg/config/sem_dedup_config.yaml
+++ b/tutorials/peft-curation-with-sdg/config/sem_dedup_config.yaml
@@ -13,7 +13,7 @@ embedding_max_mem_gb: 20
 
 # Clustering configuration
 clustering_save_loc: "clustering_results"
-n_clusters: 1000
+n_clusters: 20
 seed: 1234
 max_iter: 100
 kmeans_with_cos_dist: false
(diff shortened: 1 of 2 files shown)


ryantwolf:
@VibhuJawa @avem-nv continuing the conversation from [earlier](https://github.com/NVIDIA/NeMo-Curator/pull/157#discussion_r1687215245), what do you think of this? As far as I understand the algorithm, we should set the number of clusters equal to 1 for this small of a dataset (12,224 samples). Let me know what y'all think.

VibhuJawa:
> @VibhuJawa @avem-nv continuing the conversation from [earlier](https://github.com/NVIDIA/NeMo-Curator/pull/157#discussion_r1687215245), what do you think of this? As far as I understand the algorithm, we should set the number of clusters equal to 1 for this small of a dataset (12,224 samples). Let me know what y'all think.

I think we can reduce it to lesser number of clusters for sure. 

I think the only thing to keep in mind that we are doing 
 O(n**2) computations for similarity matrices IIRC so maybe just move it to 10 to keep it reasonable  for folks  who might experiment with more data.  1 works too, no strong opinion.

VibhuJawa (approved):
LGTM. I think `1`, `10` and `20` are all reasonable estimates. Will let @avem-nv  confirm though.

PR #168, opened 2024-07-24
Add fineweb classifier

VibhuJawa:
## Description
This PR adds fineweb clasifier. 

## Usage

```
python3 fineweb_edu_classifier_example.py --input-data-dir input_data --output-data-dir "test_classifier"
```

## Checklist


VibhuJawa:
Merge after we switch this repo to `0.0.3` https://github.com/rapidsai/crossfit/releases/tag/0.0.3  , PIP process is still ongoing.

examples/fineweb_edu_classifier_example.py
```diff
@@ -0,0 +1,140 @@
...
+from transformers import AutoConfig, AutoModelForSequenceClassification
+
+from nemo_curator.datasets import DocumentDataset
+from nemo_curator.utils.distributed_utils import get_client
+from nemo_curator.utils.script_utils import ArgumentHelper
+
+EDU_CLASSIFIER_PATH = "HuggingFaceTB/fineweb-edu-classifier"
+
+
+class FinewebEduClassifier(HFModel):
```
sarahyurick, line 32: Do we want to integrate this into https://github.com/NVIDIA/NeMo-Curator/blob/main/nemo_curator/modules/distributed_data_classifier.py?

sarahyurick (commented):
Runs fine for me. I was just wondering why we are adding this to Curator? Also, will this be added to the docs somewhere?

examples/fineweb_edu_classifier_example.py
VibhuJawa: Not right now, This just serves as an example on how to use it. Unsure if we want to combine it there ? 

@arhamm1 , Any opinions ?

VibhuJawa:
> Runs fine for me. I was just wondering why we are adding this to Curator? Also, will this be added to the docs somewhere?

@sarahyurick . The scope is to provide an example of using a model which we dont release but maintain

nemo_curator/utils/script_utils.py
```diff
@@ -289,6 +289,14 @@ def add_arg_model_path(self, help="The path to the model file"):
             required=False,
         )
 
+    def add_arg_max_mem_gb_classifier(self):
+        self.parser.add_argument(
+            "--max-mem-gb-classifier",
+            default=16,
+            type=int,
+            help="Maximum memory to use in GB",
```
ryantwolf, line 297: Nit: specify GPU memory.

examples/fineweb_edu_classifier_example.py
```diff
@@ -0,0 +1,140 @@
...
+    def __init__(self, path_or_name, max_mem_gb="16", autocast=False):
```
ryantwolf, line 33: The argparse has the type of `max_mem_gb` as an integer, while this has it as a string. If both are fine in crossfit, that's great, but I think we should be consistent in which one we use for Curator. I vote for integer, but if string is better for some reason go for that instead.

examples/fineweb_edu_classifier_example.py
```diff
@@ -0,0 +1,140 @@
...
+        return model
+
+    @staticmethod
+    def configure_forward(model, autocast=True):
+        original_forward = model.forward
+
+        def custom_forward(*args, **kwargs):
+            if autocast:
+                with torch.cuda.amp.autocast():
+                    output = original_forward(*args, **kwargs)
```
ryantwolf, line 51: Why are we using this method of autocast? It's deprecated according to [this](https://pytorch.org/docs/stable/amp.html) and we already use `torch.autocast` in the [distributed data classifiers](https://github.com/NVIDIA/NeMo-Curator/blob/main/nemo_curator/modules/distributed_data_classifier.py#L104).

examples/fineweb_edu_classifier_example.py
```diff
@@ -0,0 +1,140 @@
...
+            return output.logits.squeeze(-1).float()
+
+        model.forward = custom_forward
+        return model
+
+    def load_config(self):
+        return AutoConfig.from_pretrained(self.path_or_name)
+
+
+def run_pipeline(
```
ryantwolf, line 61: For the sake of demonstration, how do you feel about wrapping this pipeline (and the class above) in a subclass of `DistributedDataClassifier`?

examples/fineweb_edu_classifier_example.py
ryantwolf: Yeah this is a tricky one. On one hand I think having an example of how people can extend our code to their own model is important. On the other, this model is super relevant and probably should be included in NeMo Curator. Perhaps we do both?

ryantwolf (changes requested):
Looks good overall, thanks for adding this. Just some minor nits and potential reformatting.

examples/fineweb_edu_classifier_example.py
arhamm1: We are good to have an example right now, based on interest we can also get it integrated

examples/fineweb_edu_classifier_example.py
```diff
@@ -0,0 +1,140 @@
+# Copyright (c) 2024, NVIDIA CORPORATION.  All rights reserved.
```
sarahyurick, line 1: Should move this into `examples/classifiers` since https://github.com/NVIDIA/NeMo-Curator/pull/172 was merged.

sarahyurick (dismissed):

examples/fineweb_edu_classifier_example.py
VibhuJawa: Yes, we should.

VibhuJawa: Resolved

examples/fineweb_edu_classifier_example.py
VibhuJawa: I changed my mind based on our latest meetings, I think this should be a full blown member of the classifier models we have. Reformatted code accordingly.

examples/fineweb_edu_classifier_example.py
VibhuJawa: Fixed

examples/fineweb_edu_classifier_example.py
VibhuJawa: Fixed

nemo_curator/utils/script_utils.py
VibhuJawa: Fixed

nemo_curator/classifiers/aegis.py
```diff
@@ -199,7 +199,7 @@ def __init__(
         self.keep_raw_pred = keep_raw_pred
 
         try:
-            model = AegisHFModel(config=config)
+            model = AegisHFModel(config=config, max_mem_gb=48)
```
ryantwolf, line 202: Any particular reason you are setting this again here? If not, I'd prefer to keep it at default. If we need to change it later, I can see finding this hardcoded value a bit tricky potentially.

nemo_curator/utils/distributed_utils.py
```diff
@@ -633,3 +636,16 @@ def get_network_interfaces() -> List[str]:
...
+def get_gpu_memory_info() -> List[dict]:
+    """
+    Gets the total and available memory for each GPU on the machine
+
+    Returns:
+        A list of dictionaries containing the total and available memory for each GPU on the machine
+    """
+    client = get_current_client()
+    if client is None:
+        return {}
```
ryantwolf, line 650: Nit: Return an empty list instead of an empty dict (assuming the return type is accurate)

nemo_curator/classifiers/domain.py
```diff
@@ -36,10 +37,15 @@ class DomainModelConfig:
...
 class DomainModel(HFModel):
-    def __init__(self, config: DomainModelConfig, autocast: bool = False):
+    def __init__(
+        self, config: DomainModelConfig, autocast: bool = False, max_mem_gb=None
+    ):
         self.config = config
         self.autocast = autocast
-        super().__init__(self.config.model)
+        if max_mem_gb is None:
+            max_mem_gb = _get_suggest_memory_for_classifier()
```
ryantwolf, line 46: I like this memory method. Can you add a docstring to the `DomainClassifer` and `QualityClassifier` and explain what the max memory does and how it will be calculated if it is set to `None`?

nemo_curator/classifiers/fineweb_edu.py
```diff
@@ -0,0 +1,122 @@
...
+    _get_suggest_memory_for_classifier,
+    _run_classifier_helper,
+)
+from nemo_curator.datasets import DocumentDataset
+
+FINEWEB_EDU_IDENTIFIER = "HuggingFaceTB/fineweb-edu-classifier"
+
+
+class FinewebEduModel(HFModel):
+    def __init__(self, path_or_name, max_mem_gb, autocast=False):
```
ryantwolf, line 34: Nit: have `max_mem_gb` be defaulted to `None` to match the convention of the other classifiers.

nemo_curator/classifiers/fineweb_edu.py
```diff
@@ -0,0 +1,122 @@
...
+
+class FineWebEduClassifier(DistributedDataClassifier):
+    def __init__(
+        self,
+        filter_by=None,
+        batch_size=256,
+        text_field: str = "text",
+        pred_column="fineweb-edu-score",
+        int_column="fineweb-edu-score-int",
+        max_chars=-1,
```
ryantwolf, line 72: Can we do this now? I imagine this causes the truncation to happen during the tokenization step instead of the character level. If so, should we change the other classifiers?

nemo_curator/classifiers/fineweb_edu.py
```diff
@@ -0,0 +1,122 @@
...
+        return AutoConfig.from_pretrained(self.path_or_name)
+
+
+class FineWebEduClassifier(DistributedDataClassifier):
+    def __init__(
+        self,
+        filter_by=None,
+        batch_size=256,
+        text_field: str = "text",
+        pred_column="fineweb-edu-score",
```
ryantwolf, line 70: The way I had `AegisClassifier` was that `pred_column` was the final processed output and `raw_pred_column` was an intermediate result. Is `"fineweb-edu-score"` meant to be the final processed output and `"fineweb-edu-score-int"` a more raw form? If so, I think it would be good to follow the pattern in Aegis.

nemo_curator/utils/script_utils.py
```diff
@@ -289,6 +289,15 @@ def add_arg_model_path(self, help="The path to the model file"):
             required=False,
         )
 
+    def add_arg_max_mem_gb_classifier(self):
+        self.parser.add_argument(
+            "--max-mem-gb-classifier",
+            default=None,
+            type=int,
+            help="Maximum GPU memory to use in GB for the classifier",
+            default="Uses the Ttotal GPU memory - 4GB, for the classifier",
```
ryantwolf, line 298: Nit: "Ttotal" -> "total"

ryantwolf (changes requested):
Overall looks good, just a few minor concerns.

nemo_curator/classifiers/aegis.py
VibhuJawa: Did not want to mess with default. Ideally i would want `Aegis` to fall back to `max memory_db` default too (which you can control) via a flag than hard coding it , what do you think ?

nemo_curator/classifiers/domain.py
VibhuJawa: Cool, will do . Let me add it to the docstring.

nemo_curator/classifiers/fineweb_edu.py
VibhuJawa: So it is a decimal score which they round.  I dont want it to say raw, was trying to follow the classifier. Dont think `raw` tells the right story.  So i went with `score` and `score-int`, can change it to say `fineweb-edu-int-score` if thats better. 

```
result = {
    "text": text,
    "score": score,
    "int_score": int(round(max(0, min(score, 5)))),
}
```

See below: https://huggingface.co/HuggingFaceFW/fineweb-edu-classifier

nemo_curator/classifiers/fineweb_edu.py
VibhuJawa: We should always have been able to do this. Did this fail earlier ? 

I think we added this to other classifiers because Bo had a pipeline setup like that, i might be completely wrong though. Maybe @sarahyurick  knows. 

We always skip it and that should work too. 

https://github.com/rapidsai/crossfit/blob/9e52a904e101e4d2ac86a89f4ab7d7926470a8be/crossfit/op/tokenize.py#L117-L118

nemo_curator/utils/script_utils.py
VibhuJawa: Resolved by https://github.com/NVIDIA/NeMo-Curator/pull/168/commits/e889f51a9f8f1fbe77f9ce256cb378c0f3083e32

nemo_curator/classifiers/fineweb_edu.py
VibhuJawa: Resolved by: https://github.com/NVIDIA/NeMo-Curator/pull/168/commits/e889f51a9f8f1fbe77f9ce256cb378c0f3083e32

nemo_curator/classifiers/domain.py
VibhuJawa: Resolved by adding doc-string to https://github.com/NVIDIA/NeMo-Curator/pull/168/commits/e889f51a9f8f1fbe77f9ce256cb378c0f3083e32

nemo_curator/utils/distributed_utils.py
VibhuJawa: Fixed this : https://github.com/NVIDIA/NeMo-Curator/pull/168/commits/e889f51a9f8f1fbe77f9ce256cb378c0f3083e32

nemo_curator/classifiers/aegis.py
VibhuJawa: Waiting on your followup here. Have resolved everything else.

nemo_curator/classifiers/fineweb_edu.py
VibhuJawa: Have added docstring, hopefully that makes it easier to understand.

VibhuJawa:
@ryantwolf , Thanks for the careful review. Made relevant changes and followed up with questions. Let me know what else to change.

nemo_curator/classifiers/aegis.py
ryantwolf: Oh if you mean introducing a variable in the constructor for `AegisClassifier` like the other classifiers, then yeah I agree we should do that.

nemo_curator/classifiers/fineweb_edu.py
ryantwolf: No it didn't fail, I just didn't realize this was an option.

nemo_curator/classifiers/fineweb_edu.py
ryantwolf: Oh gotcha, so it just outputs 2 scores. Makes sense.

ryantwolf (commented):
Ok looks good. Just need to figure out the AEGIS memory thing.

VibhuJawa:
> Ok looks good. Just need to figure out the AEGIS memory changes 

@ryantwolf , Pushed the memory changes  https://github.com/NVIDIA/NeMo-Curator/pull/168/commits/18d93da0d6193796339d1b89b457c88766d64a7d


Please take a another look

nemo_curator/classifiers/fineweb_edu.py
VibhuJawa: Yeah, lets do a followup. Filing an issue track.  

https://github.com/NVIDIA/NeMo-Curator/issues/213

nemo_curator/classifiers/aegis.py
VibhuJawa: Resolved by : https://github.com/NVIDIA/NeMo-Curator/commit/18d93da0d6193796339d1b89b457c88766d64a7d

sarahyurick (commented):
LGTM, thanks! Found a minor bug and opened a PR for it.

PR #179, opened 2024-08-05
Fix 177. Improve score_document and _line_statistics performance

miguelusque:
Hi!

I have noticed a couple of small changes in the code that can improve its performance:

By refactoring score_document as suggested, it will run between 2x and 3x faster, and will reduce by half the memory footprint.

By refactoring '_line_statistics' as suggested, it can run about 21% faster.

Hope it helps!

Regards,
Miguel

### DIFF
--- a/nemo_curator/filters/code.py
+++ b/nemo_curator/filters/code.py
@@ -13,9 +13,9 @@
 # limitations under the License.
 
 import csv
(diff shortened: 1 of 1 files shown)


ryantwolf (approved):
Great catch, thank you for improving this!

PR #184, opened 2024-08-05
Speed up _line_statistics

miguelusque:
Speeding up this method, while removing statistics package

### DIFF
--- a/tests/test_filters.py
+++ b/tests/test_filters.py
@@ -765,7 +765,9 @@ class TestCodeFilters:
                 "code_meta.csv",
             )
         )
-        per_extension_filter = PerExtensionFilter("c++", "cpp", metadata_file=metadata_file)
+        per_extension_filter = PerExtensionFilter(
+            "c++", "cpp", metadata_file=metadata_file
+        )
         filters = ScoreFilter(per_extension_filter)
         filtered_data = filters(dataset)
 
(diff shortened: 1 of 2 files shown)


ryantwolf:
@miguelusque thanks for doing this. Do you mind adding some more unit tests for this filter? Given that you are changing the underlying way it calculates the mean, I want to ensure that the behavior is the same.

ryantwolf (changes requested):
@miguelusque thanks for doing this. Do you mind adding some more unit tests for this filter? Given that you are changing the underlying way it calculates the mean, I want to ensure that the behavior is the same.

miguelusque:
> @miguelusque thanks for doing this. Do you mind adding some more unit tests for this filter? Given that you are changing the underlying way it calculates the mean, I want to ensure that the behavior is the same.

Sure thing! Will do it by tomorrow. Thanks!

miguelusque:
Hi @ryantwolf ,

Just added some tests. Thanks!

tests/test_filters.py
ryantwolf, line 1: Looks like the style check is failing on this file.

tests/test_filters.py
```diff
@@ -776,6 +775,16 @@ def test_per_extension_filter(self):
             expected_data, filtered_data
         ), f"Expected {expected_data} but got {filtered_data}"
 
+        tests = [["", (0, 0.0)],
```
ryantwolf, line 778: Can you use `@pytest.mark.parametrize` for these tests instead of writing the for loop? You can take a look at the fuzzy dedup tests [here](https://github.com/NVIDIA/NeMo-Curator/blob/main/tests/test_fuzzy_dedup.py#L118) for pointers, it's not too big of a change.

ryantwolf (changes requested):
Thanks for adding them! Just a few changes and we should be good.

tests/test_filters.py
miguelusque: Done!

ryantwolf (approved):
Thank you!

PR #190, opened 2024-08-07
HuggingFace support for Quality Classifier

sarahyurick:
Closes https://github.com/NVIDIA/NeMo-Curator/issues/167

### DIFF
--- a/nemo_curator/classifiers/base.py
+++ b/nemo_curator/classifiers/base.py
@@ -12,12 +12,17 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 import os
+from dataclasses import dataclass
 
 os.environ["RAPIDS_NO_INITIALIZE"] = "1"
 from abc import ABC, abstractmethod
 from typing import List
 
+import torch
+import torch.nn as nn
+from huggingface_hub import PyTorchModelHubMixin
 from crossfit import op
+from transformers import AutoModel
 
 from nemo_curator.datasets import DocumentDataset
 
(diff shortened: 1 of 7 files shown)


nemo_curator/classifiers/quality.py
```diff
@@ -12,22 +12,21 @@
...
+    HFCustomModel,
     _run_classifier_helper,
 )
 from nemo_curator.datasets import DocumentDataset
 
+QUALITY_IDENTIFIER = "nvidia/quality-classifier-deberta"
+
 
 @dataclass
 class QualityModelConfig:
```
sarahyurick, line 32: There's a lot of repeated/identical logic for `DomainModelConfig` vs `QualityModelConfig`, `DomainModel` vs `QualityModel`, and `DomainClassifier` vs `QualityClassifier`. I was wondering how much of this I should try to consolidate versus keeping them separate?

sarahyurick:
Hi @VibhuJawa @ryantwolf this is ready for review.

nemo_curator/classifiers/base.py
```diff
@@ -78,6 +83,32 @@ def get_labels(self) -> List[str]:
         return self.labels
 
 
+class HFCustomModel(nn.Module, PyTorchModelHubMixin):
```
ryantwolf, line 86: Can you rename this to be something like `HFDeberta`?

nemo_curator/classifiers/quality.py
```diff
@@ -36,108 +35,28 @@ class QualityModelConfig:
...
-        if self.autocast:
-            with torch.autocast(device_type="cuda"):
-                return self._forward(batch)
-        else:
-            return self._forward(batch)
-
-
 class QualityModel(HFModel):
-    def __init__(self, config, out_dim=None, model_path=None, autocast=False):
+    def __init__(self, config: dataclass, autocast: bool = False):
```
ryantwolf, line 39: Shoot I meant to do this in my PR, but can you change the type annotation here from `dataclass` to `QualityModelConfig`? Same with the domain classifier.

ryantwolf (commented):
Thanks for the quick turnaround on this! I just have two small nits.

nemo_curator/classifiers/quality.py
ryantwolf: I think you've done a good job so far. We'll keep it with what you have for now.

nemo_curator/classifiers/quality.py
```diff
@@ -36,108 +35,28 @@ class QualityModelConfig:
     max_len = 512
```
ryantwolf, line 35: This should be updated too. This model has `max_len = 1024`

sarahyurick:
@ryantwolf updated, thanks!

ryantwolf (approved):
Thanks Sarah!

nemo_curator/classifiers/domain.py
```diff
@@ -37,40 +35,14 @@ class DomainModelConfig:
     max_len = 512
```
VibhuJawa, line 35: Minor non blocking NIT

@ryantwolf / @sarahyurick , Are we sure the `max_len` for domain is `512` and `quality` is `1024`. Might be worth double checking this.

VibhuJawa (approved):
LGTM, thanks for working on this Sarah. 

Have added a non blocking nit, which i dont think is related to your PR

nemo_curator/classifiers/domain.py
ryantwolf: Double checked, this is correct.

PR #193, opened 2024-08-08
Add support for parallel data curation

shuoyangd:
## Description
This PR adds support for parallel data curation. Namely:

- A new dataset class `ParallelDataset` that supports loading and writing parallel data in simple bitext format.
- A new `ScoreFilter` subclass `ParallelScoreFilter` that allows application of existing monolingual filters on parallel data while maintaining the alignment of sentence/document pairs.
- A new `ScoreFilter` subclass `JointScoreFilter` that allows implementation of filters that takes both fields of the parallel sentence/document pairs.
- New heuristic filters: `HistogramFilter` and `LengthRatioFilter`.
- Adding model-based filters with quality estimation models: `QualityEstimationFilter`.
- Support for two families of quality estimation models: [`comet`](https://huggingface.co/Unbabel/wmt20-comet-qe-da) and [`cometoid`](https://huggingface.co/collections/marian-nmt/cometoid-wmt23-metrics-66903bb137eadb9c5768d5f2).
- A tutorial for parallel data curation.
- Tests accompanying new features.

Joint work at [MTMA 2024](https://mtma24.github.io/) with @nverma1.

## Usage
See `tutorials/bitext_cleaning/main.py`.

## Checklist


nemo_curator/modules/filter.py
```diff
@@ -204,3 +206,161 @@ def __call__(self, dataset: DocumentDataset) -> DocumentDataset:
...
+            )
+        else:
+            bool_mask = scores.apply(self.filter_obj.keep_document, meta=(None, bool))
+        if self.invert:
+            bool_mask = ~bool_mask
+
+        return ParallelDataset(dataset.df[bool_mask])
+
+
+class ParallelScoreFilter:
```
ryantwolf, line 287: Correct me if I'm wrong, but is this the same as applying two `ScoreFilter`s one after the other and just changing the field they apply to?

nemo_curator/filters/heuristic_filter.py
```diff
@@ -633,3 +638,119 @@ def score_document(self, text):
...
+
+    def __init__(self, max_ratio=3, src_lang="en", tgt_lang="en"):
+        super().__init__()
+        self._max_ratio = max_ratio
+        self._src_word_splitter = get_word_splitter(src_lang)
+        self._tgt_word_splitter = get_word_splitter(tgt_lang)
+        self._name = "length_ratio"
+
+    def score_document(self, bitext_tuple):
+        src_len = len(self._src_word_splitter(bitext_tuple.iloc[0].strip()))
```
ryantwolf, line 657: I don't like the `.iloc[0]` pattern. It seems fragile and error prone, but I see why you did it. It's the result of a poor design inherited from the `ScoreFilter` with `DocumentFilter` paradigm. I'm going to suggest you abandon the previous design for something better. I'll explain here but let me know if I need to clarify anything.

Instead of having two separate classes like `JointScoreFilter` that accepts a `DocumentFilter`, have a single base class called `BitextFilter` (or whatever name you prefer) that your other filters inherit from. The `JointFilter` definition would look something like this:

```python
class BitextFilter:
  def __init__():
    # Roughly same init as the JointScoreFilter

  def __call__():
    # Same call as the JointScoreFilter

  @abstractmethod
  def score_document():
    pass

  @abstractmethod
  def keep_document():
    pass

```

Then, all the filters like this inherit from `BitextFilter` directly. This also gives them access to the `src_field` and `tgt_field` as member variables. I know it's a bit of a change, but I'm trying to learn from the headaches I've run into with our design and make it easier for you and your users. I plan on porting the old filters to a similar design too, I just haven't gotten around to it.

nemo_curator/utils/file_utils.py
```diff
@@ -23,8 +23,6 @@
 import pandas as pd
 from dask import delayed
 
-from nemo_curator.utils.distributed_utils import single_partition_write_with_filename
```
ryantwolf, line 26: Undo this change, this function is still being used here.

tests/test_read_simple_bitext.py
```diff
@@ -0,0 +1,63 @@
...
+        ds = ParallelDataset.read_simple_bitext(
+            src_input_files=[src_file],
+            tgt_input_files=[tgt_file],
+            src_lang="de",
+            tgt_lang="en",
+            backend="pandas",
+        )
+
+        for idx, (src_line, tgt_line) in enumerate(zip(open(src_file), open(tgt_file))):
+            assert ds.df["src"].compute()[idx] == src_line.rstrip("\n")
```
ryantwolf, line 44: Nit: instead of calling compute multiple times, call it once and then do all the assert statements.

tutorials/bitext_cleaning/README.md
```diff
@@ -0,0 +1,21 @@
...
+
+The dataset contains a small training set of ~150k sentences, and ~2k dev and test sets. This tutorial only downloads and filters the training set.
+
+## Walkthrough
+
+This tutorial highlights several bitext-specific functionalities within NeMo Curator's API, including:
+1. The ParallelDataset Class
+2. Length Ratio Filtering via the LengthRatioFilter and the JointScoreFilter classes
+3. Histogram-based Language ID Filtering via the HistogramFilter and the ParallelScoreFilter class.
+4. (Optional and only on GPU) Model-based filtering using quality estimation model (`Unbabel/wmt20-comet-qe-da`) and the JointScoreFiletr class.
```
ryantwolf, line 13: Nit: "JointScoreFiletr" -> "JointScoreFilter"

tutorials/bitext_cleaning/main.py
```diff
@@ -0,0 +1,132 @@
...
+
+    dataset = curation_steps(bitext_dataset)
+    print("Executing the pipeline...")
+    dataset = dataset.persist()
+
+    print(f"Original dataset length: {len(bitext_dataset.df)}")
+    print(f"After dataprep: {len(dataset.df)}")
+    print("Writing the results to disk...")
+
+    # raise NotImplementedError("writing not finished yet, filters not checked")
```
ryantwolf, line 108: Remove this commented out code.

tutorials/bitext_cleaning/README.md
ryantwolf, line 1: Ahhh I love this! Thanks for adding a tutorial!

ryantwolf (changes requested):
Such good work! I have a few additional requests around documentation. Can you create one or more documentation pages about curating parallel datasets in our `docs/user-guide/`? You can see how it currently is rendered [here](https://docs.nvidia.com/nemo-framework/user-guide/latest/datacuration/index.html). Also, please add all the classes and functions you expect a user to use in our [API reference](https://github.com/NVIDIA/NeMo-Curator/tree/main/docs/user-guide/api)

nemo_curator/datasets/doc_dataset.py
```diff
@@ -252,3 +256,67 @@ def _read_json_or_parquet(
         raise TypeError("File input must be a string or list.")
 
     return raw_data
+
+
+class ParallelDataset(DocumentDataset):
```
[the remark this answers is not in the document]
shuoyangd: Yes. We'd still want the flexibility of being able to load in json or parquet format. I'm already aware of some downstream use cases that will depend on json.

nemo_curator/utils/distributed_utils.py
```diff
@@ -295,6 +298,91 @@ def read_pandas_pickle(file, add_filename=False) -> pd.DataFrame:
...
+
+    df_src = df.read_table(src_input_file, names=["src"], quoting=csv.QUOTE_NONE)
+    df_tgt = df.read_table(tgt_input_file, names=["tgt"], quoting=csv.QUOTE_NONE)
+    assert len(df_src) == len(
+        df_tgt
+    ), f"We assume the source and target file would have the same number of lines, but got {len(df_src)} and {len(df_tgt)}."
+    df_combined = df.concat([df_src, df_tgt], axis=1)
+    df_combined["src_lang"] = src_lang
+    df_combined["tgt_lang"] = tgt_lang
+    df_combined["seg_id"] = pd.Series(range(len(df_combined))).astype(str)
```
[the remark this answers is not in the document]
shuoyangd: Yeah looking at this now I think we can probably remove this one. I don't think this ever gets used.

nemo_curator/filters/heuristic_filter.py
shuoyangd: Can you clarify on why `.iloc[0]` is fragile? My exposure to `DataFrame`-like objects is probably too little to understand.

If I understand correctly, your comment basically proposes two ideas:

- unify filter objects inheriting `DocumentFilter` and filter modules like `ScoreFilter`, `JointScoreFilter`, etc.
- unify `JointScoreFilter` and `ParallelScoreFilter` into a uniform base class for bitext filtering.

As far as I understood it, I actually don't quite like either of the ideas. Let me explain why.

Basically, `JointScoreFilter` and `ParallelScoreFilter` has different logics. Let's say each sentence pair is denoted as `(s, t)`.

- What `JointScoreFilter` accomplishes is to compute score with a function `f(s, t)`, then filter according to the return value of the function.
- What `ParallelScoreFilter` accomplishes is to compute *two* scores with function `f(s)`, `g(t)` (note that `f` and `g` are not necessarily the same function). Then, we first see if `s` and `t` will be retained. If any of these are not retained, we'll have to throw the other side as well to maintain the sentence alignment of the data.

This should already tell you that `JointScoreFilter` and `ParallelScoreFilter` will have different logics in the `__call__` function. Hence unifying them will result in some awkward design.

I also like the idea of decoupling the high-level filtering logic (in `ScoreFilter`, `JointScoreFilter`, etc.) and the specific filtering functions (`f(s, t)`, `f(s)` and `g(t)` mentioned above). What this enables me is to re-use any the monolingual filtering functions implemented previously in `ParallelScoreFilter`, which is great.

Let me know if that makes sense.

nemo_curator/modules/filter.py
shuoyangd: Yeah.. I just realized that I could have just apply monolingual filter twice, on source and target fields respectively.

I still think it's nice to have a wrapper class that does that for the user, instead of having users do that manually on parallel data, but I could probably simplify the implementation somewhat.

nemo_curator/utils/distributed_utils.py
```diff
@@ -521,6 +624,19 @@ def write_to_disk(df, output_file_dir, write_to_filename=False, output_type="jso
...
+                output_file_dir + "records" + f".{df['src_lang'].iloc[0]}"
+            )
+            tgt_output_file_path = (
+                output_file_dir + "records" + f".{df['tgt_lang'].iloc[0]}"
+            )
+            with open(src_output_file_path, "w") as src_out, open(
+                tgt_output_file_path, "w"
+            ) as tgt_out:
+                for src, tgt in zip(df["src"], df["tgt"]):
+                    src_out.write(src + os.linesep)
```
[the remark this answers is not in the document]
shuoyangd: Good call.. do you think I could just call `df.compute()` right after `elif` on L627?

shuoyangd (commented):
Thanks for the very thorough review! I left a few questions and comments. mainly concerned about the refactoring you proposed. Meanwhile I'll proceed with the lower-level changes.

nemo_curator/modules/filter.py
ryantwolf: If you think it's nice to have a wrapper class then I'm fine keeping it. Do you think you could change the implementation to just internally create and use two `ScoreFilter` modules? At a glance this seems to be possible, but let me know if it's not

nemo_curator/filters/heuristic_filter.py
ryantwolf: > - unify filter objects inheriting DocumentFilter and filter modules like ScoreFilter, JointScoreFilter, etc.

> ...

> What this enables me is to re-use any monolingual filtering functions implemented previously in ParallelScoreFilter, which is great.

This makes sense. I'm fine keeping the `DocumentFilter` logic separate from the `ScoreFilter` and `JointScoreFilter` modules because of this.

> - unify JointScoreFilter and ParallelScoreFilter into a uniform base class for bitext filtering.

Sorry for not being clear, this was not my intention. I agree they need to be separate since they are fundamentally two different operations.

> Can you clarify on why .iloc[0] is fragile? My exposure to DataFrame-like objects is probably too little to understand.

This is now my only concern with this filter. The fragility comment comes from the fact that it's easy to accidentally change the order of the items that get passed in here by changing the order the `fields` list gets populated in the `JointScoreFilter` class. If you could somehow have them be different arguments in the function, or identify them by unique names that would be preferable to the vague index.

Also, you mentioned that you liked the `DocumentFilter` to be separate from `ScoreFilter` since it allows you to reuse the filtering logic across methods. However, in this case it's harming you since this `DocumentFilter` method cannot be used with a `ScoreFilter`. You might want to define a new base class for this.

nemo_curator/utils/distributed_utils.py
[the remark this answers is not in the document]
ryantwolf: I think you need to put this in a map partitions call then call compute on it like is done earlier in the function [here](https://github.com/shuoyangd/NeMo-Curator/blob/main/nemo_curator/utils/distributed_utils.py#L605)

setup.py
```diff
@@ -42,6 +42,9 @@
         "dask[complete]>=2021.7.1",
         "distributed>=2021.7.1",
         "dask-mpi>=2021.11.0",
+        "tqdm==4.66.4",
```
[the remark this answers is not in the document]
shuoyangd: We can remove for now and add back if someone runs into issues (in my own experience there are often back-compatibility issues with `transformers` package).

tests/test_read_simple_bitext.py
shuoyangd: Interestingly, I get errors when I call it only once:

```
TypeError: Trying to convert dd.Scalar<series-..., dtype=bool> to a boolean value. Because Dask objects are lazily evaluated, they cannot be converted to a boolean value or used in boolean conditions like if statements. Try calling .compute() to force computation prior to converting to a boolean value or using in a conditional statement.
```

ryantwolf (changes requested):
Cool, just a couple of nits. I'll set up a dedicated call to walk through some of the model stuff though, I think that might need a bit more work.

nemo_curator/download/__init__.py
```diff
@@ -47,17 +47,7 @@
     "import_downloader",
     "import_extractor",
     "import_iterator",
-    "download_common_crawl",
```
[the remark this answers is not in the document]
ryantwolf: this still should be addressed.

nemo_curator/utils/distributed_utils.py
```diff
@@ -456,12 +459,25 @@ def single_partition_write_with_filename(df, output_file_dir, output_type="jsonl
...
             else:
                 raise ValueError(f"Unknown output type: {output_type}")
 
     return success_ser
 
 
+def _single_partition_write_to_simple_bitext(out_df, output_file_path):
+    src_output_file_path = output_file_path + f".{out_df['src_lang'].iloc[0]}"
+    tgt_output_file_path = output_file_path + f".{out_df['tgt_lang'].iloc[0]}"
+    with open(src_output_file_path, "w+") as src_out, open(
```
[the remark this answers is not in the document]
ryantwolf: Method has been changed. Resolving this.

nemo_curator/filters/bitext_filter.py
```diff
@@ -0,0 +1,144 @@
...
+        self.tgt_field = tgt_field
+        self.metadata_fields = metadata_fields
+        self.score_field = score_field
+        self.score_type = score_type
+        self.invert = invert
+
+    def __call__(
+        self,
+        dataset: ParallelDataset,
+        metadata_field_name_mapping: Dict[str, str] = {},
```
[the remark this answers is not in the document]
ryantwolf: Done

nemo_curator/filters/bitext_filter.py
```diff
@@ -0,0 +1,144 @@
...
+        kwargs["src"] = df[self.src_field]
+        kwargs["tgt"] = df[self.tgt_field]
+        for field_name in self.metadata_fields:
+            arg_name = metadata_field_name_mapping.get(field_name, field_name)
+            kwargs[arg_name] = df[field_name]
+
+        return self._score_bitext(**kwargs)
+
+    @abstractmethod
+    def _score_bitext(self, src, tgt, **kwargs):
```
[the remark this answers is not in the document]
ryantwolf: Fixed.

requirements/requirements_bitext.txt
ryantwolf, line 1: Can you delete this file now that we use the pyproject.toml?

ryantwolf (approved):
So excited to finally have this in!

PR #199, opened 2024-08-10
Add option to skip false positive checks during Fuzzy Deduplication

ayushdg:
## Description

The existing Minhash based Fuzzy Dedup algo has a computationally expensive false positive check comprising of the `map_buckets`, `jaccard_shuffle` & `jaccard_compute` stages.

This PR aims to allow an alternative codepath allowing skipping these computationally expensive operations and directly going from LSH buckets to graph creation, allowing significant perf gains at the cost of ~1-2% tokens depending on the chose initial config and dataset.

TODO:
Add perf results

## Usage

```python
fuzzy_dup_config = FuzzyDuplicatesConfig(false_positive_check=False)
...
```
## Checklist

- [ ] The documentation is up to date with these changes (Will be handled in a followup)

### DIFF
--- a/nemo_curator/modules/config.py
+++ b/nemo_curator/modules/config.py
@@ -84,9 +84,15 @@ class FuzzyDuplicatesConfig(BaseConfig):
             raise ValueError(
                 "Finding fuzzy duplicates requires a cache directory accessible via all workers to store intermediates"
             )
-        if not self.false_positive_check:
-            raise NotImplementedError(
-                "Skipping false positive checks is not supported at the moment"
+        if self.false_positive_check:
+            warnings.warn(
+                "Identifying false positives during the Minhash deduplication is computationally expensive."
+                " For improved performance consider setting this to False"
+            )
+        if not self.false_positive_check and self.char_ngrams < 20:
+            warnings.warn(
+                "Using a small char_ngrams value might lead to a large number of false positives during deduplication."
+                " Using a value for char_ngrams atleast 20 is recommended."
             )
         if self.num_anchors <= 0:
             raise ValueError("Number of anchors must be greater than 0")
--- a/nemo_curator/modules/config.py
+++ b/nemo_curator/modules/config.py
@@ -91,7 +91,7 @@ class FuzzyDuplicatesConfig(BaseConfig):
             )
         if not self.false_positive_check and self.char_ngrams < 20:
             warnings.warn(
-                "Using a small char_ngrams value might lead to a large number of false positives during deduplication."
+                "Using a small char_ngrams value might lead to a large number (~5%) of false positives during deduplication."
                 " Using a value for char_ngrams atleast 20 is recommended."
             )
         if self.num_anchors <= 0:
--- a/nemo_curator/modules/config.py
+++ b/nemo_curator/modules/config.py
@@ -92,7 +92,7 @@ class FuzzyDuplicatesConfig(BaseConfig):
         if not self.false_positive_check and self.char_ngrams < 20:
             warnings.warn(
                 "Using a small char_ngrams value might lead to a large number (~5%) of false positives during deduplication."
-                " Using a value for char_ngrams atleast 20 is recommended."
+                " Using a value of at least 20 for char_ngrams is recommended."
             )
         if self.num_anchors <= 0:
             raise ValueError("Number of anchors must be greater than 0")
--- a/nemo_curator/modules/fuzzy_dedup.py
+++ b/nemo_curator/modules/fuzzy_dedup.py
@@ -434,32 +434,44 @@ class FuzzyDuplicates:
             id_fields=[self.config.id_field],
             profile_dir=self.config.profile_dir,
         )
-        self.map_buckets = _MapBuckets(
-            id_fields=[self.config.id_field],
-            text_field=self.config.text_field,
-            logger=self._logger,
-            num_anchors=self.config.num_anchors,
-        )
-        self.jaccard_shuffle = _Shuffle(
-            id_fields=[self.config.id_field],
-            text_field=self.config.text_field,
-            logger=self._logger,
-            profile_dir=self.config.profile_dir,
-        )
-        self.jaccard_compute = JaccardSimilarity(
-            id_field=self.config.id_field,
-            text_field=self.config.text_field,
-            ngram_width=self.config.char_ngrams,
-            anchor_id_fields=[
-                f"anchor_{i}_{self.config.id_field}"
-                for i in range(self.config.num_anchors)
-            ],
+
+        if self.config.false_positive_check:
+            self.map_buckets = _MapBuckets(
+                id_fields=[self.config.id_field],
+                text_field=self.config.text_field,
+                logger=self._logger,
+                num_anchors=self.config.num_anchors,
+            )
+            self.jaccard_shuffle = _Shuffle(
+                id_fields=[self.config.id_field],
+                text_field=self.config.text_field,
+                logger=self._logger,
+                profile_dir=self.config.profile_dir,
+            )
+            self.jaccard_compute = JaccardSimilarity(
+                id_field=self.config.id_field,
+                text_field=self.config.text_field,
+                ngram_width=self.config.char_ngrams,
+                anchor_id_fields=[
+                    f"anchor_{i}_{self.config.id_field}"
+                    for i in range(self.config.num_anchors)
+                ],
+            )
+        else:
+            self.buckets_to_edges = _BucketsToEdges(
+                cache_dir=self.config.cache_dir,
+                id_fields=self.config.id_field,
+                logger=self._logger,
+            )
+
+        jaccard_pairs_fname = (
+            "jaccard_similarity_results.parquet"
+            if self.config.false_positive_check
+            else "_edges.parquet"
         )
         self.connected_components = ConnectedComponents(
             cache_dir=self.config.cache_dir,
-            jaccard_pairs_path=os.path.join(
-                self.config.cache_dir, "jaccard_similarity_results.parquet"
-            ),
+            jaccard_pairs_path=os.path.join(self.config.cache_dir, jaccard_pairs_fname),
             id_column=self.config.id_field,
             convert_str_ids=False,
             jaccard_threshold=self.config.jaccard_threshold,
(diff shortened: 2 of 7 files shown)


nemo_curator/modules/fuzzy_dedup.py
```diff
@@ -475,62 +490,189 @@ def __call__(self, dataset: DocumentDataset):
...
+            buckets_df.groupby(self.bucket_field)[self.str_id_name]
+            .agg(list)
+            .list.sort_values()
+        )
+        bucket_docs = grouped_buckets.to_arrow().to_pylist()
+        edges = []
+        # Create pairs of all documents within a bucket since they are near duplicates
+        # Effectively create a edge list of all near duplicate documents
+        for bucket_doc in bucket_docs:
+            edges.extend(combinations(bucket_doc, 2))
```
yury-tokpanov, line 646: Is it possible to do something else here? It seems like this step generates a lot of edges. We probably don't need to have all possible combinations.

ayushdg: Thanks for the ayushdg/Nemo-Curator#2 for a much better alternative which improve perf at least 10x from my testing. I could not retain the original commit history since the commit in that pr was not signed off as required by the repo, so I added you as a co-author to another commit.

nemo_curator/modules/config.py
```diff
@@ -84,9 +84,15 @@ def __post_init__(self):
...
-            raise NotImplementedError(
-                "Skipping false positive checks is not supported at the moment"
+        if self.false_positive_check:
+            warnings.warn(
+                "Identifying false positives during the Minhash deduplication is computationally expensive."
+                " For improved performance consider setting this to False"
+            )
+        if not self.false_positive_check and self.char_ngrams < 20:
+            warnings.warn(
+                "Using a small char_ngrams value might lead to a large number of false positives during deduplication."
```
ryantwolf, line 94: It could be useful adding the exact false positive rate we observed. Also good to add to the docs, but that can come later since I think an overhaul of the fuzzy dedup docs is already on the plan.

nemo_curator/modules/fuzzy_dedup.py
```diff
@@ -475,62 +490,200 @@ def __call__(self, dataset: DocumentDataset):
...
+        )
+        bucket_docs = grouped_buckets.to_arrow().to_pylist()
+        edges = []
+        # Create pairs of all documents within a bucket since they are near duplicates
+        # Effectively create a edge list of all near duplicate documents
+        for bucket_doc in bucket_docs:
+            edges.extend(pairwise(bucket_doc))
+        edges = pd.DataFrame(edges, columns=self.output_ids)
+        edges = pa.Table.from_pandas(edges)
+        result_df = cudf.DataFrame.from_arrow(edges)
```
ryantwolf, line 651: Why do you go through pyarrow instead of using `cudf.from_pandas`?

ryantwolf (commented):
Looks good overall, just some minor points

nemo_curator/modules/fuzzy_dedup.py
ayushdg: In the grand scheme of things this accounts for a small portion of the method runtime so it's probably insignificant, but for some reason `cudf.from_pandas` is a bit slower than `pandas->arrow->cuDF`. 
Here's some results I got:
```
Time taken from_pandas: 3.225353717803955s
Time taken from_pandas: 3.197599172592163s
Time taken from_pandas: 3.3601927757263184s
Time taken from_pandas: 3.8164901733398438s
Time taken from_pandas: 3.725322961807251s

Time taken from_arrow: 2.4147517681121826s
Time taken from_arrow: 1.6933174133300781s
Time taken from_arrow: 1.6882858276367188s
Time taken from_arrow: 1.3348863124847412s
Time taken from_arrow: 1.6437954902648926s
```

Here's the script:
```python
df_size = 10_000_000
df = pd.DataFrame({"a": np.random.randint(0, np.iinfo("int64").max, df_size), "b": np.random.randint(0, np.iinfo("int64").max, df_size)})
df["a"] = "dataset-1-" + df["a"].astype(str)
df["b"] = "dataset-1-" + df["b"].astype(str)

def f1(df):
    return cudf.from_pandas(df)

def f2(df):
    arrow_df = pa.Table.from_pandas(df)
    return cudf.DataFrame.from_arrow(arrow_df)

n_iter = 5

for i in range(n_iter):
    st = time.time()
    res = f1(df)
    print(f"Time taken from_pandas: {time.time()-st}s")

for i in range(n_iter):
    st = time.time()
    res = f2(df)
    print(f"Time taken from_arrow: {time.time()-st}s")
```

nemo_curator/modules/config.py
ayushdg: Just added a (~5%) in the warning to give users an estimate. The docs overhaul can contain more info on the actual false positive rates.

nemo_curator/modules/config.py
```diff
@@ -84,9 +84,15 @@ def __post_init__(self):
...
-            raise NotImplementedError(
-                "Skipping false positive checks is not supported at the moment"
+        if self.false_positive_check:
+            warnings.warn(
+                "Identifying false positives during the Minhash deduplication is computationally expensive."
+                " For improved performance consider setting this to False"
+            )
+        if not self.false_positive_check and self.char_ngrams < 20:
+            warnings.warn(
+                "Using a small char_ngrams value might lead to a large number (~5%) of false positives during deduplication."
```
yury-tokpanov, line 94: Theoretical FP rate with respect to Jaccard similarity can be computed like this, given the parameters of minhash LSH: https://github.com/ekzhu/datasketch/blob/5512549871d29c55b3cd8e99a79fcbd14859b77d/datasketch/lsh.py#L12

The issue with Jaccard similarity is that it itself is just a proxy for document similarity, quality of which depends on the chosen size and type of n-grams. Are there any studies that would compare Jaccard similarity with, say, edit similarity? Or is there even a theoretical way to do that?

We did some comparisons of minhash LSHs deduplication with word-based 5-grams vs 13-grams, and we found that FP rates with respect to Jaccard similarity was comparable with a theoretical value, but edit similarity FP rate was much smaller with 13-grams.

ayushdg: Agreed. From my understanding the formula helps estimate FP's based on the number of hashes and buckets, but as you mentioned doesn't take into account ngram sizes since there is no standard on what is the baseline ngram size for overlap (w.r.t jaccard) or equivalent edit distance that makes 2 document duplicates.

>Are there any studies that would compare Jaccard similarity with, say, edit similarity? Or is there even a theoretical way to do that?

I need to check but am not immediately aware of any studies/approach.

For the time being the 5% number quoted here is empirical evidence from some of our internal experiments at smaller ngram sizes (1 word/5char ngrams) where skipping the false positive check leads to an additional 5% documents being marked as duplicate. The % goes down when using larger ngrams (20 chars or above).

nemo_curator/modules/config.py
```diff
@@ -84,9 +84,15 @@ def __post_init__(self):
...
+                " Using a value for char_ngrams atleast 20 is recommended."
```
sarahyurick, line 95:
```suggestion
                " Using a value of at least 20 for char_ngrams is recommended."
```

nemo_curator/modules/fuzzy_dedup.py
```diff
@@ -475,62 +490,200 @@ def __call__(self, dataset: DocumentDataset):
...
+        print(f"Stage{stage_num}: Starting Minhash + LSH computation")
         minhashLSH = Sequential([self.minhash, self.lsh])
         buckets_df = minhashLSH(dataset)
-        print("Stage1: Minhash + LSH complete!")
+        print(f"Stage{stage_num}: Minhash + LSH complete!")
+        stage_num += 1
+
+        if self.config.false_positive_check:
+            # Map buckets to lower cardinality distribution
+            print(f"Stage{stage_num} (False Postive Check): Starting Map_Buckets")
```
sarahyurick, line 502:
```suggestion
            print(f"Stage{stage_num} (False Positive Check): Starting Map_Buckets")
```

nemo_curator/modules/fuzzy_dedup.py
```diff
@@ -475,62 +490,200 @@ def __call__(self, dataset: DocumentDataset):
...
+        else:
+            # Map buckets to lower cardinality distribution
+            print(f"Stage{stage_num}: Starting LSH Buckets to Graph edgelist")
+            self.buckets_to_edges(buckets_df)
+            print(f"Stage{stage_num}: Starting LSH Buckets to Graph edgelist Complete!")
+            stage_num += 1
 
         # Connected components across buckets
-        print("Stage5: Connected Components across buckets")
+        print("Stage{stage_num}: Connected Components across buckets")
```
sarahyurick, line 559:
```suggestion
        print(f"Stage{stage_num}: Connected Components across buckets")
```

nemo_curator/modules/fuzzy_dedup.py
```diff
@@ -475,62 +490,200 @@ def __call__(self, dataset: DocumentDataset):
...
         cc_path = os.path.join(self.config.cache_dir, "connected_components.parquet")
         self.connected_components.cc_workflow(cc_path)
-        print("Stage5: Connected Components across buckets complete!")
+        print("Stage{stage_num}: Connected Components across buckets complete!")
```
sarahyurick, line 562:
```suggestion
        print(f"Stage{stage_num}: Connected Components across buckets complete!")
```

tests/test_fuzzy_dedup.py
```diff
@@ -372,6 +372,47 @@ def test_num_anchors(self, large_fuzzy_dedup_data, num_anchors, tmpdir):
...
+            false_positive_check=False,
+            num_anchors=2,
+            jaccard_threshold=0.39,
+        )
+        fuzzy_duplicates = FuzzyDuplicates(config=config)
+        result = fuzzy_duplicates(fuzzy_dedup_data)
+        result_df = result.df.compute()
+        # Drop non duplicated docs
+        result_df = result_df[result_df.group.duplicated(keep=False)]
+        result_df = result_df.groupby("group").id.collect()
```
sarahyurick, line 406:
```suggestion
        result_df = result_df.groupby("group").id.agg(list)
```
like in `test_fuzzy_dedup`?

sarahyurick (commented):
LGTM, added some nits.

nemo_curator/modules/fuzzy_dedup.py
ryantwolf: Strange, thanks for the details though.

ryantwolf (approved):
Good on my end.

sarahyurick (approved):
LGTM.

PR #217, opened 2024-08-26
Remove extra `parse_args()` in classifier examples

sarahyurick:
Fixes minor bug discovered when reviewing https://github.com/NVIDIA/NeMo-Curator/pull/168.


VibhuJawa (approved):
Thanks for catching this @sarahyurick

PR #224, opened 2024-09-03
[Tutorials] Fix API key missing errors in the SDG tutorial

Maghoumi:
## Description
- Fixes an issue where the OpenAI client would fail to initialize if an API key is not provided.
- Add links to the NeMo FW tutorial for model fine-tuning.

## Checklist

### DIFF
--- a/tutorials/peft-curation-with-sdg/main.py
+++ b/tutorials/peft-curation-with-sdg/main.py
@@ -278,7 +278,7 @@ def run_pipeline(args, jsonl_fp):
     llm_client = AsyncOpenAIClient(
         AsyncOpenAI(
             base_url="https://integrate.api.nvidia.com/v1",
-            api_key=args.api_key,
+            api_key=args.api_key or "",
             timeout=args.api_timeout,
         )
     )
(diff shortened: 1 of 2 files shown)


tutorials/peft-curation-with-sdg/main.py
```diff
@@ -278,7 +278,7 @@ def run_pipeline(args, jsonl_fp):
     llm_client = AsyncOpenAIClient(
         AsyncOpenAI(
             base_url="https://integrate.api.nvidia.com/v1",
-            api_key=args.api_key,
+            api_key=args.api_key or "",
```
ryantwolf, line 281: I understand this was done to prevent an error, but shouldn't this raise an error? I imagine it failing silently is worse. Unless there is a more descriptive error that gets raised from this as a result.

ryantwolf (commented):
One clarification needed.

tutorials/peft-curation-with-sdg/main.py
Maghoumi: No silent failures here, this is actually preventing undesired bevavior :)

Note that in the preceding lines, if a key is not provided, or if the `synth-xyz` arguments are not set, the code will display appropriate messages (or errors). SDG would be skipped in all such cases.

Specifically, in case where the user does not provide an API key through the command line, they will see a message indicating SDG will be disabled and the rest of the code ensures everything works as intended. However, if the `api_key` value is `None`, the OpenAI client will start looking for the `OPENAI_API_KEY` environment variable, which is not desired (build.nvidia.com keys are different from OpenAI's, which can lead to confusing authentication errors.). My simple modification ensures that only the key that the user explicitly provides through the CLI would be used.

ryantwolf (approved):
Thank you!

PR #229, opened 2024-09-04
Clean up ID logic

sarahyurick:
Closes https://github.com/NVIDIA/NeMo-Curator/issues/150

Removes hardcoded `adlr_id` instances in favor of requiring the user to specify an ID themselves.


nemo_curator/utils/script_utils.py
```diff
@@ -497,11 +497,11 @@ def parse_gpu_dedup_args(description: str) -> argparse.ArgumentParser:
         argumentHelper.parser.add_argument(
             "--input-json-id-field",
             type=str,
-            default="adlr_id",
+            required=True,
             help="The name of the field within each json object of the jsonl "
             "file that assigns a unqiue ID to each document. "
             "Can be created by running the script "
-            "'./prospector/add_id.py' which adds the field 'adlr_id' "
+            "'./prospector/add_id.py' which adds the field "
```
ayushdg, line 504: Nit: can we also update the path to add_id in the comment here to the correct dir.

ayushdg (commented):
Minor nit, overall lgtm!

Quick note to @ryantwolf and @yyu22, with this change the `id` field is now a required flag when calling the CLI scripts. If CI/other scripts are relying on the default `adlr_id` we should update it to explicitly provide the ID column now.

nemo_curator/utils/script_utils.py
sarahyurick: Updated, thanks!

ayushdg (approved):
lgtm!

tests/test_download.py
```diff
@@ -124,7 +124,7 @@ def test_incorrect_snapshot_order_news(self):
             start_snapshot = "2021-10"
             urls = get_common_crawl_urls(start_snapshot, end_snapshot, news=True)
 
-    @pytest.mark.flaky(reruns=4)
+    @pytest.mark.flaky(reruns=5)
```
ryantwolf, line 127: Honestly, while you're here, if you don't mind deleting this test for now I think it's for the best. The flaky hasn't been much help. We can mock it later, but it's more hassle than it's worth at this point.

ryantwolf (approved):
Looks good, we just need to update the tests. Also I left another tangential comment that you can address now if you'd like or later.

tests/test_download.py
sarahyurick: Sure, can remove in a separate PR.

sarahyurick: Opened https://github.com/NVIDIA/NeMo-Curator/pull/235.

PR #231, opened 2024-09-05
Add ability to build using RAPIDS nightly

praateekmahajan:
## Description
We wish to build nightly of curator using nightly of RAPIDS. (https://github.com/NVIDIA/NeMo-Curator/issues/133)
We add a simple conditional on an environment variable `IS_NIGHTLY` that decides the path correctly. The user needs to provide the correct `index-url`.

## Usage
```sh
IS_NIGHTLY=1 pip install --extra-index-url=https://pypi.anaconda.org/rapidsai-wheels-nightly/simple ".[cuda12x]"
IS_NIGHTLY=0 pip install --extra-index-url=https://pypi.nvidia.com --no-cache-dir ".[cuda12x]"
```

### Dockerfile to test this
```
FROM rapidsai/ci-conda:cuda12.5.1-ubuntu22.04-py3.10

RUN conda create -y --name rapids -c conda-forge -c nvidia \
    python=3.10 \
    cuda-toolkit=12.5


RUN mamba run -n rapids pip install cython pytest --no-cache-dir
RUN mamba run -n rapids IS_NIGHTLY=0 pip install --extra-index-url= https://pypi.nvidia.com --no-cache-dir ".[cuda12x]"
```
### Questions for reviewer
1. Should our nightly RAPIDS_VERSION be similar as non-nightly, i.e`24.10` vs `24.8`
1. `strtobool(..)` or just `os.get(..)` or something else?
    - problem with `os.get(..)` is that it might not be obvious to the user why `IS_NIGHTLY=0` also installs nightly version since from CLI it's interpreted as a string and `if "0"` will resolve to true.
    - con with strtobool is it'll get deprecated in 3.12 and maybe an overkill?

## Checklist


setup.py
```diff
@@ -12,13 +12,18 @@
...
 from setuptools import setup, find_packages
 import pathlib
 
+from distutils.util import strtobool
+
 here = pathlib.Path(__file__).parent.resolve()
 
 long_description = (here / "README.md").read_text(encoding="utf-8")
 
+RAPIDS_VERSION = "24.10.0a0" if strtobool(os.getenv("IS_NIGHTLY", "False")) else "24.8"
```
bdice, line 25: I would suggest pinning to `==24.8.*` for stable builds and `>=24.10.0a0,<=24.10` for nightly builds. That's the recommendation of the [RAPIDS install guide](https://docs.rapids.ai/install), as well.

PR #234, opened 2024-09-07
Upgrade crossfit version

VibhuJawa:
## Description
This PR upgrades the crossfit version being used in NeMo-Curator. 

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -65,7 +65,7 @@ setup(
         "presidio-anonymizer==2.2.351",
         "usaddress==0.5.10",
         "nemo_toolkit[nlp]>=1.23.0",
-        "crossfit @ git+https://github.com/rapidsai/crossfit.git@0cc2993",
+        "crossfit @ git+https://github.com/rapidsai/crossfit.git@00.0.5",
         # Numpy 2.0 breaks with spacy https://github.com/explosion/spaCy/issues/13528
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -37,7 +37,7 @@
         "Programming Language :: Python :: 3.10",
     ],
     packages=find_packages(),
-    python_requires=">=3.10, <3.11",
+    python_requires=">=3.10, <3.12",
```
ayushdg, line 40: Does py3.11 just work? We should probably also add testing to our CI matrix before enabling this change. wdyt?

ryantwolf, line 40: Can we leave this out for now? We don't have a platform to test Python 3.11 right now. I agree we should make the switch soon, but I want to have some test plan built out first just in case.

ryantwolf (changes requested):
One comment about python version.

setup.py
VibhuJawa: Yeah, it just works , can leave it out though.  Let me revert

VibhuJawa: Agreed

ryantwolf: haha Ayush you barely beat me to it

VibhuJawa: Reverted

ryantwolf (approved):

ayushdg (approved):

PR #235, opened 2024-09-09
Remove flaky PyTest

sarahyurick:
Removes changes from https://github.com/NVIDIA/NeMo-Curator/pull/218. Since `test_uneven_common_crawl_range` is very flaky, we skip it for now. I will also open an issue for possibly re-adding it in the future.

### DIFF
--- a/tests/test_download.py
+++ b/tests/test_download.py
@@ -124,6 +124,22 @@ class TestDownload:
             start_snapshot = "2021-10"
             urls = get_common_crawl_urls(start_snapshot, end_snapshot, news=True)
 
+    @pytest.mark.skip(reason="Skipping until we figure out how to get this to a non flaky state")
+    def test_uneven_common_crawl_range(self):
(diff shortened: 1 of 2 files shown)


tests/test_download.py
```diff
@@ -124,22 +124,6 @@ def test_incorrect_snapshot_order_news(self):
             start_snapshot = "2021-10"
             urls = get_common_crawl_urls(start_snapshot, end_snapshot, news=True)
 
-    @pytest.mark.flaky(reruns=5)
```
VibhuJawa, line 127: Maybe we just mark it as skip if we want to add it back ? 

@pytest.mark.skip(reason="Skipping till we figure out how to get this to a non flaky state")

sarahyurick: Thanks! Sounds good to me.

VibhuJawa (approved):
LGTM

PR #237, opened 2024-09-09
Enabled nightly build using RAPIDS nightly

praateekmahajan:
## Description
We wish to build nightly of curator using nightly of RAPIDS. (https://github.com/NVIDIA/NeMo-Curator/issues/133)
We add a simple conditional on an environment variable `RAPIDS_NIGHTLY` that decides the path correctly. The user needs to provide the correct `index-url`.

## Usage
```sh
RAPIDS_NIGHTLY=1 pip install --extra-index-url=https://pypi.anaconda.org/rapidsai-wheels-nightly/simple ".[cuda12x]"
RAPIDS_NIGHTLY=0 pip install --extra-index-url=https://pypi.nvidia.com --no-cache-dir ".[cuda12x]"
```

### Dockerfile to test this
```
FROM rapidsai/ci-conda:cuda12.5.1-ubuntu22.04-py3.10

RUN conda create -y --name rapids -c conda-forge -c nvidia \
    python=3.10 \
    cuda-cudart \
    libcufft \
    libcublas \
    libcurand \
    libcusparse \
    libcusolver



RUN mamba run -n rapids \
    pip install cython pytest --no-cache-dir
RUN mamba run -n rapids \
    RAPIDS_NIGHTLY=0 pip install --extra-index-url= https://pypi.nvidia.com --no-cache-dir ".[cuda12x]"
```
### Questions for reviewer
1. `strtobool(..)` or just `os.get(..)` or something else?
    - problem with `os.get(..)` is that it might not be obvious to the user why `IS_NIGHTLY=0` also installs nightly version since from CLI it's interpreted as a string and `if "0"` will resolve to true.
    - con with strtobool is it'll get deprecated in 3.12 and maybe an overkill?

## Checklist


setup.py
```diff
@@ -15,6 +15,7 @@
 from setuptools import setup, find_packages
 import pathlib
 from itertools import chain
+from distutils.util import strtobool
```
ayushdg, line 18: I would like to avoid `distutils` since the package is deprecated in python 3.10 and will be removed completely in 3.12. 

I'm open to using any of our existing dependencies or writing a quick utility function to mimic this behavior.
relevant links:
https://peps.python.org/pep-0632/#migration-advice
https://stackoverflow.com/questions/77941994/what-is-setuptools-alternative-to-the-deprecated-distutils-strtobool
https://danielms.site/zet/2023/pythons-distutil-strtobool-replacement/

VibhuJawa (commented):
Can we also add a small section about rapids nightly in https://github.com/NVIDIA/NeMo-Curator/tree/main?tab=readme-ov-file#install-nemo-curator too please

PR #247, opened 2024-09-16
Make id field optional for filtering

ryantwolf:
## Description

Makes the id field optional for the `filter_docments` utility. It is only used if the user wants to attach each score to the corresponding id when writing out scores separately, so this does not need to be specified most of the time.

## Usage

N/A
## Checklist


sarahyurick (approved):
LGTM

PR #251, opened 2024-09-18
Add profiling / time in different stages of fuzzy dedup

praateekmahajan:
## Usage

```python
```
## Checklist

### DIFF
--- a/nemo_curator/modules/fuzzy_dedup.py
+++ b/nemo_curator/modules/fuzzy_dedup.py
@@ -14,13 +14,14 @@
 
 from __future__ import annotations
 
+import logging
 import math
 import os
 import time
 import warnings
 from datetime import datetime
 from itertools import pairwise
-from typing import List, Tuple, Union
+from typing import List, Optional, Tuple, Union
 
 import cudf
 import cugraph.dask as dcg
@@ -184,7 +185,7 @@ class MinHash:
         ):
             result.to_parquet(write_path, write_index=False, overwrite=True)
         self._logger.info(
-            f"Minhash signature computation for dataset took {time.time() - t0}s complete at {write_path}"  # noqa:E501
+            f"Time taken for Minhash signature computation = {time.time() - t0}s and output written at {write_path}"
         )
         return DocumentDataset(
             dask_cudf.read_parquet(write_path, blocksize="2GB", aggregate_files=True)
@@ -205,7 +206,7 @@ class LSH:
         logger: Union[logging.LoggerAdapter, str] = "./",
         id_fields: Union[str, list] = "id",
         minhash_field: str = "_minhash_signature",
-        profile_dir: str = None,
+        profile_dir: Optional[str] = None,
     ):
         """
         Parameters
@@ -380,7 +381,9 @@ class LSH:
             f"lsh-profile-{datetime.now().strftime('%Y%m%d_%H%M%S')}.html",
         ):
             self.lsh(write_path=write_path, df=df)
-        self._logger.info(f"Computing and writing buckets took {time.time() - t0} s")
+        self._logger.info(
+            f"Time taken for LSH = {time.time() - t0}s and output written at {write_path}"
+        )
 
         buckets_df = dask_cudf.read_parquet(write_path, split_row_groups=False)
         return DocumentDataset(buckets_df)
@@ -462,6 +465,7 @@ class FuzzyDuplicates:
                 cache_dir=self.config.cache_dir,
                 id_fields=self.config.id_field,
                 logger=self._logger,
+                profile_dir=self.config.profile_dir,
             )
 
         jaccard_pairs_fname = (
@@ -475,6 +479,8 @@ class FuzzyDuplicates:
             id_column=self.config.id_field,
             convert_str_ids=False,
             jaccard_threshold=self.config.jaccard_threshold,
+            logger=self._logger,
+            profile_dir=self.config.profile_dir,
         )
 
     def __call__(self, dataset: DocumentDataset):
@@ -489,6 +495,10 @@ class FuzzyDuplicates:
         DocumentDataset containing IDs of all documents and the corresponding duplicate group
         they belong to. Documents in the same group are near duplicates.
         """
+        if self.config.false_positive_check and not self.config.cache_dir:
+            msg = "cache_dir needs to be specified when false_positive_check is enabled"
+            raise ValueError(msg)
+
         # Minhash + LSH
         stage_num = 1
         print(f"Stage{stage_num}: Starting Minhash + LSH computation")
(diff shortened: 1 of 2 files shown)


nemo_curator/modules/fuzzy_dedup.py
```diff
@@ -500,15 +510,26 @@ def __call__(self, dataset: DocumentDataset):
...
-            ddf_mapped_buckets_w_anchors = self.map_buckets.map_buckets_with_anchors(
-                documents_df=dataset.df, buckets_df=buckets_df.df
-            )
+            t0 = time.time()
             mapped_buckets_w_anchors_path = os.path.join(
                 self.config.cache_dir, "anchor_docs_with_bk.parquet"
             )
-            ddf_mapped_buckets_w_anchors.to_parquet(
-                mapped_buckets_w_anchors_path, write_index=False
+            with performance_report_if(
```
praateekmahajan, line 517: For this profiling do we need _all_ of in this context or just the `to_parquet` part? Same questions below

ayushdg: In general we need to wrap `performance_report` around any sections of code that trigger a computation. For most modules (such as `minhash/jaccard_compute` etc. everything is lazy until the `to_parquet` call, so wrapping just that call is sufficient to capture the whole profile. 

On the other hand, stages like `LSH` have a `persist` call in the code which trigger a computation. In this case wrapping just the `to_parquet` call will only wrap everything _after_ the `persist` since there has already been some computation before persist not captured by profiling.

praateekmahajan: Good call out on intermediate _persists_, however if you see around Line 1466 where we call `from_dask_cudf_edgelist` and other methods within the `cc_workflow`. (or even in this method where we call `map_buckets_with_anchors`). Should we wrap these since it won't be obviously clear if within these methods there are intermediate actions happening (such as persist or writing to disk)

nemo_curator/modules/fuzzy_dedup.py
```diff
@@ -1410,49 +1461,60 @@ def _run_connected_components(
...
+            )
+            result = dcg.weakly_connected_components(G)
+            del G
+            max_partitions = min(32, result.npartitions)
+            n_components = len(
+                result[["labels"]].drop_duplicates(split_out=max_partitions)
+            )
+            num_labels = len(result)
+            print("# of groups", n_components)
+            print("# of docs removed", num_labels - n_components)
```
praateekmahajan, line 1496: @ayushdg do you have recommendation on logging these using logger, and if you have preference of a certain format. In other places we are following the format
> Time taken for **{NAME}** = **{TIME}** and output written at **{PATH}**

nemo_curator/modules/fuzzy_dedup.py
```diff
@@ -1410,49 +1461,60 @@ def _run_connected_components(
...
-            deduped_encoded_jaccard_path, blocksize="1GB", aggregate_files=True
-        )
-        df = df[df["jaccard"] == 1].reset_index(drop=True)
+        t0 = time.time()
+        with performance_report_if(
+            self.profile_dir,
+            f"connected-components-run-{datetime.now().strftime('%Y%m%d_%H%M%S')}.html",
+        ):
+
+            Comms.initialize(p2p=True)
```
VibhuJawa, line 1470: @ayushdg , should we remove this here ?  (Unrelated to the PR)
```suggestion
            Comms.initialize(p2p=True)
```

VibhuJawa (approved):
LGTM to me, thanks for pushing through this @praateekmahajan

nemo_curator/modules/fuzzy_dedup.py
ayushdg: Let's hold off on this update here. We can open a followup that sets this based on the cluster protocol. `True` for ucx and `False` for tcp

VibhuJawa: Lets discuss this on that separate PR.

nemo_curator/modules/semantic_dedup.py
```diff
@@ -200,20 +205,36 @@ def create_embeddings(
...
+            with performance_report_if(
+                self.config.profile_dir,
+                f"embedding-creator-{datetime.now().strftime('%Y%m%d_%H%M%S')}.html",
+            ):
+                write_to_disk(
+                    embedding_ddf,
+                    self.embedding_output_dir,
+                    write_to_filename=self.write_to_filename,
+                    output_type="parquet",
+                )
```
VibhuJawa, line 219: I think we might want to move this up to line `207`, just to ensure that 

` embedding_ddf = self.create_embeddings(dataset.df, self.input_column)` is added to the profile code trace.

nemo_curator/modules/semantic_dedup.py
```diff
@@ -304,63 +328,74 @@ def __call__(self, embeddings_dataset: DocumentDataset):
...
-        del embeddings_df
+            embeddings_df.to_parquet(
+                clustering_output_dir,
+                index=False,
+                partition_on="nearest_cent",
+            )
+            # TODO question for reviewer do we need a logger time here too? if so what "step" would that be for?
+            self.logger.info(
+                f"Saved embeddings by nearest center to {clustering_output_dir}"
+            )
```
VibhuJawa, line 397: What do you mean by logger time ?

VibhuJawa (changes requested):
Minor nits

nemo_curator/modules/semantic_dedup.py
praateekmahajan: Moved it in, but for future note, we tried profiling inside and out, and it doesn't matter where as long as an action / execution is happening within the `with profile(..)` context

nemo_curator/modules/semantic_dedup.py
praateekmahajan: I mean, in most places where we're logging the time taken at each "micro-stage" within a bigger pipeline so that we can closely observe the slower / faster parts. Given there is so much happening in `ClusteringModel.__call__`, similar to the profiling question above, which parts should we time? I've already added logs for `KMeans.fit` and `KMeans.predict`.

For now I'm calling the step `Assigning distance to each embedding` but pls let me know if that's not accurate.

nemo_curator/modules/fuzzy_dedup.py
```diff
@@ -488,6 +494,10 @@ def __call__(self, dataset: DocumentDataset):
         DocumentDataset containing IDs of all documents and the corresponding duplicate group
         they belong to. Documents in the same group are near duplicates.
         """
+        if self.config.false_positive_check and not self.config.cache_dir:
+            msg = "cache_dir needs to be specified when false_positive_check is enabled"
+            raise ValueError(msg)
+
```
ayushdg, line 500: While the individual components don't necessarily require the `cache_dir` to be present we set up the `FuzzyDuplicates` module to always require it enforced in the config: https://github.com/NVIDIA/NeMo-Curator/pull/251/files#diff-a4ab81e29b653f18212be5078fac527ad2aab530b988ab5ae8d330de59bf30dfL86.

We can probably skip the message here.

nemo_curator/modules/fuzzy_dedup.py
ayushdg: I don't have a preference but I agree we should log this in the logger with some standard prefix like `Result ...`

nemo_curator/modules/semantic_dedup.py
```diff
@@ -199,21 +204,37 @@ def create_embeddings(
...
+                f"embedding-creator-{datetime.now().strftime('%Y%m%d_%H%M%S')}.html",
+            ):
+                embedding_ddf = self.create_embeddings(dataset.df, self.input_column)
+                write_to_disk(
+                    embedding_ddf,
+                    self.embedding_output_dir,
+                    write_to_filename=self.write_to_filename,
+                    output_type="parquet",
+                )
+            dd = DocumentDataset(
```
ayushdg, line 220: Minor nit: From  dask conventions it's very common to use something like `from dask import dataframe as dd` so I prefer not having it as a variable name and using something like `ddf` instead.

ayushdg (changes requested):
Thanks a lot for standardizing this across the deduplication modules. Minor nits but overall lgtm!

nemo_curator/modules/fuzzy_dedup.py
```diff
@@ -499,15 +506,26 @@ def __call__(self, dataset: DocumentDataset):
...
-            )
+            t0 = time.time()
             mapped_buckets_w_anchors_path = os.path.join(
                 self.config.cache_dir, "anchor_docs_with_bk.parquet"
             )
-            ddf_mapped_buckets_w_anchors.to_parquet(
-                mapped_buckets_w_anchors_path, write_index=False
+            with performance_report_if(
+                self.config.profile_dir,
+                f"map_buckets-{datetime.now().strftime('%Y%m%d_%H%M%S')}.html",
```
VibhuJawa, line 515: Minor nit `%Y%m%d_%H%M%S` , maybe we change it to `PROFILE_TIMESTAMP_FORMAT` and change it to a constant.

praateekmahajan: Made the change!

VibhuJawa (approved):
LGTM, thanks for working on this, lets get it in .

ayushdg (approved):

PR #253, opened 2024-09-18
Add GPU CI/CD

sarahyurick:
This PR enables gpuCI for our PyTests marked with `@pytest.mark.gpu`.

To trigger it, a user has to add the `gpuci` label to their PR for the GPU tests to run. Only users with write access are able to add labels to PRs.

If more commits are added to the PR, the `gpuci` label has to be removed and re-added to run the GPU tests again. This is for security reasons; for example, a random user could open a PR, a user with write access adds the `gpuci` label, the GPU tests run safely, and then the user pushes malicious code. As long as no one with write access re-adds the `gpuci` label, it won't run at all.


.github/workflows/auto-label.yml
```diff
@@ -0,0 +1,30 @@
+name: Auto-Label for Write Access and Python Files
```
sarahyurick, line 1: This file checks if a PR is (1) created by a user with write access and (2) modifying at least 1 Python file. If both are true, it automatically adds the `gpuci` label to automatically kick off our GPU CI.

.github/workflows/gpuci.yml
```diff
@@ -0,0 +1,41 @@
+name: "GPU CI/CD"
```
sarahyurick, line 1: Used [NeMo's cicd-main.yml](https://github.com/NVIDIA/NeMo/blob/main/.github/workflows/cicd-main.yml) as a reference for creating this.

.github/workflows/gpuci.yml
```diff
@@ -0,0 +1,41 @@
...
+
+on:
+  pull_request:
+    branches:
+      - 'main'
+      - 'r**'
```
sarahyurick, line 7: This is from NeMo, meaning the workflow will also trigger for PRs targeting any branch that starts with the letter r followed by any characters (for example, we have `r0.3.0`, `r0.4.0`, and `r0.5.0`).

I personally don't think we need to check for this case but wanted to see what others think?

EDIT: It looks like we also have branches called `rywolf/*`, so this could be changed to `r0**`.

.github/workflows/gpuci.yml
```diff
@@ -0,0 +1,41 @@
...
+    types: [ labeled, synchronize ]
```
sarahyurick, line 8: `synchronize` makes it so that our GPU CI is run every time a new commit is added (provided the `gpuci` label is still there). This makes it more convenient for us, however it may cause security risks, and NeMo does not use it.

It probably makes sense for us not to use `synchronize`, but I wanted to include it here for consideration.

.github/workflows/gpuci.yml
```diff
@@ -0,0 +1,41 @@
...
+
+jobs:
+  gpu-test:
+    runs-on: self-hosted
+    if: ${{ github.event.label.name == 'gpuci' || github.event_name == 'synchronize' }}
```
sarahyurick, line 13: See above comment about `synchronize`.

.github/workflows/gpuci.yml
```diff
@@ -0,0 +1,41 @@
...
+
+    steps:
+    - name: Run nvidia-smi test
+      run: |
+        whoami
+        nvidia-smi
```
sarahyurick, line 19: Not strictly needed, but a nice way to make sure our GPU runner is okay.

.github/workflows/gpuci.yml
```diff
@@ -0,0 +1,41 @@
...
+
+    - name: Checkout the PR code
+      uses: actions/checkout@v3
+
+    - name: Set up Python 3.10
+      uses: actions/setup-python@v4
+      with:
+        python-version: '3.10'
```
sarahyurick, line 27: I guess we could have this mimic our existing [test.yml](https://github.com/NVIDIA/NeMo-Curator/blob/main/.github/workflows/test.yml#L31-L34) logic if we wanted.

.github/workflows/gpuci.yml
```diff
@@ -0,0 +1,41 @@
...
+
+    - name: Install NeMo-Curator and PyTest
+      run: |
+        pip install wheel cython
```
sarahyurick, line 31: Might be a silly question, but do we need `wheel`? I saw it in [test.yml](https://github.com/NVIDIA/NeMo-Curator/blob/main/.github/workflows/test.yml#L42) which is why I added it.

.github/workflows/gpuci.yml
```diff
@@ -0,0 +1,41 @@
...
+        RAPIDS_NIGHTLY=1 pip install --extra-index-url=https://pypi.anaconda.org/rapidsai-wheels-nightly/simple ".[cuda12x]"
```
sarahyurick, line 32: I think it makes sense to use RAPIDS Nightly Builds here, but please feel free to comment if not? I just took this command straight from our README.

Alternatively, we could just do `pip install --extra-index-url https://pypi.nvidia.com ".[cuda12x]"`.

Also, maybe it would make sense to directly point to [requirements_rapids_nightly.txt](https://github.com/NVIDIA/NeMo-Curator/blob/main/requirements/requirements_rapids_nightly.txt)/[requirements_cuda12x.txt](https://github.com/NVIDIA/NeMo-Curator/blob/main/requirements/requirements_cuda12x.txt) and [requirements.txt](https://github.com/NVIDIA/NeMo-Curator/blob/main/requirements/requirements.txt) instead?

.github/workflows/gpuci.yml
```diff
@@ -0,0 +1,41 @@
...
+        pip install pytest
+    
+    - name: Verify installation
+      run: |
+        pip list
```
sarahyurick, line 37: Not strictly necessary but nice for debugging.

.github/workflows/gpuci.yml
```diff
@@ -0,0 +1,41 @@
...
+
+    - name: Run PyTests with GPU mark
+      run: |
+        python -m pytest -m gpu
```
sarahyurick, line 41: We run our GPU CI tests only on PyTests with the `pytest.mark.gpu` decorator. We already have quite a few of them in our repository:

- [exact_dedup_data](https://github.com/NVIDIA/NeMo-Curator/blob/main/tests/test_exact_dedup.py#L25)
- [TestSemDuplicates](https://github.com/NVIDIA/NeMo-Curator/blob/main/tests/test_semdedup.py#L51)
- [TestMetadataSep](https://github.com/NVIDIA/NeMo-Curator/blob/main/tests/test_seperate_by_metadata.py#L43)
- [TestMinhashes](https://github.com/NVIDIA/NeMo-Curator/blob/main/tests/test_fuzzy_dedup.py#L92)
- [TestLSH](https://github.com/NVIDIA/NeMo-Curator/blob/main/tests/test_fuzzy_dedup.py#L153)
- [TestFuzzyDuplicates](https://github.com/NVIDIA/NeMo-Curator/blob/main/tests/test_fuzzy_dedup.py#L210)
- [test_extract_partitioning_index](https://github.com/NVIDIA/NeMo-Curator/blob/main/tests/test_fuzzy_dedup.py#L463)

sarahyurick (commented):
Need to fix trailing whitespaces in `gpuci.yml` and should make the spacing in `auto-label.yml` consistent with `gpuci.yml`.

Also, maybe I should rename it to `auto-label-gpuci.yml`.

.github/workflows/gpuci.yml
sarahyurick: I also don't know about including `--no-cache-dir`?

.github/workflows/auto-label.yml
```diff
@@ -0,0 +1,30 @@
...
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      continue-on-error: true
+    - name: Verify Python files are modified
+      id: python-files
+      run: |
+        git diff --name-only ${{ github.event.pull_request.base.sha }} ${{ github.event.pull_request.head.sha }} | grep '.py' || echo 'no-python'
+      continue-on-error: true
+    - name: Add label if conditions are met
+      if: ${{ steps.write-access.outputs.permission == 'write' && steps.python-files.outputs.result != 'no-python' }}
+      uses: actions-ecosystem/action-add-labels@v1
```
sarahyurick, line 26: :(
```
actions-ecosystem/action-add-labels@v1 is not allowed to be used in NVIDIA/NeMo-Curator. Actions in this workflow must be: within a repository that belongs to your Enterprise account, created by GitHub, verified in the GitHub Marketplace, or matching the following: NVIDIA/gitlab-answer-app/*, aws-actions/configure-aws-credentials@*, cinotify/github-action@*, codecov/codecov-action@*, conda-forge/automerge-action@*, conda-incubator/setup-miniconda@*, convictional/trigger-workflow-and-wait@*, crazy-max/ghaction-docker-meta@*, dangoslen/changelog-enforcer@*, EndBug/add-and-commit@*, jacobtomlinson/gha-anaconda-package-version@*, jacobtomlinson/gha-find-replace@*, peter-evans/create-pull-request@*, peter-evans/dockerhub-description@*, pre-commit/action@*, pypa/gh-action-pypi-publish@*, release-drafter/release-drafter@*, ruby/setup-ruby@*, softprops/action-gh-release@*, test-summary/action@*, the-coding-turtle/ga-yaml-parser@*, devcontainers/action@*, tj-actions/changed-files@*, nrwl/las...
```
I might be able to use something else (like `peter-evans/create-pull-request@*`?) or use a GitHub API approach like:
```
- name: Add label if conditions are met
  if: ${{ steps.write-access.outputs.permission == 'write' && steps.python-files.outputs.result != 'no-python' }}
  run: |
    curl -X POST \
      -H "Authorization: token ${{ secrets.GITHUB_TOKEN }}" \
      -H "Accept: application/vnd.github.v3+json" \
      https://api.github.com/repos/${{ github.repository }}/issues/${{ github.event.pull_request.number }}/labels \
      -d '{"labels":["Run GPU CI"]}'
```

.github/workflows/gpuci.yml
ayushdg: If a regex works maybe something like `r[0-9]*`? R followed by any number

.github/workflows/gpuci.yml
ayushdg: @praateekmahajan might have run into this sooner, but iirc wheel is needed for pycld2 or fastext(don't recall which on) to install as expected.

.github/workflows/gpuci.yml
ayushdg: `--no-cache-dir` is optional, it's useful on machines with limited storage, but not a big issue either way.

Curious on others thoughts here, but I prefer running GPU tests without Rapids nightly. I see pytests on pr's as a way to check if we broke any gpu functionality. 
As a followup ( I can assist if needed) we can set up a nightly or weekly job that runs these same pytests with Rapids nightlies to identify any upstream breakages.

README.md
```diff
@@ -131,9 +131,9 @@ pip install --extra-index-url https://pypi.nvidia.com nemo-curator[cuda12x]
     pip install --extra-index-url https://pypi.nvidia.com ".[cuda12x]"
     ```
 
-#### Using Nightly Dependencies for Rapids
+#### Using Nightly Dependencies for RAPIDS
 
-You can also install NeMo Curator using the Rapids nightly, to do so you can set the environment variable `RAPIDS_NIGHTLY=1`.
+You can also install NeMo Curator using the [RAPIDS Nightly Builds](https://anaconda.org/rapidsai-nightly). To do so, you can set the environment variable `RAPIDS_NIGHTLY=1`.
```
ayushdg, line 136: I would prefer linking to the pip nightly index for RAPIDS or the rapids website with nightly install instructions here rather than the conda channel.

.github/workflows/gpuci.yml
praateekmahajan: I don't think we need `wheel`, or atleast I haven't run into it yet when building a container

Let's try removing it and then see?

.github/workflows/gpuci.yml
```diff
@@ -0,0 +1,43 @@
...
+
+    - name: Set up Python 3.10
+      uses: actions/setup-python@v4
+      with:
+        python-version: '3.10'
+
+    - name: Install NeMo-Curator and PyTest
+      run: |
+        sudo apt-get update
+        sudo apt-get install -y build-essential cmake g++ python3-dev
```
praateekmahajan, line 32: Maybe this works but something to be careful about is which `gcc` (or `g++`) it ends up using, since one doesn't work for fasttext==0.9.2 (https://github.com/facebookresearch/fastText/issues/1281)

README.md
```diff
@@ -131,9 +131,9 @@ pip install --extra-index-url https://pypi.nvidia.com nemo-curator[cuda12x]
     pip install --extra-index-url https://pypi.nvidia.com ".[cuda12x]"
     ```
 
-#### Using Nightly Dependencies for Rapids
+#### Using Nightly Dependencies for RAPIDS
```
praateekmahajan, line 134: Thanks for the cleanup here @sarahyurick

.github/workflows/gpuci.yml
sarahyurick: I don't think that would match to the periods in something like `r0.5.0`.

sarahyurick: But maybe something like `r[0-9].*`?

.github/workflows/gpuci.yml
sarahyurick: Cool, looks like it's able to work without it.

.github/workflows/auto-label.yml
sarahyurick: This feature is mostly for convenience. I can add it in a future PR if desired.

sarahyurick:
Thanks @ayushdg and @praateekmahajan ! I have it working now, but had to comment out one of our existing GPU tests which was failing. It looks like the others are running though.

To summarize, a user has to add the `gpuci` label to their PR for the GPU tests to run. Only users with write access are able to add labels to PRs.

If more commits are added to the PR, the `gpuci` label has to be removed and re-added to run the GPU tests again. This is for security reasons; for example, a random user could open a PR, we add the `gpuci` label, the GPU tests run safely, and then the user pushes malicious code. As long as none of us re-add the `gpuci` label, it won't run at all.

.github/workflows/gpuci.yml
```diff
@@ -0,0 +1,41 @@
...
+    - name: Run nvidia-smi test
+      run: |
+        whoami
+        nvidia-smi
+
+    - name: Checkout the PR code
+      uses: actions/checkout@v3
+
+    - name: Set up Python 3.10
+      uses: actions/setup-python@v4
```
ko3n1g, line 25: Hey folks!

Are you planning to migrate to a containerized test later? With our non-ephemeral infra I have a bit of worries with pure python installs. This can lead to nasty unexpected behavior. Workflows don’t run in isolation, thus concurrently running workflows can override each others environment. A containerized workflow (either building freshly or inheriting from NeMo) would be much safer & easier to debug.

.github/workflows/gpuci.yml
ko3n1g: I would recommend two patterns (besides main), and actually good reminder to fix it for NeMo:

* `r[0-9].[0-9].[0-9]` for release branches
* `[rv][0-9].[0-9].[0-9]rc[0-9]` for pre-release branches (like we had with NeMo 2.0.0rc1)

These are rules you could also use for branch protection. I learned the hard way that GH doesn’t use regex but fnmatch, which is quite limited. So we need two rules to cover both cases.

.github/workflows/auto-label.yml
ko3n1g: Ah another thing: This didn’t work because using secrets.GITHUB_TOKEN is not permitted to trigger other workflows. GH disallows that to prevent spam. You would need to use your personal access token (PAT). However, since we’re in a public environment, you would need to limit that to a protected branch (via environments). It’s still doable (from branch main, react to all issue change events), but a bit more tricky. As long as the PAT is not exposed to all branches, this is fine

.github/workflows/auto-label.yml
sarahyurick: Opened https://github.com/NVIDIA/NeMo-Curator/issues/259 for now.

.github/workflows/auto-label.yml
sarahyurick: Thanks! Added this comment to https://github.com/NVIDIA/NeMo-Curator/issues/259 for now.

.github/workflows/gpuci.yml
sarahyurick: Thanks! Left a comment about this in https://github.com/NVIDIA/NeMo-Curator/issues/259 as a reminder.

.github/workflows/_build_container.yml
```diff
@@ -0,0 +1,65 @@
...
+
+      - name: Pull cache images
+        run: |
+          docker pull nemoci.azurecr.io/nemo_curator_container:${{ github.event.pull_request.number || 'buildcache' }} || true
+          docker pull nemoci.azurecr.io/nemo_curator_container:buildcache || true
+
+      - name: Build and push
+        uses: docker/build-push-action@v5
+        with:
+          file: Dockerfile
```
ko3n1g, line 44: This file needs to be created (on the root directory of the repository). Something simple like this will do the trick: 

```dockerfile
FROM python:3.10

WORKDIR /opt

ARG CURATOR_COMMIT
RUN <<"EOF" bash -exu
git clone https://github.com/NVIDIA/NeMo-Curator.git
cd NeMo-Curator
git checkout $CURATOR_COMMIT
pip install cython pytest setuptools
pip install --extra-index-url https://pypi.nvidia.com ".[cuda12x]"
EOF
```

It will build a self-contained image that we can later instantiate and run as a container

.github/workflows/_build_container.yml
```diff
@@ -0,0 +1,65 @@
...
+          push: true
+          build-args: |
+            MAX_JOBS=32
+            ALIGNER_COMMIT=${{ github.event.pull_request.head.sha || github.sha }}
```
ko3n1g, line 48:
```suggestion
            CURATOR_COMMIT=${{ github.event.pull_request.head.sha || github.sha }}
```

.github/workflows/gpuci.yml
```diff
@@ -0,0 +1,61 @@
...
+      run: |
+        docker exec nemo-curator-container apt-get update
+        docker exec nemo-curator-container apt-get install -y python3.10 python3.10-venv python3.10-dev
+        docker exec nemo-curator-container python3.10 -m venv /opt/py310
+        docker exec nemo-curator-container /opt/py310/bin/pip install --upgrade pip
+
+    - name: Install NeMo-Curator and PyTest dependencies
+      run: |
+        docker exec nemo-curator-container /opt/py310/bin/pip install cython pytest setuptools pip --upgrade
+        docker exec nemo-curator-container /opt/py310/bin/pip install --extra-index-url https://pypi.nvidia.com ".[cuda12x]"
```
ko3n1g, line 50: Can be dropped now since we installed everything before into an image

.github/workflows/gpuci.yml
```diff
@@ -0,0 +1,61 @@
...
+
+    - name: Verify installation
+      run: |
+        pip list
```
ko3n1g, line 54:
```suggestion
        docker exec nemo-curator-container pip list
```

.github/workflows/gpuci.yml
```diff
@@ -0,0 +1,61 @@
...
+
+    - name: Set up Docker Buildx
+      uses: docker/setup-buildx-action@v3
+
+    - name: Pull the Docker image built in the previous step
+      run: docker pull nemoci.azurecr.io/nemo_curator_container:buildcache
+
+    - name: Run Docker container with Python 3.10
+      run: |
+        docker run --name nemo-curator-container -d nemoci.azurecr.io/nemo_curator_container:buildcache bash -c "sleep infinity"
```
ko3n1g, line 38:
```suggestion
        docker run --name nemo-curator-container -d nemoci.azurecr.io/nemo_curator_container:${{ github.run_id }} bash -c "sleep infinity"
```
Very important note: The image tag `buildcache` is only meant to be used for speeding up builds as it provides a cache manifest. The source code might be outdated (which is okay for a cache image since it will lead to a cache miss at the relevant layer). But for running your test suite you want to use the produced image which you can find by the image tag `${{ github.run_id }}`

.github/workflows/gpuci.yml
```diff
@@ -0,0 +1,61 @@
...
+        docker exec nemo-curator-container /opt/py310/bin/pip install cython pytest setuptools pip --upgrade
+        docker exec nemo-curator-container /opt/py310/bin/pip install --extra-index-url https://pypi.nvidia.com ".[cuda12x]"
+
+    - name: Verify installation
+      run: |
+        pip list
+
+    - name: Run PyTests with GPU mark
+      run: |
+        docker exec nemo-curator-container /opt/py310/bin/pytest -m gpu
```
ko3n1g, line 58:
```suggestion
        docker exec nemo-curator-container pytest -m gpu
```

.github/workflows/gpuci.yml
```diff
@@ -0,0 +1,61 @@
...
+    if: ${{ github.event.label.name == 'gpuci' }}
+
+    steps:
+    - name: Run nvidia-smi test
+      run: |
+        whoami
+        nvidia-smi
+
+    - name: Checkout the PR code
+      uses: actions/checkout@v4
```
ko3n1g, line 28: Can also be dropped (we run everything inside the container where the source code already exists - so no need to checkout again)

ko3n1g (changes requested):
Hey @sarahyurick, 
thanks for picking up our templates! 
It's drafted some snippets that should help finalizing this PR. 
So looks overall great to me, I think we can finalize it very soon

.github/workflows/gpuci.yml
```diff
@@ -0,0 +1,47 @@
...
+    steps:
+    - name: Remove existing container if it exists
+      run: |
+        if [ "$(docker ps -aq -f name=nemo-curator-container)" ]; then
+            docker rm -f nemo-curator-container
+        fi
+
+    - name: Run Docker container with Python 3.10
+      run: |
+        docker run --gpus all --name nemo-curator-container -d nemoci.azurecr.io/nemo_curator_container:${{ github.run_id }} bash -c "sleep infinity"
```
ko3n1g, line 30:
```suggestion
        docker run --gpus all --runtime=nvidia --name nemo-curator-container -d nemoci.azurecr.io/nemo_curator_container:${{ github.run_id }} bash -c "sleep infinity"
```

ko3n1g: That should sovel the missing cuda driver issue

Dockerfile
```diff
@@ -0,0 +1,15 @@
+FROM nvcr.io/nvidia/pytorch:24.07-py3
```
sarahyurick, line 1: [PyTorch Release 24.07](https://docs.nvidia.com/deeplearning/frameworks/pytorch-release-notes/rel-24-07.html)

sarahyurick: This is what NeMo uses in their [Dockerfile.ci](https://github.com/NVIDIA/NeMo/blob/main/Dockerfile.ci#L17) file. Happy to use a different version if people have other preferences.

VibhuJawa: My preference is RAPIDS CI container too.  We are not really tip of the spear when it comes to using Pytorch features but are for RAPIDS so probably using a barebones Rapids CI container will be better to make it similar to other RAPIDS builds.

sarahyurick: Thanks all! Updated.

praateekmahajan (approved):
This is great and helps us catch issues much earlier! Thank you for pushing through this!

Dockerfile
```diff
@@ -0,0 +1,26 @@
...
+  libcusparse \
+  libcusolver
+
+RUN bash -exu <<EOF
+  git clone $FORKED_REPO_URL
+  cd NeMo-Curator
+  git fetch origin $CURATOR_COMMIT --depth=1
+  git checkout $CURATOR_COMMIT
+  source activate rapids
+  conda install -y -c conda-forge cython pytest setuptools
```
praateekmahajan, line 23: Do you need it to be a `conda` install?

I think you can do `pip install cython pytest`. You probably also don't need `setuptools`

Dockerfile
```diff
@@ -0,0 +1,26 @@
+FROM rapidsai/ci-conda:cuda12.5.1-ubuntu22.04-py3.10
```
ayushdg, line 1: Nit:
```suggestion
ARG CUDA_VER=12.5.1
ARG LINUX_VER=ubuntu22.04
ARG PYTHON_VER=3.10
FROM rapidsai/ci-conda:cuda${CUDA_VER}-${LINUX_VER}-py${PYTHON_VER}
```

Dockerfile
```diff
@@ -0,0 +1,26 @@
...
+
+WORKDIR /opt
+
+ARG FORKED_REPO_URL
+ARG CURATOR_COMMIT
+
+RUN conda create -y --name rapids -c conda-forge -c nvidia \
```
ayushdg, line 8:
```suggestion
RUN conda create -y --name curator -c conda-forge -c nvidia \
```

Dockerfile
```diff
@@ -0,0 +1,26 @@
...
+
+RUN bash -exu <<EOF
+  git clone $FORKED_REPO_URL
+  cd NeMo-Curator
+  git fetch origin $CURATOR_COMMIT --depth=1
+  git checkout $CURATOR_COMMIT
+  source activate rapids
+  conda install -y -c conda-forge cython pytest setuptools
+  pip install --upgrade pip
+  pip install --extra-index-url https://pypi.nvidia.com ".[cuda12x]"
```
ayushdg, line 25: Potentially install `.[all]` to also include image. (Image doesn't have tests right now, but since this is a generic docker file in the repo, folks might use it outside CI to build containers.

ayushdg (changes requested):
Minor suggestions

.github/workflows/gpuci.yml
```diff
@@ -0,0 +1,47 @@
...
+        docker run --gpus all --runtime=nvidia --name nemo-curator-container -d nemoci.azurecr.io/nemo_curator_container:${{ github.run_id }} bash -c "sleep infinity"
+
+    - name: Check GPUs
+      run: |
+        whoami
+        docker exec nemo-curator-container nvidia-smi
+
+    - name: Verify installations
+      run: |
+        docker exec nemo-curator-container /bin/bash -c "source activate rapids && pip list"
```
ayushdg, line 39: Probably update the environment name to Curator here as well if it's updated in the docker file.

sarahyurick:
@ko3n1g is disconnecting our runner for a bit. Should be back up tomorrow, then I can rerun the checks and merge.

Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+ARG CUDA_VER=12.5.1
+ARG LINUX_VER=ubuntu22.04
+ARG PYTHON_VER=3.10
+FROM rapidsai/ci-conda:cuda${CUDA_VER}-${LINUX_VER}-py${PYTHON_VER}
+
+WORKDIR /opt
+
+# Needed to navigate to and pull the forked repository's changes
+ARG FORKED_REPO_URL
+ARG CURATOR_COMMIT
```
ko3n1g, line 12: Please move these two lines below the expensive `RUN conda...` command.

Background:
Docker invalidates cache of `RUN` layers whenever a build-arg `ARG` was modified. This is, b/c behind the scenes docker mounts build-args as environment variables into each `RUN` process and so if the `ARG` changed the checksum of that layer differs to cache resulting in a cache break.

We want to define `ARG` as close as possible to the actual usage.

Better caching leads to not only quicker build-time but also to less storage pressure on the host

ko3n1g (changes requested):
This is important since it impacts other projects that use this infra

Dockerfile
sarahyurick: Thank you! Updated.

ko3n1g (approved):

PR #266, opened 2024-09-25
Expand RMM options for Python API

sarahyurick:
Closes https://github.com/NVIDIA/NeMo-Curator/issues/260


nemo_curator/utils/distributed_utils.py
```diff
@@ -137,6 +154,31 @@ def get_client(
...
+            unused memory held by the pool will be released at the next synchronization point.
+            Can be an integer (bytes), float (fraction of total device memory), string (like "5GB" or "5000M") or None.
+            By default, this feature is disabled.
+            Note: This size is a per-worker configuration, and not cluster-wide.
+        rmm_track_allocations: For GPU-based clusters only. If True, wraps the memory resource used by each worker with a rmm.mr.TrackingResourceAdaptor,
+            which tracks the amount of memory allocated.
+            Note: This option enables additional diagnostics to be collected and reported by the Dask dashboard.
+            However, this is significant overhead associated with this and it should only be used for debugging and memory profiling.
+        cluster_kwargs: For GPU-based clusters only. Additional keyword arguments for the LocalCUDACluster configuration.
+            See API documentation https://docs.rapids.ai/api/dask-cuda/nightly/api/ for all possible parameters.
```
sarahyurick, line 181: I just chose all of the RMM-related parameters from https://docs.rapids.ai/api/dask-cuda/nightly/api/ and copied the descriptions from there too.

I'm happy to add/remove any parameters that people might think would be helpful for our users, and delegate all remaining ones to `cluster_kwargs`. For example, I've never personally used `rmm_track_allocations` so maybe we don't want to explicitly list it here... On the flip side, `CUDA_VISIBLE_DEVICES` is a very useful one and might make sense to include here instead of just letting `cluster_kwargs` catch it.

nemo_curator/utils/distributed_utils.py
```diff
@@ -106,6 +117,12 @@ def get_client(
     rmm_pool_size="1024M",
     enable_spilling=True,
     set_torch_to_use_rmm=False,
+    rmm_async=True,
+    rmm_maximum_pool_size=None,
+    rmm_managed_memory=False,
+    rmm_release_threshold=None,
+    rmm_track_allocations=False,
+    **cluster_kwargs,
```
ryantwolf, line 125: Can you also pass `cluster_kwargs` to the CPU `LocalCluster` path incase folks want to tweak those too? Update the docstring as well when you do to include that note. Unless there's a reason you only wanted to pass on them to GPU clusters.

ryantwolf (changes requested):
One small comment. I'll leave the decision of what args to explicitly call out up to y'all RAPIDS folks. My guiding principle would be that anything we have ever had to set should be there at least.

nemo_curator/utils/distributed_utils.py
sarahyurick: Works for me! Updated.

nemo_curator/utils/distributed_utils.py
ayushdg: I agree that we should keep this list as small as possible only keeping essential options. 
I would avoid explicit kwargs for `rmm_track_allocations` and possibly `rmm_maximum_pool_size`.

I'll let you @VibhuJawa comment on whether `rmm_managed_memory` is needed or not.

nemo_curator/utils/distributed_utils.py
```diff
@@ -154,6 +175,32 @@ def get_client(
             host to enable out-of-memory computation, i.e., computing on objects that occupy more memory than is available on the GPU.
         set_torch_to_use_rmm: For GPU-based clusters only. Sets up the PyTorch memory pool to be the same as the RAPIDS memory pool.
             This helps avoid OOM errors when using both PyTorch and RAPIDS on the same GPU.
+        rmm_async: For GPU-based clusters only. Initializes each worker with RMM and sets it to use RMM's asynchronous allocator.
+            See rmm.mr.CudaAsyncMemoryResource for more info.
```
ayushdg, line 179: Not valid for this repo. can probably remove.

nemo_curator/utils/distributed_utils.py
```diff
@@ -154,6 +175,32 @@ def get_client(
             host to enable out-of-memory computation, i.e., computing on objects that occupy more memory than is available on the GPU.
         set_torch_to_use_rmm: For GPU-based clusters only. Sets up the PyTorch memory pool to be the same as the RAPIDS memory pool.
             This helps avoid OOM errors when using both PyTorch and RAPIDS on the same GPU.
+        rmm_async: For GPU-based clusters only. Initializes each worker with RMM and sets it to use RMM's asynchronous allocator.
```
ayushdg, line 178: We should either link to RMM or simplify the explanation here to say async memory pool to not confuse users.

VibhuJawa (approved):
LGTM

VibhuJawa (approved):
LGTM

ryantwolf (approved):
Good with me, thanks!

PR #270, opened 2024-09-30
Remove `DASK_DATAFRAME__QUERY_PLANNING` environment variable

sarahyurick:
Since https://github.com/NVIDIA/NeMo-Curator/pull/139 is merged, we don't need to be explicitly setting `DASK_DATAFRAME__QUERY_PLANNING=False` anymore.


ryantwolf (approved):
You caught all the ones I could find too. Thanks!

PR #273, opened 2024-10-01
Pin Rapids to 24.8 for 0.5.0 Release 

praateekmahajan:
## Description
Since our release is planned for 2nd week of October, and [Rapids might do a 24.10 release by 10th Oct](https://docs.rapids.ai/maintainers/)


ayushdg (approved):

ryantwolf (approved):

requirements/requirements.txt
```diff
@@ -14,7 +14,6 @@ justext==3.0.1
 lxml_html_clean
 mwparserfromhell==0.6.5
 nemo_toolkit[nlp]>=1.23.0
-numpy<2
```
ryantwolf, line 17: Do you just want to have `numpy` without any pin instead of removing it entirely?

ayushdg (changes requested):
We should also hold off on merging until we can run some gpu tests locally + Nemo ci (if it has any tests using spacy features). Just to ensure nothing on the gpu+ space end changed.

ayushdg (approved):
Is the plan to add an upper pin to spacy for now and investigate removing the numpy pin completely in a followup?

PR #275, opened 2024-10-01
Fix enabling spilling by enabling it on client process

VibhuJawa:
In RAPIDS version 24.08, there was a bug that has been fixed in version 24.10 (see PR: [rapidsai/dask-cuda#1369](https://github.com/rapidsai/dask-cuda/pull/1369)). This PR temporarily enables compatibility to allow us to run on 24.08.

We should remove this when we pin to 24.10 rapids


praateekmahajan (approved):
LGTM!

ayushdg (approved):

PR #285, opened 2024-10-08
Add blocksize to `DocumentDataset.read_*` that uses `dask_cudf.read_*`

praateekmahajan:
This PR introduces a new codepath that leverages `dask_cudf.read_json / read_parquet` directly rather than our existing `from_map` implementation. 

> [!IMPORTANT]  
> The newer implementation supports fewer use-cases compared to original, however it supports the most frequently used implementations (i.e jsonl and parquet files without add_filename)

## New code gotchya's
1. `blocksize` > `filesize` doesn't work when `backend='cpu'` and `filetype='jsonl'` since pandas doesn't support reading multiple files together.
2. If underlying data has inconsistent schema / metadata the read might fail. See https://github.com/dask/dask/issues/11595

## Existing (fpp implementtion) unexpected behavior
As we added more tests for existing code, uncovered a few cases where it doesn't work as expected
1. `filename` column can't be selected when `backend='pandas'`, `filetype='parquet'` and `add_columns=True`
3. Specifying `input_meta` with `pandas` `jsonl` didn't select only subset of columns. That behavior has been fixed.

## Differences between Old vs New
| Discussion Points | New Implementation | Existing Implementation |
|---|------|---------|
| Underlying API | `dask_cudf.read_*` | `dd.from_map(read_single_partition)` |
| backend | Works with pandas / cudf | Works with pandas / cudf |
| filetype | Only supports jsonl and parquet | Supports **`json`**, along with `jsonl` and `parquet`
| add_filename | Only when filetype is `jsonl`  | Supports all filetypes |
| input_meta | Returns only columns in `input_meta.keys()` | Returns only columns in `input_meta.keys()` |
| meta as `**kwarg` | Not required as the first file is used to parse the schema | Is required otherwise can result in OOM (see benchmark row 2 below) | 
| Inconsitent Schema | Doesn't always work (see https://github.com/dask/dask/issues/11595) | Is supported |

## Follow up Issues
1. https://github.com/NVIDIA/NeMo-Curator/issues/427
2. https://github.com/dask/dask/issues/11595
3. https://github.com/dask/dask/issues/6575
4. https://github.com/rapidsai/cudf/pull/17554/files

## Benchmarking
Reading 6000 files of ~25mb each, i.e ~145gb over 8GPUs

| add_filename | partition_size | input_meta | Using `dask.read_json` #285 | Providing meta in <br> `dask.from_map` #291 |
|--------|--------|--------|--------|---------|
| False | 2gb | Specified | 24.9 s ± 330 ms | 25.9 s ± 520 ms 
| False | 2gb | None | 24.9 s ± 470 ms  |  OOM
| True | 2gb | Specified | 55 s ± 177 ms  | 53.2 s ± 350 ms per loop
| True | 2gb | None | 54.8 s ± 248 ms  | 64s ± 289 ms per loop

| Using dask.read_json #285 | Providing meta in dask.from_map #291 |
|-----|-----|
image | image |
|First two are `add_filename=False`, latter two are `True` where  we see a lower utilization  | The first one is `add_filename=False`, and the latter are `True` where we see a lower utilization

## Usage

```python
```
## Checklist


nemo_curator/utils/distributed_utils.py
```diff
@@ -261,6 +261,20 @@ def _set_torch_to_use_rmm():
     torch.cuda.memory.change_current_allocator(rmm_torch_allocator)
 
 
+def select_and_sort_columns(
+    df: Union[dd.DataFrame, dask_cudf.DataFrame],
+    columns: List[str],
+    add_filename: bool,
+) -> Union[dd.DataFrame, dask_cudf.DataFrame]:
+    # TODO : Reviewer TAL if filetype check is needed
```
sarahyurick, line 269: Originally I added the "json" and "jsonl" check because `columns` is already added as a kwarg to `read_parquet`, so I didn't want to filter twice in the Parquet case.

nemo_curator/datasets/doc_dataset.py
```diff
@@ -43,7 +43,8 @@ def read_json(
         cls,
         input_files: Union[str, List[str]],
         backend: str = "pandas",
-        files_per_partition: int = 1,
+        files_per_partition: Optional[int] = None,
+        blocksize: Optional[str] = "1gb",
```
sarahyurick, line 47: As you mentioned it might be nice to add some of the information from https://github.com/rapidsai/cudf/blob/81cd4a00f8ae0dcde359ac11b53a9b3d855e50e2/docs/dask_cudf/source/best_practices.rst#use-parquet to NeMo Curator's best practices.

Also might be of interest for you: https://github.com/rapidsai/cudf/pull/17250

nemo_curator/utils/distributed_utils.py
praateekmahajan: That makes sense. It was originally missing `pickle` but I didn't realize about column selection already for `parquet`. I'll exclude `parquet` in that case. WDYT?

sarahyurick: IIRC `pickle` does not go through this file path, only JSON and Parquet.

nemo_curator/utils/distributed_utils.py
```diff
@@ -343,13 +357,81 @@ def read_single_partition(
...
-        if add_filename and "filename" not in columns:
-            columns.append("filename")
-        df = df[columns]
+    print(f"Reading with {read_kwargs=}", flush=True)
+    return select_and_sort_columns(df, columns, add_filename)
 
-    df = df[sorted(df.columns)]
-    return df
+
+def read_data_cudf_blocksize(
```
sarahyurick, line 364: Can we not do it for the CPU case too? CPU Dask has `blocksize` too.

nemo_curator/utils/distributed_utils.py
```diff
@@ -343,13 +357,81 @@ def read_single_partition(
...
+                if isinstance(input_meta, str)
+                else input_meta
+            )
+        if add_filename:
+            read_kwargs["include_path_column"] = add_filename
+
+    elif file_type == "parquet":
+        if add_filename:
+            msg = "add_filename and blocksize cannot be set at the same time for parquet files"
+            raise ValueError(msg)
```
sarahyurick, line 392: I think it would be nice to look further into Parquet support for `include_path_column`. In the meantime, is there any way we could support `add_filename` too? It's an unfortunate trade-off here otherwise.

praateekmahajan: Yea, I've raised it with the `dask_cudf` team, once we get consensus we can create a feature request to support  https://github.com/dask/dask/pull/9441

nemo_curator/utils/distributed_utils.py
praateekmahajan: Oh I just noticed that `read_pandas_pickle` already handles the `column` case, but small nit that it doesn't sort the columns the same way as in for json / parquet path.

praateekmahajan: Lmk if we need the sorting behavior, if so we'll need a variation of `select_and_sort_columns` to be called after our `read_*` method

sarahyurick: IMO we don't have to worry about the Pickle case, thanks.

docs/user-guide/bestpractices.rst
```diff
@@ -24,6 +24,17 @@ Handling GPU Out-of-Memory (OOM) Errors
 NeMo Curator is designed to be scalable with large amounts of text data, but OOM errors occur when the available GPU memory is insufficient for a given task.
 To help avoid these issues and ensure efficient processing, here are some strategies for managing memory usage and mitigating OOM challenges.
 
+Controlling Partition Sizes
```
praateekmahajan, line 27: @sarahyurick added here. Let me know if we should add something specific from https://github.com/rapidsai/cudf/blob/81cd4a00f8ae0dcde359ac11b53a9b3d855e50e2/docs/dask_cudf/source/best_practices.rst#use-parquet

nemo_curator/utils/distributed_utils.py
```diff
@@ -267,6 +269,23 @@ def _set_torch_to_use_rmm():
     torch.cuda.memory.change_current_allocator(rmm_torch_allocator)
 
 
+def select_columns(
+    df: Union[dd.DataFrame, dask_cudf.DataFrame],
+    columns: List[str],
+    filetype: Literal["jsonl", "json", "parquet"],
+    add_filename: bool,
+) -> Union[dd.DataFrame, dask_cudf.DataFrame]:
+    # We exclude parquet because the parquet readers already support column selection
```
praateekmahajan, line 278: @sarahyurick I'd love a third eye here, as the tests `test_read_data_select_columns` are failing when using `read_data_fpp`

(When add_filename=True)

tests/test_read_data.py
```diff
@@ -0,0 +1,397 @@
...
+@pytest.fixture
+def mock_multiple_jsonl_files(tmp_path):
+    file_paths = []
+    for file_id in range(NUM_FILES):
+        jsonl_file = tmp_path / f"test_{file_id}.jsonl"
+        with open(jsonl_file, "w") as f:
+            for record_id in range(NUM_RECORDS):
+                # 100 rows are ~5kb
+                f.write(
+                    f'{{"id": "id_{file_id}_{record_id}", "text": "A longish string {file_id}_{record_id}"}}\n'
```
praateekmahajan, line 21: Observed an interesting thing here with pandas when `"id" : f"{file_id}_{record_id}"` then `pd.read_json(.., lines=True).dtypes["id"]` returns `int`

tests/test_read_data.py
```diff
@@ -0,0 +1,397 @@
...
+@pytest.mark.parametrize(
+    "backend,file_type",
+    [
+        pytest.param("cudf", "jsonl", marks=pytest.mark.gpu),
+        pytest.param("cudf", "parquet", marks=pytest.mark.gpu),
+        ("pandas", "jsonl"),
+        pytest.param(
+            "pandas",
+            "parquet",
+            marks=pytest.mark.xfail(
```
praateekmahajan, line 264: Marked it as `xfail` but this is something that we should look at. I would imagine that the errors in `test_read_data_select_columns` are related, but I could be wrong (or just tired looking at the same errors)

nemo_curator/utils/distributed_utils.py
```diff
@@ -267,6 +269,23 @@ def _set_torch_to_use_rmm():
...
+    if filetype in ["jsonl", "json"] and columns is not None:
+        if add_filename and "filename" not in columns:
+            columns.append("filename")
+        df = df[columns]
+
+    df = df[sorted(df.columns)]
```
praateekmahajan, line 284: The sorting also isn't respected under the `read_files_fpp` path. Removing the sorting still results in `test_read_data_select_columns` failures but far fewer than when we sort.

tests/test_read_data.py
```diff
@@ -0,0 +1,397 @@
...
+            ("parquet", False, func)
+            for func in ["read_data_blocksize", "read_data_fpp"]
+        ],
+        *[("parquet", True, "read_data_fpp")],
+    ],
+)
+@pytest.mark.parametrize(
+    "cols_to_select", [None, ["id"], ["text", "id"], ["id", "text"]]
+)
+def test_read_data_select_columns(
```
praateekmahajan, line 322: These are the tests that are failing right now

tests/test_read_data.py
```diff
@@ -0,0 +1,397 @@
...
+    df = func(
+        input_files=mock_multiple_jsonl_files,
+        backend=backend,
+        file_type="jsonl",
+        add_filename=False,
+        input_meta=input_meta,
+        columns=None,
+        **read_kwargs,
+    )
+
```
praateekmahajan, line 391: Please TAL here as the behavior of when we pass `input_meta` results in different outputs depending on `fpp` vs `blocksize`

nemo_curator/utils/distributed_utils.py
```diff
@@ -349,13 +368,99 @@ def read_single_partition(
...
+                "Use files_per_partition if blocksize exceeds average file size"
+            )
+        read_func = dd.read_json
+        read_kwargs["lines"] = True
+        if input_meta is not None:
+            if backend == "cudf":
+                # To save GPU memory, we prune columns while reading, and keep only those that are
+                # specified in the input_meta
+                read_kwargs["prune_columns"] = True
+
```
VibhuJawa, line 401: Can you just add a note about this upstream.

nemo_curator/datasets/doc_dataset.py
```diff
@@ -229,8 +229,9 @@ def _read_json_or_parquet(
     input_files: Union[str, List[str]],
```
VibhuJawa, line 229: @rjzamora , Curious if you have thoughts around `_read_json_or_parquet` . 

Now that we control the IO code path directly in `dask_cudf`, i am curious if we can just offload the details of this function there directly than doing it here.

VibhuJawa (commented):
Overall the implementation looks good to me but i am wondering if we should just try to upstream a bunch of the logic in this and related PRs to `dask_cudf` now that we control that code path more cleanly.

docs/user-guide/bestpractices.rst
```diff
@@ -24,6 +24,17 @@ Handling GPU Out-of-Memory (OOM) Errors
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+You should consider using ``files_per_partition`` or ``blocksize`` when reading data. This can help reduce the memory load by processing large datasets in smaller chunks.
+
+#. ``blocksize`` argument is available for ``jsonl`` and ``parquet`` files. But for `parquet` files it's only available when ``add_filename=False``
```
VibhuJawa, line 32: @praateekmahajan , Have we filed issues to `cuDF` to support these features long term. 

1. Supporting `add_file_name` to `cuDF` with multiple files (to decrease that perf delta)
2. Blocksize+`add_file_name` argument for `cuDF`.

docs/user-guide/bestpractices.rst
```diff
@@ -24,6 +24,17 @@ Handling GPU Out-of-Memory (OOM) Errors
...
+
+#. For ``blocksize``, the recommendation is to use 1/32 of the total GPU memory. For example, if you have a GPU with 32GB of memory, you can set the blocksize to ``1gb``.
+
```
VibhuJawa, line 35: Now that https://github.com/rapidsai/cudf/pull/17250/files should we also note that by default we will move to `1/32` by default if nothing is provided. 

What do you think ?

docs/user-guide/bestpractices.rst
praateekmahajan: Discussed offline, i think there is confusion, we need to resurrect the issue https://github.com/dask/dask/issues/6575

docs/user-guide/bestpractices.rst
praateekmahajan: Discussed offline, we can't support 1/32 as an input value because `dask_cudf.read_parquet` supports it but not `read_json` (or either of cpu implementations)

tests/test_read_data.py
sarahyurick: Okay, I think I understand what is being demonstrated here.

Even though Pandas `read_data_fpp` and Pandas/cuDF `read_data_blocksize` cannot prune columns while reading, can't we still have it prune the columns after reading? I know that doesn't save on I/O but then at least the user is getting the resulting DataFrame that they would expect.

tests/test_read_data.py
```diff
@@ -0,0 +1,396 @@
...
+@pytest.mark.parametrize(
+    "backend",
+    [
+        "pandas",
+        pytest.param("cudf", marks=pytest.mark.gpu),
+    ],
+)
+@pytest.mark.parametrize(
+    "file_type,add_filename,function_name",
```
sarahyurick, line 307: Instead of `function_name` I would make this a bool and call it `read_with_blocksize` or `read_with_files_per_partition`.

nemo_curator/utils/distributed_utils.py
```diff
@@ -349,13 +368,99 @@ def read_single_partition(
...
+        raise ValueError(msg)
+
+    with dask.config.set({"dataframe.backend": backend}):
+        df = read_func(input_files, blocksize=blocksize, **read_kwargs, **kwargs)
+        if postprocessing_func is not None:
+            df = postprocessing_func(df)
+        return select_columns(df, columns, file_type, add_filename)
+
+
+def read_data_fpp(
```
sarahyurick, line 434:
```suggestion
def read_data_files_per_partition(
```
I would really prefer to replace anywhere that says "fpp" with the full "files_per_partition".

nemo_curator/utils/distributed_utils.py
sarahyurick: Wanted to clarify what the goal is here: if `add_filename=True`, then a "filename" column is added to the DataFrame. Then the user prunes to only include `columns` (and the user's `columns` does not "filename"). Is the desired behavior to include the "filename" column anyway or not?

nemo_curator/utils/distributed_utils.py
```diff
@@ -267,6 +269,23 @@ def _set_torch_to_use_rmm():
...
 
 
+def select_columns(
+    df: Union[dd.DataFrame, dask_cudf.DataFrame],
+    columns: List[str],
+    filetype: Literal["jsonl", "json", "parquet"],
+    add_filename: bool,
+) -> Union[dd.DataFrame, dask_cudf.DataFrame]:
+    # We exclude parquet because the parquet readers already support column selection
+    if filetype in ["jsonl", "json"] and columns is not None:
```
sarahyurick, line 279: Your comment in the README which says:

"`df.columns` does not show `filename` when `columns` are provided, `filetype="parquet"`, and `add_filename=True`"

is because of this?

tests/test_read_data.py
sarahyurick: This is demonstrating the note in the README:
"`filename` column cannot be selected when `backend="pandas"`, `filetype="parquet"` and `add_columns=True`"
?

This is saying that the "filename" column does not get added when `backend="pandas"` and `file_type="parquet"`? Or are you just saying it gets dropped by the `select_columns` function?

docs/user-guide/bestpractices.rst
sarahyurick: Thanks, maybe it would also make since to link to the Dask cuDF documentation directly for more information.

nemo_curator/utils/distributed_utils.py
```diff
@@ -388,7 +493,8 @@ def read_data(
     input_files: Union[str, List[str]],
     file_type: str = "pickle",
     backend: Literal["cudf", "pandas"] = "cudf",
-    files_per_partition: int = 1,
+    blocksize: Optional[str] = None,
+    files_per_partition: Optional[int] = 1,
     add_filename: bool = False,
     input_meta: Union[str, dict] = None,
     columns: Optional[List[str]] = None,
```
sarahyurick, line 500: Can you elaborate a bit more on a comment from the README: a slowdown when `columns` are selected and the file type is JSON?

sarahyurick (changes requested):
Sorry for the delay in getting back to this. Added some questions and suggestions, thanks.

nemo_curator/utils/distributed_utils.py
praateekmahajan: The desired behavior would be to include `filename`, as the user might just do `add_filename=True` along with set of columns that doesn't have `filename` as part. 

Assuming df has `id` and `text`
| add_filename | columns | output |
|-----------|--------|--------------|
False | None | ['id', 'text'] |
True | None | ['filename', 'id', 'text'] |
False | ['id'] | ['id']
True | ['id'] | ['filename', 'id'] |
False | ['filename', 'id'] | KeyError
True | ['filename', 'id'] | ['filename', 'id']

nemo_curator/utils/distributed_utils.py
praateekmahajan: No, it was a very interesting issue now that we figured thanks to @rjzamora 

> You are basically running into a case where from_map is trying (but failing) to be "smart". When the callable argument to from_map has a columns= argument in its signature, we end up with a [FromMapProjectable](https://github.com/dask/dask-expr/blob/77d0f895238b4c90063e40b75ffb0e45564afd7c/dask_expr/io/io.py#L297)  expression. This is because dask-expr assumes that having this argument means that your operation supports column projection (and that passing columns=["text", "id"] means that the output of that operation must contain those exact columns.
If you don't want this behavior, you should call the argument something else. e.g.:
def read_single_partition_pq(files: List[str], io_columns: Optional[List[str]] = None):

nemo_curator/utils/distributed_utils.py
praateekmahajan: I haven't benchmarked it, so I can remove that comment, but I remember long ago seeing a noticeable slowdown when you pass in `columns`. But memory is fuzzy now, so I'll remove it, just something to keep in mind.

tests/test_read_data.py
praateekmahajan: `df.columns` doesn't return `filename` as a column, and I believe that's because `filename` is a reserved field in the pyarrow schema. However if we were to change our `filename` field to `file_name` / `path` things would work as expected. So if it sounds good I'll create a followup issue and do the renaming in the next PR (and also remove the xfail)

tests/test_read_data.py
praateekmahajan: I think in pytest when we use bools, it becomes harder to understand which test has failed. For instance here a test will be read as `cudf-parquet-True-read_data_fpp` , and the bool for `add_filename` is already undescriptive. Renaming even `function_name` to blocksize would be even harder to parse `cudf-parquet-True-True`. It's a small nuance, but I would prefer keeping it the same way. If you feel strongly I can change, but I thought I'll share the intent

tests/test_read_data.py
```diff
@@ -0,0 +1,396 @@
...
+
+
+@pytest.mark.parametrize(
+    "backend",
+    [
+        "pandas",
+        pytest.param("cudf", marks=pytest.mark.gpu),
+    ],
+)
+@pytest.mark.parametrize("function_name", ["read_data_blocksize", "read_data_fpp"])
```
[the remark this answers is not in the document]
praateekmahajan: Same as above, however I could change it to

```python

@paremetrize("read_kwargs", [
   {"files_per_partition": 1, "blocksize" : None}, 
   {"files_per_partition": None, "blocksize" : "128MiB"}
])
```

And then below instead of `func` I can call `read_data`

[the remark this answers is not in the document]
praateekmahajan: But then the naming is still slightly ugly i.e `pandas-read_kwargs-1-input_meta-2` rather than `pandas-read_data_blocksize-input_meta-2`

tests/test_read_data.py
praateekmahajan: That's a good point. Would you want to do a `select_columns([input_meta.keys()])` when input_type=='jsonl' (since input_meta in it's current setting only supports `json`)? We can do that, though I worry that it'll become convolved along with `columns` (and maybe even `add_filename`)

tests/test_read_data.py
```diff
@@ -0,0 +1,544 @@
...
+        ],
+    ],
+)
+def test_read_data_different_columns(
+    mock_multiple_jsonl_files_different_cols,
+    mock_multiple_parquet_files_different_cols,
+    backend,
+    file_type,
+    read_kwargs,
+):
```
praateekmahajan, line 519: In this test case we try to read json / parquet directories which have some files containing `id / text` columns while some containing `text / meta / id` columns (we try to mimic RPV1 dataset). 

We try reading with both APIs of blocksize and FPP. 

Our goal is to read only `id` and `text` fields since those are common across all files. However we notice that 3/4 of the blocksize test cases ie `read_kwargs-5/6/7` fail (except when backend=='pandas' and filetype=='parquet') and I suspect that's because `dd.read_json` / `dd.read_parquet` doesn't nicely seem to support mismatched schemas.

Would be amazing if I can get a third eye here since that's the only case left for this PR. Worst case we can mark it as `xfail` until `dd` can support mismatched schemas (in which case for RPV1 like datasets we'll have to go down the RPV1 path :( )

tests/test_read_data.py
sarahyurick: I think as long as we are thoroughly checking all the conditions we expect to have to do this, it should be okay. Happy to discuss this offline if anything is unclear here.

tests/test_read_data.py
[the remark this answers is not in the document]
sarahyurick: Oh I see, in that case yeah it's up to you. Thanks.

tests/test_read_data.py
sarahyurick: Fair enough, feel free to keep as is. Thanks.

nemo_curator/utils/distributed_utils.py
sarahyurick: Sounds good, thank you.

docs/user-guide/bestpractices.rst
```diff
@@ -24,6 +24,17 @@ Handling GPU Out-of-Memory (OOM) Errors
 NeMo Curator is designed to be scalable with large amounts of text data, but OOM errors occur when the available GPU memory is insufficient for a given task.
 To help avoid these issues and ensure efficient processing, here are some strategies for managing memory usage and mitigating OOM challenges.
 
+Controlling Partition Sizes
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+You should consider using ``files_per_partition`` or ``blocksize`` when reading data. This can help reduce the memory load by processing large datasets in smaller chunks.
```
sarahyurick, line 30:
```suggestion
The user should consider using ``files_per_partition`` or ``blocksize`` when reading data. This can help reduce the memory load by processing large datasets in smaller chunks.
```

docs/user-guide/bestpractices.rst
sarahyurick, line 32:
```suggestion
#. The ``blocksize`` argument is available for ``jsonl`` and ``parquet`` files. However, for `parquet` files, it is currently only available when ``add_filename=False``.
```

docs/user-guide/bestpractices.rst
```diff
@@ -24,6 +24,17 @@ Handling GPU Out-of-Memory (OOM) Errors
...
+
+#. ``blocksize`` argument is available for ``jsonl`` and ``parquet`` files. But for `parquet` files it's only available when ``add_filename=False``
+
+#. For ``blocksize``, the recommendation is to use 1/32 of the total GPU memory. For example, if you have a GPU with 32GB of memory, you can set the blocksize to ``1gb``.
```
sarahyurick, line 34:
```suggestion
#. For the ``blocksize`` parameter, the recommendation is to use 1/32 of the total GPU memory. For example, if you have a GPU with 32GB of memory, you can set ``blocksize="1GB"``.
```

tests/test_read_data.py
```diff
@@ -0,0 +1,544 @@
...
+        input_meta=None,
+        columns=list(cols_to_select) if cols_to_select else None,
+        **read_kwargs,
+    )
+    if not cols_to_select:
+        cols_to_select = ["id", "text"]
+
+    assert list(df.columns) == list(df.head().columns)
+    if not add_filename:
+        # assert list(df.columns) == sorted(cols_to_select)
```
sarahyurick, line 446:
```suggestion
```

tests/test_read_data.py
```diff
@@ -0,0 +1,544 @@
...
+        assert list(df.columns) == sorted(cols_to_select)
+    else:
+        # assert list(df.columns) == sorted(cols_to_select + ["filename"])
```
sarahyurick, line 449:
```suggestion
```

tests/test_read_data.py
```diff
@@ -0,0 +1,544 @@
...
+    file_type,
+    read_kwargs,
+):
+
+    read_kwargs_cp = read_kwargs.copy()
+    read_kwargs_cp["columns"] = ["id", "text"]
+    if file_type == "jsonl":
+        input_files = mock_multiple_jsonl_files_different_cols
+        read_kwargs_cp["input_meta"] = {"id": "str", "text": "str"}
+        # read_kwargs_cp["meta"] = {"id": "str", "text": "str"}
```
sarahyurick, line 526:
```suggestion
```

tests/test_read_data.py
```diff
@@ -0,0 +1,544 @@
...
+    [
+        pytest.param("cudf", "jsonl", marks=pytest.mark.gpu),
+        pytest.param("cudf", "parquet", marks=pytest.mark.gpu),
+        ("pandas", "jsonl"),
+        pytest.param(
+            "pandas",
+            "parquet",
+            # marks=pytest.mark.xfail(
+            #     reason="filename column inaccessible with pandas backend and parquet"
+            # ),
```
sarahyurick, line 348:
```suggestion
```

sarahyurick (changes requested):
I think I am still getting lost in some of the ongoing work that will be delegated to other PRs. After this is merged, can you please open issues for each use case, along with MREs?

Can you also confirm if there is any functionality that currently works with main that will be broken by this PR? It sounds like not(?) but want to confirm before merging with the release happening soon.

tests/test_read_data.py
praateekmahajan: @sarahyurick I made the changes in the last commit however the `test_io` now fails due to since the test does

```python
            dataset = DocumentDataset.read_json(
                temp_file.name, input_meta='{"id": "float"}'
            )

        output_meta = str({col: str(dtype) for col, dtype in dataset.df.dtypes.items()})

        expected_meta = (
            "{'date': 'datetime64[ns, UTC]', 'id': 'float64', 'text': 'object'}"
        )
```

I was going to change it but just wanted to confirm if it's okay. FWIW, that test only passes in pandas backend case, but would've failed if backend would've been cudf.

nemo_curator/utils/distributed_utils.py
```diff
@@ -313,9 +332,13 @@ def read_single_partition(
             read_kwargs["dtype"] = (
                 ast.literal_eval(input_meta) if type(input_meta) == str else input_meta
             )
+            # because pandas doesn't support `prune_columns`, it'll always return all columns even when input_meta is specified
+            # to maintain consisntency we explicitly set `io_columns` here
```
sarahyurick, line 336:
```suggestion
            # because pandas doesn't support `prune_columns`, it'll always return all columns even when input_meta is specified
            # to maintain consistency we explicitly set `io_columns` here
```

nemo_curator/utils/distributed_utils.py
```diff
@@ -344,18 +367,133 @@ def read_single_partition(
...
+    **kwargs,
+) -> dd.DataFrame:
+
+    read_kwargs = dict()
+
+    postprocessing_func: Optional[Callable[[dd.DataFrame], dd.DataFrame]] = None
+    if file_type == "jsonl":
+        warnings.warn(
+            "If underlying jsonl data doesn't have consistent schema, reading with blocksize will fail."
+            "Use files_per_partition approach."
```
sarahyurick, line 397:
```suggestion
            "If underlying JSONL data does not have a consistent schema, reading with blocksize will fail. "
            "Please use files_per_partition approach instead."
```

nemo_curator/utils/distributed_utils.py
```diff
@@ -344,18 +367,133 @@ def read_single_partition(
...
+        )
+
+        if backend == "panads":
+            warnings.warn(
+                "Pandas backend with blocksize cannot read multiple JSONL files into a single partition. "
+                "Use files_per_partition if blocksize exceeds average file size"
```
sarahyurick, line 403:
```suggestion
        if backend == "pandas":
            warnings.warn(
                "Pandas backend with blocksize cannot read multiple JSONL files into a single partition. "
                "Please use files_per_partition if blocksize exceeds average file size."
```

nemo_curator/utils/distributed_utils.py
```diff
@@ -344,18 +367,133 @@ def read_single_partition(
...
+
+            read_kwargs["include_path_column"] = add_filename
+            read_kwargs["path_converter"] = extract_filename
+            postprocessing_func = lambda df: df.rename(columns={"path": "filename"})
+
+    elif file_type == "parquet":
+        if backend == "cudf" and not DASK_CUDF_PARQUET_READ_INCONSISTENT_SCHEMA:
+            warnings.warn(
+                "If underlying parquet data doesn't have consistent schema, reading with blocksize will fail."
+                "Update underlying rapids package to 25.02+ or use files_per_partition approach."
```
sarahyurick, line 436:
```suggestion
                "If underlying Parquet data does not have consistent schema, reading with blocksize will fail. "
                "Please update underlying RAPIDS package to version 25.02 or higher, or use files_per_partition approach instead."
```

nemo_curator/utils/distributed_utils.py
```diff
@@ -344,18 +367,133 @@ def read_single_partition(
...
+            )
+        elif backend == "pandas":
+            warnings.warn(
+                "If underlying parquet data doesn't have consistent column order, reading with blocksize might fail."
+                "Use files_per_partition approach."
```
sarahyurick, line 441:
```suggestion
                "If underlying Parquet data does not have a consistent column order, reading with blocksize might fail. "
                "Please use files_per_partition approach instead."
```

nemo_curator/utils/distributed_utils.py
```diff
@@ -344,18 +367,133 @@ def read_single_partition(
...
+            )
+
+        if add_filename:
+            msg = "add_filename and blocksize cannot be set at the same time for parquet files"
```
sarahyurick, line 445:
```suggestion
            msg = "add_filename and blocksize cannot be set at the same time for Parquet files."
```

nemo_curator/utils/distributed_utils.py
```diff
@@ -344,18 +367,133 @@ def read_single_partition(
...
+            raise ValueError(msg)
+        read_func = dd.read_parquet
+        read_kwargs["columns"] = columns
+        # In dask_cudf >= 24.12, aggregate_files is not required, but we've kept here until
+        # it gets in dask (pandas) as well
+        read_kwargs["aggregate_files"] = True
+    else:
+        msg = f"Reading with blocksize is only supported for jsonl and parquet files, not {file_type=}"
```
sarahyurick, line 453:
```suggestion
        msg = f"Reading with blocksize is only supported for JSONL and Parquet files, not {file_type=}"
```

sarahyurick (changes requested):
Added a couple more nits, then should be good on my end. Thanks!

tests/test_read_data.py
```diff
@@ -0,0 +1,586 @@
...
+    "backend",
+    [
+        "pandas",
+        pytest.param("cudf", marks=pytest.mark.gpu),
+    ],
+)
+def test_read_data_blocksize_add_filename_parquet(mock_multiple_parquet_files, backend):
+    with pytest.raises(
+        ValueError,
+        match="add_filename and blocksize cannot be set at the same time for parquet files",
```
sarahyurick, line 324:
```suggestion
        match="add_filename and blocksize cannot be set at the same time for Parquet files",
```

sarahyurick (approved):
https://github.com/NVIDIA/NeMo-Curator/pull/434 should fix the gpuCI build errors, then we can make sure all GPU tests pass, then merge!

nemo_curator/datasets/doc_dataset.py
VibhuJawa: Discussed offline

PR #302, opened 2024-10-15
[REVIEW] Speedup Connected Components

VibhuJawa:
This pull request includes several changes to the `nemo_curator/modules/fuzzy_dedup.py` file, focusing on removing the `convert_str_ids` functionality, optimizing performance, and improving logging. 

The most important changes are:

#### Removal of `convert_str_ids` functionality:

* Removed the `convert_str_ids` parameter and its associated logic from the `__init__` method and other methods in `nemo_curator/modules/fuzzy_dedup.py`. [[1]](diffhunk://#diff-b836d89c298548dd47480ece278782f030d04e2beb421c59c0e546718aede6c3L472) [[2]](diffhunk://#diff-b836d89c298548dd47480ece278782f030d04e2beb421c59c0e546718aede6c3L1408) [[3]](diffhunk://#diff-b836d89c298548dd47480ece278782f030d04e2beb421c59c0e546718aede6c3L1418) [[4]](diffhunk://#diff-b836d89c298548dd47480ece278782f030d04e2beb421c59c0e546718aede6c3L1482-R1483) [[5]](diffhunk://#diff-b836d89c298548dd47480ece278782f030d04e2beb421c59c0e546718aede6c3L1592-L1607) [[6]](diffhunk://#diff-b836d89c298548dd47480ece278782f030d04e2beb421c59c0e546718aede6c3L1633-R1650)

This is done because now we have longstrings support in `cuDF` so we no longer need to convert string to int ids

#### Performance optimizations:

* Decreased the block size for reading parquet files in `_write_dedup_parsed_id`  [[1]](diffhunk://#diff-b836d89c298548dd47480ece278782f030d04e2beb421c59c0e546718aede6c3L1592-L1607) to a lesser value to allow scaling of drop_duplicates (which has a big memory overhead 16x+ ) to prevent OOMs, this will allow us to run CC at larger scales without requiring more hardware.

* Increased the   chuck size in `_write_encoded_jaccard_pair` methods to improve merge performance, as with large base chunks, we have bigger transfers so the throughput of transfer is better on TCP  [[2]](diffhunk://#diff-b836d89c298548dd47480ece278782f030d04e2beb421c59c0e546718aede6c3L1633-R1650) 

* Updated the `_run_connected_components` method to initialize `Comms` with `p2p=False` 

Merge Improvements: 
  - This PR optimizes the merge process by using an index-based approach instead of the previous batched method, while maintaining the broadcast merge.
  - The new method reduces shuffles to 2*num_batches - 1 through indexing. 
  - The only additional operation is setting the index on the `ddf_id` column.

Main: 22m 10s 
PR: 444.85 s 

image

Dask Profiles:
[cc_profiles.zip](https://github.com/user-attachments/files/17374101/cc_profiles.zip)

#### Logging improvements:

* Added start time logging in the `cc_workflow` method and end-to-end time logging for the workflow. [[1]](diffhunk://#diff-b836d89c298548dd47480ece278782f030d04e2beb421c59c0e546718aede6c3R1428) [[2]](diffhunk://#diff-b836d89c298548dd47480ece278782f030d04e2beb421c59c0e546718aede6c3R1439)

**Verify Equal Results:**

```python3
ddf_1 = dask_cudf.read_parquet("/raid/vjawa/rpv2_debug_cache_pull_302/connected_components.parquet")
ddf_1 = ddf_1.repartition(npartitions=4).sort_values(by=['id', 'group'])
len(ddf_1)
```
376321911

```python3
ddf_2 = dask_cudf.read_parquet("/raid/vjawa/rpv2_debug_cache/connected_components.parquet")
ddf_2 = ddf_2.repartition(npartitions=4).sort_values(by=['id', 'group'])

len(ddf_2)
```
376321911

#### Check same ids
```python3
merged_df = ddf_1[['id']].merge(ddf_2[['id']], on='id', how='inner')
len(merged_df)
```
376321911

CC: @ayushdg 

## Checklist

### DIFF
--- a/nemo_curator/modules/fuzzy_dedup.py
+++ b/nemo_curator/modules/fuzzy_dedup.py
@@ -469,7 +469,6 @@ class FuzzyDuplicates:
             cache_dir=self.config.cache_dir,
             jaccard_pairs_path=os.path.join(self.config.cache_dir, jaccard_pairs_fname),
             id_column=self.config.id_field,
-            convert_str_ids=False,
             jaccard_threshold=self.config.jaccard_threshold,
             logger=self._logger,
             profile_dir=self.config.profile_dir,
@@ -1405,7 +1404,6 @@ class ConnectedComponents:
         cache_dir: str,
         jaccard_pairs_path: str,
         id_column="id",
-        convert_str_ids=False,
         jaccard_threshold: float = 0.8,
         logger: Union[logging.LoggerAdapter, str] = "./",
         profile_dir: Optional[str] = None,
@@ -1415,7 +1413,6 @@ class ConnectedComponents:
         self.id_column = id_column
         self.left_id = f"{id_column}_x"
         self.right_id = f"{id_column}_y"
-        self.convert_str_ids = convert_str_ids
         self.jaccard_threshold = jaccard_threshold
         self.profile_dir = profile_dir
         if isinstance(logger, str):
@@ -1593,18 +1590,6 @@ class ConnectedComponents:
                 aggregate_files=True,
             )
             id_columns = [self.id_column]
-            if self.convert_str_ids:
-                ddf = ddf.map_partitions(
-                    self._convert_str_id_pair_to_int,
-                    meta={
-                        "dataset_id_x": "uint32",
-                        "doc_id_x": "int64",
-                        "dataset_id_y": "uint32",
-                        "doc_id_y": "int64",
-                    },
-                )
-                id_columns = ["dataset_id", "doc_id"]
-
             unique_docs = ddf.map_partitions(
                 ConnectedComponents._get_unique_ids_per_partition, id_columns=id_columns
             )
@@ -1630,32 +1615,16 @@ class ConnectedComponents:
             ddf_id = dask_cudf.read_parquet(
                 dedup_parsed_id_path, blocksize="2GB", aggregate_files=True
             )
-            ddf_id = ddf_id.persist()
-            len(ddf_id)
             ddf = dask_cudf.read_parquet(
                 self.jaccard_pairs_path,
                 blocksize="1GB",
                 aggregate_files=True,
             )
-            id_columns = [self.id_column]
-            if self.convert_str_ids:
-                ddf = ddf.map_partitions(
-                    self._convert_str_id_pair_to_int,
-                    meta={
-                        "jaccard": "float32",
-                        "dataset_id_x": "uint32",
-                        "doc_id_x": "int64",
-                        "dataset_id_y": "uint32",
-                        "doc_id_y": "int64",
-                    },
-                )
-                id_columns = ["dataset_id", "doc_id"]
-
             self._merge_and_write(
                 ddf=ddf,
                 ddf_id=ddf_id,
                 output_path=output_path,
-                id_columns=id_columns,
+                id_column=self.id_column,
             )
         self._logger.info(
             f"Time taken for Encoding Jaccard Pairs = {time.time() - t0}s and output written at {output_path}"
@@ -1667,60 +1636,23 @@ class ConnectedComponents:
         ddf: dask_cudf.DataFrame,
         ddf_id: dask_cudf.DataFrame,
         output_path: str,
-        id_columns: Union[str, List[str]],
+        id_column: str,
     ) -> None:
         st = time.time()
         # Ensure 'id_columns' is a list
-        if isinstance(id_columns, str):
-            id_columns = [id_columns]
-
-        # Check if there are multiple id_columns
-        if len(id_columns) == 1:
-            # Single id_column case
-            ddf_id = ddf_id.set_index(id_columns)
-            for tag in ["x", "y"]:
-                pair_ids = [f"{id_columns[0]}_{tag}"]
-                # Merge 'ddf' with 'ddf_id' to map ids to uids
-                ddf = ddf.merge(
-                    ddf_id,
-                    left_on=pair_ids,
-                    right_index=True,
-                    how="inner",
-                    broadcast=True,
-                )
-                ddf = ddf.drop(columns=pair_ids)
-                ddf = ddf.rename(columns={"uid": f"{self.id_column}_{tag}"})
-        else:
-            # Multiple id_columns case
-            # Combine id_columns into a single string column in 'ddf_id'
-            ddf_id["id_combined"] = ""
-            for id_c in id_columns:
-                ddf_id["id_combined"] = (
-                    ddf_id["id_combined"] + "-" + ddf_id[id_c].astype("str")
-                )
-
-            ddf_id = ddf_id.drop(columns=id_columns)
-            ddf_id = ddf_id.set_index("id_combined")
-            for tag in ["x", "y"]:
-                pair_ids = [f"{id_col}_{tag}" for id_col in id_columns]
-                # Combine pair_ids into a single string column in 'ddf'
-                ddf["pair_id"] = ""
-                for pair_id in pair_ids:
-                    ddf["pair_id"] = ddf["pair_id"] + "-" + ddf[pair_id].astype("str")
-
-                ddf = ddf.drop(columns=pair_ids)
-                # Merge 'ddf' with 'ddf_id' to map ids to uids
-                ddf = ddf.merge(
-                    ddf_id,
-                    left_on="pair_id",
-                    right_index=True,
-                    how="inner",
-                    broadcast=True,
-                )
-                # Drop the used columns
-                ddf = ddf.drop(columns=["pair_id"])
-                ddf = ddf.rename(columns={"uid": f"{self.id_column}_{tag}"})
-        # Select the required columns
+        ddf_id = ddf_id.set_index(id_column)
+        for tag in ["x", "y"]:
+            pair_id = f"{id_column}_{tag}"
+            # Merge 'ddf' with 'ddf_id' to map ids to uids
+            ddf = ddf.merge(
+                ddf_id,
+                left_on=pair_id,
+                right_index=True,
+                how="inner",
+                broadcast=True,
+            )
+            ddf = ddf.drop(columns=pair_id)
+            ddf = ddf.rename(columns={"uid": f"{self.id_column}_{tag}"})
         ddf = ddf[[self.left_id, self.right_id, "jaccard"]]
         ddf.to_parquet(output_path, write_index=False)
 
--- a/nemo_curator/modules/fuzzy_dedup.py
+++ b/nemo_curator/modules/fuzzy_dedup.py
@@ -1674,29 +1674,59 @@ class ConnectedComponents:
         if isinstance(id_columns, str):
             id_columns = [id_columns]
 
-        # Optimize 'ddf_id' by setting 'id_columns' as the index
-        ddf_id = ddf_id.set_index(id_columns)
+        # Check if there are multiple id_columns
+        if len(id_columns) == 1:
+            # Single id_column case
+            ddf_id = ddf_id.set_index(id_columns)
+            for tag in ["x", "y"]:
+                pair_ids = [f"{id_columns[0]}_{tag}"]
+                # Merge 'ddf' with 'ddf_id' to map ids to uids
+                ddf = ddf.merge(
+                    ddf_id,
+                    left_on=pair_ids,
+                    right_index=True,
+                    how="inner",
+                    broadcast=True,
+                )
+                ddf = ddf.drop(columns=pair_ids)
+                ddf = ddf.rename(columns={"uid": f"{self.id_column}_{tag}"})
+        else:
+            # Multiple id_columns case
+            # Combine id_columns into a single string column in 'ddf_id'
+            ddf_id["id_combined"] = ""
+            for id_c in id_columns:
+                ddf_id["id_combined"] = (
+                    ddf_id["id_combined"] + "-" + ddf_id[id_c].astype("str")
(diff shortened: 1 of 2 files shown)


nemo_curator/modules/fuzzy_dedup.py
```diff
@@ -1652,18 +1666,52 @@ def _write_encoded_jaccard_pair(self, dedup_parsed_id_path):
...
+
+            ddf = ddf.drop(columns=pair_ids)
+            ddf = ddf.rename(columns={"uid": f"{self.id_column}_{tag}"})
+        ddf = ddf[[self.left_id, self.right_id, "jaccard"]]
+        ddf.to_parquet(output_path, write_index=False)
+
+        et = time.time()
+        print(f"Merge and write completed in {et - st} seconds", flush=True)
+
     def _batched_merge_and_write(
```
VibhuJawa, line 1715: TODO: Remove this function

nemo_curator/modules/fuzzy_dedup.py
```diff
@@ -1428,16 +1428,30 @@ def __init__(
             self._logger = logger
 
     def cc_workflow(self, output_path):
+        print("Writing deduped parsed id")
+        st = time.time()
         deduped_parsed_id_path = self._write_dedup_parsed_id()
+        print(f"Time taken for writing deduped parsed id = {time.time()-st}s")
```
praateekmahajan, line 1434: We don't need to duplicate these since `_write_dedup_parsed_id` already has the logger. See line 1633

nemo_curator/modules/fuzzy_dedup.py
```diff
@@ -1428,16 +1428,30 @@ def __init__(
...
+        print("Writing encoded jaccard pair")
+        st = time.time()
         encoded_jaccard_pair_path = self._write_encoded_jaccard_pair(
             deduped_parsed_id_path
         )
+        print(f"Time taken for writing encoded jaccard pair = {time.time()-st}s")
```
praateekmahajan, line 1440: Same here, 1677

nemo_curator/modules/fuzzy_dedup.py
```diff
@@ -1428,16 +1428,30 @@ def __init__(
...
+        print("Writing deduped encoded jaccard pair")
+        st = time.time()
         deduped_encoded_jaccard_path = self._write_dedup_encoded_jaccard_pair(
             encoded_jaccard_pair_path
         )
+        print(
```
praateekmahajan, line 1446: Same, 1447

nemo_curator/modules/fuzzy_dedup.py
```diff
@@ -1428,16 +1428,30 @@ def __init__(
...
+            f"Time taken for writing deduped encoded jaccard pair = {time.time()-st}s"
+        )
+        print("Running connected components")
+        st = time.time()
         cc_path = self._run_connected_components(
             deduped_encoded_jaccard_path, deduped_parsed_id_path, output_path
         )
+        print(f"Time taken for running connected components = {time.time()-st}s")
```
praateekmahajan, line 1454: Same at 1518

nemo_curator/modules/fuzzy_dedup.py
```diff
@@ -1652,52 +1666,52 @@ def _write_encoded_jaccard_pair(self, dedup_parsed_id_path):
...
+                broadcast=True,
             )
 
+            ddf = ddf.drop(columns=pair_ids)
+            ddf = ddf.rename(columns={"uid": f"{self.id_column}_{tag}"})
+        ddf = ddf[[self.left_id, self.right_id, "jaccard"]]
+        ddf.to_parquet(output_path, write_index=False)
+
+        et = time.time()
+        print(f"Merge and write completed in {et - st} seconds", flush=True)
```
praateekmahajan, line 1713: Can change this to 
```
self._logger.info(
            f"Time taken for merge and write = {time.time() - t0}s and output written at {output_path}"
        )
```

nemo_curator/modules/fuzzy_dedup.py
```diff
@@ -1652,52 +1666,52 @@ def _write_encoded_jaccard_pair(self, dedup_parsed_id_path):
...
+            pair_ids = [f"{id_col}_{tag}" for id_col in id_columns]
+
+            # Merge 'ddf' with 'ddf_id' to map ids to uids
+            ddf = ddf.merge(
+                ddf_id,
+                left_on=pair_ids,
+                right_index=True,
+                how="inner",
+                # shuffle="p2p",
+                broadcast=True,
```
praateekmahajan, line 1704: Is there a limit / recommendation on broadcast merge, given that now `ddf_id` which I believe is the smaller of the two will be unbounded, whereas in the original implementation it is batched? I worry if it'll OOM at certain thresholds

VibhuJawa: Discussed this with Praateek on a call.  Couple of clarifications: 

1. It was previously also a broadcast merge so nothing changes there. 
2. We have 2*num_batches -1 shuffles with this index once and forget approach 
3. The only extra set is set_index on `ddf_id` column

nemo_curator/modules/fuzzy_dedup.py
```diff
@@ -1634,7 +1635,7 @@ def _write_encoded_jaccard_pair(self, dedup_parsed_id_path):
             len(ddf_id)
             ddf = dask_cudf.read_parquet(
                 self.jaccard_pairs_path,
-                blocksize="256MB",
+                blocksize="1GB",
```
VibhuJawa, line 1638: We can probably scale this up a bit more (I kept it conservative). 

CC: @ayushdg

nemo_curator/modules/fuzzy_dedup.py
```diff
@@ -1651,52 +1651,83 @@ def _write_encoded_jaccard_pair(self, dedup_parsed_id_path):
...
+        output_path: str,
+        id_columns: Union[str, List[str]],
+    ) -> None:
+        st = time.time()
+        # Ensure 'id_columns' is a list
+        if isinstance(id_columns, str):
+            id_columns = [id_columns]
+
+        # Check if there are multiple id_columns
+        if len(id_columns) == 1:
```
VibhuJawa, line 1678: @ayushdg , Do we have use cases of `convert_str_ids = True` or we can just nuke it ?  

I have added support for now but wonder if we should just remove it .

nemo_curator/modules/fuzzy_dedup.py
```diff
@@ -1451,13 +1448,13 @@ def _run_connected_components(
...
 
-            Comms.initialize(p2p=True)
+            Comms.initialize(p2p=False)
             df = dask_cudf.read_parquet(
                 deduped_encoded_jaccard_path, blocksize="1GB", aggregate_files=True
             )
             df = df[df["jaccard"] == 1].reset_index(drop=True)
 
-            labels_df = dask_cudf.read_parquet(deduped_parsed_id_path)
+            labels_df = dask_cudf.read_parquet(deduped_parsed_id_path, blocksize="1GB", aggregate_files=True)
```
VibhuJawa, line 1457: This was causing errors as we had too little files and it some caused unequal distribution of work.  Testing again to verify this is robust

nemo_curator/modules/fuzzy_dedup.py
```diff
@@ -1438,6 +1436,7 @@ def cc_workflow(self, output_path):
         cc_path = self._run_connected_components(
             deduped_encoded_jaccard_path, deduped_parsed_id_path, output_path
         )
+        self._logger.info(f"End to End time in cc_workflow = {time.time() - st}s")
```
praateekmahajan, line 1439: Can this also be in the same `Time taken for end to end cc_workflow was 12.345s`

nemo_curator/modules/fuzzy_dedup.py
```diff
@@ -1589,22 +1580,10 @@ def _write_dedup_parsed_id(self):
             ddf = dask_cudf.read_parquet(
                 self.jaccard_pairs_path,
                 columns=[self.left_id, self.right_id],
-                blocksize="1GB",
+                blocksize="512MB",
```
praateekmahajan, line 1583: Why did we have to change this? Was this causing slowdowns?

nemo_curator/modules/fuzzy_dedup.py
praateekmahajan: Similar to above, can you write why we decrease the above blocksize but increased this?

nemo_curator/modules/fuzzy_dedup.py
```diff
@@ -1479,9 +1480,7 @@ def _run_connected_components(
             labels_df = labels_df.merge(
                 result, left_on=["uid"], right_on=["vertex"], how="inner"
             )
-            id_columns = (
```
praateekmahajan, line 1482: In the PR can you also document that we've removed `self.convert_str_ids` and it's behavior now is that of `False` and why so, will help our future selves.

nemo_curator/modules/fuzzy_dedup.py
VibhuJawa: Yeah so this is the first dedup application before encoding and dedup on strings is a expensive operation (often with a  `16x` overhead in terms of memory). 

While testing this workflow for larger scales (>40 TB) , this is the step that blew up in memory for me . 

Long term we should run `ConnectedComponents` on as few nodes as possible as this is a communication heavy stage for maximum TCO.  This step helps ensure that .

nemo_curator/modules/fuzzy_dedup.py
VibhuJawa: Good call out will change

nemo_curator/modules/fuzzy_dedup.py
VibhuJawa: Removed so resolved

nemo_curator/modules/fuzzy_dedup.py
VibhuJawa: Done

nemo_curator/scripts/fuzzy_deduplication/connected_components.py
```diff
@@ -32,15 +32,14 @@ def main(args):
...
-
     client = get_client(**ArgumentHelper.parse_client_args(args))
 
     components_stage = ConnectedComponents(
         cache_dir=args.cache_dir,
         jaccard_pairs_path=args.jaccard_pairs_path,
         id_column=args.input_json_id_field,
-        convert_str_ids=True,
         jaccard_threshold=args.jaccard_threshold,
+        logger=args.log_dir,
```
praateekmahajan, line 42: Let's add the profile_dir here too?

praateekmahajan (approved):
LGTM, would wait for @ayushdg to also TAL! 
Great speedup 🎊

nemo_curator/scripts/fuzzy_deduplication/connected_components.py
VibhuJawa: Done

VibhuJawa:
@ayushdg , Please take a look. Lets land this in soon. Have addressed all the changes

nemo_curator/modules/fuzzy_dedup.py
```diff
@@ -1566,20 +1561,12 @@ def _write_dedup_encoded_jaccard_pair(self, encoded_jaccard_pair_path):
                 transform_divisions=False,
                 align_dataframes=False,
             )
-            ddf.to_parquet(output_path, write_index=False)
+            ddf.to_parquet(output_path, write_index=False, overwrite=True)
```
praateekmahajan, line 1564: Nice to see we're adding overwrites, though I thought we discussed not adding them sicne the `fs` removal might cause a slowdown. See [this](https://github.com/orgs/NVIDIA/projects/34/views/13?pane=issue&itemId=82492890) and hence [this](https://github.com/NVIDIA/NeMo-Curator/pull/321)

praateekmahajan: cc @ayushdg (in case we're adding overwrites) I'll also add some in #321

VibhuJawa: So the only place we should not add this is the place is when false_positive check is False and we write  approximately `num_input_text_files//num_workers * num_output_buckets` files which is what overwhelms removing a bunch of small files in OS.  

Everywhere else if fair game and we should do it. 
 
https://github.com/NVIDIA/NeMo-Curator/blob/1cc6d7a06946b5e244865ee907b214efccbf6a75/nemo_curator/modules/fuzzy_dedup.py#L1176-L1192

n_input_text_partitions/num_workers)

nemo_curator/modules/fuzzy_dedup.py
```diff
@@ -1451,7 +1448,7 @@ def _run_connected_components(
             self.profile_dir, "connected-components-run"
         ):
 
-            Comms.initialize(p2p=True)
+            Comms.initialize(p2p=False)
```
ayushdg, line 1451: Thanks for this change!

nemo_curator/modules/fuzzy_dedup.py
ayushdg: I think we're aligned here based on the updates in #321/#330. 

Side note: The dynamics of overwriting/removing files might change when we start testing with cloud paths rather than local filesystems, but will leave that discussion for later.

ayushdg (approved):
Thanks!

PR #311, opened 2024-10-18
Specify `columns` when reading files with `DocumentDataset`

sarahyurick:
Closes https://github.com/NVIDIA/NeMo-Curator/issues/180.

All of these will now work:

(1) [Pandas](https://pandas.pydata.org/docs/reference/api/pandas.read_json.html) and [cuDF](https://docs.rapids.ai/api/cudf/legacy/user_guide/api_docs/api/cudf.read_json/) `read_json` do not support a `columns` parameter, so we read in the entire DataFrame and then remove unwanted columns behind the scenes.
```
dataset = DocumentDataset.read_json(dataset_path, columns=["col1", "col2"])
```
(2) [Pandas](https://pandas.pydata.org/docs/reference/api/pandas.read_parquet.html) and [cuDF](https://docs.rapids.ai/api/cudf/latest/user_guide/api_docs/api/cudf.read_parquet/) `read_parquet` both support a `columns` parameter, so we are able to take advantage of this functionality.
```
dataset = DocumentDataset.read_parquet(dataset_path, columns=["col1", "col2"])
```
(3) [Pandas](https://pandas.pydata.org/docs/reference/api/pandas.read_pickle.html) `read_pickle` (there is no cuDF `read_pickle`) does not support a `columns` parameter, so we read in the entire DataFrame and then remove unwanted columns behind the scenes.
```
dataset = DocumentDataset.read_pickle(dataset_path, columns=["col1", "col2"])
```
(4) Following [cudf.read_json](https://docs.rapids.ai/api/cudf/legacy/user_guide/api_docs/api/cudf.read_json/), you can specify `dtype` and `prune_columns=True` to only return the columns mentioned in the `dtype` argument. Note that Pandas does not support `prune_columns`.
```
dataset = DocumentDataset.read_json(
    dataset_path,
    dtype={"col1": str, "col2": str},
    prune_columns=True,
    backend="cudf",
)
```

### DIFF
--- a/nemo_curator/datasets/doc_dataset.py
+++ b/nemo_curator/datasets/doc_dataset.py
@@ -88,6 +88,8 @@ class DocumentDataset:
         backend="pandas",
         files_per_partition=1,
         add_filename=False,
+        columns: List[str] = None,
+        **kwargs,
     ):
         return cls(
             read_data(
@@ -96,6 +98,8 @@ class DocumentDataset:
                 backend=backend,
                 files_per_partition=files_per_partition,
                 add_filename=add_filename,
+                columns=columns,
+                **kwargs,
             )
         )
 
--- a/nemo_curator/datasets/doc_dataset.py
+++ b/nemo_curator/datasets/doc_dataset.py
@@ -43,7 +43,7 @@ class DocumentDataset:
         files_per_partition: int = 1,
         add_filename: bool = False,
         input_meta: Union[str, dict] = None,
-        columns: List[str] = None,
+        columns: Optional[List[str]] = None,
         **kwargs,
     ):
         return cls(
@@ -66,7 +66,7 @@ class DocumentDataset:
         backend="pandas",
         files_per_partition=1,
         add_filename=False,
-        columns: List[str] = None,
+        columns: Optional[List[str]] = None,
         **kwargs,
     ):
         return cls(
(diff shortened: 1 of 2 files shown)


nemo_curator/datasets/doc_dataset.py
```diff
@@ -43,6 +43,8 @@ def read_json(
         files_per_partition: int = 1,
         add_filename: bool = False,
         input_meta: Union[str, dict] = None,
+        columns: List[str] = None,
```
praateekmahajan, line 46: nit on the type hint, it should be `Optional[List[str]]` given default is `None`

sarahyurick: Thanks, updated!

nemo_curator/utils/distributed_utils.py
```diff
@@ -271,6 +273,7 @@ def read_single_partition(
         add_filename: Whether to add a "filename" column to the DataFrame.
         input_meta: A dictionary or a string formatted as a dictionary, which outlines
             the field names and their respective data types within the JSONL input file.
+        columns: If not None, only these columns will be read from the file.
```
ryantwolf, line 276: It would be great if you could elaborate on how each of the file formats handles this parameter. Something simple like "Parquet files and cudf json files will get a performance bump from this."

ryantwolf (commented):
One minor thing, then we should be good.

nemo_curator/utils/distributed_utils.py
sarahyurick: Updated, thanks. The comment (and PR) only includes Parquet gains for now, since there is some refactoring around handling JSON files that @praateekmahajan and I are planning to work on.

ryantwolf (approved):
Good with me, thanks!

PR #312, opened 2024-10-18
Assert that there is at least 1 worker

sarahyurick:
Closes https://github.com/NVIDIA/NeMo-Curator/issues/216.


VibhuJawa (approved):
LGTM, thanks for fixing this @sarahyurick

PR #317, opened 2024-10-22
Write to file without including "filename" column

sarahyurick:
Closes https://github.com/NVIDIA/NeMo-Curator/issues/293.

### DIFF
--- a/nemo_curator/utils/distributed_utils.py
+++ b/nemo_curator/utils/distributed_utils.py
@@ -468,13 +468,19 @@ def process_all_batches(
     )
 
 
-def single_partition_write_with_filename(df, output_file_dir, output_type="jsonl"):
+def single_partition_write_with_filename(
+    df,
+    output_file_dir,
+    keep_filename_column=False,
+    output_type="jsonl",
+):
     """
     This function processes a DataFrame and writes it to disk
 
     Args:
         df: A DataFrame.
         output_file_dir: The output file path.
+        keep_filename_column: Whether to keep or drop the "filename" column, if it exists.
         output_type="jsonl": The type of output file to write.
     Returns:
         If the DataFrame is non-empty, return a Series containing a single element, True.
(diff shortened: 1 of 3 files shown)


nemo_curator/utils/distributed_utils.py
```diff
@@ -524,16 +536,24 @@ def single_partition_write_with_filename(df, output_file_dir, output_type="jsonl
...
 
     return success_ser
 
 
-def write_to_disk(df, output_file_dir, write_to_filename=False, output_type="jsonl"):
+def write_to_disk(
+    df,
+    output_file_dir,
+    write_to_filename=False,
+    keep_filename_column=False,
```
sarahyurick, line 554: Interested in others' opinions here. I think it makes the most sense to default to false here, and automatically drop the "filename" column unless specified otherwise. It saves storage space and the column itself means nothing to the user who can already see the filename after it is written.

I know a lot of our classifier scripts, etc. use the `write_to_disk`, so those will all now automatically default to `keep_filename_column=False`. Don't know if we should consider adding it to `argparse` for more user control on the CLI side, or if folks think we should be good without.

ryantwolf: I agree, defaulting to `False` is good.

VibhuJawa: Agreed, this behavior makes a lot more sense

ayushdg (approved):

ryantwolf (approved):

nemo_curator/utils/distributed_utils.py
```diff
@@ -468,13 +468,19 @@ def process_all_batches(
     )
 
 
-def single_partition_write_with_filename(df, output_file_dir, output_type="jsonl"):
+def single_partition_write_with_filename(
+    df,
+    output_file_dir,
+    keep_filename_column=False,
+    output_type="jsonl",
```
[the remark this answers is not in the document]
sarahyurick: Thanks @VibhuJawa ! I've added them to https://github.com/NVIDIA/NeMo-Curator/pull/318, hope that's okay.

PR #318, opened 2024-10-22
Skip reading files with incorrect extension

sarahyurick:
Closes https://github.com/NVIDIA/NeMo-Curator/issues/214.


ayushdg:
We might need to expand the list of extensions since some files are format like `.json.gz`.
I wonder if an alternative could be to expand `get_all_files_paths_under` to also filter on an extension. That way users can specify what extension they want to filter on. 
I'm hoping that #50 will make things easier in this regard.

sarahyurick:
Thanks @ayushdg ! I like your idea of having it in `get_all_files_paths_under` so I changed it to use that instead.

Also, I agree with you about `.json.gz`. I think it is outside the scope of this PR, but I have added it to #50 for now.

nemo_curator/utils/distributed_utils.py
```diff
@@ -415,15 +421,30 @@ def read_data(
             df = df.to_backend("cudf")
 
     elif file_type in ["json", "jsonl", "parquet"]:
+        assert len(input_files) > 0
+
+        input_extensions = {os.path.splitext(f)[-1] for f in input_files}
+        if len(input_extensions) != 1:
+            raise RuntimeError(
```
sarahyurick, line 428: An example of when we would expect this `RuntimeError` is for:
```
doc = DocumentDataset.read_json(in_files)
```
Where `in_files` is a string path to a directory with multiple JSONL files and a CRC file. Since the CRC file is not explicitly being filtered out, we raise the error.

nemo_curator/utils/file_utils.py
```diff
@@ -45,7 +46,32 @@ def expand_outdir_and_mkdir(outdir):
...
+            warnings.warn(f"Skipping read for file: {file}")
+
+    return filtered_files
+
+
+def get_all_files_paths_under(
+    root: str,
+    recurse_subdirectories: bool = True,
+    followlinks: bool = False,
+    filter_by: Optional[Union[str, List[str]]] = None,
```
sarahyurick, line 73: All of these examples work:
(1)
```
input_files = get_all_files_paths_under(in_files, filter_by="jsonl")
input_dataset = DocumentDataset.read_json(input_files)
```
(2)
```
input_files = get_all_files_paths_under(in_files, filter_by=["jsonl"])
input_dataset = DocumentDataset.read_json(input_files)
```
(3)
```
# Returns a list containing only .jsonl, .parquet, and .csv files
input_files = get_all_files_paths_under(in_files, filter_by=["jsonl", "parquet", "csv"])
```

nemo_curator/datasets/doc_dataset.py
```diff
@@ -65,10 +66,10 @@ def read_json(
...
     def read_parquet(
         cls,
-        input_files,
-        backend="pandas",
-        files_per_partition=1,
-        add_filename=False,
+        input_files: Union[str, List[str]],
+        backend: str = "pandas",
+        files_per_partition: int = 1,
+        add_filename: bool = False,
```
ayushdg, line 72: Thanks a lot for adding annotations here. Could we also add some docstrings here as well?

nemo_curator/utils/file_utils.py
```diff
@@ -45,7 +46,32 @@ def expand_outdir_and_mkdir(outdir):
     return outdir
 
 
-def get_all_files_paths_under(root, recurse_subdirectories=True, followlinks=False):
+def filter_files_by_extension(
+    files_list: str,
```
ayushdg, line 50:
```suggestion
    files_list: List[str],
```

nemo_curator/utils/file_utils.py
```diff
@@ -45,7 +46,32 @@ def expand_outdir_and_mkdir(outdir):
...
+    if isinstance(filter_by, str):
+        filter_by = [filter_by]
+
+    file_extensions = [s if s.startswith(".") else "." + s for s in filter_by]
+
+    for file in files_list:
+        if file.endswith(tuple(file_extensions)):
+            filtered_files.append(file)
+        else:
+            warnings.warn(f"Skipping read for file: {file}")
```
ayushdg, line 64: I wonder if this might get too noisy in some cases. I'm leaning towards warning once if we have to skip, but not for every file we skip.

nemo_curator/utils/distributed_utils.py
ayushdg: We can leave this as is for now. 
In theory there might be cases where a user filters by `[.json, .jsonl]` using the file filter, but will raise errors here. In practice I expect it to be unlikely so we can wait an see if there is any user feedback around this.

ayushdg (dismissed):
Thanks for the changes. Overall changes lgtm! Minor nits/comments.

As a followup it might make sense to track updating tutorials/notebooks to use this newer filter arg in the api but not required for this pr.

sarahyurick:
Thanks @ayushdg ! Updated.

nemo_curator/datasets/doc_dataset.py
```diff
@@ -35,7 +36,7 @@ def __len__(self):
     def persist(self):
         return DocumentDataset(self.df.persist())
 
-    def head(self, n=5):
+    def head(self, n: int = 5):
```
praateekmahajan, line 39:
```suggestion
    def head(self, n: int = 5) -> ??:
```
since we're adding the annotation let's add the return type too

nemo_curator/datasets/doc_dataset.py
```diff
@@ -65,13 +79,25 @@ def read_json(
     @classmethod
     def read_parquet(
         cls,
-        input_files,
-        backend="pandas",
-        files_per_partition=1,
-        add_filename=False,
+        input_files: Union[str, List[str]],
+        backend: str = "pandas",
```
praateekmahajan, line 83: nit, but both here and line 46 IIUC the type is `Literal["pandas", "cudf"]`

nemo_curator/utils/file_utils.py
```diff
@@ -45,7 +46,44 @@ def expand_outdir_and_mkdir(outdir):
     return outdir
 
 
-def get_all_files_paths_under(root, recurse_subdirectories=True, followlinks=False):
+def filter_files_by_extension(
```
praateekmahajan, line 49: return type hint of List[str]

nemo_curator/utils/file_utils.py
```diff
@@ -45,7 +46,44 @@ def expand_outdir_and_mkdir(outdir):
...
+    files_list: List[str],
+    filter_by: Union[str, List[str]],
```
praateekmahajan, line 51: nit : I think have a more verbose arg name might be beneficial here such as `keep_extensions` or `extensions_list`, since `filter` is rather open-ended, and when I first read through I thought it is a list of filters on filename (e.g. UDF like `lambda file_name : True if .... else False`)

praateekmahajan: nit2 : having return type hint

nemo_curator/utils/file_utils.py
```diff
@@ -45,7 +46,44 @@ def expand_outdir_and_mkdir(outdir):
...
+    file_extensions = [s if s.startswith(".") else "." + s for s in filter_by]
+
+    for file in files_list:
+        if file.endswith(tuple(file_extensions)):
+            filtered_files.append(file)
+        else:
+            warning_flag = True
+
+    if warning_flag:
+        warnings.warn(f"Skipped at least one file due to unmatched file extension(s).")
```
praateekmahajan, line 76: instead of tracking `warning_flag` as a boolean, you can just do `if len(files_list) != len(filtered_files): warnings.warn(...)`

nemo_curator/utils/file_utils.py
```diff
@@ -45,7 +46,44 @@ def expand_outdir_and_mkdir(outdir):
...
+
+    return filtered_files
+
+
+def get_all_files_paths_under(
+    root: str,
+    recurse_subdirectories: bool = True,
+    followlinks: bool = False,
+    filter_by: Optional[Union[str, List[str]]] = None,
```
praateekmahajan, line 85: similar nit on arg name as above

nemo_curator/utils/file_utils.py
```diff
@@ -329,7 +377,7 @@ def separate_by_metadata(
     return delayed(reduce)(merge_counts, delayed_counts)
 
 
-def parse_str_of_num_bytes(s, return_str=False):
+def parse_str_of_num_bytes(s: str, return_str: bool = False):
```
praateekmahajan, line 380: nit : return type hint

nemo_curator/datasets/doc_dataset.py
```diff
@@ -65,13 +79,25 @@ def read_json(
...
+        files_per_partition: int = 1,
+        add_filename: bool = False,
         columns: Optional[List[str]] = None,
         **kwargs,
     ):
+        """
```
praateekmahajan, line 89: nit typehint on line 88

nemo_curator/datasets/doc_dataset.py
```diff
@@ -49,6 +50,19 @@ def read_json(
         columns: Optional[List[str]] = None,
         **kwargs,
     ):
```
praateekmahajan, line 52: type hint on return

nemo_curator/utils/file_utils.py
sarahyurick: Good point, thanks!

nemo_curator/utils/file_utils.py
sarahyurick: Great idea, thanks!

sarahyurick:
Thank you @praateekmahajan ! I have addressed all your comments.

nemo_curator/utils/distributed_utils.py
```diff
@@ -262,10 +268,10 @@ def _set_torch_to_use_rmm():
 
 
 def read_single_partition(
-    files,
-    backend="cudf",
-    filetype="jsonl",
-    add_filename=False,
+    files: List[str],
+    backend: str = "cudf",
```
praateekmahajan, line 272: super nit, I think it should be `Literal["cudf", "pandas"]`

nemo_curator/utils/distributed_utils.py
```diff
@@ -415,15 +428,30 @@ def read_data(
...
+        assert len(input_files) > 0
+
+        input_extensions = {os.path.splitext(f)[-1] for f in input_files}
+        if len(input_extensions) != 1:
+            raise RuntimeError(
+                "All files being read must have the same file type. "
+                "Please check your input directory or list of files to ensure this. "
+                "To generate a list of files with a given file type in your directory, "
+                "please use the nemo_curator.utils.file_utils.get_all_files_paths_under "
+                "function with the filter_by parameter."
```
praateekmahajan, line 440:
```suggestion
                "function with the `keep_extensions` parameter."
```

praateekmahajan (approved):
LGTM! Thanks for adding type hint as well. (left two small nits on comment / typehint)

PR #320, opened 2024-10-23
Japanese support for `get_word_splitter`

sarahyurick:
Closes https://github.com/NVIDIA/NeMo-Curator/issues/257.

### DIFF
--- a/nemo_curator/utils/text_utils.py
+++ b/nemo_curator/utils/text_utils.py
@@ -23,16 +23,32 @@ from itertools import groupby
 
 def get_word_splitter(language):
     language = language.lower()
+
     if language == "zh":
         import jieba
 
-        return jieba.cut
+        def jieba_splitter(text):
+            return list(jieba.cut(text))
+
+        return jieba_splitter
+
+    elif language == "ja":
+        import MeCab
+
+        def mecab_splitter(text):
+            mecab = MeCab.Tagger()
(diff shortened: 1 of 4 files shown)


nemo_curator/utils/text_utils.py
```diff
@@ -23,16 +23,33 @@
...
     if language == "zh":
         import jieba
 
-        return jieba.cut
+        def jieba_splitter(text):
+            return list(jieba.cut(text))
+
+        return jieba_splitter
+
+    elif language == "ja":
```
ryantwolf, line 35: Do you mind updating the docstrings in the filters where the `get_word_splitter` is used to explain the splitting strategy we do for each language?

requirements/requirements.txt
```diff
@@ -12,6 +12,7 @@ in-place==0.5.0
 jieba==0.42.1
 justext==3.0.1
 lxml_html_clean
+mecab-python3==1.0.9
```
ryantwolf, line 15: Any reason for the hard pin? If not, I'd prefer to just keep it at `mecab-python3`

ryantwolf (changes requested):
With documentation updates, this should be good to go.

sarahyurick:
Thanks @ryantwolf ! Updated.

ryantwolf (approved):
Great, thanks!

PR #321, opened 2024-10-23
Add more comments for clearing Fuzzy Dedup Cache

praateekmahajan:
## Description
If a user doesn't clear cache between FuzzyDedup runs they might end up using stale data.
`dask` provides an `overwrite` argument but we avoid using that, so that GPU time is not wasted in clearing up the filesystem.

## Checklist

### DIFF
--- a/docs/user-guide/gpudeduplication.rst
+++ b/docs/user-guide/gpudeduplication.rst
@@ -184,7 +184,7 @@ Python API
     from nemo_curator import FuzzyDuplicatesConfig
 
     config = FuzzyDuplicatesConfig(
-        cache_dir="/path/to/dedup_outputs",
+        cache_dir="/path/to/dedup_outputs", # must be cleared between runs
         id_field="my_id",
         text_field="text",
         seed=42,


docs/user-guide/gpudeduplication.rst
```diff
@@ -184,7 +184,7 @@ Python API
     from nemo_curator import FuzzyDuplicatesConfig
 
     config = FuzzyDuplicatesConfig(
-        cache_dir="/path/to/dedup_outputs",
+        cache_dir="/path/to/dedup_outputs", # must be cleared between runs
```
praateekmahajan, line 187: Line 254 has the tip already, can we add somewhere else too?

ayushdg: I believe other PR's are tackling some of the documentation changes. The only other place I can think of is: https://github.com/NVIDIA/NeMo-Curator/blob/main/examples/fuzzy_deduplication.py

praateekmahajan:
Closing in favor of #330

PR #323, opened 2024-10-24
Add `nemo_curator.__version__`, `--version`, and `-v` displays

sarahyurick:
Closes https://github.com/NVIDIA/NeMo-Curator/issues/132.


nemo_curator/__init__.py
```diff
@@ -29,3 +29,6 @@
 # See https://github.com/NVIDIA/NeMo-Curator/issues/33
 # This also happens when reading and writing to files
 dask.config.set({"dataframe.convert-string": False})
+
+# See setup.py
+__version__ = "0.4.0"
```
ayushdg, line 34: Thanks for the adding this! One concern I have is needing to manually update this to stay in sync with setup.py. I wonder if there's a way to populate this from our `setup.py` file or perhaps there's a standard canonical approach many python packages follow. I need to read up a bit on this. Curious if perhaps @vyasr has some suggestions here since I'm not too familiar with python versioning/packaging semantics.

vyasr: Yeah there are standard tools here. There are a couple of things I would look at:
1. If your goal is only to make sure that __init__.py and setup.py stay in sync, I would look at [updating this package to specify more of the metadata in pyproject.toml instead of setup.py](https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html), at which point you can use [setuptools' dynamic metadata support to read the version from __init__.py](https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html#dynamic-metadata).
2. If you want to go one step further and extract the version based on git tags ([it looks like you have some](https://github.com/NVIDIA/NeMo-Curator/tags)) then once you complete step one I would look at https://github.com/pypa/setuptools-scm. It will pull the version from your git tags.

pyproject.toml
```diff
@@ -12,16 +12,157 @@
...
-[tool.black]
-line_length = 88
 
 [tool.pytest.ini_options]
 markers = [
     "gpu: marks tests as GPU tests (deselect with '-m \"not gpu\"')"
 ]
+
+[tool.setuptools.dynamic]
+version = { attr = "nemo_curator.__version__" }
```
sarahyurick, line 164: Dynamically fetches the version from `nemo_curator/__init__.py`.

pyproject.toml
```diff
@@ -12,16 +12,157 @@
...
+    "resiliparse",
+    "spacy>=3.6.0, <3.8.0",
+    "unidic-lite==1.0.8",
+    "usaddress==0.5.10",
+    "warcio==1.7.4",
+    "zstandard==0.18.0",
+]
+dynamic = ["version"]
+
+[project.optional-dependencies]
```
sarahyurick, line 72: Curious for opinions on this new implementation. Instead of having the `RAPIDS_NIGHTLY` flag, the user installs `.[image_nightly]` or `.[all_nightly`.

pyproject.toml
```diff
@@ -12,16 +12,157 @@
...
+image_nightly = [
+    "nvidia-dali-cuda120",
+    "nvidia-nvjpeg2k-cu12",
+    "timm>=1.0.8",
+    "nemo_curator[cuda12x_nightly]",
+]
+# Installs all of the above with Stable RAPIDS
+all = [
+    "nemo_curator[image]",
+]
```
sarahyurick, line 108: Question: aren't installing `image` and `all` the same thing, since `image` needs GPU modules? Or am I misunderstanding something?

pyproject.toml
```diff
@@ -12,16 +12,157 @@
...
+    { name = "Ryan Wolf", email = "rywolf@nvidia.com" },
+    { name = "Sarah Yurick", email = "syurick@nvidia.com" },
+]
+classifiers = [
+    "Development Status :: 3 - Alpha",
+    "Programming Language :: Python :: 3",
+    "Programming Language :: Python :: 3.10",
+]
+requires-python = ">=3.10, <3.11"
+dependencies = [
```
sarahyurick, line 40: According to [setuptools documentation](https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html#dynamic-metadata), I could dynamically link `dependencies` and `optional-dependencies` to requirements files, but this is currently in beta. Instead, I just list them all in here and delete the txt files.

Happy to hear other preferences.

sarahyurick (commented):
TODO:

- [ ] Fix `AttributeError: module 'cuda.ccudart' has no attribute '__pyx_capi__'`
- [x] Update to [correct version](https://github.com/NVIDIA/NeMo-Curator/blob/main/setup.py#L58)
- [ ] Re-add [MeCab](https://github.com/NVIDIA/NeMo-Curator/blob/main/requirements/requirements.txt#L15) requirement
- [ ] Edit README
- [ ] Use setuptools-scm

pyproject.toml
sarahyurick: @ryantwolf ?

ryantwolf: they are the same. I just made them different names for once we add more modalities and extras.

nemo_curator/scripts/classifiers/quality_classifier_inference.py
```diff
@@ -29,7 +31,9 @@
 
 
 def main():
-    args = ArgumentHelper.parse_distributed_classifier_args().parse_args()
+    args = ArgumentHelper.parse_distributed_classifier_args(
```
ryantwolf, line 34: Instead of modifying every script to have this, can we just make it a prefix to all descriptions in the class?

pyproject.toml
ryantwolf: fine with me

ryantwolf (changes requested):
Can you also try creating and uploading a pip package with this? I want to see if any of that flow breaks with this. Message me if you need help testing that.

nemo_curator/__init__.py
sarahyurick: Thanks! I ended up only doing step 1 for now. When I tried to implement step 2 (see [this commit](https://github.com/NVIDIA/NeMo-Curator/pull/323/commits/df3550e7b609398840d214a1a06451bc2349fefb)), I was having trouble getting it to grab any tags, both locally and on GitHub CI.

Locally, it would be `nemo_curator-0.0.post181+g5c03c7c.d20241106`, where 181 is the total number of commits we have in the NeMo Curator repository, `5c03c7c` is the latest commit hash, and `20241106` is the date. Since it was `nemo_curator-0.0`, it was not capturing the 0.5.0 tag.

I saw similar behavior on GitHub CI, see an [example run](https://github.com/NVIDIA/NeMo-Curator/actions/runs/11712284148/job/32622597310) here. If you navigate to "Install NeMo-Curator and pytest", you will see `Collecting awscli>=1.22.55 (from nemo_curator==0.0.post195+g74dbc65)`, where this time 195 is the total number of commits in the NeMo Curator repository plus the number of commits in this PR at the time.

For now, I think it should be okay to just have the version in a single location, in our `pyproject.toml` file.

nemo_curator/scripts/classifiers/quality_classifier_inference.py
sarahyurick: Done, thanks! Added to `ArgumentHelper`'s `__init__` function.

pyproject.toml
vyasr: I definitely think that optional dependencies are a cleaner solution for this.

pyproject.toml
vyasr: I'm not aware of any reason to prefer the requirements files unless you specifically wanted to distribute them.

sarahyurick:
Thanks @ryantwolf ! Added your suggestions and tested a pip release. Everything looks okay from my end.

cc @ayushdg

nemo_curator/__init__.py
sarahyurick: Yes when I did `git fetch --tags` they were there. Very strange behavior IMO, I tried various things but couldn't get it to work at all for me.

ryantwolf (approved):
LGTM, thanks!

ayushdg (approved):

nemo_curator/__init__.py
vyasr: Strange. Anyway, if you're happy with the current state we're good. If you want help setting up setuptools-scm or similar later I can take a look another time.

PR #326, opened 2024-10-25
Add codepath for computing buckets without int conversion

### LINKED ISSUE #67: [BUG] Fuzzy deduplication fails on datasets with no duplicates
**Describe the bug**

When attempting to run fuzzy deduplication on a dataset that has no duplicates, the code errors out.

1) Clone the repo
2) Run the TinyStories tutorial
3) Run `examples/fuzzy_deduplication.py` on the dataset under `tutorials/tinystories/data/jsonl/val/`

**Expected behavior**

The code should not crash.

**Environment details**
Using NVIDIA pytorch:24.04-py3 image

ayushdg:
## Description

PR has 2 enhancements:
1. Improves performance for cases where users want to skip the fp check by skipping conversion of bucket_id's to integers, only needed by `map_buckets` and following steps in the fpcheck path.
2. Improves error messages/cases where the data contained no duplicates. Fixes #67.

## Usage

```python
        lsh = LSH(
            ..., # same params as earlier
            buckets_as_int=False, # or true if planning to go via FP check.
        )
```
## Checklist


nemo_curator/modules/fuzzy_dedup.py
```diff
@@ -362,21 +369,42 @@ def lsh(
...
+                    f"No duplicate documents found for buckets: {value_vars}"
+                )
+            else:
+                self._logger.info(f"Wrote data for buckets: {value_vars}")
+
+        if buckets_isempty:
+            self._logger.info("No duplicate documents found during LSH")
+            import shutil
+
+            shutil.rmtree(write_path)
```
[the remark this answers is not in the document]
ayushdg: Agreed. Decided to go via this route for now (since other places also use shutil). Aligned that the refactor to be more remote friendly should leverage fsspec utilities where possible.

nemo_curator/modules/fuzzy_dedup.py
```diff
@@ -493,6 +559,11 @@ def __call__(self, dataset: DocumentDataset):
         minhashLSH = Sequential([self.minhash, self.lsh])
         buckets_df = minhashLSH(dataset)
         print(f"Stage{stage_num}: Minhash + LSH complete!")
+        if buckets_df is None:
+            print(
+                f"Stage{stage_num}: No potential duplicate documents found during LSH"
+            )
+            return None
```
praateekmahajan, line 566: Should this return None or an empty `DocumentDataset` with no id's

ayushdg: I prefer returning `None`. Empty DocumentDatasets might lead to unexplained errors downstream that might be tougher to debug/understand. Happy to hear counter points. 
One thing that comes up from this is that I might update the `examples/FuzzyDedup.py` to handle the case where the result returned was `None`

praateekmahajan: Makes sense, but then for `Sequential` I think we might want to handle that behavior too?

ayushdg: I haven't seen `Sequential` being used directly with `FuzzyDuplicates` since the results cannot be processed downstream by any of the other modules without using to filter out the duplicates first. I'm not sure how to handle this use case. But longer term, we would probably want to add a `FuzzyDeduplicate` class that calls Fuzzy Duplicates and also handles removal.

nemo_curator/modules/fuzzy_dedup.py
```diff
@@ -261,6 +262,7 @@ def __init__(
         num_hashes: int,
         num_buckets: int,
         buckets_per_shuffle: int = 1,
+        buckets_as_int: bool = False,
```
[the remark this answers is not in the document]
ayushdg: I think it's a good suggestion. We can update the docstrings to indicate that it writes out data in a format required by false positive_check if set to true.

nemo_curator/modules/fuzzy_dedup.py
```diff
@@ -343,48 +357,98 @@ def lsh(
...
+        if are_buckets_empty:
+            self._logger.info("No duplicate documents found during LSH")
+            if os.path.exists(write_path):
+                import shutil
+
+                shutil.rmtree(write_path)
+
+        return are_buckets_empty
+
+    def _write_bucket_parquet(
```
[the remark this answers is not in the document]
ayushdg: There is a non-zero cost to checking if the buckets are empty or not. I've tried to write `check_empty_buckets` in a way that it breaks on the first file of any where it finds non empty data, but this might be slow for large network based filesystems. It should however be faster than the current approach of persisting the data first and converting to int. 

Once a non empty bucket is detected, that setting is persisted through the next set of iterations so the check is skipped in future iterations.

nemo_curator/modules/fuzzy_dedup.py
```diff
@@ -406,40 +420,90 @@ def lsh(
...
+                    )
+                    continue
+                bucket_start_id = end_id + 1
+                are_buckets_empty = False
 
             # Workaround for dtype mismatches with empty partitions
-            dtypes = df2.dtypes.to_dict()
-            df2 = df2.map_partitions(lambda x: x.astype(dtypes))
+            # dtypes = df2.dtypes.to_dict()
+            # df2 = df2.map_partitions(lambda x: x.astype(dtypes))
```
VibhuJawa, line 439: If we are commenting out this code, probably remove all these lines.

nemo_curator/modules/fuzzy_dedup.py
```diff
@@ -406,40 +420,90 @@ def lsh(
...
+                warnings.warn(
+                    f"Output path {write_path} already exists and will be overwritten"
+                )
+            df.to_parquet(write_path, write_index=False, overwrite=True)
+        else:
+            df.to_parquet(
+                write_path,
+                write_index=False,
+                overwrite=are_buckets_empty,
+                append=not are_buckets_empty,
```
VibhuJawa, line 480: Should we add `ignore_divisions=True` here ? 
Context: 
https://github.com/dask/dask/issues/8145#issuecomment-1112382518

VibhuJawa (approved):
Mostly looks good to me. 

I have two nitpicks but the logic seems good to me.

sarahyurick (approved):
LGTM, thanks!

Issue #211, opened 2024-08-19
[BUG] Semdedup Embedding Restart not working cleanly


VibhuJawa:
**Describe the bug**

Currently our semdedup restart mechanism for embedding is not working cleanly. 

 This is because of following (` add_filename=False`) 

https://github.com/NVIDIA/NeMo-Curator/blob/3a31ab13137e43fd8c1ffd40e07c52606a852acb/nemo_curator/scripts/semdedup/compute_embeddings.py#L62-L64

And write to filename is False

https://github.com/NVIDIA/NeMo-Curator/blob/3a31ab13137e43fd8c1ffd40e07c52606a852acb/nemo_curator/scripts/semdedup/compute_embeddings.py#L78

And `get_remaining_files`  by default cant handle comparing files with different extensions.  

 https://github.com/NVIDIA/NeMo-Curator/blob/3a31ab13137e43fd8c1ffd40e07c52606a852acb/nemo_curator/utils/file_utils.py#L66-L80

sarahyurick:
Happy to pair on this at some point; in general there are a couple of things I have been thinking should be refactored with `DocumentDataset`'s read and write functions.

See: https://github.com/NVIDIA/NeMo-Curator/issues/50, https://github.com/NVIDIA/NeMo-Curator/issues/180, https://github.com/NVIDIA/NeMo-Curator/issues/293...

VibhuJawa:
@sarahyurick , I think given your PRs , you should probably just take this on. Happy to provide input as needed. Let me know what you think.

sarahyurick:
I'm not sure I can reproduce this. I ran:
```
python compute_embeddings.py \
    --input-data-dir "my_data" \
    --input-file-type "jsonl" \
    --input-file-extension "jsonl" \
    --config-file "semdedup_config.yaml"
```
where `my_data` is a directory with 2 JSONL files. In `semdedup_config.yaml`, I specified a different directory as the `cache_dir` where the 2 resulting Parquet files were written. When I rerun without changing anything, there are no errors.

LMK if there is anything else I should be setting or changing, otherwise we can close this issue.

sarahyurick:
NVM, the issue is that it should not rerun if the embeddings are already present.

sarahyurick:
Closed by https://github.com/NVIDIA/NeMo-Curator/pull/327.

PR #327, opened 2024-10-25
Semantic deduplication improvements

sarahyurick:
- [x] Closes https://github.com/NVIDIA/NeMo-Curator/issues/198.
- [x] Closes https://github.com/NVIDIA/NeMo-Curator/issues/211.


sarahyurick:
Hi @VibhuJawa this is ready for review.

These changes enable the user to do `compute_embeddings.py` and output the results to JSONL files via `write_to_filename=True`, which fixes #211. The `clustering.py` script is modified to read either JSONL or Parquet files, and still outputs Parquet files either way.

Also, I was wondering if we should remove `--input_file_extension` and `--input_file_type` in the CLI scripts in favor of having the user specify it in the configuration file. Currently, if the user is running the steps individually (`compute_embeddings.py` then `clustering.py`), they have to specify the arguments; if they are only initializing an object from the `SemDedup` class, it grabs the `input_file_type` from the YAML file.

docs/user-guide/semdedup.rst
```diff
@@ -41,14 +41,15 @@ Semantic deduplication in NeMo Curator can be configured using a YAML file. Here
...
-    id_col_type: "int"
+    id_col_type: "int" # or "str" if the `add_id` module was used
     input_column: "text"
 
     # Embeddings configuration
     embeddings_save_loc: "embeddings"
     embedding_model_name_or_path: "sentence-transformers/all-MiniLM-L6-v2"
     embedding_batch_size: 128
     embedding_max_mem_gb: 25
+    input_file_type: "jsonl" # or "parquet"
```
VibhuJawa, line 52: I dont think we should keep here as this is shared among modules. 

I think we should probably also get rid of `input_column`, `id_col_name`, `id_col_type`, what do you think ?

nemo_curator/modules/config.py
```diff
@@ -151,6 +152,7 @@ class SemDedupConfig(BaseConfig):
     embedding_model_name_or_path: str = "sentence-transformers/all-MiniLM-L6-v2"
     embedding_batch_size: int = 128
     embedding_max_mem_gb: int = 25
+    input_file_type: str = "parquet"
```
VibhuJawa, line 155: Same here probably

nemo_curator/modules/semantic_dedup.py
```diff
@@ -171,6 +177,9 @@ def __init__(
         self.model = EmbeddingCrossFitModel(self.embeddings_config)
         self.write_embeddings_to_disk = write_embeddings_to_disk
         self.write_to_filename = write_to_filename
+        if input_file_type == "json":
+            input_file_type = "jsonl"
+        self.output_file_type = input_file_type.lower()
```
VibhuJawa, line 182: We should not force the same `input_file_type` and `output_file_type`.  We should just stick with `parquet` internally to help performance.

nemo_curator/modules/semantic_dedup.py
```diff
@@ -216,13 +225,22 @@ def __call__(self, dataset: DocumentDataset) -> DocumentDataset:
...
                     self.embedding_output_dir,
                     write_to_filename=self.write_to_filename,
-                    output_type="parquet",
+                    output_type=self.output_file_type,
+                )
+
+            if self.output_file_type == "jsonl":
+                embedding_files = get_all_files_paths_under(self.embedding_output_dir)
+                ddf = DocumentDataset(
+                    dask_cudf.read_json(embedding_files, blocksize="2GB")
```
VibhuJawa, line 234: We should probably move to loading from `DocumentDataset.from_parquet` , also lets to write stuff to `jsonl` for intermediate structures

tutorials/peft-curation-with-sdg/config/sem_dedup_config.yaml
```diff
@@ -10,6 +10,7 @@ embeddings_save_loc: "embeddings"
 embedding_model_name_or_path: "sentence-transformers/all-MiniLM-L6-v2"
 embedding_batch_size: 128
 embedding_max_mem_gb: 20
+input_file_type: "jsonl"
```
VibhuJawa, line 13: Same here, lets not add config (and actively remove config) which we have support for here. 

We also probably remove `embedding_max_mem_gb `

VibhuJawa (changes requested):
We should probably just use `parquet` as a internal data structure.  

Lets chat more towards what a good long term solution is for this

sarahyurick:
@VibhuJawa thanks! I think I have addressed all of your comments. I basically just expanded the `get_remaining_files` function to use a helper function called `_update_filetype` which enables us to compare `input_files` and `completed_files` even when they have different file extensions. Let me know what you think!

nemo_curator/modules/semantic_dedup.py
```diff
@@ -99,9 +98,7 @@ def _mean_pooling(self, model_output, attention_mask):
 class EmbeddingCrossFitModel(HFModel):
     def __init__(self, config: EmbeddingConfig):
         self.config = config
-        super().__init__(
-            self.config.model_name_or_path, max_mem_gb=self.config.max_mem_gb
-        )
+        super().__init__(self.config.model_name_or_path)
```
VibhuJawa, line 101: We should pass in max_mem_gb similar to how we do here by letting folks configure it automatically if not provided.  

https://github.com/NVIDIA/NeMo-Curator/blob/36fcf50cee12ccd3e85b204f7ef8c4f62c84aa51/nemo_curator/classifiers/fineweb_edu.py#L32-L37

nemo_curator/utils/script_utils.py
```diff
@@ -556,10 +573,12 @@ def parse_semdedup_args(
         argumentHelper = ArgumentHelper(parser)
         argumentHelper.add_distributed_args()
         if add_input_args:
-            argumentHelper.add_arg_input_data_dir(required=True)
+            argumentHelper.add_arg_input_data_dir()
```
VibhuJawa, line 576: Why did we drop `required=True` here ?

VibhuJawa (changes requested):
Thanks a lot for making this change and cleaning up the code.  Always great to see us decrease the number of configs a user has to keep track off. 

Have requested changes around memory estimation, everything else looks good to me.

nemo_curator/modules/semantic_dedup.py
```diff
@@ -123,7 +120,6 @@ class EmbeddingCreator:
     def __init__(
         self,
         embedding_model_name_or_path: str,
-        embedding_max_mem_gb: str,
```
VibhuJawa, line 126: We can probably allow this but set it to None to match other classifiers.

nemo_curator/utils/script_utils.py
sarahyurick: So for our CLI scripts, the input data directory is only needed for step 1 of 3, but some of the other input arguments (ID field, ID datatype) are used in subsequent steps. I figured since `parse_semdedup_args` is used for all 3 steps, removing the requirement is okay.

sarahyurick:
@VibhuJawa thanks! Let me know what you think.

examples/semdedup_example.py
```diff
@@ -72,7 +72,7 @@ def main(args):
 
 
 def attach_args():
-    parser = ArgumentHelper.parse_semdedup_args(add_input_args=True)
+    parser = ArgumentHelper.parse_semdedup_args()
```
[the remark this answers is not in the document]
sarahyurick: I removed it because all of the CLI scripts need at least one of
```
argumentHelper.add_arg_input_data_dir()
argumentHelper.add_arg_input_file_extension()
argumentHelper.add_arg_input_file_type()
argumentHelper.add_arg_input_text_field()
argumentHelper.add_arg_id_column()
argumentHelper.add_arg_id_column_type()
```
which are all input arguments.

nemo_curator/scripts/semdedup/README.md
```diff
@@ -30,7 +30,7 @@ Please edit `config/sem_dedup_config.yaml` to configure the pipeline and run it
 
 4) Extract deduplicated data
     ```sh
-    python extract_dedup_data.py --config-file "$CONFIG_FILE"
+    semdedup_extract_dedup_ids --id-column "my_id" --id-column-type "str" --config-file "$CONFIG_FILE"
```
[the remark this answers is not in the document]
sarahyurick: Updated to call it `semdedup_extract_unique_ids`.

examples/semdedup_example.py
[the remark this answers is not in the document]
VibhuJawa: Makes sense, thanks

VibhuJawa (approved):
LGTM

PR #330, opened 2024-10-28
Fix Jaccard Shuffle Error when Result is Empty + Overwrite using fuzzy_dedup cache

praateekmahajan:
## Fix Jaccard Shuffle Error

Check `len(merged_df) == 0`, if so won't bother. 

## Overwrite cache
Vibhu in #302 added
1. connected_components.parquet
2. final_dedup_encoded_jaccard_pair.parquet

This PR adds
1. encoded_jaccard_pair
2. dedup_parsed_id
3. anchor_docs_with_bk
4. jaccard_similarity_results.parquet

These were already being overwritten 
1. _buckets.parquet
2. _minhashes.parquet
3. _edges.parquet

We are overwritten this
1. shuffled_docs.parquet
## Usage

```python
```
## Checklist

### DIFF
--- a/nemo_curator/modules/fuzzy_dedup.py
+++ b/nemo_curator/modules/fuzzy_dedup.py
@@ -513,7 +513,7 @@ class FuzzyDuplicates:
                     )
                 )
                 ddf_mapped_buckets_w_anchors.to_parquet(
-                    mapped_buckets_w_anchors_path, write_index=False
+                    mapped_buckets_w_anchors_path, write_index=False, overwrite=True
                 )
             self._logger.info(
                 f"Time taken for Map_buckets : {time.time() - t0}s and output written at {mapped_buckets_w_anchors_path}"
@@ -557,6 +557,7 @@ class FuzzyDuplicates:
                     jaccard_pairs_path,
                     write_index=False,
                     write_metadata_file=False,
+                    overwrite=True,
                 )
                 self._logger.info(
                     f"Time taken for Jaccard Similarity = {time.time()-t0}s and output written at {jaccard_pairs_path}"
@@ -1144,16 +1145,21 @@ class _Shuffle:
                 try:
                     # NOTE: If we have more text-df partitions than bucket-map
                     # partitions, we are more likely to see an OverflowError
+
+                    subset_merged_df = merge_left_to_shuffled_right(
+                        subset_text_df,
+                        subset_bucket_df,
+                        merge_on,
+                    )
+
                     output_df = text_bytes_aware_shuffle(
-                        df=merge_left_to_shuffled_right(
-                            subset_text_df,
-                            subset_bucket_df,
-                            merge_on,
-                        ),
+                        df=subset_merged_df,
                         partition_on=partition_on,
                         text_column=self.text_field,
                         num_workers=num_workers,
                     )
+                    if output_df is None:
+                        continue
                 except OverflowError as err:
                     # We encountered an overflow error!
                     # Let's try again with less text data
@@ -1615,7 +1621,9 @@ class ConnectedComponents:
             unique_docs["uid"] = np.uint64(1)
             unique_docs["uid"] = unique_docs["uid"].cumsum()
             unique_docs["uid"] = unique_docs["uid"] - 1
-            unique_docs.to_parquet(dedup_parsed_id_path, write_index=False)
+            unique_docs.to_parquet(
+                dedup_parsed_id_path, write_index=False, overwrite=True
+            )
         self._logger.info(
             f"Time taken for Dedup Parsed Id = {time.time() - t0}s and output written at {dedup_parsed_id_path}"
         )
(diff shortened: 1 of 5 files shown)


nemo_curator/modules/fuzzy_dedup.py
```diff
@@ -1691,7 +1700,12 @@ def _batched_merge_and_write(
 
             subset_ddf = subset_ddf[[self.left_id, self.right_id, "jaccard"]]
             output_batch_path = os.path.join(output_path, f"{batch_id}.parquet")
-            subset_ddf.to_parquet(output_batch_path, write_index=False)
+            if batch_id == 0:
+                subset_ddf.to_parquet(
+                    output_batch_path, write_index=False, overwrite=True
+                )
+            else:
+                subset_ddf.to_parquet(output_batch_path, write_index=False, append=True)
```
praateekmahajan, line 1708: Can the reviewer TAL at this closely

tests/test_fuzzy_dedup.py
```diff
@@ -71,6 +71,24 @@ def large_fuzzy_dedup_data():
...
+def shuffle_fail_fuzzy_dedup_data():
+    df = cudf.DataFrame(
+        {
+            "id": [1, 2, 300, 4, -1],
+            "text": [
+                "A test string",
+                "A different test string",
+                "A different object",
+                "Something completely else that doesn't match",
+                "The quick black cat jumps over the lazy dog",
```
praateekmahajan, line 84: Added this test where there is only a group of first three documents (at a higher jaccard similarity only first two are grouped)

nemo_curator/utils/fuzzy_dedup_utils/shuffle_utils.py
```diff
@@ -169,6 +173,8 @@ def text_bytes_aware_shuffle(df, partition_on, text_column, num_workers=None):
     output_col = "_partitions"
 
     df = df.persist()
+    if len(df) == 0:
+        return None
```
praateekmahajan, line 177: If we remove this conditional, then the test we added below fails

nemo_curator/modules/fuzzy_dedup.py
```diff
@@ -1144,16 +1145,22 @@ def _batched_merge_and_write(
...
-                            subset_bucket_df,
-                            merge_on,
-                        ),
+                        df=subset_merged_df,
                         partition_on=partition_on,
                         text_column=self.text_field,
                         num_workers=num_workers,
                     )
+                    if output_df is None:
+                        text_part_offset += parts_per_text_batch_use
```
praateekmahajan, line 1162: Reviewer please TAL at this

ayushdg: I think the logic makes sense here to increment the `text_part_offset`. It might also make sense to call `update_restart_offset` here to update the txt file to restart from the next chunk if this job is resumed.

nemo_curator/modules/fuzzy_dedup.py
ayushdg: Thinking out loud: If a prev job wrote `encoded_jaccard_pair` with `batch_0.parquet`, `batch_1.parquet` ... `batch_n.parquet`. In theory the script would need to overwrite all batches rather than append for the batch_id != 0 case. Cases like these can however be problematic:  Is a prev job wrote more batches than the current one is trying to write we only overwrite a few batch files but don't touch the extra batches written by a previous job. In this case removing the whole directory where `encoded_jaccard_pair` will be written is cleaner wdyt?

nemo_curator/utils/fuzzy_dedup_utils/shuffle_utils.py
```diff
@@ -151,7 +153,9 @@ def get_shuffle_partition_info(
     return shuffle_part_ids
 
 
-def text_bytes_aware_shuffle(df, partition_on, text_column, num_workers=None):
+def text_bytes_aware_shuffle(
+    df, partition_on: str, text_column: str, num_workers: Optional[int] = None
+) -> Optional[None]:
```
ayushdg, line 158: Thanks for adding annotations in many parts of the codebase that's missing them 😄

nemo_curator/utils/fuzzy_dedup_utils/shuffle_utils.py
ayushdg: Yup that testcase makes sense to me.

ayushdg (commented):
Added minor comments. But overall lgtm!

nemo_curator/modules/fuzzy_dedup.py
ayushdg: Actually now that I've looked at #302 this codepath is going away, so ignore my comments.

praateekmahajan: Good catch @sarahyurick, looking at the code, `overwrite` only removes when path is directory (it has `isdir` check) https://github.com/dask/dask/blob/7ab0c93cfca0af130c778f9c1baec29a1f10b616/dask/dataframe/io/parquet/core.py#L860

I guessed it does for filepaths too, but that doesn't make sense, now that I think of it. I saw a similar pattern in lsh ([lines here](https://github.com/NVIDIA/NeMo-Curator/blob/dc727828bed491933fddfa166cc4c5ad865cfba7/nemo_curator/modules/fuzzy_dedup.py#L358-L365)) from where I was _inspired_, but the path there seems to be directory path (tho I'm intrigued how are filenames decided)

Any how looks like after the merge this doesn't need to be addressed but thanks for catching this

nemo_curator/modules/fuzzy_dedup.py
praateekmahajan: Removed the continue logic based on @VibhuJawa 's suggestion so this isn't needed anymore. Thanks for catching that there is more to the control flow than what I caught

ayushdg (approved):
Thanks a lot!

VibhuJawa (approved):
LGTM

sarahyurick (approved):

PR #334, opened 2024-10-29
Modify classifier scripts

sarahyurick:
Closes https://github.com/NVIDIA/NeMo-Curator/issues/333.


ryantwolf (approved):
Good catch, thanks!

PR #352, opened 2024-11-12
build: Improve caching

ko3n1g:
## Description

This PR improves the container image build time by separating the environment installation into a separate layer.

Since we copy Curators code via https (instead of docker's `COPY`), we move this into a separate build stage from which we selectively copy over the pyproject.toml for env installation. Since dependencies only change with pyproject.toml, this caches nicely. Afterwards we copy the actual source code and pip install again. The second pip install will be quick since all deps are already satisfied.

## Usage

```python
```
## Checklist


.github/workflows/gpuci.yml
```diff
@@ -11,12 +11,25 @@ on:
     # If new commits are added, the "gpuCI" label has to be removed and re-added to rerun gpuCI
     types: [ labeled ]
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
```
ko3n1g, line 16: This cancels runs of a PR if a new commit is added to a PR. It does not cancel "foreign runs" (runs of other PRs)

Dockerfile
```diff
@@ -15,19 +35,21 @@ RUN conda create -y --name curator -c conda-forge -c nvidia \
...
-# Needed to navigate to and pull the forked repository's changes
-ARG FORKED_REPO_URL
-ARG CURATOR_COMMIT
+RUN \
+  --mount=type=bind,source=/opt/NeMo-Curator/nemo_curator/__init__.py,target=nemo_curator/__init__.py,from=curator-update \
+  --mount=type=bind,source=/opt/NeMo-Curator/pyproject.toml,target=pyproject.toml,from=curator-update \
+  source activate curator && \
+  pip install ".[all]"
+
+COPY --from=curator-update /opt/NeMo-Curator/ /opt/NeMo-Curator/
```
sarahyurick, line 48: Ah okay, I see. By splitting the repository logic and the installation logic into 2 separate stages, it is allowing us to take advantage of caching in the environment setup portion (for example, the libcu* installations)?

sarahyurick (approved):
LGTM, thank you!

Dockerfile
ko3n1g: Exactly! This layer will only be build if the files `__init__.py` or `pyproject.toml` change. Otherwise it'll use the cached instance

PR #356, opened 2024-11-12
Add `build-test-publish-wheel` CI file

sarahyurick:
Re-opening https://github.com/NVIDIA/NeMo-Curator/pull/353.

### DIFF
--- a/.github/workflows/build-test-publish-wheel.yml
+++ b/.github/workflows/build-test-publish-wheel.yml
@@ -15,7 +15,6 @@
 name: Build, test, and publish a PyPi wheel (to testpypi)
 
 on:
-  pull_request:
   push:
     branches:
       - 'main'
--- a/.github/workflows/build-test-publish-wheel.yml
+++ b/.github/workflows/build-test-publish-wheel.yml
@@ -1,0 +1,48 @@
+# Copyright (c) 2020-2021, NVIDIA CORPORATION.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+name: Build, test, and publish a PyPi wheel (to testpypi)
+
+on:
+  pull_request:
+    branches:
+      - 'main'
+      - '[rv][0-9].[0-9].[0-9]'
+      - '[rv][0-9].[0-9].[0-9]rc[0-9]'
+
+defaults:
+  run:
+    shell: bash -x -e -u -o pipefail {0}
+
+jobs:
+  build-test-publish-wheel:
+    uses: NVIDIA/NeMo-FW-CI-templates/.github/workflows/_build_test_publish_wheel.yml@v0.7.0
+    with:
+      image-name: nemo_curator_container
+      dockerfile: Dockerfile
+      image-label: nemo-curator
+      build-args: |
+        IMAGE_LABEL=nemo-curator
+        REPO_URL=https://github.com/${{ github.repository }}.git
+        CURATOR_COMMIT=${{ github.sha }}
+      prune-filter-timerange: 24h
+      dry-run: true
+      python-package: nemo_curator
+      container-workdir: /opt/NeMo-Curator/
+      environment: public
+    secrets:
+      TWINE_USERNAME: ${{ secrets.TWINE_USERNAME }}
+      TWINE_PASSWORD: ${{ secrets.TWINE_PASSWORD }}
+      SLACK_WEBHOOK_ADMIN: ${{ secrets.SLACK_WEBHOOK_ADMIN }}
+      SLACK_WEBHOOK: ${{ secrets.SLACK_WEBHOOK }}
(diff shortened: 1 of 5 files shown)


.github/workflows/build-test-publish-wheel.yml
```diff
@@ -0,0 +1,48 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+name: Build, test, and publish a PyPi wheel (to testpypi)
+
+on:
+  pull_request:
```
sarahyurick, line 18:
```suggestion
  push:
```
Will change this back to `push` if the tests succeed.

ko3n1g:
Can you also modify `nemo_curator/__init__.py` like this? https://github.com/NVIDIA/NeMo/blob/main/nemo/__init__.py#L16

nemo_curator/package_info.py
```diff
@@ -0,0 +1,40 @@
...
+MINOR = 6
+PATCH = 0
+PRE_RELEASE = "rc0"
+DEV = "dev1"
+
+# Use the following formatting: (major, minor, patch, pre-release)
+VERSION = (MAJOR, MINOR, PATCH, PRE_RELEASE, DEV)
+
+__shortversion__ = ".".join(map(str, VERSION[:3]))
+__version__ = __shortversion__ + "".join(VERSION[3:])
```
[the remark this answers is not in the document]
ayushdg: I'm not completely familiar with versioning semantics but based on what I see from [pypa](https://packaging.python.org/en/latest/specifications/version-specifiers/#developmental-releases), it lines up that we would want the `0.6.0rcN.devN` approach.

nemo_curator/package_info.py
```diff
@@ -0,0 +1,36 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+MAJOR = 0
+MINOR = 6
+PATCH = 0
+PRE_RELEASE = "rc0"
```
ayushdg, line 19: Again I'm not sure what the convention is, but I prefer leaving this empty `""` and just having `DEV` until code freeze or QA.

.github/workflows/build-test-publish-wheel.yml
ayushdg: Our dev cycle typically consists of making a `0.n.0` branch around code freeze, and publishing the final release wheel later. Does it make sense to modify the process slightly to create `rc` branches instead?

I wonder if updating the action to trigger on something like is another alternative that works here:
```
on:
  release:
    types: [created]
```

nemo_curator/package_info.py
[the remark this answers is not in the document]
sarahyurick: From Oliver:
"
According to PEP 440, the version number of a Python package looks like this: `[N!]N(.N)*[{a|b|rc}N][.postN][.devN]`.

Curator has the `dev` segment but not `rc` (like the other NeMo libraries do). I like that you use `dev` and I could see some use cases in the future where this segment would become very handy. I would like to use this for NeMo Aligner and NeMo itself too. However, `rc` expresses pre-releases so we definitely have to have this too.

`devN` increases (probably with each push to `main`) until a new pre-release resets it.
"

sarahyurick:
Hi @ko3n1g and @ayushdg I have added the changes, please let me know what you think.

nemo_curator/package_info.py
ko3n1g: No, both segments (prerelease and dev) are filled. For the actual prerelease workflow that tags a commit, we‘ll set dev to "". And for the actual release workflow that tags a commit we‘ll set both dev and prelease to ""

.github/workflows/build-test-publish-wheel.yml
```diff
@@ -0,0 +1,48 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+name: Build, test, and publish a PyPi wheel (to testpypi)
+
+on:
+  push:
```
ko3n1g, line 18:
```suggestion
on:
  pull_request:
  push:
```

ko3n1g (changes requested):
test

nemo_curator/__init__.py
```diff
@@ -12,22 +12,54 @@
...
-    NemoDeployClient,
-    OpenAIClient,
-)
-from .utils.distributed_utils import get_client, get_network_interfaces
+    # Dask will automatically convert the list score type
+    # to a string without this option.
+    # See https://github.com/NVIDIA/NeMo-Curator/issues/33
+    # This also happens when reading and writing to files
+    dask.config.set({"dataframe.convert-string": False})
+except ImportError:
```
ryantwolf, line 23: Why do we need these import guards?

ryantwolf (commented):
Want to dig deeper on the import guards before merging.

nemo_curator/__init__.py
sarahyurick: We were seeing a Dask `ModuleNotFoundError` in the build and test CI, example here: https://github.com/NVIDIA/NeMo-Curator/actions/runs/11812742210/job/32908443159.

But I agree, we shouldn't have the try/except. Just debugged and figured out that it was an issue with our pyproject.toml file dynamically fetching the version from `nemo_curator/__init__.py` - switching it to fetch from the `package_info.py` file should work. Will push changes shortly.

ryantwolf: Thank you!

ryantwolf (approved):
Looks good now that the import guard is gone.

ko3n1g (approved):

PR #363, opened 2024-11-13
Multilingual Domain Classifier

sarahyurick:
Awaiting Hugging Face release.

Update: https://huggingface.co/nvidia/multilingual-domain-classifier


nemo_curator/classifiers/multilingual_domain.py
```diff
@@ -0,0 +1,136 @@
...
+        return model.eval()
+
+    def load_tokenizer(self):
+        return AutoTokenizer.from_pretrained(DOMAIN_IDENTIFIER)
+
+    def load_config(self):
+        return AutoConfig.from_pretrained(DOMAIN_IDENTIFIER)
+
+
+class MultilingualDomainClassifier(DistributedDataClassifier):
```
sarahyurick, line 67: It might be cool to just integrate this with our `DomainClassifier` instead of creating a whole separate class for the multilingual case, like doing a `multilingual` parameter or something. I plan to experiment more with this after the Hugging Face release.

nemo_curator/classifiers/domain.py
```diff
@@ -28,19 +28,31 @@
...
+    identifier: str = DOMAIN_IDENTIFIER
+    base_model: str = DOMAIN_BASE_MODEL
+    fc_dropout: float = 0.2
+    max_len: int = 512
+
+
+@dataclass
+class MultilingualDomainModelConfig:
+    identifier: str = MULTILINGUAL_DOMAIN_IDENTIFIER
+    base_model: str = MULTILINGUAL_DOMAIN_BASE_MODEL
```
praateekmahajan, line 47: nit : but given `DomainModelConfig` is a dataclass, do we need `MultilingualDomainModelConfig`? Since in it's current state `MultilingualDomainModelConfig` is just `DomainModelConfig` with non-default values

sarahyurick: Updated, thanks!

docs/user-guide/api/classifiers.rst
```diff
@@ -12,4 +12,4 @@ Classifiers
     :members:
 
 .. autoclass:: nemo_curator.classifiers.AegisClassifier
-    :members:
\ No newline at end of file
+    :members:
```
ryantwolf, line 15: Do you know what the change was here?

nemo_curator/classifiers/domain.py
```diff
@@ -49,30 +53,32 @@ def __init__(
...
-    for running on multi-node, multi-GPU setups to enable fast and efficient inference on large datasets.
+    DomainClassifier is a specialized classifier designed for domain classification tasks,
+    utilizing the NVIDIA Domain Classifier (https://huggingface.co/nvidia/domain-classifier)
+    and the NVIDIA Multilingual Domain Classifier (https://huggingface.co/nvidia/multilingual-domain-classifier) models.
+    This class is optimized for running on multi-node, multi-GPU setups to enable fast and efficient inference on large datasets.
 
     Attributes:
         filter_by (list[str], optional): The classes to filter the dataset by.
                                          If None, all classes will be included. Defaults to None.
+        multilingual (bool): If True, enable domain classification across 52 languages.
```
ryantwolf, line 81: It's worth calling out that this will use a separate model. Probably good to link to the model used in each case directly like you do in the overall docstring.

ryantwolf (approved):
Two nits, but nothing blocking.

docs/user-guide/api/classifiers.rst
sarahyurick: I think this is just adding an empty line to the bottom of the file (there previously wasn't one).

nemo_curator/classifiers/domain.py
sarahyurick: Updated, thanks.

ryantwolf (changes requested):
Leaving this while awaiting refactor discussed to separate the domain classifier into two classes (one for English and one for Multilingual).

ryantwolf (approved):
Clever implementation of the two classes!

PR #369, opened 2024-11-14
ci: Add `copyright-check` workflow

ko3n1g:
## Usage

```python
```
## Checklist

### DIFF
--- a/.github/workflows/copyright-check.yml
+++ b/.github/workflows/copyright-check.yml
@@ -1,0 +1,22 @@
+# Copyright (c) 2024, NVIDIA CORPORATION.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+name: Copyright check
+
+on:
+  pull_request:
+
+jobs:
+  copyright-check:
+    uses: NVIDIA/NeMo-FW-CI-templates/.github/workflows/_copyright_check.yml@v0.2.0


sarahyurick (commented):
I have a few questions about how this is supposed to work. Some are a bit nitpicky but want to make sure:

- It looks like it is only supposed to check Python files. However, I see some places where [__init__.py](https://github.com/NVIDIA/NeMo-Curator/blob/ko3n1g/ci/copyright-check/nemo_curator/pii/recognizers/__init__.py) files are skipped. Is this intentional?
- Can we also have it check YAML/YML files and shell scripts?
- There are a couple Python files where the year is 2023 (see [example](https://github.com/NVIDIA/NeMo-Curator/blob/ko3n1g/ci/copyright-check/nemo_curator/pii/algorithm.py)) instead of 2024. Is there a way we can check for this too? I imagine when 2025 hits we will want to update too?
- There are other instances, such as [here](https://github.com/NVIDIA/NeMo-Curator/blob/ko3n1g/ci/copyright-check/.github/workflows/build-test-publish-wheel.yml), where the copyright is even more out of date.
- Sometimes the files say `Copyright (c) 2024, NVIDIA CORPORATION.  All rights reserved.` and sometimes they just say `Copyright (c) 2024, NVIDIA CORPORATION.`. Is there any way we can do an exact string match to ensure that all copyrights are identical?
- Should we make sure the Dockerfile has the copyright?

ko3n1g:
> I have a few questions about how this is supposed to work. Some are a bit nitpicky but want to make sure:
> 
> * It looks like it is only supposed to check Python files. However, I see some places where [**init**.py](https://github.com/NVIDIA/NeMo-Curator/blob/ko3n1g/ci/copyright-check/nemo_curator/pii/recognizers/__init__.py) files are skipped. Is this intentional?
> * Can we also have it check YAML/YML files and shell scripts?
> * There are a couple Python files where the year is 2023 (see [example](https://github.com/NVIDIA/NeMo-Curator/blob/ko3n1g/ci/copyright-check/nemo_curator/pii/algorithm.py)) instead of 2024. Is there a way we can check for this too? I imagine when 2025 hits we will want to update too?
> * There are other instances, such as [here](https://github.com/NVIDIA/NeMo-Curator/blob/ko3n1g/ci/copyright-check/.github/workflows/build-test-publish-wheel.yml), where the copyright is even more out of date.
> * Sometimes the files say `Copyright (c) 2024, NVIDIA CORPORATION.  All rights reserved.` and sometimes they just say `Copyright (c) 2024, NVIDIA CORPORATION.`. Is there any way we can do an exact string match to ensure that all copyrights are identical?
> * Should we make sure the Dockerfile has the copyright?

Maybe best to sync with Pablo on this, he's in charge of the copyright script.

.github/workflows/copyright-check.yml
```diff
@@ -0,0 +1,22 @@
...
+# limitations under the License.
+
+name: Copyright check
+
+on:
+  pull_request:
+
+jobs:
+  copyright-check:
+    uses: NVIDIA/NeMo-FW-CI-templates/.github/workflows/_copyright_check.yml@v0.2.0
```
ko3n1g, line 22:
```suggestion
    uses: NVIDIA/NeMo-FW-CI-templates/.github/workflows/_copyright_check.yml@v0.14.0
```

This version allows to check other file types too.

pablo-garay:
image

Sorry i was writing reply for this & got into something else & missed it

It looks like it is only supposed to check Python files. However, I see some places where init.py files are skipped. Is this intentional?
Yes, init.py files are empty so we skip them
Can we also have it check YAML/YML files and shell scripts?
Just created a PR to allow this
There are a couple Python files where the year is 2023 (see example) instead of 2024. Is there a way we can check for this too? I imagine when 2025 hits we will want to update too? There are other instances, such as here, where the copyright is even more out of date.
I am not sure about this actually . I had a few talks and we didnt see this a needed requirement
Sometimes the files say Copyright (c) 2024, NVIDIA CORPORATION. All rights reserved. and sometimes they just say Copyright (c) 2024, NVIDIA CORPORATION.. Is there any way we can do an exact string match to ensure that all copyrights are identical?
I guess we can do this / it's doable
Should we make sure the Dockerfile has the copyright?
Not sure, need to check on this / if this needed. And which exact Dockerfile are we talking about?

pablo-garay:
Sorry i was writing reply for this & got into something else & missed it

It looks like it is only supposed to check Python files. However, I see some places where init.py files are skipped. Is this intentional?
**Yes, init.py files are empty so we skip them**
Can we also have it check YAML/YML files and shell scripts?
**Just created a PR to allow this**
There are a couple Python files where the year is 2023 (see example) instead of 2024. Is there a way we can check for this too? I imagine when 2025 hits we will want to update too? There are other instances, such as here, where the copyright is even more out of date.
**I am not sure about this actually . I had a few talks and we didnt see this a needed requirement so far**
Sometimes the files say Copyright (c) 2024, NVIDIA CORPORATION. All rights reserved. and sometimes they just say Copyright (c) 2024, NVIDIA CORPORATION.. Is there any way we can do an exact string match to ensure that all copyrights are identical?
**I guess we can do this / it's doable**
Should we make sure the Dockerfile has the copyright?
**Not sure, need to check on this / if this needed. And which exact Dockerfile are we talking about?**

.github/workflows/copyright-check.yml
sarahyurick: Thanks! From https://github.com/NVIDIA/NeMo-FW-CI-templates/pull/57, I am thinking we should set `additional-find-args` to `.yml`, `.yaml`, and `.sh`.

sarahyurick:
Thanks @pablo-garay ! A couple more comments:

- A few of our `__init__.py` files are not empty. How should we handle this?
- Thanks for https://github.com/NVIDIA/NeMo-FW-CI-templates/pull/57!
- I will check with the rest of the team about the year mismatches.
- I am referring to this Dockerfile: https://github.com/NVIDIA/NeMo-Curator/blob/main/Dockerfile. I think we should add the copyright, although since there is only 1 Dockerfile in this case I am okay with it not being part of the CI.

PR #383, opened 2024-11-19
Allow users to write to single file

sarahyurick:
Enables something like:
```
input_path = "input.jsonl"
output_path = "output.jsonl"

doc = DocumentDataset.from_pandas(pd.read_json(input_path, lines=True))
doc.to_json(output_path)
```


nemo_curator/utils/distributed_utils.py
```diff
@@ -617,56 +617,68 @@ def write_to_disk(
...
         write_to_filename: Boolean representing whether to write the filename using the "filename" column.
         keep_filename_column: Boolean representing whether to keep or drop the "filename" column, if it exists.
         output_type: The type of output file to write. Can be "jsonl" or "parquet".
 
     """
-    if write_to_filename and "filename" not in df.columns:
+
+    # output_path is a file name
+    if isinstance(output_path, str) and output_path.endswith((".jsonl", ".parquet")):
+        _write_to_jsonl_or_parquet(df.compute(), output_path, output_type)
```
VibhuJawa, line 629: Couple of things:

1. Directories can also end with `.parquet` so do not think this logic works.
2. Should not call `.compute` and write via client (Too much overhead to bring stuff to client)
3. The only recommended way to write to a single file is repartitioning to a single partition and relying on dask to write 


QQ. Why are we doing this ,  this increases run time by a factor of num_workers ?

VibhuJawa (changes requested):
Left a comment asking a bunch of clarifications

nemo_curator/utils/distributed_utils.py
praateekmahajan: @VibhuJawa it [came from here](https://docs.google.com/document/d/19lnrlC7k9lR0PM_xB6M7JXQCa-qRG97USmUPwqe_4u4/edit?tab=t.0#heading=h.dwj89051ed5v)

Though now that I read the requirements more clearly, I don't know if we should support it too, since any distributed data processing system like dask (or spark), will write using multiple workers, hence writing to the same file will always be inefficient since you'll have to perform `.compute() / .collect()` (or `repartition(1)`) to get the data to the one node which will perform the write. 

In my previous job we also had a need for something similar, where we made it a) hard b) very explicit.

An example could be 

```python
DocumentDataset.to_json(..) # write distributed fashion
DocumentDataset.to_single_file(filetype=...) # OOM is possible
```

Let me know what you think?


Also for @VibhuJawa, assuming two things
1. client specs is same as any workers spec
2. the operation after repartition(1) is write (no intermediate map_partition)

Then isn't `.repartition(1)` and `.compute(..)` bottlenecked by the worker (or client)'s memory since the data will need to fit in that one single node, i.e has potential to OOM?

sarahyurick: Yeah, I agree using `compute()` is bad, but I also think it makes sense to allow users to write to a single file if they really want to do it. I imagine this would only happen if the dataset they are working with is pretty small anyway.

I like Praateek's idea of calling it a separate function like `to_single_file` or perhaps adding a parameter to the `to_json` and `to_parquet` function which the user has to explicitly set to write a single file.

VibhuJawa: > client specs is same as any workers spec

Yes. 

> I like Praateek's idea of calling it a separate function like to_single_file or perhaps adding a parameter to the to_json and to_parquet function which the user has to explicitly set to write a single file.

Yeah, we can have `to_single_file` , i think that is still tenable, we should have a warning though if we are doing a repartiton from npartitons > 1 to a single partition.

ayushdg: Looking through the original discussion where this came up I see the comment: `with small datasets with a single partition, this save API is difficult to use`.

Maybe for @Maghoumi use case, just enabling the `to_single_file` method by default if ddf is a single partition (no matter how big or small) is good enough? Is that the case?

nemo_curator/datasets/doc_dataset.py
```diff
@@ -37,6 +37,21 @@ def __len__(self) -> int:
     def persist(self) -> "DocumentDataset":
         return DocumentDataset(self.df.persist())
 
+    def compute(self) -> "DocumentDataset":
+        return DocumentDataset(self.df.compute())
```
praateekmahajan, line 41: If we do this, then the `df` becomes a `pd.DataFrame / cudf.DataFrame` type, isn't?  While `self.df` is expected to be `dd.DataFrame`?

nemo_curator/utils/distributed_utils.py
```diff
@@ -617,54 +617,72 @@ def write_to_disk(
...
+        output_path: The output file path.
         write_to_filename: Boolean representing whether to write the filename using the "filename" column.
         keep_filename_column: Boolean representing whether to keep or drop the "filename" column, if it exists.
         output_type: The type of output file to write. Can be "jsonl" or "parquet".
 
     """
-    if write_to_filename and "filename" not in df.columns:
+
+    # output_path is a file name
+    if isinstance(output_path, str) and output_path.endswith(".jsonl"):
```
praateekmahajan, line 628: @VibhuJawa made a point that `output_path` can end with `.parquet`, the same argument is true for any `.xyz` (including `.jsonl`) since those are legible characters for a directory name, so IMO that's not a strong argument to avoid the check for one extension while not the other.

nemo_curator/datasets/doc_dataset.py
sarahyurick: Oh yeah that's fair. I'll remove this, mainly just thought it would be nice to have the `repartition` function.

nemo_curator/utils/distributed_utils.py
sarahyurick: IMO if a user is specifying to write to "filename.jsonl" it's pretty clear that they want to write to a single file, not a directory. I was wondering if there is some type of use case I am missing here?

I think the difference with "filename.parquet" is that Parquet is more flexible here - you can use it for both the single partition and multi partition case. With this PR, the former case will look like:
```
filename.parquet
    part.0.parquet
```
while the latter will look like:
```
filename.parquet
    part.0.parquet
    part.1.parquet
    ...
```
which are both easily readable with `DocumentDataset.read_parquet`. Meanwhile, if you used `filename.jsonl` and wrote to multiple partitions, you wouldn't be able to read it back with `DocumentDataset.read_json` because the file format would look like:
```
filename.jsonl
    0.part
    1.part
    ...
```
And you would get a directory-related error. So IMO the `.jsonl` format should only write out a single partition `.jsonl` file.

nemo_curator/utils/distributed_utils.py
Maghoumi: Sarah and I had a quick discussion offline. She told me the new way this is done is such that if the dataframe is a single partition, then it writes to a single file. If it's not a single partition, then a runtime warning is issued, the output path is treated as a directory and the data set will be written in multiple files (one for each partition).

Having read this discussion and the one below, I think this is the most reasonable way of handling it and won't require adding another API such as `to_single_file`.

nemo_curator/utils/distributed_utils.py
praateekmahajan: Thanks for explaining in great detail! This makes sense.

praateekmahajan (approved):

nemo_curator/datasets/doc_dataset.py
```diff
@@ -37,6 +37,18 @@ def __len__(self) -> int:
...
+        self,
+        divisions: Optional[List[int]] = None,
+        npartitions: Optional[int] = None,
+        partition_size: Optional[Union[int, str]] = None,
+        freq: Optional[str] = None,
+        force: bool = False,
+    ) -> "DocumentDataset":
+        return DocumentDataset(
+            self.df.repartition(divisions, npartitions, partition_size, freq, force)
+        )
```
ayushdg, line 50: It might be worth exploring `functools.wraps` here to wrap df's repartition method. IIuc it might allow inheriting the function signature and docstrings, while not having to redefine all args.
According to LLM's
```python
    @wraps(dd.DataFrame.repartition)  # Ensures the docstring and signature are inherited
    def repartition(self, *args, **kwargs):
        return self.__class__(self.df.repartition(*args, **kwargs))
```

nemo_curator/utils/distributed_utils.py
```diff
@@ -617,54 +617,72 @@ def write_to_disk(
...
+        if df.npartitions == 1:
+            _write_to_jsonl_or_parquet(df.compute(), output_path, output_type)
```
ayushdg, line 630: Instead of `df.compute()` I wonder if `df.map_partitions(_write_to_jsonl).compute()` or other equivalent like `df.get_partition(0)...` might work in this case. It avoids the latency of pulling the dataframe from one of the worker processes to the client process and instead uses the worker to write out to a single file instead.

ayushdg (commented):
Left small suggestions around `compute` and nit around `repartition`

nemo_curator/datasets/doc_dataset.py
sarahyurick: Thanks!

nemo_curator/utils/distributed_utils.py
sarahyurick: Very nice, thank you. Updated.

sarahyurick:
Thanks @ayushdg ! Updated, LMK what you think.

ayushdg (approved):
lgtm, thanks!

PR #384, opened 2024-11-19
Global cache variable for exact, fuzzy, and semantic deduplication

sarahyurick:


nemo_curator/modules/semantic_dedup.py
```diff
@@ -448,31 +453,38 @@ def __init__(
...
+                "No cache directory specified; please use initialize_cache_directory"
+            )
+        else:
+            self.emb_by_clust_dir = os.path.join(
+                get_cache_directory(), "clustering", "embs_by_nearest_center"
+            )
+            self.sorted_clusters_dir = os.path.join(
+                get_cache_directory(), "clustering", "sorted"
+            )
+            self.output_dir = os.path.join(get_cache_directory(), "clustering")
```
sarahyurick, line 483:
```suggestion
            self.output_dir = os.path.join(get_cache_directory(), "duplicates")
```
?

nemo_curator/cache.py
```diff
@@ -0,0 +1,34 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from nemo_curator.utils.file_utils import expand_outdir_and_mkdir
+
+# Global variable to store the cache directory
+_global_cache_dir = None
```
Maghoumi, line 18: Instead of a global variable, I think we should the singleton design pattern here, where there is a `Cache` class that creates the static singleton instance (if it doesn't exist), or returns an existing one (if it was already created).

Subsequently, the init and get methods can be nicely wrapped inside that class.

Maghoumi: @sarahyurick please let me know what you think and whether my comment makes sense to you

sarahyurick: Thank you! Yes, that makes sense to me. Sorry I haven't gotten back to this earlier, but I expect to have more changes ready next week.

nemo_curator/modules/semantic_dedup/clusteringmodel.py
```diff
@@ -47,13 +47,15 @@ def add_dist_to_cents(
...
+# Clustering module
 class ClusteringModel:
     def __init__(
         self,
         id_column: str,
         max_iter: int,
         n_clusters: int,
-        clustering_output_dir: str,
+        cache_dir: Optional[str] = None,
+        clustering_save_loc: str = "clustering_results",
```
sarahyurick, line 58: We deprecate `clustering_output_dir` in favor of `cache_dir` and `clustering_save_loc` for `ClusteringModel`, which matches the logic in the `SemDedup` class.

nemo_curator/modules/semantic_dedup/embeddings.py
```diff
@@ -114,7 +114,8 @@ def __init__(
         self,
         embedding_model_name_or_path: str,
         embedding_batch_size: int,
-        embedding_output_dir: str,
+        cache_dir: Optional[str] = None,
+        embeddings_save_loc: str = "embeddings",
```
sarahyurick, line 118: We deprecate `embedding_output_dir` in favor of `cache_dir` and `embeddings_save_loc` for `EmbeddingCreator`, which matches the logic in the `SemDedup` class.

nemo_curator/modules/semantic_dedup/semanticclusterleveldedup.py
```diff
@@ -51,31 +52,52 @@ def __init__(
 
         Args:
             n_clusters (int): Number of clusters.
-            emb_by_clust_dir (str): Directory containing embeddings by cluster.
-            sorted_clusters_dir (str): Directory containing sorted clusters.
```
sarahyurick, line 55: We deprecate `emb_by_clust_dir` and `sorted_clusters_dir` in favor of `cache_dir` and `embeddings_save_loc` for `SemanticClusterLevelDedup`, which matches the logic in the `SemDedup` class.

sarahyurick (commented):
This PR allows the user to use `Cache(cache_dir=...)` instead of having to specify the `cache_dir` at every step of the deduplication pipelines.

However, the user does not have to use `Cache` if they want to keep using the modules the same way as before. Both ways work, with preference if the user specifies the `cache_dir` field in a deduplication module.

The only hard changes are with the semantic deduplication fields, see below.

Maghoumi:
Thanks so much for working on this change. What I like about this now is that it gives users the option to either use the same cache directory for anything that requires caching, or provide a specific directory if they don't want to re-use the same cache.

The cache class implementation is functional but not thread-safe. I don't think that's a blocking problem for this PR.

I didn't run the samples/tutorials, but I assume the change has been thoroughly verified?

sarahyurick:
> Thanks so much for working on this change. What I like about this now is that it gives users the option to either use the same cache directory for anything that requires caching, or provide a specific directory if they don't want to re-use the same cache.
> 
> The cache class implementation is functional but not thread-safe. I don't think that's a blocking problem for this PR.
> 
> I didn't run the samples/tutorials, but I assume the change has been thoroughly verified?

Thanks! Yes, I tried to make as few breaking changes as possible. The examples and tutorials should all reflect those changes.

Maghoumi (approved):

config/sem_dedup_config.yaml
```diff
@@ -11,13 +11,11 @@ write_embeddings_to_disk: true
 # Clustering configuration
 clustering_save_loc: "clustering_results"
 n_clusters: 1000
-seed: 1234
 max_iter: 100
 kmeans_with_cos_dist: false
 
 # Semdedup configuration
 which_to_keep: "hard"
-largest_cluster_size_to_process: 100000
```
sarahyurick, line 20: IIRC, I removed `seed` and `largest_cluster_size_to_process` from the semantic dedupe parameters because I couldn't find them being used anywhere.

PR #386, opened 2024-11-21
Fuzzy Dedup: Make skipping the False positive check the default

ayushdg:
## Description

This PR updates the FuzzyDedup modules making `false_positive_check=False` and `char_ngrams=24` the default params going forward.

Motivation:
1. Internal testing and external research has shown that depending on the values of `num_buckets` and `hashes_per_bucket`, the false positive rate can be minimized based on intended jaccard similarity. 
2. The defaults of `20` buckets and `13` hashes per bucket has a low false positive rate for documents with similarity < 0.8 and internal testing has shown a difference of 1-2% when skipping the fp check. 
3. Increasing the default minhash length to `24` which roughly corresponds to 5 word shingles, results in removal rates which are very similar to 5 char ngrams with the fp check enabled since 5 char ngrams are prone to more false positives.
4. The false positive check is computationally expensive and has significantly higher memory requirements. It uses requires significantly higher cache in memory keeping an additional copy of the corpus in addition to minhashes and buckets. 

These new defaults should produce similar results to the previous defaults while being significantly less compute intensive and faster.

Will update tutorials in a followup PR.

## Usage

```python
```
## Checklist


nemo_curator/scripts/fuzzy_deduplication/README.md
```diff
@@ -1,99 +1,2 @@
...
-          --output-dir /path/to/dedup_output \
-          --cache-dir /path/to/cc_cache \
-          --jaccard-threshold 0.8
-          # --scheduler-file /path/to/file.json
-    ```
-
-> [!TIP]
-> When using these scripts in a multi-node environment (like Slurm, K8's etc.) it is recommended to start up a Dask cluster prior to execution and connect to the existing cluster via the `--scheduler-address` or `--scheduler-file` flag.
-> Use the `--help` flag to view all possible CLI options for the scripts and details on what they do.
+> The up to date documentation on running the Fuzzy Deduplication scripts can be found in the [NeMo-User guide](https://docs.nvidia.com/nemo-framework/user-guide/latest/datacuration/gpudeduplication.html#id4). It is recommended to use the Python API directly rather than the CLI scripts for most cases.
```
sarahyurick, line 2:
```suggestion
> The up to date documentation on running the fuzzy deduplication scripts can be found in the [NeMo Curator User Guide](https://docs.nvidia.com/nemo-framework/user-guide/latest/datacuration/gpudeduplication.html#id4). It is recommended to use the Python API directly rather than the CLI scripts for most cases.
```

nemo_curator/modules/config.py
```diff
@@ -67,47 +67,67 @@ class FuzzyDuplicatesConfig(BaseConfig):
...
-    char_ngrams: int = 5
+    char_ngrams: int = 24
     num_buckets: int = 20
     hashes_per_bucket: int = 13
     use_64_bit_hash: bool = False
     buckets_per_shuffle: int = 1
 
-    false_positive_check: bool = True
+    false_positive_check: bool = False
     # Only required for fp check
```
sarahyurick, line 77:
```suggestion
    # Only required for false positive check
```
Nit, but I prefer spelling out `false_positive` each time. Going to be making a similar comment about "fpp" on https://github.com/NVIDIA/NeMo-Curator/pull/285 but I think it can get pretty difficult to modify our codebase otherwise.

nemo_curator/modules/config.py
```diff
@@ -67,47 +67,67 @@ class FuzzyDuplicatesConfig(BaseConfig):
...
+    parts_per_worker: Optional[int] = None
+    bucket_parts_per_worker: Optional[int] = None
 
     def __post_init__(self):
         self.num_hashes = self.num_buckets * self.hashes_per_bucket
-        if self.cache_dir is None:
-            raise ValueError(
-                "Finding fuzzy duplicates requires a cache directory accessible via all workers to store intermediates"
-            )
+        fp_defaults = {
```
sarahyurick, line 86:
```suggestion
        false_positive_defaults = {
```

nemo_curator/modules/config.py
```diff
@@ -67,47 +67,67 @@ class FuzzyDuplicatesConfig(BaseConfig):
...
-                "Using a small char_ngrams value might lead to a large number (~5%) of false positives during deduplication."
-                " Using a value of at least 20 for char_ngrams is recommended."
-            )
-        if self.num_anchors <= 0:
-            raise ValueError("Number of anchors must be greater than 0")
-        if self.num_anchors > 2:
-            warnings.warn(
-                "Using a higher number of anchor docs might lead to higher memory footprint and might impact performance",
-                category=UserWarning,
+            for arg, default in fp_defaults.items():
```
sarahyurick, line 98:
```suggestion
            for arg, default in false_positive_defaults.items():
```

nemo_curator/modules/config.py
```diff
@@ -67,47 +67,67 @@ class FuzzyDuplicatesConfig(BaseConfig):
...
+                )
+            if not 0 <= self.jaccard_threshold <= 1:
+                raise ValueError("Jaccard Threshold must be between [0,1]")
+        else:
+            if self.char_ngrams < 20:
+                warnings.warn(
+                    "Using a small char_ngrams value might lead to a large number (~5%) of false positives during deduplication."
+                    " Using a value of at least 20 for char_ngrams is recommended."
+                )
+            unused_fp_args = [
```
sarahyurick, line 116:
```suggestion
            unused_false_positive_args = [
```

nemo_curator/modules/config.py
```diff
@@ -67,47 +67,67 @@ class FuzzyDuplicatesConfig(BaseConfig):
...
+                arg for arg in fp_defaults.keys() if getattr(self, arg) is not None
```
sarahyurick, line 117:
```suggestion
                arg for arg in false_positive_defaults.keys() if getattr(self, arg) is not None
```

nemo_curator/modules/config.py
```diff
@@ -67,47 +67,67 @@ class FuzzyDuplicatesConfig(BaseConfig):
...
+            ]
+            if unused_fp_args:
```
sarahyurick, line 119:
```suggestion
            if unused_false_positive_args:
```

nemo_curator/modules/config.py
```diff
@@ -67,47 +67,67 @@ class FuzzyDuplicatesConfig(BaseConfig):
...
+                warnings.warn(
+                    f"False positive check is disabled. Unused arguments {unused_fp_args} will be ignored",
```
sarahyurick, line 121:
```suggestion
                    f"False positive check is disabled. Unused arguments {unused_false_positive_args} will be ignored",
```

sarahyurick (commented):
I think making the default `false_positive_check=False` makes sense. Added some nits, otherwise are there any other files that should be updated? Maybe:

- https://github.com/NVIDIA/NeMo-Curator/blob/main/config/fuzzy_dedup_config.yaml
- https://github.com/NVIDIA/NeMo-Curator/tree/main/tutorials/zyda2-tutorial/1_fuzzy_dedup
- https://github.com/NVIDIA/NeMo-Curator/blob/main/docs/user-guide/gpudeduplication.rst

?

ayushdg:
closing in favor of #498 to simplify conflict resolution.

PR #387, opened 2024-11-22
Fix GPU error messages for fuzzy deduplication

sarahyurick:
- [x] Raise a clear and readable error if the Dask client is not GPU-based
- [x] Raise a clear and readable error if the dataset’s backend is not cuDF


sarahyurick:
cc @Maghoumi please let me know if these error messages look more how you would expect them.

nemo_curator/modules/fuzzy_dedup/bucketstoedges.py
```diff
@@ -124,6 +124,12 @@ def buckets_to_edges(
         return result_df
 
     def __call__(self, dataset: DocumentDataset) -> DocumentDataset:
+        if "cudf" not in str(type(dataset.df)):
```
Maghoumi, line 127: Since this pattern is repeated, it might be better to move to a separate helper (e.g. `checks.py`) so it can be re-used in different places.

Something like: `ensure_cudf(dataset, module="buckets_to_edges")` which would print something like:
```
A Dask DataFrame with cuDF backend is required to run 'buckets_to_edges'. Please convert using....
```

Maghoumi (approved):

nemo_curator/modules/fuzzy_dedup/bucketstoedges.py
sarahyurick: Thanks, I see what you mean but for now I'm not sure whether it's worth adding a new script for this. I will ask others to review as well, thank you.

nemo_curator/datasets/doc_dataset.py
```diff
@@ -38,6 +38,9 @@ def __len__(self) -> int:
     def persist(self) -> "DocumentDataset":
         return DocumentDataset(self.df.persist())
 
+    def to_backend(self, backend: Optional[str] = None) -> "DocumentDataset":
+        return DocumentDataset(self.df.to_backend(backend))
+
```
ayushdg, line 43: Does it make sense to use the `ToBackend` module here or add some of the same checks as the __init__ method from there?

cc: @ryantwolf if you have thoughts.

nemo_curator/modules/fuzzy_dedup/bucketstoedges.py
```diff
@@ -124,6 +124,12 @@ def buckets_to_edges(
         return result_df
```
ayushdg, line 124: Alternatively this class since it follows the same `__call__` pattern can probably inherit from BaseModule?

nemo_curator/modules/fuzzy_dedup/bucketstoedges.py
ayushdg, line 127: Can call the `is_cudf_type` utility instead of the explicit check here instead.

nemo_curator/modules/fuzzy_dedup/connectedcomponents.py
```diff
@@ -58,7 +58,7 @@ def __init__(
         else:
             self._logger = logger
 
-    def cc_workflow(self, output_path):
+    def __call__(self, output_path):
```
ayushdg, line 61: The main reason this was not kept as the standard `__call__` method is because those typically have an assumption of `DocDataset` in `DocDataset` out. Since this doesn't follow the same pattern, it cannot be chained well in `Sequential`. Not sure if we should revert back to the old name or expand the definition of what this method can do.

nemo_curator/modules/fuzzy_dedup/fuzzyduplicates.py
```diff
@@ -155,6 +155,11 @@ def identify_duplicates(
         DocumentDataset containing IDs of all documents and the corresponding duplicate group
         they belong to. Documents in the same group are near duplicates.
         """
+        if "cudf" not in str(type(dataset.df)):
```
ayushdg, line 158: Same comment as above

nemo_curator/modules/fuzzy_dedup/fuzzyduplicates.py
```diff
@@ -189,14 +194,14 @@ def identify_duplicates(
                     mapped_buckets_w_anchors_path, write_index=False, overwrite=True
                 )
             self._logger.info(
-                f"Time taken for Map_buckets : {time.time() - t0}s and output written at {mapped_buckets_w_anchors_path}"
+                f"Time taken for Map_Buckets : {time.time() - t0}s and output written at {mapped_buckets_w_anchors_path}"
             )
 
-            print(f"Stage {stage_num} (False Postive Check): Map_Buckets Complete!")
+            print(f"Stage {stage_num} (False Positive Check): Map_Buckets complete!")
```
ayushdg, line 200: Lol thanks for updating.

nemo_curator/modules/fuzzy_dedup/lsh.py
```diff
@@ -272,6 +272,12 @@ def _write_bucket_parquet(
         return wrote_buckets, are_buckets_empty
 
     def __call__(self, dataset: DocumentDataset) -> DocumentDataset:
+        if "cudf" not in str(type(dataset.df)):
```
ayushdg, line 275: Since this is a public class with a `__call__` method can default to inheriting from BaseModule instead.

nemo_curator/modules/fuzzy_dedup/minhash.py
```diff
@@ -202,6 +203,12 @@ def __call__(self, dataset: DocumentDataset) -> Union[str, DocumentDataset]:
         -------
         DocumentDataset containing IDs of all documents and the corresponding MinHash Signature
         """
+        if "cudf" not in str(type(dataset.df)):
```
ayushdg, line 206: Same

ayushdg (changes requested):
Minor nits around the explicit cudf type checks and some questions around inheriting some of these from the BaseModule

sarahyurick:
Should not be needed after Ray refactor.

PR #407, opened 2024-12-02
Add pin to NeMo Toolkit

sarahyurick:
Closes https://github.com/NVIDIA/NeMo-Curator/issues/399

### DIFF
--- a/requirements/requirements.txt
+++ b/requirements/requirements.txt
@@ -13,7 +13,7 @@ jieba==0.42.1
 justext==3.0.1
 lxml_html_clean
 mwparserfromhell==0.6.5
-nemo_toolkit[nlp]<2.0.0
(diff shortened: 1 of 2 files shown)


requirements/requirements.txt
```diff
@@ -13,7 +13,7 @@ jieba==0.42.1
 justext==3.0.1
 lxml_html_clean
 mwparserfromhell==0.6.5
-nemo_toolkit[nlp]>=1.23.0
+nemo_toolkit[nlp]<2.0.0
```
ayushdg, line 16: could we keep both the lower and upper pin in this case? Highly unlikely but want to avoid installation if there's an older version of toolkit than 1.23.0

sarahyurick: Done, thanks.

ayushdg (approved):
lgtm, Thanks!

PR #409, opened 2024-12-03
Create separate files for each deduplication class

sarahyurick:
Re-opening https://github.com/NVIDIA/NeMo-Curator/pull/389 because of merge issues.


ayushdg:
cc: @VibhuJawa Could you take a look for semdedup as well?

nemo_curator/scripts/semdedup/clustering.py
```diff
@@ -18,10 +18,10 @@
 
 import dask_cudf
 
+from nemo_curator import ClusteringModel
```
VibhuJawa, line 21: We should keep this as semnatic dedup specific import (because clustering can be generic). So something like:

VibhuJawa (commented):
Reviwed the semdedup functions, my only nit pick is to keep the module level import for stuff like `ClusteringModel `, `EmbeddingCreator` etc because they can also be generic in nature so it is a easier mental model.

nemo_curator/scripts/semdedup/clustering.py
sarahyurick: It looks like `EmbeddingCreator`, `ClusteringModel`, and `SemanticClusterLevelDedup` are currently available as generic imports in NeMo Curator: https://github.com/NVIDIA/NeMo-Curator/blob/main/docs/user-guide/api/deduplication.rst

Should I change it to only be importable via `nemo_curator.semantic_dedup` instead?

ayushdg (approved):

nemo_curator/scripts/semdedup/clustering.py
sarahyurick: Hi @VibhuJawa for now I am going to merge as is (as I have several PRs related to dedupe that I would like to update with this PR), and I can create a separate PR if we decide to change the import logic. Thanks!

PR #421, opened 2024-12-11
Add `tests/test_classifiers.py` PyTests

sarahyurick:
Closes https://github.com/NVIDIA/NeMo-Curator/issues/405.
Closes https://github.com/NVIDIA/NeMo-Curator/issues/419.

### DIFF
--- a/.github/workflows/gpuci.yml
+++ b/.github/workflows/gpuci.yml
@@ -78,7 +78,7 @@ jobs:
       # and then the directory where the PyTests are located
     - name: Run PyTests with GPU mark
       run: |
-        docker exec -e HUGGING_FACE_API_TOKEN=${{ secrets.HUGGING_FACE_API_TOKEN }} nemo-curator-container pytest -m gpu --rootdir /opt/NeMo-Curator /opt/NeMo-Curator/tests
+        docker exec nemo-curator-container pytest -m gpu --rootdir /opt/NeMo-Curator /opt/NeMo-Curator/tests
 
       # After running `docker stop`, the container remains in an exited state
       # It is still present on our system and could be restarted with `docker start`
--- a/.github/workflows/gpuci.yml
+++ b/.github/workflows/gpuci.yml
@@ -78,7 +78,7 @@ jobs:
       # and then the directory where the PyTests are located
     - name: Run PyTests with GPU mark
       run: |
-        docker exec nemo-curator-container pytest -m gpu --rootdir /opt/NeMo-Curator /opt/NeMo-Curator/tests
+        docker exec -e HUGGING_FACE_API_TOKEN=${{ secrets.HUGGING_FACE_API_TOKEN }} nemo-curator-container pytest -m gpu --rootdir /opt/NeMo-Curator /opt/NeMo-Curator/tests
 
       # After running `docker stop`, the container remains in an exited state
       # It is still present on our system and could be restarted with `docker start`
--- a/tests/test_classifiers.py
+++ b/tests/test_classifiers.py
@@ -260,8 +260,8 @@ def test_prompt_task_complexity_classifier(gpu_client):
     expected_pred["contextual_knowledge"] = round(
         expected_pred["contextual_knowledge"], 3
     )
-    result_pred["creativity_scope"] = round(result_pred["creativity_scope"], 3)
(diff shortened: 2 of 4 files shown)


.github/workflows/gpuci.yml
```diff
@@ -78,7 +78,7 @@ jobs:
       # and then the directory where the PyTests are located
     - name: Run PyTests with GPU mark
       run: |
-        docker exec nemo-curator-container pytest -m gpu --rootdir /opt/NeMo-Curator /opt/NeMo-Curator/tests
+        docker exec -e HUGGING_FACE_API_TOKEN=${{ secrets.HUGGING_FACE_API_TOKEN }} nemo-curator-container pytest -m gpu --rootdir /opt/NeMo-Curator /opt/NeMo-Curator/tests
```
sarahyurick, line 81: We need to add a GitHub secret for this to work.

tests/test_classifiers.py
```diff
@@ -0,0 +1,265 @@
...
+        "nvidia/Aegis-AI-Content-Safety-LlamaGuard-Defensive-1.0",
+        "nvidia/Aegis-AI-Content-Safety-LlamaGuard-Permissive-1.0",
+    ],
+)
+def test_aegis_classifier(gpu_client, aegis_variant):
+    from nemo_curator.classifiers import AegisClassifier
+
+    hugging_face_api_token = os.getenv("HUGGING_FACE_API_TOKEN")
+    if not hugging_face_api_token:
+        pytest.fail("HUGGING_FACE_API_TOKEN environment variable not set")
```
sarahyurick, line 104: We grab the Hugging Face token here instead of using a `@pytest.fixture`. The reason for this is that if we used a PyTest fixture and the test failed for whatever reason, the Hugging Face token would be exposed in the logs.

.github/workflows/gpuci.yml
sarahyurick: cc @ryantwolf

VibhuJawa (approved):
This looks good to me. Thanks a ton for adding these tests @sarahyurick . 

I think we should be good to merge this once we resolve how to pass in HF Secret

tests/test_classifiers.py
VibhuJawa: If there is not a short term fix here, we can probably skip this test using `skipif` . 

```
@pytest.mark.skipif(
    not os.getenv("HUGGING_FACE_API_TOKEN"),
    reason="HUGGING_FACE_API_TOKEN environment variable is not set"
)
```

tests/test_classifiers.py
```diff
@@ -0,0 +1,265 @@
...
+            "Travel_and_Transportation",
+        ]
+    )
+
+    assert result_pred.equals(expected_pred)
+
+
+@pytest.mark.skip(
+    reason="Skipping until https://github.com/NVIDIA/NeMo-Curator/pull/361 is merged"
+)
```
sarahyurick, line 214: TODO: Remove this.

tests/test_classifiers.py
```diff
@@ -0,0 +1,265 @@
...
+    result_pred = result_dataset.df.compute()["content_pred"]
+
+    expected_pred = cudf.Series(["Online Comments"])
+
+    assert result_pred.equals(expected_pred)
+
+
+@pytest.mark.skip(
+    reason="Skipping until https://github.com/NVIDIA/NeMo-Curator/pull/364 is merged"
+)
```
sarahyurick, line 234: TODO: Remove this.

tests/test_classifiers.py
```diff
@@ -0,0 +1,323 @@
...
+            "no_label_reason": [0.0],
+            "number_of_few_shots": [0],
+            "prompt_complexity_score": [0.2783],
+            "reasoning": [0.0632],
+            "task_type_1": ["Code Generation"],
+            "task_type_2": ["Text Generation"],
+            "task_type_prob": [0.767],
+            "text": text,
+        }
+    )
```
sarahyurick, line 253: This passes on my remote machine (both my conda setup and my Docker setup) but is failing on our gpuCI. I think perhaps it is related to floating point errors on different machines? I will probably just add some rounding estimations so that it passes both locally for me and here on GitHub.

sarahyurick: cc @VibhuJawa

Issue #417, opened 2024-12-10
PII Modifier should support documents greater than pre-configured length


praateekmahajan:
**Is your feature request related to a problem? Please describe.**

Under the hood Pii Modifier uses Presidio (which uses spacy I believe).
Currently if the documents are very long (I believe 2,000,000 chars in length) then presidio errors out. See https://github.com/microsoft/presidio/issues/957#issuecomment-1333654029

To solve this we might need to modify the `nlp` object that `Presidio` is using or create a new class that allows the modification of character limit.

For now to avoid this I've done

```python
    dataset = DocumentDataset(
        dataset.df[dataset.df['text'].map(
            lambda x : len(x) < 2_000_000, meta={"text" : int}
        )]
    )
```

**Describe the solution you'd like**
support a `max_char_length` as an arg e.g. `PiiModifier(max_char_length=3_000_000)`

**Describe alternatives you've considered**

**Additional context**

PR #423, opened 2024-12-11
Dockerfile should use the correct nvidia channel to install libcu*

praateekmahajan:
## Description
Realised that while we're installing cuda-12.5.1 in the CI image, the conda installations are all over the place and some of them are being picked up from conda-forge rather than nvidia. (IIUC -c A -c B, A takes precedence over B, and in our case conda-forge might have older versions of packages)

Secondly if we wish to use UDFs on `dask_cudf` we need `NVVM`. The place where I needed it was hack for #417

## Checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -29,7 +29,7 @@ LABEL "nemo.library"=${IMAGE_LABEL}
 WORKDIR /opt
 
 # Install the minimal libcu* libraries needed by NeMo Curator
-RUN conda create -y --name curator -c nvidia/label/cuda-${CUDA_VER} -c conda-forge
+RUN conda create -y --name curator -c nvidia/label/cuda-${CUDA_VER} -c conda-forge \
   python=3.10 \
   cuda-cudart \
   libcufft \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -29,14 +29,15 @@ LABEL "nemo.library"=${IMAGE_LABEL}
 WORKDIR /opt
 
 # Install the minimal libcu* libraries needed by NeMo Curator
-RUN conda create -y --name curator -c conda-forge -c nvidia \
+RUN conda create -y --name curator -c nvidia/label/cuda-${CUDA_VER} -c conda-forge
   python=3.10 \
```
sarahyurick, line 33: Looks like gpuCI is currently throwing an error here.

sarahyurick (changes requested):

Dockerfile
praateekmahajan: Apparently build arg's can't be in `RUN`, so had to create env var, which is ugly

sarahyurick: Oh lol, interesting. LGTM!

sarahyurick (approved):

PR #431, opened 2024-12-13
Add `TrafilaturaExtractor` class

sarahyurick:
Closes https://github.com/NVIDIA/NeMo-Curator/issues/400.


tests/test_download.py
```diff
@@ -82,6 +91,17 @@ def test_resiliparse_extract_text(self):
...
         assert result == expected
 
+    def test_trafilatura_extract_text(self, html_string):
+        algorithm = TrafilaturaExtractor()
+        stop_words = get_stop_list_dict()
+        result = algorithm.extract_text(html_string, stop_words["ENGLISH"])
+
+        expected = [
+            "Let's keep this paragraph: either came does last new took taken making became from.",
+        ]
```
sarahyurick, line 101: Trafilatura has a really bad bug where it is returning the string twice. I can double check all of my logic, but in the case that this is a Trafilatura-specific issue, I am debating whether I should add our own exact deduplication code into this.

(Trafilatura actually has their own `deduplicate` parameter [here](https://github.com/adbar/trafilatura/blob/b0107793454f7c6334d87f1228e0ff24d19c02b7/trafilatura/core.py#L365), but it did not fix this use case.)

Possible related issue: https://github.com/adbar/trafilatura/issues/634

sarahyurick: I have not scoped to see what conditions cause this.

sarahyurick: Another issue: https://github.com/adbar/trafilatura/issues/768

sarahyurick: Solved by adding support for these extraction and deduplication parameters: https://trafilatura.readthedocs.io/en/latest/settings.html.

docs/user-guide/download.rst
```diff
@@ -128,15 +131,15 @@ You can choose to modify the HTML text extraction algorithm used in ``download_c
   if __name__ == "__main__":
       main()
 
-Above, we changed the extraction algorithm from the default ``JusTextExtractor``.
+Above, we changed the extraction algorithm from the default ``JusTextExtractor``. **Note:** Please see the Trafilatura documentation `here <https://trafilatura.readthedocs.io/en/latest/settings.html>`_
```
ryantwolf, line 134: Why should the user see the Trafilatura documentation?

nemo_curator/download/commoncrawl.py
```diff
@@ -200,6 +203,98 @@ def extract_text(self, html, stop_words):
         return result
 
 
+class TrafilaturaExtractor(HTMLExtractorAlgorithm):
```
ryantwolf, line 206: Can you add a class level docstring explaining what trafilatura is?

tests/test_download.py
ryantwolf: This seems like an issue customers might run into. Have you looked in papers and such for good default values? If Trafilatura's defaults aren't good (or aren't used by most researchers), it could be good to substitute our own.

ryantwolf (commented):
A few minor things, looks good overall.

tests/test_download.py
sarahyurick: Yeah, I can look into it. At the very least, I think setting `deduplicate=True` as the default in NeMo Curator might be a good idea. Otherwise, I think we could make sure to emphasize these parameters in the documentation.

nemo_curator/download/commoncrawl.py
```diff
@@ -200,6 +235,117 @@ def extract_text(self, html, stop_words):
...
+        )
+        if self.max_tree_size:
+            trafilatura_config["DEFAULT"]["MAX_TREE_SIZE"] = str(self.max_tree_size)
+        trafilatura_config["DEFAULT"]["MIN_DUPLCHECK_SIZE"] = str(
+            self.min_duplcheck_size
+        )
+        trafilatura_config["DEFAULT"]["MAX_REPETITIONS"] = str(self.max_repetitions)
+
+        # Recommended to set deduplicate=True
+        self.extract_kwargs.setdefault("deduplicate", True)
```
ryantwolf, line 322: Can you mention in the docstring that this happens?

sarahyurick: Updated, thanks!

ryantwolf (approved):

PR #432, opened 2024-12-16
Updating the Quick Example

stsfaroz:
Import the download_common_crawl function from nemo_curator

## Description
The Quick Example is missing the import statement, so it has been added.


ayushdg (approved):
Thanks for the fix!

PR #436, opened 2024-12-17
[WIP] Add RAPIDS Nightly to GPU CI

praateekmahajan:
## Usage

```python
```
## Checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,6 +6,7 @@ ARG PYTHON_VER=3.10
 ARG IMAGE_LABEL
 ARG REPO_URL
 ARG CURATOR_COMMIT
+ARG BUILD_TYPE=stable
 
 FROM rapidsai/ci-conda:cuda${CUDA_VER}-${LINUX_VER}-py${PYTHON_VER} as curator-update
 # Needed to navigate to and pull the forked repository's changes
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -48,15 +51,23 @@ RUN \
 --mount=type=bind,source=/opt/NeMo-Curator/pyproject.toml,target=/opt/NeMo-Curator/pyproject.toml,from=curator-update \
   cd /opt/NeMo-Curator && \
   source activate curator && \
-  pip install ".[all]"
+  if [ "$BUILD_TYPE" = "nightly" ]; then \
+    pip install ".[all_nightly]"; \
+  else \
+    pip install ".[all]"; \
+  fi
```
sarahyurick, line 58: We might need to check if the nightly can use the same caching logic as stable, or if nightly should just always build a fresh non-cached container.

praateekmahajan: +1, wanted to flag this explicitly as we do some curator-update optimization above, and I want to ensure if branching doesn't mess with it.
cc @ko3n1g

ko3n1g: Do these optimizations come from the base-image?

ko3n1g: I mean, if they do even if the base-image tag is unchanged b/c they overwrite it we should extend this build-workflow so that we can force pull the base image. In all other cases we don't need to be worried about caching i believe

ko3n1g:
Ok so here my suggestion:

* We run the gpuci workflow on nightly schedule (via github workflows cron trigger)
* We use `use-cache: false` to make sure we build from scratch with latest dependencies (merge #448 before)
* It doesn't matter whether we build two containers or one with two venvs. Initially I thought it would be easier for debugging we if we had a single container, but two containers (like you already prepared) have the benefit that size / container is smaller which come in handy next year
* Tip: `if: ${{ github.event.label.name == 'gpuci' || github.ref == 'refs/heads/main' }}` needs to be changed to `if: ${{ github.event.label.name == 'gpuci' || github.ref == 'refs/heads/main' || github. event_name == 'schedule' }}`

PR #445, opened 2024-12-20
Support the new minhash 25.02 api

praateekmahajan:
## Description
Solves https://github.com/NVIDIA/NeMo-Curator/issues/426 and piggy backs on https://github.com/NVIDIA/NeMo-Curator/pull/442/files

## Usage

```python
```
## Checklist

### DIFF
--- a/nemo_curator/_compat.py
+++ b/nemo_curator/_compat.py
@@ -39,10 +39,15 @@ try:
 except (ImportError, TypeError):
     CURRENT_CUDF_VERSION = parse_version("24.10.0")
 
-# TODO remove this once 24.12.0 becomes the base version of cudf in nemo-curator
-MINHASH_PERMUTED_AVAILABLE = CURRENT_CUDF_VERSION >= parse_version("24.12.0") or (
-    CURRENT_CUDF_VERSION.is_prerelease
-    and CURRENT_CUDF_VERSION.base_version >= "24.12.0"
+# TODO remove this once 25.02 becomes the base version of cudf in nemo-curator
+
+# minhash in < 24.12 used to have a minhash(txt) api which was deprecated in favor of
+# minhash(a, b) in 25.02 (in 24.12, minhash_permuted(a,b) was introduced)
+MINHASH_DEPRECATED_API = (
+    CURRENT_CUDF_VERSION.base_version < parse_version("24.12").base_version
+)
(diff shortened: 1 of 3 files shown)


nemo_curator/_compat.py
```diff
@@ -39,10 +39,15 @@
...
-    and CURRENT_CUDF_VERSION.base_version >= "24.12.0"
+# TODO remove this once 25.02 becomes the base version of cudf in nemo-curator
+
+# minhash in < 24.12 used to have a minhash(txt) api which was deprecated in favor of
+# minhash(a, b) in 25.02 (in 24.12, minhash_permuted(a,b) was introduced)
+MINHASH_DEPRECATED_API = (
+    CURRENT_CUDF_VERSION.base_version < parse_version("24.12").base_version
+)
+MINHASH_PERMUTED_AVAILABLE = (CURRENT_CUDF_VERSION.major == 24) & (
+    CURRENT_CUDF_VERSION.minor == 12
```
sarahyurick, line 50: Wait I'm confused, why are we doing it for 24.10 and 24.12 instead of for 24.12 and 25.02?

praateekmahajan: Maybe the code [block here helps](https://github.com/NVIDIA/NeMo-Curator/pull/445/files#diff-b836d89c298548dd47480ece278782f030d04e2beb421c59c0e546718aede6c3R176-R196)



```python
if MINHASH_DEPRECATED_API:
     # this is 24.10, the variable is named minhash deprecated api, 
     # because now cudf only needs four args, rather than two args
    return ser.str.minhash(seeds=seeds, width=char_ngram)
else:
    if MINHASH_PERMUTED_AVAILABLE: 
        # this is 24.12 because in this version the four arg function is called minhash_permuted
        return ser.str.minhash_permuted(
            a=seeds_a, b=seeds_b, seed=seeds[0][0], width=char_ngram
        )
    else:
        # this is the 25.02 case where it's neither the deprecated case neither permuted is available
        return ser.str.minhash(a=seeds_a, b=seeds_b, seed=seeds[0][0], width=char_ngram)
```

sarahyurick: Ah ok I see, thanks! So are we waiting until the next NeMo Curator release (which will have 24.12 as the stable RAPIDS version) before removing `MINHASH_DEPRECATED_API`?

praateekmahajan: Yes that's correct! And 25.02+ we wouldn't need any of these variables

PR #446, opened 2024-12-20
Bug fix in dockerfile ARG vs ENV var

praateekmahajan:
## Description
Since we have two `FROM`s in our Dockerfile, to reference the first build arg (CUDA_VER) we need to redeclare it as an arg again.

 

## Usage

```python
```
## Checklist


sarahyurick (approved):
Thanks!

PR #449, opened 2024-12-20
Change filename column name to file_name

praateekmahajan:
## Description
Resolves https://github.com/NVIDIA/NeMo-Curator/issues/427

## Usage

```python
```
## Checklist

### DIFF
--- a/nemo_curator/datasets/doc_dataset.py
+++ b/nemo_curator/datasets/doc_dataset.py
@@ -64,7 +64,7 @@ class DocumentDataset:
             input_files: The path of the input file(s).
             backend: The backend to use for reading the data.
             files_per_partition: The number of files to read per partition.
-            add_filename: Whether to add a "filename" column to the DataFrame.
+            add_filename: Whether to add a "file_name" column to the DataFrame.
             input_meta: A dictionary or a string formatted as a dictionary, which outlines
                 the field names and their respective data types within the JSONL input file.
             columns: If not None, only these columns will be read from the file.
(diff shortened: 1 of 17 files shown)


nemo_curator/datasets/doc_dataset.py
```diff
@@ -64,7 +64,7 @@ def read_json(
             input_files: The path of the input file(s).
             backend: The backend to use for reading the data.
             files_per_partition: The number of files to read per partition.
-            add_filename: Whether to add a "filename" column to the DataFrame.
+            add_filename: Whether to add a "file_name" column to the DataFrame.
```
sarahyurick, line 67: What do you think about changing `add_filename` to `add_file_name`, `keep_filename_column` to `keep_file_name_column`, `write_to_filename` to `write_to_file_name`, etc.?

praateekmahajan: Aah I did think about it but I personally felt the cons of renaming arguments in function signature >> pros of having alignment between the actual column name, specifically where backward compatibility, changing across scripts / tutorials / user behavior.
If you think the that the consistency between functions argument name and the `df` column name matters, I'll be happy to change it

sarahyurick: Okay yeah, I don't have a very strong preference for it so feel free to keep as is, unless others have different opinions.

In this case, my only other comment is if you could make sure there aren't any lingering scripts or examples/tutorials that try to access a "filename" column when it should now be "file_name"?

VibhuJawa (approved):
LGTM, thanks for fixing it. I think `file_name ` is innocuous  enough for the benefit we are getting.

nemo_curator/datasets/doc_dataset.py
praateekmahajan: I took a quick look and couldn't find anything, found one remaining comment line that I've added https://github.com/NVIDIA/NeMo-Curator/pull/449/commits/1126461bb7f8b5d4acce91fb68dbf2e8829d7fe2

PR #452, opened 2024-12-21
ci: Update release.yml

ko3n1g:
## Usage

```python
```
## Checklist

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -25,10 +25,14 @@ on:
         required: true
         default: true
         type: boolean
+      version-bump-branch:
+        description: Branch for version bump
+        required: true
+        type: string
 
 jobs:
   release:
-    uses: NVIDIA/NeMo-FW-CI-templates/.github/workflows/_release_library.yml@v0.17.4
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -43,6 +47,7 @@ jobs:
       container-workdir: /opt/NeMo-Curator
       library-name: NeMo Curator
       dry-run: ${{ inputs.dry-run }}
+      version-bump-branch: ${{ inputs.version-bump-branch }}
```
sarahyurick, line 50: Could you please elaborate on how this works? My guess is that this is for handling automatic updates to the version by creating a new branch(?), but looking into https://github.com/NVIDIA/NeMo-FW-CI-templates/tree/main it's not very clear to me how this works.

ko3n1g: Oh sorry for the late response! This is mostly for pre-releases: A pre-release runs most of the time on a commit older than top of tree (due to lag of internal CI). Therefore, the input `release-ref` will be a commit and not a branch.

Part of doing a pre-release is attempting a version bump via PR. If `release-ref` is a commit, the automation lacks the information of the target branch to apply the version bump. In most cases this is `main`, but for more flexibility we made this an input

ko3n1g:
Outdated

PR #457, opened 2024-12-23
Clean up internal column logic in `_run_classifier_helper` function

sarahyurick:
Closes https://github.com/NVIDIA/NeMo-Curator/issues/425.

### DIFF
--- a/nemo_curator/classifiers/base.py
+++ b/nemo_curator/classifiers/base.py
@@ -122,32 +122,27 @@ def _run_classifier_helper(
 ) -> "dask_cudf.DataFrame":
 
     if prob_col:
-        keep_prob = True
+        df[prob_col] = 0
     else:
-        keep_prob = False
         prob_col = "_prob"
 
-    df["sliced_text"] = df[text_field].str.slice(0, max_chars)
-
     columns_to_keep_list = df.columns.to_list()
-    columns_to_keep_list.remove("sliced_text")
 
     classifier_pipe = op.Sequential(
-        op.Tokenizer(model, cols=["sliced_text"], tokenizer_type="default"),
+        op.Tokenizer(
+            model, cols=[text_field], tokenizer_type="default", max_chars=max_chars
+        ),
         op.Predictor(
             model,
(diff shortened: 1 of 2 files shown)


nemo_curator/classifiers/base.py
```diff
@@ -121,44 +121,29 @@ def _run_classifier_helper(
...
+            model, cols=[text_field], tokenizer_type="default", max_chars=max_chars
+        ),
         op.Predictor(
             model,
             sorted_data_loader=True,
             batch_size=batch_size,
-            pred_output_col=prob_internal_col,
+            pred_output_col=prob_col,
         ),
+        op.Labeler(labels, cols=[prob_col], suffix=label_col),
```
sarahyurick, line 141: Using `suffix=label_col` depends on https://github.com/rapidsai/crossfit/pull/106.

nemo_curator/classifiers/base.py
```diff
@@ -121,44 +121,29 @@ def _run_classifier_helper(
     prob_col: str = None,
 ) -> "dask_cudf.DataFrame":
 
-    keep_prob = prob_col is not None
-    prob_internal_col = "_prob"
-    # TODO: Make crossfit handle this cleanly
-    pred_internal_col = "labels"
-    df["sliced_text"] = df[text_field].str.slice(0, max_chars)
+    if prob_col:
+        df[prob_col] = 0
```
sarahyurick, line 125: I wanted CrossFit to internally create the `prob_col`, but I was having trouble getting this to work. Specifically, I was able to get the entire classification pipeline to work, but then at the end `prob_col` would be dropped somewhere and not returned with the result.

VibhuJawa (approved):
LGTM. Thanks for making these changes

nemo_curator/classifiers/base.py
sarahyurick: I have opened an issue here: https://github.com/rapidsai/crossfit/issues/108. Please LMK if there is anything else I can clarify there. Thanks!

PR #458, opened 2024-12-30
Fix image curation on latest RAPIDS

VibhuJawa:
This PR fixes image curation on latest RAPIDS to ensure compatibilty

Also fixes: https://github.com/NVIDIA/NeMo-Curator/issues/453


nemo_curator/utils/cudf_utils.py
VibhuJawa, line 1: This was removed in favor of using crossfit directly: https://github.com/rapidsai/crossfit/blob/9c3871246408adbf6f6aabfe0b6e79315ecb7eb1/crossfit/backend/cudf/series.py#L81

This is because the list column creation was broken due to multiple upstream RAPIDS changes.  Using crossfit ensures we just that code cleanly. Otherwise we will have to duplicate the whole thing here. 

@ryantwolf ,  @praateekmahajan  Let me know your thoughts.

nemo_curator/datasets/image_text_pair_dataset.py
```diff
@@ -79,7 +79,11 @@ def from_webdataset(cls, path: str, id_col: str):
             path (str): The path to the WebDataset-like format on disk or cloud storage.
             id_col (str): The column storing the unique identifier for each record.
         """
-        metadata = dask_cudf.read_parquet(path)
+        metadata = dask_cudf.read_parquet(path, split_row_groups=False)
+        # TODO: This is a hack to ensure that the number of partitions is not combined
+        # and remain the same as the number of shards.
+        # DEBUG: Why is this happening?
+        metadata = metadata.repartition(npartitions=metadata.npartitions)
```
praateekmahajan, line 86: Curious why was this an issue / where was it caught?

praateekmahajan (approved):
LGTM except the one comment

nemo_curator/utils/cudf_utils.py
ryantwolf: Good with me

ryantwolf (approved):
Thanks for catching this. I do want to double check though, have you verified that the partition layout is the same as before?

nemo_curator/datasets/image_text_pair_dataset.py
VibhuJawa: Let me file an issue for this to dask_cudf/dask (will merge after a file an issue and link it here) .

VibhuJawa:
> Thanks for catching this. I do want to double check though, have you verified that the partition layout is the same as before?

Yup. but i still want to file a MRE (to link the right issue etc) , moving to draft for a bit

nemo_curator/datasets/image_text_pair_dataset.py
VibhuJawa: Resolved here: https://nvidia.slack.com/archives/C07P8N0LSGK/p1735755385631169

VibhuJawa:
@ryantwolf , @praateekmahajan , I think this is good to merge now. Synced internally with Rick, setting blocksize=None prevents runtime io fusion. This now works cleanly now. 

Please give a final look.

PR #463, opened 2025-01-02
Run GPU Tests on Rapids Nightly at midnight UTC

praateekmahajan:
## Description
Sets up a cron schedule (UTC midnight) to build a fresh container with nightly dependencies of RAPIDs  and then run tests along with them.

1. Adds schedule to gpuci.yml
2. Adds `ARG BUILD_TYPE` in `Dockerfile` to have one dockerfile for stable / nighlty (alternatively we could move it to two separate nightly; [previous discussion](https://github.com/NVIDIA/NeMo-Curator/pull/436/files#r1889168872)) 
## Usage

```python
```
## Checklist


.github/workflows/gpuci.yml
```diff
@@ -1,6 +1,8 @@
 name: "GPU CI/CD"
 
 on:
+  schedule:
+    - cron: '0 0 * * *'  # scheduled trigger for nightly builds, runs at midnight UTC every day
```
praateekmahajan, line 5: We should perhaps time it to whenever the nightly cudf / cuml packages are published

.github/workflows/gpuci.yml
```diff
@@ -21,27 +23,30 @@ concurrency:
...
+      dockerfile: ${{ github.event_name == 'schedule' && 'Dockerfile.nightly' || 'Dockerfile' }}
+      image-label: ${{ github.event_name == 'schedule' && 'nemo-curator-nightly' || 'nemo-curator' }}
       build-args: |
-        IMAGE_LABEL=nemo-curator
+        IMAGE_LABEL=${{ github.event_name == 'schedule' && 'nemo-curator-nightly' || 'nemo-curator' }}
         REPO_URL=https://github.com/${{ github.repository }}.git
         CURATOR_COMMIT=${{ github.sha }}
+        BUILD_TYPE=${{ github.event_name == 'schedule' && 'nightly' || 'stable' }}
       prune-filter-timerange: 24h
+      use_cache: ${{ github.event_name != 'schedule' }}
```
sarahyurick, line 40: Would you mind elaborating on how this section works?

sarahyurick (commented):
Really great PR so far, thanks! Please let me know if there is anything I can help with here.

.github/workflows/gpuci.yml
ko3n1g: In case timing was important, you could use a `workflow_dispatch` event and trigger this workflow via https through the other repo that builds the nightly container

ko3n1g (approved):
Looks like a great change and some cool learnings too (conditional GHA syntax). Left a suggestion on how to address the potential timing issue, but already approve since I think this also works in the current state

.github/workflows/gpuci.yml
praateekmahajan: Yes, to begin with `${{ }}}` is the [syntax for expressions](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/evaluate-expressions-in-workflows-and-actions#example).

And then whenever the statement `github.event_name == 'schedule'` is `True` we know we're dealing with nightly (as it runs on schedule) instead of the merge-to-main / PR labeling stuff.

Now going over the individual build args that are passed to Dockerfile
1. IMAGE_LABEL = `nemo-curator-nightly` or `nemo-curator`
2. REPO_URL = `github.com/NVIDIA/NeMo-Curator`
3. CURATOR_COMMIT = `commit_sha`
4. BUILD_TYPE = `nightly` or `stable`

Apart from build args we also don't want to use cache for nightly when building docker images hence we use the conditional.

.github/workflows/gpuci.yml
praateekmahajan: That's a good point, and the step within should help us know the version of our depdenencies

```
    - name: Verify installations
      run: |
        docker exec nemo-curator-container pip list
```

.github/workflows/gpuci.yml
```diff
@@ -21,27 +23,30 @@ concurrency:
...
     # This is the tag on our Azure runner found in Actions -> Runners -> Self-hosted runners
     # It has 2 A100 GPUs
     runs-on: self-hosted-azure
-    # "run-gpu-tests" job is run if the "gpuci" label is added to the PR
-    if: ${{ github.event.label.name == 'gpuci' || github.ref == 'refs/heads/main' }}
+    # "run-gpu-tests" job is run if the "gpuci" label is added to the PR / merged to main / scheduled run
+    if: ${{ github.event.label.name == 'gpuci' || github.ref == 'refs/heads/main' || github.event_name == 'schedule' }}
 
     steps:
       # If something went wrong during the last cleanup, this step ensures any existing container is removed
```
praateekmahajan, line 52: @ko3n1g I'm curious where is the name of the docker container set to `nemo-curator-container` even in the existing `gpuci.yml`? 

I thought it might be `${IMAGE_LABEL}-container` but I couldn't see it in the code?

ko3n1g: The container is created here: https://github.com/NVIDIA/NeMo-Curator/blob/main/.github/workflows/gpuci.yml#L60

praateekmahajan: Aah thanks, I don't know how I missed that lol. Any thoughts on naming it differently between nightly vs stable

ko3n1g: Yes sure it doesn't harm to suffix the nightly container and avoid potential race conditions

.github/workflows/gpuci.yml
sarahyurick: Got it, thanks!

.github/workflows/gpuci.yml
```diff
@@ -57,33 +66,43 @@ jobs:
...
+
+      # TODO Add notification step for nightly failures
+      # - name: Notify on failure
+      #   if: failure() && github.event_name == 'schedule'
+      #   uses: slackapi/slack-github-action@v1
+      #   with:
+      #     channel-id: 'xyz'
+      #     slack-message: "Nightly GPU tests failed! Check: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}"
+      #   env:
+      #     SLACK_BOT_TOKEN: ${{ secrets.SLACK_BOT_TOKEN }} # add secret somewhere
```
sarahyurick, line 108: Just wanted to confirm if there is anywhere else the logs are viewable (besides Slack for failures)? Maybe even a way for me to subscribe to look at it every time it is run?

praateekmahajan: I believe we should be able to see the logs here https://github.com/NVIDIA/NeMo-Curator/actions/workflows/gpuci.yml

I'm not sure how we can subscribe tbh.

sarahyurick: I see, thanks! I was also wondering is there something blocking the Slack notifications atm?

sarahyurick (approved):

PR #477, opened 2025-01-09
Support `dask_expr` migration into `dask.dataframe`

rjzamora:
## Description
Minor changes to use `dask.dataframe.dask_expr` for `dask>2024.12.1`.

See also:
- https://github.com/rapidsai/cudf/pull/17704
- https://github.com/rapidsai/dask-cuda/pull/1424

## Checklist


ayushdg (approved):
Thanks for the fix. We've finally transitioned to newer dask releases in CI. It'll probably be a month or so before we're pinned to a newer dask where we can remove the explicit `import dask_expr` statements.

Issue #142, opened 2024-07-08
Running into OOM with add id


yyu22:
**Describe the bug**

Running the add id module of curator runs into ooms even with small batch size, e.g., 32. 
The dataset for adding ID is a single snapshot of Red Pajama v2 dataset, which is about 4 TB in size.
Job was run on 10 cpu nodes. Each cpu node has 96 cores and 176 GB memory

```
Jun 25 13:48:18.323459 942129 slurmstepd   0x155552de2d40: error: Detected 7 oom_kill events in StepId=1127164.0. Some of the step tasks have been OOM Killed.
srun: error: cpu-00046: task 5: Out Of Memory
srun: Terminating StepId=1127164.0
Jun 25 13:48:19.899767 2590557 slurmstepd   0x155552de2d40: error: Detected 1 oom_kill event in StepId=1127164.0. Some of the step tasks have been OOM Killed.
srun: error: cpu-00017: task 2: Terminated
srun: error: cpu-00038: task 3: Terminated
srun: error: cpu-00050: task 6: Terminated
Jun 25 13:48:20.991455 2567860 slurmstepd   0x155552de2d40: error: Detected 2 oom_kill events in StepId=1127164.0. Some of the step tasks have been OOM Killed.
srun: Force Terminated StepId=1127164.0
```

some observations:

- Memory usage is extremely unbalanced across the nodes

```
cpu-00009              total        used        free      shared  buff/cache   available
Mem:            176          14         144           0          18         160
cpu-00042              total        used        free      shared  buff/cache   available
Mem:            176          79          88           0           8          94
cpu-00046              total        used        free      shared  buff/cache   available
Mem:            176         113          61           0           2          61
cpu-00082              total        used        free      shared  buff/cache   available
Mem:            176          13         145           0          17         160
cpu-00050              total        used        free      shared  buff/cache   available
Mem:            176          74          78           0          23          99
cpu-00019              total        used        free      shared  buff/cache   available
Mem:            176          72          38           0          65         101
cpu-00087              total        used        free      shared  buff/cache   available
Mem:            176          55         106           0          15         119
cpu-00086              total        used        free      shared  buff/cache   available
Mem:            176          90          80           0           6          84
cpu-00020              total        used        free      shared  buff/cache   available
Mem:            176          36         101           0          39         138
cpu-00002              total        used        free      shared  buff/cache   available
Mem:            176         156           2           0          17          18
```

[...]

PR #479, opened 2025-01-10
Enable ADD ID to work with CPU/GPU both

VibhuJawa:
This PR fixes: https://github.com/NVIDIA/NeMo-Curator/issues/473


sarahyurick:
Do you know if this could help fix https://github.com/NVIDIA/NeMo-Curator/issues/142?

VibhuJawa:
> Running into OOM with add id #142

I think if we just do the `start_index=None` it should be scalable as we pass it to `_add_id_fast `

tests/test_add_id.py
```diff
@@ -18,26 +18,37 @@
...
-def list_to_dataset(documents, col_name="text", npartitions=2):
+
+def list_to_dataset(documents, col_name="text", npartitions=2, backend="pandas"):
     data = {col_name: documents}
     pdf = pd.DataFrame(data)
-
-    return DocumentDataset(dd.from_pandas(pdf, npartitions=npartitions))
+    ddf = dd.from_pandas(pdf, npartitions=npartitions)
+    if backend == "cudf" and is_unavailable(cudf):
+        pytest.skip("cuDF is not installed or importable.")
```
sarahyurick, line 32: Might be better to raise an error here instead of skipping.

tests/test_add_id.py
```diff
@@ -56,6 +67,8 @@ def test_basic_id(self, single_partition_dataset):
                 "doc_id-0000000004",
             ]
         )
+        if is_cudf_available and isinstance(actual_ids, cudf.Series):
+            actual_ids = actual_ids.to_pandas()
```
sarahyurick, line 71: Can we make sure we are actually entering this `if` block in the GPU case? Maybe:
```
if is_cudf_available:
    assert isinstance(actual_ids, cudf.Series)
    actual_ids = actual_ids.to_pandas()
```
?

tests/test_add_id.py
VibhuJawa: Agreed: https://github.com/NVIDIA/NeMo-Curator/pull/479/commits/dc3a13836d9fbfc72d4ec614df9dc19b192c2ff9

tests/test_add_id.py
VibhuJawa: So we actually run this cell for both cuDF and Pandas.  So we cant really ` assert isinstance(actual_ids, cudf.Series)`. 

We check `is_cudf_available` above now for GPU cases so i think that case is checked so i think we should be fine.

ryantwolf (approved):
Wow, didn't realize it all worked OOTB. Good find, glad to have it be faster.

PR #480, opened 2025-01-10
Add `nemo-toolkit` dependency to gpuCI

sarahyurick:
Closes https://github.com/NVIDIA/NeMo-Curator/issues/376.


pyproject.toml
```diff
@@ -107,20 +109,14 @@ image_nightly = [
     "timm>=1.0.8",
     "nemo_curator[cuda12x_nightly]",
 ]
-# Installs bitext curation modules
-bitext = [
-    "huggingface-hub",
-    "tqdm",
-    "transformers",
-    "nemo_curator[cuda12x]",
-]
```
sarahyurick, line 116: Looking at https://github.com/NVIDIA/NeMo-Curator/pull/193 it does not look like the `cuda12x` dependencies are strictly needed here? Also, I believe all of `huggingface-hub`, `tqdm`, and `transformers` are already being installed either directly or indirectly.

sarahyurick:
Closing this as we no longer need to rely on NeMo dependencies :)

PR #484, opened 2025-01-17
Create `check_dask_cwd` function

sarahyurick:
Closes https://github.com/NVIDIA/NeMo-Curator/issues/201.

### DIFF
--- a/nemo_curator/utils/distributed_utils.py
+++ b/nemo_curator/utils/distributed_utils.py
@@ -16,6 +16,7 @@ from __future__ import annotations
 import ast
 import os
 import shutil
+import subprocess
 
 import dask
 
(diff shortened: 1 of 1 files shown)


nemo_curator/utils/distributed_utils.py
```diff
@@ -1014,6 +1018,17 @@ def get_current_client():
         return None
 
 
+def check_dask_cwd(file_list):
```
praateekmahajan, line 1021: nit : type hint

nemo_curator/utils/distributed_utils.py
```diff
@@ -1014,6 +1018,17 @@ def get_current_client():
...
+    if any(not os.path.isabs(file_path) for file_path in file_list):
+        dask_cwd = list(get_current_client().run(os.getcwd).values())[0]
```
praateekmahajan, line 1023: should there be another chekc here than ensures all the workers have the same `cwd`?

sarahyurick: Thanks! LMK what you think.

praateekmahajan (approved):
LGTM! If you get a chance it'll be add to add unittest as well for the success / failure cases

PR #493, opened 2025-01-21
ci: Pin twine to `6.0.1`

ko3n1g:
## Usage

```python
```
## Checklist

### DIFF
--- a/.github/workflows/build-test-publish-wheel.yml
+++ b/.github/workflows/build-test-publish-wheel.yml
@@ -20,6 +20,7 @@ on:
       - 'main'
       - '[rv][0-9].[0-9].[0-9]'
       - '[rv][0-9].[0-9].[0-9]rc[0-9]'
+  pull_request:
     
 defaults:
   run:
--- a/.github/workflows/build-test-publish-wheel.yml
+++ b/.github/workflows/build-test-publish-wheel.yml
@@ -33,7 +33,7 @@ jobs:
       dry-run: true
       python-package: nemo_curator
       environment: public
-      python-version: 3.11
+      python-version: 3.10
     secrets:
       TWINE_USERNAME: ${{ secrets.TWINE_USERNAME }}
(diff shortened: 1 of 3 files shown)


.github/workflows/build-test-publish-wheel.yml
```diff
@@ -20,7 +20,8 @@ on:
       - 'main'
       - '[rv][0-9].[0-9].[0-9]'
       - '[rv][0-9].[0-9].[0-9]rc[0-9]'
-
+  pull_request:
+
```
sarahyurick, line 24: Looks like there is extra whitespace somewhere here.

pyproject.toml
```diff
@@ -73,6 +73,7 @@ dependencies = [
     "zstandard==0.18.0",
 ]
 dynamic = ["version"]
+license-files = ["LICENSE*"]
```
sarahyurick, line 76:
```suggestion
license-files = ["LICENSE"]
```
maybe this would work? Or `"LICEN[CS]E*"`? Or maybe order matters?

.github/workflows/build-test-publish-wheel.yml
sarahyurick, line 24:
```suggestion
  pull_request:

```

ko3n1g: I added the `pull_request` event trigger only for debug purposes. I wouldn't recommend pushing to testpypi per commit, as (test)pypi has a max project size which is more likely to be saturated if we push per commit

ko3n1g:
Successful run: https://github.com/NVIDIA/NeMo-Curator/actions/runs/13093602114/job/36533080873?pr=493

praateekmahajan:
You might have a trailing whitespace in the yml file causing CI to fail, could you run pre-commit on that file?

sarahyurick (approved):
LGTM, thanks!

PR #495, opened 2025-01-23
Update to support Python 3.12

ryantwolf:
## Description

Updates NeMo Curator to remove the upper bound on the Python version. Eventually, we can raise the lower bound, but in the meantime while the CI is still transitioning I've left it at 3.10. All docs should be updated, please let me know if I missed something.

Another minor change I did was a request from @ericharper to reorder the author list. He noticed that the first author appears as the point of contact on the pypi page (https://pypi.org/project/nemo-curator/) and he prefers that I be the point of contact over Joseph since he is no longer involved in the project.

## Usage

N/A
## Checklist

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -19,7 +19,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest]
-        python-version: ["3.10"]
+        python-version: ["3.12"]
     steps:
       - uses: actions/checkout@v4
       - name: Optionally free up space on Ubuntu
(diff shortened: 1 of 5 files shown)


VibhuJawa (approved):
LGTM

ayushdg (approved):
lgtm! I would prefer keeping the lower bound around for a couple of releases to allow non container users to still use with py3.10.
Do you think it's worth expanding the CPU unit test matrix here to both 3.10 and 3.12?

sarahyurick (approved):

ryantwolf:
> Do you think it's worth expanding the CPU unit test matrix here to both 3.10 and 3.12?

Yes, I'll open a separate PR to add that.

PR #497, opened 2025-01-24
Update model nomenclature

sarahyurick:


ryantwolf (approved):
LGTM, but can you check with Lawrence if our documentation should refer to them with the "NemoCurator" prefix in our headings and make a follow up PR to revert that if needed?

PR #498, opened 2025-01-28
Update fuzzy deduplication to skip false positive checks as the default

ayushdg:
## Description
This PR updates the FuzzyDedup modules making `false_positive_check=False` and `char_ngrams=24` the default params going forward.

Motivation:

Internal testing and external research has shown that depending on the values of `num_buckets` and `hashes_per_bucket`, the false positive rate can be minimized based on intended jaccard similarity.
The defaults of 20 buckets and 13 hashes per bucket has a low false positive rate for documents with similarity < 0.8 and internal testing has shown a difference of 1-2% when skipping the fp check.
Increasing the default minhash length to 24 which roughly corresponds to 5 word shingles, results in removal rates which are very similar to 5 char ngrams with the fp check enabled since 5 char ngrams are prone to more false positives.
The false positive check is computationally expensive and has significantly higher memory requirements. It uses requires significantly higher cache in memory keeping an additional copy of the corpus in addition to minhashes and buckets.
These new defaults should produce similar results to the previous defaults while being significantly less compute intensive and faster.

Will update tutorials in a followup PR.

## Usage

```python
```
## Checklist


examples/fuzzy_deduplication.py
```diff
@@ -69,14 +69,12 @@ def main(args):
...
-            char_ngrams=5,
+            char_ngrams=24,
             num_buckets=20,
             hashes_per_bucket=13,
             use_64_bit_hash=False,
-            buckets_per_shuffle=5,
-            false_positive_check=True,
-            num_anchors=2,
-            jaccard_threshold=0.8,
+            buckets_per_shuffle=5,  # set to a smaller value if encountering OOM's during LSH.
```
sarahyurick, line 76:
```suggestion
            buckets_per_shuffle=5,  # set to a smaller value if encountering OOMs during LSH
```

nemo_curator/modules/config.py
```diff
@@ -67,47 +67,69 @@ class FuzzyDuplicatesConfig(BaseConfig):
...
 
-    false_positive_check: bool = True
-    # Only required for fp check
-    num_anchors: int = 2
-    jaccard_threshold: float = 0.8
-    bucket_mapping_blocksize: int = 256
-    parts_per_worker: int = 1
-    bucket_parts_per_worker: int = 8
+    false_positive_check: bool = False
+    # Only required for fale positive check
```
sarahyurick, line 77:
```suggestion
    # Only required for false positive check
```

sarahyurick (commented):
Generally LGTM, thanks! Just wanted to ask if https://github.com/NVIDIA/NeMo-Curator/blob/main/docs/user-guide/gpudeduplication.rst or any tutorials need to be updated?

sarahyurick (approved):
Thanks!

sarahyurick:
These are the tutorials which will potentially need updating:

- https://github.com/NVIDIA/NeMo-Curator/tree/main/tutorials/zyda2-tutorial
- https://github.com/NVIDIA/NeMo-Curator/tree/main/tutorials/dapt-curation
- https://github.com/NVIDIA/NeMo-Curator/tree/main/tutorials/single_node_tutorial
- https://github.com/NVIDIA/NeMo-Curator/tree/main/tutorials/pretraining-data-curation

PR #499, opened 2025-01-28
Removal logic for Exact / Fuzzy Dedup

praateekmahajan:
### Removal Description
1. We implement ~left-anti join using a broadcast merge. This allows us to scale even when right is greater than memory per node.
2. We observe that the performance varies as num partitions (or partition sizes vary). 
    - The fastest being where right is exactly one partition, in which case we broadcast all of right to all the workers, and perform a merge (resulting in least transfer).
3.  One downside of this approach is when right is bigger than system memory per node, then the merge operation will have to spill, and that won't be possible, hence resulting in a stuck state where [dask pauses workers once it reaches 80% capacity](https://distributed.dask.org/en/latest/worker-memory.html#pause-worker) in which case user will have to implement a different merge logic.
 
### Class Description
1. Implements an abstract class called `Deduplicator` (name suggestions welcome)
        1. Lives in `nemo_curator._deduplicator.py`
        7. Implements identify / remove / __call__
        8. __call__ now accepts a boolean called `perform_removal` (alternative API's also accepted)  
2. Fuzzy / Exact Duplicates extend this class and the `__call__` is now renamed as `identify` since the `__call__` is implemented in base class 
## Usage

```python
dataset = DocumentDataset.read_parquet(...)
exact_dedup = ExactDeduplicator(..)
duplicates = exact_dedup(dataset) # or exact_dedup.identify(exact_dedup)

exact_dedup.remove(dataset)
```
## Checklist

### DIFF
--- a/nemo_curator/_deduplicator.py
+++ b/nemo_curator/_deduplicator.py
@@ -87,6 +87,7 @@ class Deduplicator(ABC):
         left = dataset.df
         right = duplicates.df
 
+        print(f"{left.npartitions=}, {right.npartitions=}")
         if left.npartitions < right.npartitions:
             msg = (
                 "The number of partitions in the dataset to remove duplicates from is less than the number of partitions in the duplicates dataset. "
--- a/nemo_curator/_deduplicator.py
+++ b/nemo_curator/_deduplicator.py
@@ -13,20 +13,30 @@ def _perform_removal(
     id_field: str,
     group_field: str,
 ) -> dd.DataFrame:
+    # Create a new column name for temporary ID storage during merge
     new_id_field = f"{id_field}_new"
 
     duplicates_to_remove = (
-        duplicates.map_partitions(lambda x: x[x[group_field].duplicated(keep="first")])
-        .drop(columns=group_field)
+        duplicates
+        # Redistribute data across partitions so that all duplicates are in same partition
+        .shuffle(on=[group_field], ignore_index=True)
+        # For each partition, keep only the duplicated rows (excluding first occurrence)
+        .map_partitions(lambda x: x[x[group_field].duplicated(keep="first")]).drop(
+            columns=group_field
+        )
+        # Rename the ID field to avoid conflicts in the upcoming merge
         .rename(columns={id_field: new_id_field})[[new_id_field]]
     )
+
     merge = left.merge(
         right=duplicates_to_remove,
         how="left",
-        broadcast=True,
+        broadcast=True,  # Broadcast smaller DataFrame to all partitions
         left_on=id_field,
         right_on=new_id_field,
     )
+
+    # This effectively removes all rows that were not in duplicates_to_remove
     removed_result = merge[merge[new_id_field].isna()].drop(columns=[new_id_field])
     return removed_result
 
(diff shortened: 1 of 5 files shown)


nemo_curator/_deduplicator.py
ryantwolf, line 1: Can this be moved under `nemo_curator/modules/deduplicator.py`? I think the function is something we want users to be able to access.

nemo_curator/_deduplicator.py
```diff
@@ -0,0 +1,130 @@
...
+        left_on=id_field,
+        right_on=new_id_field,
+    )
+
+    # This effectively removes all rows that were not in duplicates_to_remove
+    removed_result = merge[merge[new_id_field].isna()].drop(columns=[new_id_field])
+    return removed_result
+
+
+class Deduplicator(ABC):
```
ryantwolf, line 44: Not sure how I feel about the abstraction. I have been wanting something like this, but I worry this is not as generalizable as I'd want it to be. For example, can semantic dedupe use this? I don't believe it can since the duplicates aren't all grouped like this. Imo, if the deduplication abstraction doesn't work for all our deduplication methods I don't want to have it so we don't confuse our users. We can always refactor out the logic into a base class if we find a general solution.

nemo_curator/_deduplicator.py
```diff
@@ -0,0 +1,130 @@
...
+        self.text_field = text_field
+        self.grouped_field = grouped_field
+        self.cache_dir = cache_dir
+
+    def identify(self, *args, **kwargs):
+        """Abstract method to be implemented by concrete deduplicator classes.
+        Should implement the logic for identifying duplicates in the dataset."""
+        raise NotImplementedError
+
+    def remove(
```
ryantwolf, line 63: I think this function could just be a helper function that's defined here that exact and fuzzy dedup import and use instead of the ABC.

ryantwolf (commented):
Couple of comments about the organization. Can you make sure you update the API docs and deduplication sections in the user guide to mention all this?

> One downside of this approach is when right is bigger than system memory per node, then the merge operation will have to spill, and that won't be possible, hence resulting in a stuck state where [dask pauses workers once it reaches 80% capacity](https://distributed.dask.org/en/latest/worker-memory.html#pause-worker) in which case user will have to implement a different merge logic.

Make sure to mention this in the user guide and describe roughly when "right is bigger than system memory per node" in layman's terms. Is this something we have/will encounter? Wondering if globally deduping 240 TB would trigger this.

praateekmahajan:
@sarahyurick / @ryantwolf given both of you had thoughts about it not being an abstract class, I took your suggestions and started this https://github.com/NVIDIA/NeMo-Curator/pull/509
@ryantwolf I'll add the docs once we are clear on what the API is, so that I don't have to change them repeatedly.

Regarding your question on if we've run into a case where it's greater than system (host) memory yes. 
In one of our experiments, a 64TB dataset produced ~300gb of duplicates, and our host memory was 250gb per worker, which didn't work out for us, as CUDF during merge was spilling to host memory, and running OOM.
In that case, we'd likely recommend the user, to implement something custom where they do it at a partition level (where each partition typically is a year of CC or something similar)

It's still an open question on why dask / cudf needed to have the whole ~300gb in memory, but for now at that scale this won't work. It works wherever the duplicates size << host memory which can vary for different systems.

PR #509, opened 2025-01-31
Removal logic for fuzzy / exact (no class abstraction)

praateekmahajan:
## Description

See https://github.com/NVIDIA/NeMo-Curator/issues/529 for much more detailed analysis on the decision and followups

### Removal Description
1. We implement ~left-anti join using a broadcast merge. This allows us to scale even when right is greater than device memory per node.
2. In our exact and fuzzy deduplication, we perform a shuffle at write time as the last step in identification. When reading for removal, we read FPP=1 so that we are reading a group at a time and not incurring "another" shuffle. By "another" shuffle I mean, that in a broadcast join, right is also reshuffled based on `hash(id_col)` so forcing a shuffle again will result in a double shuffle at read time.
4. We observe that the performance varies as num partitions (or partition sizes vary). 
    - The fastest being where right is exactly one partition, in which case we broadcast all of right to all the workers, and perform a merge (resulting in least transfer).
5.  One downside of this approach is when right is bigger than system memory per node, then the merge operation will have to spill, and that won't be possible, hence resulting in a stuck state where [dask pauses workers once it reaches 80% capacity](https://distributed.dask.org/en/latest/worker-memory.html#pause-worker) in which case user will have to implement a different merge logic.

### Class Description
1. Adds a `remove` method to Fuzzy/ExactDeduplicator that calls `nemo_curator.modules.removal.remove_duplicates`
2. Moves logic from `call` to `identify_duplicates`. 
3. Adds another paratmer to class construction for `perform_removal` which by default is False, to retain old behavior. But when True `call` removes the duplicates as well.

## Usage

```python
```
## Checklist


ryantwolf (commented):
Looks good overall, one final change though can you modify `nemo_curator/modules/__init__.py` to include `nemo_curator.modules.removal.remove_duplicates` in the `__all__`? Model it after `blend_datasets` if you need a reference.

nemo_curator/modules/exact_dedup.py
```diff
@@ -135,7 +136,7 @@ def hash_documents(
             # TODO: Generalize ty using self.hash_method
             return df.apply(lambda x: md5(x.encode()).hexdigest())
 
-    def __call__(self, dataset: DocumentDataset) -> Union[DocumentDataset, str]:
+    def identify(self, dataset: DocumentDataset) -> DocumentDataset:
```
sarahyurick, line 139:
```suggestion
    def _identify(self, dataset: DocumentDataset) -> DocumentDataset:
```
Nit, but maybe call them `_identify` and `_remove` if they are not intended to be accessed by the user directly.

nemo_curator/modules/exact_dedup.py
```diff
@@ -166,10 +167,38 @@ def __call__(self, dataset: DocumentDataset) -> Union[DocumentDataset, str]:
...
+            id_field=self.id_field,
+            group_field="_hashes",
+        )
+        return DocumentDataset(result)
+
+    def __call__(self, dataset: DocumentDataset, perform_removal : bool = False) -> DocumentDataset:
+        duplicates = self.identify(dataset)
+        if perform_removal:
+            return self.remove(dataset, duplicates)
+        return duplicates
```
sarahyurick, line 204: I think it makes since to have `__call__` identify and remove the duplicates, but at the same time this change is very disruptive to the existing workflow. Would it still make sense to use `__call__` to just identify the duplicates, then have a `remove_duplicates` function in the exact/fuzzy deduplication classes to be called separately by the user?

nemo_curator/modules/removal.py
sarahyurick, line 1: Not sure that this should live in `nemo_curator/modules` since its just a single shared function, maybe it should live in `utils` instead.

nemo_curator/modules/exact_dedup.py
praateekmahajan: I agree that changing the workflow today will be disruptive, but at the same time I do think, long term we should have identify + remove be the natural state, especially given how `Sequential` works. I propose keeping it the same as proposed in the PR, however we can add a warning when `perform_removal==False` with a deprecation notice and stating that the default behavior will be changing.

Another very minor nit would be that it's slightly more work for the user to understand args for removal
```
ds = DocumentDataset.read(...)
fuzzy = FuzzyDuplicates(...)
duplicates = fuzzy(ds)
new_ds = fuzzy.remove(ds, duplicates)
```
vs
```
ds = DocumentDataset.read(...)
fuzzy = FuzzyDuplicates(...)
new_ds = fuzzy.identify_and_remove(ds) or fuzzy(ds, perform_removal=True)
```

nemo_curator/modules/exact_dedup.py
praateekmahajan: Let's keep them exposed especially since remove won't work at scales where size of duplicate >> host memory, in which case the user will need to break down identify and remove

nemo_curator/modules/exact_dedup.py
```diff
@@ -166,10 +167,42 @@ def __call__(self, dataset: DocumentDataset) -> Union[DocumentDataset, str]:
...
+        result = remove_duplicates(
+            left=dataset.df,
+            duplicates=duplicates_to_remove.df,
+            id_field=self.id_field,
+            group_field="_hashes",
+        )
+        return DocumentDataset(result)
+
+    def __call__(
+        self, dataset: DocumentDataset, perform_removal: bool = False
```
ryantwolf, line 203: Move `perform _removal` to the constructor. `__call__` should only have the dataset as input so we can easily chain it with other modules.

nemo_curator/modules/fuzzy_dedup/fuzzyduplicates.py
```diff
@@ -243,4 +243,41 @@ def __call__(self, dataset: DocumentDataset):
...
+        result = remove_duplicates(
+            left=dataset.df,
+            duplicates=duplicates_to_remove.df,
+            id_field=self.config.id_field,
+            group_field="group",
+        )
+        return DocumentDataset(result)
+
+    def __call__(
+        self, dataset: DocumentDataset, perform_removal: bool = False
```
ryantwolf, line 278: Move `perform _removal` to the config. `__call__` should only have the dataset as input so we can easily chain it with other modules.

nemo_curator/modules/semantic_dedup/clusteringmodel.py
```diff
@@ -195,9 +195,7 @@ def __call__(self, embeddings_dataset: DocumentDataset):
                 id_col=self.id_col,
                 kmeans_centroids_file=kmeans_centroids_file,
                 nearest_cent_dir=clustering_output_dir,
-                output_sorted_clusters_dir=os.path.join(
-                    self.clustering_output_dir, "sorted"
-                ),
+                vdb=os.path.join(self.clustering_output_dir, "sorted"),
```
ryantwolf, line 198: What's this change for?

ryantwolf (changes requested):
Looks good overall, just a few comments.

nemo_curator/modules/exact_dedup.py
ryantwolf: +1 to the long term we should have identify + remove be the natural state. I vote we do it as soon as next release. For now, can you add a warning if the `perform_removal` is set to False that the default value will be changing in our next release (`0.8.0`)?

docs/user-guide/gpudeduplication.rst
```diff
@@ -82,9 +82,11 @@ After ensuring your dataset has a unique ID field (or creating one with the code
     107  doc_prefix-52271  0f763a2937d57b9d96bf9f220e55f2bd
     """
 
+    deduplicated_dataset = exact_duplicates.remove(dataset, duplicate_docs)
```
sarahyurick, line 85: Should also include the `perform_removal` option above?

nemo_curator/modules/exact_dedup.py
sarahyurick: Yes, that makes sense to me. What about calling it `identify_duplicates`?

nemo_curator/utils/removal.py
sarahyurick, line 1: Maybe call it "duplicates_removal" or something similar?

VibhuJawa (changes requested):
Thanks for pushing through this . This mostly looks good to me. 

The only ask is not to Modify the `__call__` header and behavior in this release. Everything else looks great.

nemo_curator/modules/exact_dedup.py
praateekmahajan: Feels more verbose than necessary `XYZDuplicates.identify_duplicates(..)` vs `XYZDuplicates.identify`.

Not strongly opinionated tho, can change it. One argument in favor of identify_duplicates would be that in future we might want to expose `identify_documents_to_keep` in which the distinction might be necessary

 Let's just get one more persons thought

cc @ayushdg / @ryantwolf / @VibhuJawa

nemo_curator/modules/semantic_dedup/clusteringmodel.py
praateekmahajan: mistake lol. Surprised how no test failed

nemo_curator/modules/exact_dedup.py
praateekmahajan: Sounds good, initially I thought it's slightly verbose, but another argument in favor of `identify_duplicates` would be that in future we might want to expose `identify_documents_to_keep` in which the distinction might be necessary


cc @ayushdg / @ryantwolf / @VibhuJawa

docs/user-guide/gpudeduplication.rst
```diff
@@ -187,6 +197,7 @@ Python API
         cache_dir="/path/to/dedup_outputs", # must be cleared between runs
         id_field="my_id",
         text_field="text",
+        perform_removal=False, # dictates if deduplicated dataset or duplicates are returned
```
sarahyurick, line 200:
```suggestion
        perform_removal=False, # dictates if deduplicated dataset or IDs of duplicates are returned
```

docs/user-guide/gpudeduplication.rst
```diff
@@ -82,9 +87,14 @@ After ensuring your dataset has a unique ID field (or creating one with the code
     107  doc_prefix-52271  0f763a2937d57b9d96bf9f220e55f2bd
     """
 
+    deduplicated_dataset = exact_duplicates.remove(dataset, duplicate_docs)
+
+    # Users who have specified perform_removal=True can get the output directly as follows
```
sarahyurick, line 92:
```suggestion
    # Users who have specified perform_removal=True can get the output deduplicated dataset directly as follows
```

docs/user-guide/gpudeduplication.rst
```diff
@@ -244,10 +257,15 @@ Python API
     4  doc_prefix-42050    154
     """
 
+    deduplicated_dataset = fuzzy_duplicates.remove(dataset, duplicate_docs)
+
+    # Users who have specified perform_removal=True can get the output directly as follows
```
sarahyurick, line 262:
```suggestion
    # Users who have specified perform_removal=True can get the output deduplicated dataset directly as follows
```

examples/exact_deduplication.py
```diff
@@ -40,36 +39,33 @@ def main(args):
...
         id_field=dataset_id_field,
         text_field=dataset_text_field,
+        # Decides whether output of the module is deduplicated dataset or duplicates
+        # If true, you should set cache_dir for performance improvement
+        perform_removal=False,
         # cache_dir=output_dir  # Optionally write the output to disk
     )
 
-    duplicates = exact_dup(dataset=input_dataset)
+    # When perform_removal=False, it'll only call .identify_duplicates() and return the duplicates.
```
sarahyurick, line 56:
```suggestion
    # When perform_removal=False, it will only call .identify_duplicates() and return the list of duplicate IDs.
```

examples/exact_deduplication.py
```diff
@@ -40,36 +39,33 @@ def main(args):
...
+    # When perform_removal=True then exact_dup outputs dataset with the duplicates removed
+    # It'll behave by calling .identify_duplicates() and .removal() in sequence.
```
sarahyurick, line 58:
```suggestion
    # When perform_removal=True, then exact_dup outputs the dataset with the duplicates removed.
    # It will behave by calling .identify_duplicates() and .remove() in sequence.
```

examples/fuzzy_deduplication.py
```diff
@@ -68,6 +68,8 @@ def main(args):
             cache_dir=cache_dir,
             id_field=dataset_id_field,
             text_field=dataset_text_field,
+            # Decides whether output of the module is deduplicated dataset or duplicates
```
sarahyurick, line 71:
```suggestion
            # Decides whether output of the module is a deduplicated dataset or the IDs of the duplicates
```

examples/fuzzy_deduplication.py
```diff
@@ -77,26 +79,20 @@ def main(args):
             false_positive_check=False,
         )
         fuzzy_dup = FuzzyDuplicates(logger=log_dir, config=fuzzy_dedup_config)
-        duplicates = fuzzy_dup(dataset=input_dataset)
+
+        # When perform_removal=False, it'll only call .identify_duplicates() and return the duplicates.
```
sarahyurick, line 83:
```suggestion
        # When perform_removal=False, it will only call .identify_duplicates() and return the list of duplicate IDs.
```

examples/fuzzy_deduplication.py
```diff
@@ -77,26 +79,20 @@ def main(args):
...
+        # When perform_removal=True then exact_dup outputs dataset with the duplicates removed
+        # It'll behave by calling .identify_duplicates() and .removal() in sequence.
```
sarahyurick, line 85:
```suggestion
        # When perform_removal=True, then exact_dup outputs the dataset with the duplicates removed.
        # It will behave by calling .identify_duplicates() and .remove() in sequence.
```

nemo_curator/modules/config.py
```diff
@@ -44,6 +44,8 @@ class FuzzyDuplicatesConfig(BaseConfig):
         but might lead to memory pressures and related errors.
     id_field: Column in the Dataset denoting document ID.
     text_field: Column in the Dataset denoting document content.
+    perform_removal : bool, whether calling the module should remove the duplicates from
+        original dataset or just the set of duplicates
```
sarahyurick, line 48:
```suggestion
    perform_removal: Boolean value to specify whether calling the module should remove the duplicates from
        the original dataset, or return the list of IDs denoting duplicates.
```

nemo_curator/modules/config.py
```diff
@@ -131,6 +134,9 @@ def __post_init__(self):
         if not 1 <= self.buckets_per_shuffle <= self.num_buckets:
             raise ValueError("Buckets per shuffle must be between [1, num_buckets]")
 
+        if not self.perform_removal:
+            warnings.warn("In future release (0.8.0) by default will be True.")
```
sarahyurick, line 138:
```suggestion
            warnings.warn("In future releases (starting with 0.8.0) the default will be True.")
```

nemo_curator/modules/exact_dedup.py
```diff
@@ -64,9 +65,15 @@ def __init__(
             raise ValueError(
                 f"{hash_method} not in supported hash_methods. Choose a hash_method from {self.SUPPORTED_HASHES}"
             )
+
         self.hash_method = hash_method
         self.id_field = id_field
         self.text_field = text_field
+        self.perform_removal = perform_removal
+        if not self.perform_removal:
+            warnings.warn("In future release (0.8.0) by default will be True.")
```
sarahyurick, line 74:
```suggestion
            warnings.warn("In future releases (starting with 0.8.0) the default will be True.")
```

nemo_curator/modules/fuzzy_dedup/connectedcomponents.py
```diff
@@ -125,8 +125,6 @@ def _run_connected_components(
                 f"# rows in labels_df = {len(labels_df)}"
             )
             assert num_nodes == len(labels_df)
-            # Ensure all docs in the same group are in the same partition
-            labels_df = labels_df.shuffle(on=["group"], ignore_index=True)
```
praateekmahajan, line 129: @ayushdg we're doing this here

nemo_curator/modules/exact_dedup.py
```diff
@@ -166,10 +173,38 @@ def __call__(self, dataset: DocumentDataset) -> Union[DocumentDataset, str]:
         self._logger.info(
             f"Time taken for Exact Dedup Computation = {time.time() - t0}s and output written at {write_path}"
         )
-        if is_cudf_type(result):
-            import dask_cudf
+        backend = "cudf" if is_cudf_type(result) else "pandas"
+        return DocumentDataset.read_parquet(
```
ayushdg, line 177: Given that the result here is already shuffled on the hash key which cannot be avoided. Do you think it's worth leveraging that for the removal step, or do you prefer have custom removal logic at the cost of shuffling again?

nemo_curator/modules/exact_dedup.py
```diff
@@ -166,10 +173,38 @@ def __call__(self, dataset: DocumentDataset) -> Union[DocumentDataset, str]:
...
+            write_path,
+            backend=backend,
+            blocksize="1024MiB",
```
ayushdg, line 180: Somewhat related: `blocksize=None` is a valid approach when wanting to read in 1 row group per partition or 1 fpp in the dask world, while DocDataset doesn't work well when both this and app is set to None.

ayushdg, line 180: If reusing the shuffled info from this stage again, would need to read this back in as 1 fpp?

ayushdg (commented):
Left some initial comments. Would also be interested in reviewing the remove_duplicates utility once that's uploaded

VibhuJawa (approved):
LGTM

ryantwolf (approved):

nemo_curator/modules/exact_dedup.py
```diff
@@ -168,10 +177,38 @@ def call(self, dataset: DocumentDataset) -> Union[DocumentDataset, str]:
         self._logger.info(
             f"Time taken for Exact Dedup Computation = {time.time() - t0}s and output written at {write_path}"
         )
-        if is_cudf_type(result):
-            import dask_cudf
+        backend = "cudf" if is_cudf_type(result) else "pandas"
+        return DocumentDataset.read_parquet(
+            write_path,
+            backend=backend,
+            # we read with FPP=1 so that groups are read in whole (and don't exist across partitions)
```
sarahyurick, line 184:
```suggestion
            # We read with files_per_partition=1 so that groups are read in whole (and do not exist across partitions)
```

nemo_curator/modules/fuzzy_dedup/fuzzyduplicates.py
```diff
@@ -245,4 +247,41 @@ def call(self, dataset: DocumentDataset):
         print(f"Stage {stage_num}: Connected Components across buckets complete!")
         stage_num += 1
 
-        return DocumentDataset(dask_cudf.read_parquet(cc_path, split_row_groups=False))
+        return DocumentDataset.read_parquet(
+            cc_path,
+            backend="cudf",
+            # we read with FPP=1 so that groups are read in whole (and don't exist across partitions)
```
sarahyurick, line 253:
```suggestion
            # We read with files_per_partition=1 so that groups are read in whole (and do not exist across partitions)
```

sarahyurick:
Actually, I will cancel gpuCI since the last run looks good and only docs were updated.

ayushdg (approved):
Thanks lgtm!

Once we update the defaults to include removal as well,  tutorials and CI scripts will need updates as well.

PR #514, opened 2025-02-03
Enforce Dataframe Backend Checks

ryantwolf:
## Description

Refactors all modules to inherit from `nemo_curator.modules.base.Module`. With this change, the base class now checks to see if the dataframe backend for the `Module` matches what is passed in, and will throw an error prior to computation if there is a mismatch. We also prove a new module `ToBackend()` that allows easy chaining of CPU and GPU modules.

## Usage

```python
from nemo_curator import Module, Sequential, ToBackend

class CPUModule(Module):
    def __init__(self):
        super().__init__(input_backend="pandas")

    def call(self, dataset: DocumentDataset):
        dataset.df["cpu_lengths"] = dataset.df["text"].str.len()
        return dataset


class GPUModule(Module):
    def __init__(self):
        super().__init__(input_backend="cudf")

    def call(self, dataset: DocumentDataset):
        dataset.df["gpu_lengths"] = dataset.df["text"].str.len()
        return dataset

def main():
  client = get_client(cluster_type="gpu")
  # Load a dataset with a pandas CPU backend
  dataset = DocumentDataset.from_pandas(...)
  pipeline = Sequential([
    CPUModule(),
    ToBackend("cudf"),
    GPUModule(),
  ])

  curated_dataset = pipeline(dataset)
  curated_dataset.to_json(...)

if __name__ == "__main__":
  main()
```
## Checklist

### DIFF
--- a/docs/user-guide/cpuvsgpu.rst
+++ b/docs/user-guide/cpuvsgpu.rst
@@ -86,6 +86,43 @@ Even if you start a GPU dask cluster, you can't operate on datasets that use a `
 The ``DocuemntDataset`` must either have been originally read in with a ``cudf`` backend, or it must be transferred during the script.
 
 -----------------------------------------
+Moving data between CPU and GPU
+-----------------------------------------
+
+The ``ToBackend`` module provides a way to move data between CPU memory and GPU memory by swapping between pandas and cuDF backends for your dataset.
+To see how it works, take a look at this example.
+
+.. code-block:: python
+
+  from nemo_curator import Sequential, ToBackend, ScoreFilter, get_client
+  from nemo_curator.datasets import DocumentDataset
+  from nemo_curator.classifiers import DomainClassifier
+  from nemo_curator.filters import RepeatingTopNGramsFilter
+
+  def main():
+      client = get_client(cluster_type="gpu")
+
+      dataset = DocumentDataset.read_json("books.jsonl")
+      curation_pipeline = Sequential([
+          ScoreFilter(RepeatingTopNGramsFilter(n=5)),
+          ToBackend("cudf"),
+          DomainClassifier(),
+      ])
+
+      curated_dataset = curation_pipeline(dataset)
+
+      curated_dataset.to_json("curated_books.jsonl")
+
+  if __name__ == "__main__":
+      main()
+
+Let's highlight some of the important parts of this example.
+* ``client = get_client(cluster_type="gpu")``: Creates a local Dask cluster with access to the GPUs. In order to use/swap to a cuDF dataframe backend, you need to make sure you are running on a GPU Dask cluster.
+* ``dataset = DocumentDataset.read_json("books.jsonl")``: Reads in the dataset to a pandas (CPU) backend.
+* ``curation_pipeline = ...``: Defines a curation pipeline consisting of a CPU filtering step and a GPU classifier step. The ``ToBackend("cudf")`` in between moves the dataset from CPU to GPU for the classifier.
+* ``curated_dataset.to_json("curated_books.jsonl")``: Writes the dataset directly to disk from the GPU. There is no need to transfer back to the CPU before writing to disk.
+
+-----------------------------------------
 Dask with Slurm
 -----------------------------------------
 
--- a/docs/user-guide/cpuvsgpu.rst
+++ b/docs/user-guide/cpuvsgpu.rst
@@ -117,6 +117,7 @@ To see how it works, take a look at this example.
       main()
 
 Let's highlight some of the important parts of this example.
+
 * ``client = get_client(cluster_type="gpu")``: Creates a local Dask cluster with access to the GPUs. In order to use/swap to a cuDF dataframe backend, you need to make sure you are running on a GPU Dask cluster.
 * ``dataset = DocumentDataset.read_json("books.jsonl")``: Reads in the dataset to a pandas (CPU) backend.
 * ``curation_pipeline = ...``: Defines a curation pipeline consisting of a CPU filtering step and a GPU classifier step. The ``ToBackend("cudf")`` in between moves the dataset from CPU to GPU for the classifier.
--- a/docs/user-guide/cpuvsgpu.rst
+++ b/docs/user-guide/cpuvsgpu.rst
@@ -119,7 +119,7 @@ To see how it works, take a look at this example.
 Let's highlight some of the important parts of this example.
 
 * ``client = get_client(cluster_type="gpu")``: Creates a local Dask cluster with access to the GPUs. In order to use/swap to a cuDF dataframe backend, you need to make sure you are running on a GPU Dask cluster.
-* ``dataset = DocumentDataset.read_json("books.jsonl")``: Reads in the dataset to a pandas (CPU) backend.
+* ``dataset = DocumentDataset.read_json("books.jsonl")``: Reads in the dataset to a pandas (CPU) backend by default.
 * ``curation_pipeline = ...``: Defines a curation pipeline consisting of a CPU filtering step and a GPU classifier step. The ``ToBackend("cudf")`` in between moves the dataset from CPU to GPU for the classifier.
 * ``curated_dataset.to_json("curated_books.jsonl")``: Writes the dataset directly to disk from the GPU. There is no need to transfer back to the CPU before writing to disk.
 
--- a/docs/user-guide/cpuvsgpu.rst
+++ b/docs/user-guide/cpuvsgpu.rst
@@ -97,7 +97,7 @@ To see how it works, take a look at this example.
   from nemo_curator import Sequential, ToBackend, ScoreFilter, get_client
   from nemo_curator.datasets import DocumentDataset
   from nemo_curator.classifiers import DomainClassifier
-  from nemo_curator.filters import RepeatingTopNGramsFilter
+  from nemo_curator.filters import RepeatingTopNGramsFilter, NonAlphaNumericFilter
 
   def main():
       client = get_client(cluster_type="gpu")
@@ -107,6 +107,8 @@ To see how it works, take a look at this example.
           ScoreFilter(RepeatingTopNGramsFilter(n=5)),
           ToBackend("cudf"),
           DomainClassifier(),
+          ToBackend("pandas"),
+          ScoreFilter(NonAlphaNumericFilter()),
       ])
 
       curated_dataset = curation_pipeline(dataset)
@@ -120,7 +122,7 @@ Let's highlight some of the important parts of this example.
 
 * ``client = get_client(cluster_type="gpu")``: Creates a local Dask cluster with access to the GPUs. In order to use/swap to a cuDF dataframe backend, you need to make sure you are running on a GPU Dask cluster.
 * ``dataset = DocumentDataset.read_json("books.jsonl")``: Reads in the dataset to a pandas (CPU) backend by default.
-* ``curation_pipeline = ...``: Defines a curation pipeline consisting of a CPU filtering step and a GPU classifier step. The ``ToBackend("cudf")`` in between moves the dataset from CPU to GPU for the classifier.
+* ``curation_pipeline = ...``: Defines a curation pipeline consisting of a CPU filtering step, a GPU classifier step, and another CPU filtering step. The ``ToBackend("cudf")`` moves the dataset from CPU to GPU for the classifier, and the ``ToBackend("pandas")`` moves the dataset back to the CPU from the GPU for the last filter.
 * ``curated_dataset.to_json("curated_books.jsonl")``: Writes the dataset directly to disk from the GPU. There is no need to transfer back to the CPU before writing to disk.
 
 -----------------------------------------
--- a/nemo_curator/filters/doc_filter.py
+++ b/nemo_curator/filters/doc_filter.py
@@ -14,7 +14,7 @@
 
 import importlib
 from abc import ABC, abstractmethod
-from typing import Any, Union
+from typing import Any, Literal, Union
 
 from nemo_curator.filters.bitext_filter import BitextFilter
 
@@ -81,7 +81,8 @@ class DocumentFilter(ABC):
             "keep_document method must be implemented by subclasses"
         )
 
-    def get_backend(self) -> str:
+    @property
+    def backend(self) -> Literal["pandas", "cudf", "any"]:
         """
         The dataframe backend the filter operates on.
         Can be 'pandas', 'cudf', or 'any'. Defaults to 'pandas'.
--- a/nemo_curator/filters/doc_filter.py
+++ b/nemo_curator/filters/doc_filter.py
@@ -81,6 +81,15 @@ class DocumentFilter(ABC):
             "keep_document method must be implemented by subclasses"
         )
 
+    def get_backend(self) -> str:
+        """
+        The dataframe backend the filter operates on.
+        Can be 'pandas', 'cudf', or 'any'. Defaults to 'pandas'.
+        Returns:
+            str: A string representing the dataframe backend the filter needs as input
+        """
+        return "pandas"
+
     @property
     def name(self):
         return self._name
--- a/nemo_curator/modifiers/doc_modifier.py
+++ b/nemo_curator/modifiers/doc_modifier.py
@@ -26,3 +26,12 @@ class DocumentModifier(ABC):
     @abstractmethod
     def modify_document(self, text):
         pass
+
+    def get_backend(self) -> str:
+        """
+        The dataframe backend the filter operates on.
+        Can be 'pandas', 'cudf', or 'any'. Defaults to 'pandas'.
+        Returns:
+            str: A string representing the dataframe backend the filter needs as input
+        """
+        return "pandas"
--- a/nemo_curator/modifiers/doc_modifier.py
+++ b/nemo_curator/modifiers/doc_modifier.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 from abc import ABC, abstractmethod
+from typing import Literal
 
 
 class DocumentModifier(ABC):
@@ -27,11 +28,12 @@ class DocumentModifier(ABC):
     def modify_document(self, text):
         pass
 
-    def get_backend(self) -> str:
+    @property
+    def backend(self) -> Literal["pandas", "cudf", "any"]:
         """
-        The dataframe backend the filter operates on.
+        The dataframe backend the modifier operates on.
         Can be 'pandas', 'cudf', or 'any'. Defaults to 'pandas'.
         Returns:
-            str: A string representing the dataframe backend the filter needs as input
+            str: A string representing the dataframe backend the modifier needs as input
         """
         return "pandas"
--- a/nemo_curator/modules/__init__.py
+++ b/nemo_curator/modules/__init__.py
@@ -22,7 +22,7 @@ os.environ["TORCHINDUCTOR_COMPILE_THREADS"] = "1"
 from nemo_curator.utils.import_utils import gpu_only_import_from
 
 from .add_id import AddId
-from .base import Module
+from .base import BaseModule
 from .config import FuzzyDuplicatesConfig, SemDedupConfig
 from .dataset_ops import blend_datasets, Shuffle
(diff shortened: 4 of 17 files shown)


tests/test_backends.py
```diff
@@ -0,0 +1,214 @@
...
+from nemo_curator import Module, Sequential, ToBackend
+from nemo_curator.datasets import DocumentDataset
+from nemo_curator.utils.import_utils import gpu_only_import, gpu_only_import_from
+
+cudf = gpu_only_import("cudf")
+dask_cudf = gpu_only_import("dask_cudf")
+LocalCUDACluster = gpu_only_import_from("dask_cuda", "LocalCUDACluster")
+
+
+class CPUModule(Module):
```
sarahyurick, line 28: Was wondering if we can also have some type of test with at least 1 CPU and GPU module? Such as fuzzy duplicates?

docs/user-guide/cpuvsgpu.rst
```diff
@@ -85,6 +85,44 @@ To read a dataset into GPU memory, one could use the following function call.
...
+
+  def main():
+      client = get_client(cluster_type="gpu")
+
+      dataset = DocumentDataset.read_json("books.jsonl")
+      curation_pipeline = Sequential([
+          ScoreFilter(RepeatingTopNGramsFilter(n=5)),
+          ToBackend("cudf"),
+          DomainClassifier(),
+      ])
```
sarahyurick, line 110: This is very nice!

nemo_curator/modules/__init__.py
```diff
@@ -22,12 +22,14 @@
...
 from .config import FuzzyDuplicatesConfig, SemDedupConfig
 from .dataset_ops import blend_datasets, Shuffle
 from .exact_dedup import ExactDuplicates
 from .meta import Sequential
 from .modify import Modify
 from .task import TaskDecontamination
+from .to_backend import ToBackend
 
 # GPU packages
 MinHash = gpu_only_import_from("nemo_curator.modules.fuzzy_dedup.minhash", "MinHash")
```
sarahyurick, line 35: Do you think there is any way to take advantage of the module backends so that we don't need to use `gpu_only_import_from`?

tests/test_backends.py
ryantwolf: you mean with a real module, not just these dummy modules?

nemo_curator/modules/__init__.py
ryantwolf: Not quite sure what you mean. Are you asking if we could somehow determine that `MinHash` is a GPU module from its backend and then refactor this to do something different than `gpu_only_import_from`? I'm wondering what you think should be improved with these imports.

tests/test_backends.py
sarahyurick: Yes.

docs/user-guide/cpuvsgpu.rst
```diff
@@ -85,6 +85,44 @@ To read a dataset into GPU memory, one could use the following function call.
...
+
+      curated_dataset.to_json("curated_books.jsonl")
+
+  if __name__ == "__main__":
+      main()
+
+Let's highlight some of the important parts of this example.
+
+* ``client = get_client(cluster_type="gpu")``: Creates a local Dask cluster with access to the GPUs. In order to use/swap to a cuDF dataframe backend, you need to make sure you are running on a GPU Dask cluster.
+* ``dataset = DocumentDataset.read_json("books.jsonl")``: Reads in the dataset to a pandas (CPU) backend.
```
praateekmahajan, line 122: I think we should I explicitly mention the implicit default argument of `backend="pandas"`

nemo_curator/filters/doc_filter.py
```diff
@@ -81,6 +81,15 @@ def keep_document(self, scores: Any) -> bool:
             "keep_document method must be implemented by subclasses"
         )
 
+    def get_backend(self) -> str:
```
praateekmahajan, line 84:
```suggestion
    def get_backend(self) -> Literal["pandas", "cudf", "any"]:
```

(what's any though?)

nemo_curator/modules/base.py
```diff
@@ -0,0 +1,50 @@
...
+
+    @abstractmethod
+    def call(self, dataset: DocumentDataset) -> DocumentDataset:
+        raise NotImplementedError("call method must be implemented by subclasses")
+
+    def _check_backend(self, ddf):
+        if self.input_backend == "any":
+            return
+
+        backend = type(ddf._meta).__module__.split(".")[0]
```
praateekmahajan, line 41: Can we add a test case for this? And also type hints for ddf which would be `dd.DataFrame`

nemo_curator/modules/__init__.py
sarahyurick: I guess I was wondering if it would be possible to use the backend checks in this PR to help all modules be importable by users. For example if a user with CPU only tries to do an `from nemo_curator import FuzzyDuplicates` then maybe this file should raise a more readable message which specifies that the module can only be run with the GPU?

nemo_curator/modules/base.py
```diff
@@ -0,0 +1,50 @@
...
+            raise ValueError(
+                f"{input_backend} not one of the supported backends {self.SUPPORTED_BACKENDS}"
+            )
+        self.input_backend = input_backend
+
+    @abstractmethod
+    def call(self, dataset: DocumentDataset) -> DocumentDataset:
+        raise NotImplementedError("call method must be implemented by subclasses")
+
+    def _check_backend(self, ddf):
```
praateekmahajan, line 37: nit : name could be `_assert_correct_backend`

nemo_curator/filters/doc_filter.py
praateekmahajan: Seems like this could be an [abstractproperty](https://docs.python.org/3/library/abc.html#abc.abstractproperty) (now abstractmethod + property). 

And instead of the abstract class defining it, IMO the inherited class should explicitly define that property to be `pandas` or whatever

nemo_curator/modifiers/doc_modifier.py
```diff
@@ -26,3 +26,12 @@ def __init__(self):
...
         pass
+
+    def get_backend(self) -> str:
+        """
+        The dataframe backend the filter operates on.
+        Can be 'pandas', 'cudf', or 'any'. Defaults to 'pandas'.
+        Returns:
+            str: A string representing the dataframe backend the filter needs as input
+        """
+        return "pandas"
```
praateekmahajan, line 37: Same as above on type hint + abstractproperty + require explicit definition in child class rather than abstract class

nemo_curator/modules/to_backend.py
```diff
@@ -0,0 +1,25 @@
...
+from nemo_curator.modules.base import Module
+
+
+class ToBackend(Module):
+    def __init__(self, backend: str) -> None:
+        super().__init__(input_backend="any")
+        self.backend = backend
+
+    def call(self, dataset: DocumentDataset) -> DocumentDataset:
+        return DocumentDataset(dataset.df.to_backend(self.backend))
```
praateekmahajan, line 25: can add an explicit no-op if backend is the same if it's easy to check

nemo_curator/modules/to_backend.py
```diff
@@ -0,0 +1,25 @@
...
+# limitations under the License.
+
+from nemo_curator.datasets.doc_dataset import DocumentDataset
+from nemo_curator.modules.base import Module
+
+
+class ToBackend(Module):
+    def __init__(self, backend: str) -> None:
+        super().__init__(input_backend="any")
+        self.backend = backend
```
praateekmahajan, line 22: nit : type hint

nemo_curator/modules/base.py
```diff
@@ -0,0 +1,50 @@
...
+        self.name = name or self.__class__.__name__
+
+        if input_backend not in self.SUPPORTED_BACKENDS:
+            raise ValueError(
+                f"{input_backend} not one of the supported backends {self.SUPPORTED_BACKENDS}"
+            )
+        self.input_backend = input_backend
+
+    @abstractmethod
+    def call(self, dataset: DocumentDataset) -> DocumentDataset:
```
praateekmahajan, line 34: super nit : `call` could be slightly confusing for python readers, as it does overlap with the `__call__` dunder. What about we call it `fn` / `func` / `impl` / `implementation`?

nemo_curator/modules/base.py
```diff
@@ -0,0 +1,50 @@
...
+
+from abc import ABC, abstractmethod
+
+from nemo_curator.datasets import DocumentDataset
+
+
+class Module(ABC):
+    SUPPORTED_BACKENDS = ["pandas", "cudf", "any"]
+
+    def __init__(self, input_backend: str, name=None) -> None:
```
praateekmahajan, line 23: nit : type hints

tests/test_backends.py
```diff
@@ -0,0 +1,222 @@
...
+                ToBackend("cudf"),
+                GPUModule(),
+                ToBackend("pandas"),
+                CPUModule(),
+                ToBackend("cudf"),
+                GPUModule(),
+            ]
+        )
+        result = pipeline(dataset)
+        result_df = result.df.compute()
```
praateekmahajan, line 212: add a small assert here for list(df.columns) which should be `id, text, gpu_lengths, cpu_lengths`

praateekmahajan (changes requested):
This looks great. Left a review, with mostly nits around type hints and function renaming  and some class abstraction opinions.

nemo_curator/modules/__init__.py
ryantwolf: I don't think it can. In `FuzzyDuplicates` for example, the import would fail before the `FuzzyDuplicates` class would be called due to the missing `dask_cudf` import. And we'd need the `FuzzyDuplicates` class to be there in order to check its backend.

nemo_curator/modules/base.py
```diff
@@ -0,0 +1,50 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from abc import ABC, abstractmethod
+
+from nemo_curator.datasets import DocumentDataset
+
+
+class Module(ABC):
```
praateekmahajan, line 20: nit to rename this to `BaseModule`

nemo_curator/filters/doc_filter.py
ryantwolf: "any" means that the module can operate on either "pands" or "cudf" as a backend. Will update to be a property instead of a method.

> And instead of the abstract class defining it, IMO the inherited class should explicitly define that property to be pandas or whatever

The reason that I have the base class start off with a default of `pandas` is that right now I believe every DocumentFilter has a backend of pandas. In my mind, it's more convenient to set the default and have a few classes override it in the future if they want. What do you think?

nemo_curator/modules/base.py
ryantwolf: What part of this do you want to test? I'm not quite seeing which part of this you want me to isolate.

nemo_curator/modules/base.py
ryantwolf: what do you think about `_validate_correct_backend` instead? Since there's no direct `assert` statement or error thrown I think something slightly more generic is better.

nemo_curator/modules/base.py
```diff
@@ -0,0 +1,50 @@
...
+        if self.input_backend == "any":
+            return
+
+        backend = type(ddf._meta).__module__.split(".")[0]
+        if backend != self.input_backend:
+            raise ValueError(
+                f"Module {self.name} requires dataset to have backend {self.input_backend} but got backend {backend}"
+            )
+
+    def __call__(self, dataset: DocumentDataset) -> DocumentDataset:
```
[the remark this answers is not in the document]
ryantwolf: I'll leave the return type unannotated for now.

nemo_curator/modules/to_backend.py
ryantwolf: From what I can see dask already has a no-op in its implementation for pandas, I assume the cudf one is similar. I'll keep it like this for now.

nemo_curator/modules/base.py
sarahyurick: Hmm maybe something like `run`, `execute`, `dispatch`? Something that still makes it clear that we are calling the core functionality of that class.

ryantwolf:
@sarahyurick @praateekmahajan thanks for your reviews. I addressed the feedback that I immediately agreed with/understood. I left some follow up comments and questions on the other points.

docs/user-guide/cpuvsgpu.rst
```diff
@@ -85,6 +85,44 @@ To read a dataset into GPU memory, one could use the following function call.
...
+  from nemo_curator.filters import RepeatingTopNGramsFilter
+
+  def main():
+      client = get_client(cluster_type="gpu")
+
+      dataset = DocumentDataset.read_json("books.jsonl")
+      curation_pipeline = Sequential([
+          ScoreFilter(RepeatingTopNGramsFilter(n=5)),
+          ToBackend("cudf"),
+          DomainClassifier(),
```
ayushdg, line 109: It might be worth showing an example going the other way from cuDF to pandas via ToBackend in this short example as well.

nemo_curator/modules/base.py
```diff
@@ -0,0 +1,82 @@
...
+        Args:
+            dataset (DocumentDataset): The dataset to operate on
+        """
+        raise NotImplementedError("call method must be implemented by subclasses")
+
+    def _validate_correct_backend(self, ddf: dd.DataFrame):
+        if self.input_backend == "any":
+            return
+
+        backend = type(ddf._meta).__module__.split(".")[0]
```
ayushdg, line 67: Would the `is_cudf_type` utility help in this case? https://github.com/NVIDIA/NeMo-Curator/blob/c3fb61dfc37d25a916384c8484a8da2252e9e105/nemo_curator/utils/gpu_utils.py#L19-L28

nemo_curator/modules/base.py
```diff
@@ -0,0 +1,82 @@
...
+        if backend != self.input_backend:
+            raise ValueError(
+                f"Module {self.name} requires dataset to have backend {self.input_backend} but got backend {backend}"
+            )
+
```
ayushdg, line 72: In addition to `_validate_correct_backend`, it also makes sense to have something like `_validate_correct_cluster_type` that can also verify that for cuDF backends folks are using a GPU cluster. Perhaps the `getClient` api can set the cluster type in some variable? 

I realize that there is no clean way to do this today.

nemo_curator/modules/base.py
```diff
@@ -0,0 +1,82 @@
...
+        raise NotImplementedError("call method must be implemented by subclasses")
+
+    def _validate_correct_backend(self, ddf: dd.DataFrame):
+        if self.input_backend == "any":
+            return
+
+        backend = type(ddf._meta).__module__.split(".")[0]
+        if backend != self.input_backend:
+            raise ValueError(
+                f"Module {self.name} requires dataset to have backend {self.input_backend} but got backend {backend}"
```
ayushdg, line 70: Might be worth giving a hint here by mentioning that Users can use the `ToBackend` module to switch backends.

nemo_curator/modules/to_backend.py
```diff
@@ -0,0 +1,36 @@
...
+
+    def __init__(self, backend: Literal["pandas", "cudf"]) -> None:
+        """
+        Constructs a ToBackend module
+
+        Args:
+            backend (str): The backend to transfer the dataset to. Can be "pandas" or "cudf"
+        """
+        super().__init__(input_backend="any")
+        self.backend = backend
```
ayushdg, line 33: Might be worth adding the `supported_backends` list and check here as well to ensure users don't try to convert to an unsupported backend.

tests/test_backends.py
```diff
@@ -0,0 +1,437 @@
...
+        return dataset
+
+
+class AnyModule(BaseModule):
+    def __init__(self):
+        super().__init__(input_backend="any")
+
+    def call(self, dataset: DocumentDataset):
+        dataset.df["any_lengths"] = dataset.df["text"].str.len()
+        return dataset
```
ayushdg, line 69: repeated class

tests/test_backends.py
```diff
@@ -0,0 +1,437 @@
...
+
+
+@pytest.fixture
+def cpu_data():
+    df = pd.DataFrame(
```
ayushdg, line 74: nit: You could create a common fixture that returns the original data and have two fixtures that use the common one as input for generating cpu and gpu data instead of repeating the data twice.

eg: 
```python
@pytest.fixture
def raw_data():
...

@pytest.fixture
def cpu_data(raw_data):
    return ...

@pytest.fixture
def gpu_data(raw_data):
 ...
```

tests/test_backends.py
```diff
@@ -0,0 +1,437 @@
...
+        assert sorted(list(result_df.columns)) == [
+            "cpu_lengths",
+            "gpu_lengths",
+            "id",
+            "text",
+        ]
+        assert_eq(result_df["cpu_lengths"], gt_cpu_lengths)
+        assert_eq(result_df["gpu_lengths"], gt_gpu_lengths)
+
+    def test_wrong_backend(self, cpu_data):
```
ayushdg, line 239: nit: could you add the other way around as well?

tests/test_backends.py
```diff
@@ -0,0 +1,437 @@
...
+        with pytest.raises(ValueError):
+            print("client", self.client)
+            dataset, _ = cpu_data
+            pipeline = GPUModule()
+            result = pipeline(dataset)
+            _ = result.df.compute()
+
+
+@pytest.fixture
```
ayushdg, line 248: same as above

ayushdg (changes requested):

nemo_curator/modules/base.py
ayushdg: No strong opinions: open to using `run/execute/call`.

VibhuJawa: My opinion is actually to stick with `call`.  I think CrossFit call convention comes from Keras/TensorFlow (lol) Anyways, my mental model is call actually ends up calling `__call__ ` function so it lines up in my head. 

Other options can be `transform`, `process`  too.  

I slightly dont like `run`, `execute` because it has annotations of synchronous behavior in my head (but can be convinced of it too).

ryantwolf: Ok given that it's from keras/tensorflow and crossfit I'm going to keep it that way for now.

nemo_curator/modules/base.py
ryantwolf: I think it does make sense to do this. We could have a global variable `CLUSTER_TYPE` that gets set during the `get_client` call. Then, in the `BaseModule` we can just check this variable. The only problem that I see with this right now though is how to handle the case where we connect to an existing Dask cluster. We'd need to make sure that it also set some environment variable or something designating the cluster type. I think we can enforce this in our example slurm scripts, but do you mind if I do this in another PR?

ayushdg: Makes sense. Let's handle this in a followup. For cases where this variable is not set like when users are setting things up themselves without using our recommended scripts, we can keep the check lenient enough to still allow processing (maybe with a warning if preferred).

tests/test_backends.py
ryantwolf: sweet, thanks for letting me know!

sarahyurick (approved):
LGTM!

nemo_curator/modules/base.py
praateekmahajan: As someone reading this, it's not obviously clear what will be the output of `.__module__.split(".")[0]` of the type of  object's `_meta` attribute. So I would've proposed that you move this to a function and have tests for that and use that function here. 
However I see you've changed this to is_cudf_type(..) which is much more readable, and now we can unittest that directly that.

praateekmahajan (approved):
LGTM thanks for adding this!

ayushdg (approved):
lgtm! Thanks

PR #528, opened 2025-02-07
Pin Transformers version >= 4.48.0

sarahyurick:


pyproject.toml
```diff
@@ -66,8 +66,7 @@ dependencies = [
     "resiliparse",
     "sentencepiece",
     "spacy>=3.6.0, <3.8.0",
-    # TODO: Remove this pin once newer version is released
-    "transformers==4.46.3",
+    "transformers",
```
VibhuJawa, line 69: Should we pin to say not include broken version ? 

```suggestion
    "transformers!=4.47.0",
```

sarahyurick: Good point, although I think it was fixed by 4.48.0 so I might try `transformers>=4.48.0`.

VibhuJawa: I will also check if the upper bound works with Nvidia's  NeMo container. If it works, then a upper pin is great, If it does not a more permissive pin (which just prevents 1 version) might be the way to go.

ryantwolf (approved):

PR #539, opened 2025-02-11
Fix Dask Client weirdness with fuzzy deduplication PyTests

sarahyurick:
When https://github.com/NVIDIA/NeMo-Curator/pull/514 was merged, it created weirdness where some fuzzy deduplication PyTests from `test_fuzzy_dedup.py` were failing due to Dask Client crashes. When I remove the fuzzy deduplication tests from `test_backends.py`, the failures are fixed.

Alternatively, I tried renaming `test_backends.py` to `test_module_backends.py`, so that the `test_fuzzy_dedup.py` tests would run first, and that fixed them too. However, this caused the fuzzy deduplication tests in `test_module_backends.py` to fail instead.


ryantwolf (approved):
Strange that these tests interfere in this way. Glad you figured it out though.

sarahyurick:
Closing in favor of https://github.com/NVIDIA/NeMo-Curator/pull/540.

PR #565, opened 2025-02-21
Add a way to pass expected language to FastTextLangId filter

shuoyangd:
## Description

Currently, FastTextLangId filter only supports filtering by a language ID filter, but sometimes, we know what the language the data is supposed to be, and it would be a useful addition to keep only the data that matches with the expected language.

We use two-letter ISO-639 code to denote languages.

## Usage

Passing an extra argument when initializing the filter will make it check against expected language, for example:

```
FastTextLangId(model_path=FAST_TEXT_MODEL_DIR, lang=SRC_LANG)
```

If `lang` argument is not passed, it falls back to the old behavior of filtering by minimum language ID score.

`bitext_filtering` tutorial is updated to demonstrate how this is used in a pipeline.

## Checklist

(FastTextLangId filter is currently only tested with a fake emulator class. Not sure how to best cover this change with test.)

### DIFF
--- a/nemo_curator/filters/classifier_filter.py
+++ b/nemo_curator/filters/classifier_filter.py
@@ -68,14 +68,14 @@ class FastTextQualityFilter(DocumentFilter):
 
 class FastTextLangId(DocumentFilter):
 
-    def __init__(self, model_path=None, min_langid_score=0.3):
+    def __init__(self, model_path=None, min_langid_score=0.3, lang=None):
         if model_path is None:
             raise ValueError(
                 "Must provide a valid path to a FastText model "
                 "to identify languages with this filter"
             )
         self._model_path = model_path
-        self._lang_code = None
+        self._lang_code = lang
         self._cutoff = min_langid_score
         self._name = "lang_id"
 
@@ -91,14 +91,17 @@ class FastTextLangId(DocumentFilter):
             pp = text.strip().replace("\n", " ")
             label, score = model.predict(pp, k=1)
             score = score[0]
-            lang_code = label[0][-2:].upper()
+            lang_code = label[0][-2:].lower()
 
             return [score, lang_code]
 
         return df.apply(_score_document)
 
     def keep_document(self, score):
-        return score[0] >= self._cutoff
+        if self._lang_code:
+            return score[1] == self._lang_code
+        else:
+            return score[0] >= self._cutoff
 
     def _load_model(self):
         return fasttext.load_model(self._model_path)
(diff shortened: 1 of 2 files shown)


nemo_curator/filters/classifier_filter.py
```diff
@@ -68,14 +68,14 @@ def _load_model(self):
 
 class FastTextLangId(DocumentFilter):
 
-    def __init__(self, model_path=None, min_langid_score=0.3):
+    def __init__(self, model_path=None, min_langid_score=0.3, lang=None):
```
ryantwolf, line 71: Can you add a docstring for this function?

nemo_curator/filters/classifier_filter.py
```diff
@@ -91,14 +91,17 @@ def _score_document(text):
             pp = text.strip().replace("\n", " ")
             label, score = model.predict(pp, k=1)
             score = score[0]
-            lang_code = label[0][-2:].upper()
+            lang_code = label[0][-2:].lower()
```
ryantwolf, line 94: Can we keep it as upper for backwards compatibility? Or, is there some reason to prefer lower?

nemo_curator/filters/classifier_filter.py
```diff
@@ -68,14 +68,14 @@ def _load_model(self):
...
-    def __init__(self, model_path=None, min_langid_score=0.3):
+    def __init__(self, model_path=None, min_langid_score=0.3, lang=None):
         if model_path is None:
             raise ValueError(
                 "Must provide a valid path to a FastText model "
                 "to identify languages with this filter"
             )
         self._model_path = model_path
-        self._lang_code = None
+        self._lang_code = lang
```
ryantwolf, line 78: regardless of what we decide on with upper vs lower, can you also manually set `self._lang_code = lang.upper()` so the user doesn't have to remember which case to specify?

tutorials/bitext_cleaning/main.py
```diff
@@ -67,6 +72,15 @@ def filter_dataset(dataset: ParallelDataset, gpu: bool = False) -> ParallelDatas
         ]
     )
 
+    if FAST_TEXT_MODEL_DIR:
```
ryantwolf, line 75: Instead of making this a hardcoded directory, how do you feel about making this a CLI arg with argparse?

ryantwolf (commented):
Few comments, nothing major.

nemo_curator/filters/classifier_filter.py
shuoyangd: I'm trying to follow ISO-639-1 here which is 2-letter lowercase: https://en.wikipedia.org/wiki/List_of_ISO_639_language_codes

sarahyurick:
Would love to add this to our Ray refactor. If you are interested, this is how `FastTextLangId` will work after the Ray refactor: https://github.com/NVIDIA-NeMo/Curator/pull/746.

For now, I can open an issue to ensure that it gets added: https://github.com/NVIDIA-NeMo/Curator/issues/855.

sarahyurick:
Closing as out of date, and someone else has picked up https://github.com/NVIDIA-NeMo/Curator/issues/855.

PR #568, opened 2025-02-24
Add Regex Modifier

shuoyangd:
## Description

Add a modifier that performs regex replacements.

## Usage

```
regex_params = [
    {"pattern": "ö", "repl": "o"},
    {
        "pattern": "[^ !$%',-.0123456789;?ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz/:]",
        "repl": "",
    },
]

modifier = RegexModifier(regex_params)
# "Nein, es ist möglich🙃 " -> replacement ö ->
# "Nein, es ist moglich🙃 " -> remove anything other than alphanumeric characters and punctuations
# "Nein, es ist moglich " -> remove extra spaces -> "Nein, es ist moglich"
output = modifier.modify_document("Nein, es ist möglich🙃 ")  # returns "Nein, es ist moglich"
```

## Checklist

### DIFF
--- a/nemo_curator/modifiers/regex_modifier.py
+++ b/nemo_curator/modifiers/regex_modifier.py
@@ -1,0 +1,139 @@
+# Copyright (c) 2025, NVIDIA CORPORATION.  All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import re
+from typing import Dict, List
+
+from nemo_curator.modifiers import DocumentModifier
+
+__all__ = ["RegexModifier"]
+
+
+class RegexModifier(DocumentModifier):
+    """
+    A class for modifying documents using regular expressions.
+
+    This class applies a series of regex-based substitutions to an input text.
+    Each substitution rule is defined by a dictionary containing at least a
+    'pattern' and a 'repl' key, and optionally a 'count' key to limit the number
+    of substitutions.
+    """
+
+    def __init__(self, regex_params_list: List[Dict]):
+        """
+        Initialize the RegexModifier with a list of regex parameter dictionaries.
+
+        Args:
+            regex_params_list (List[Dict]): List of dictionaries where each dictionary
+                contains keys 'pattern' and 'repl' (and optionally 'count'). Each
+                dictionary defines a regex substitution rule.
+
+        Raises:
+            ValueError: If any dictionary in the list is missing the 'pattern' or 'repl' key.
+        """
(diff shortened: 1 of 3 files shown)


nemo_curator/modifiers/regex_modifier.py
```diff
@@ -0,0 +1,139 @@
...
+    'pattern' and a 'repl' key, and optionally a 'count' key to limit the number
+    of substitutions.
+    """
+
+    def __init__(self, regex_params_list: List[Dict]):
+        """
+        Initialize the RegexModifier with a list of regex parameter dictionaries.
+
+        Args:
+            regex_params_list (List[Dict]): List of dictionaries where each dictionary
```
ryantwolf, line 38: Can you give an example list? It was slightly hard to parse without looking at the code.

nemo_curator/modifiers/regex_modifier.py
```diff
@@ -0,0 +1,139 @@
...
+        Returns:
+            str: The cleaned string with extra spaces removed.
+
+        Examples:
+            >>> RegexModifier.remove_extra_spaces("abc  xyz   abc xyz")
+            'abc xyz abc xyz'
+            >>> RegexModifier.remove_extra_spaces(" abc xyz ")
+            'abc xyz'
+        """
+        output_string = " ".join(input_string.split())
```
ryantwolf, line 110: This function removes all newlines and replaces them with spaces, right? I think most folks want to preserve the newlines.

nemo_curator/modifiers/regex_modifier.py
```diff
@@ -0,0 +1,139 @@
...
+            2. Iteratively applying each regex substitution using the rules in regex_params_list.
+            3. Removing any extra spaces that might result from the substitutions.
+
+        Args:
+            text (str): The input text to modify.
+
+        Returns:
+            str: The modified text after all regex substitutions have been applied.
+        """
+        text_in = RegexModifier._add_start_end_spaces(text)
```
ryantwolf, line 74: Can you make the pre and post processing functions optional and default to not happen? They're useful, but users will likely get tripped up upon first using them if they're not aware of them being the default.

nemo_curator/modifiers/regex_modifier.py
```diff
@@ -0,0 +1,139 @@
...
+        for regex_params in self.regex_params_list:
+            text_out = re.sub(
+                pattern=regex_params["pattern"],
+                repl=regex_params["repl"],
+                string=text_in,
+                # note: this count param is the maximum number of pattern occurrences to be replaced.
+                count=regex_params.get("count", 0),
```
ryantwolf, line 81: Can you also allow users to specify the flags of the regex?

ryantwolf (commented):
This is a good modifier. We have a few other modifiers that basically do this but with hardcoded regexes. Can you find them and modify them to just inherit from this class and supply default regexes? `UrlRemover` and `MarkdownRemover` come to mind immediately, but there might be more that I missed.

ryantwolf:
Oh, could you also update the API docs at `docs/user-guide/api/modifiers.rst`?

nemo_curator/modifiers/regex_modifier.py
shuoyangd: By flags of the regex you mean [this](https://docs.python.org/3/library/re.html#flags)?

PR #570, opened 2025-02-25
Minor updates to duplicate removal

sarahyurick:
Fixes a few nits from https://github.com/NVIDIA/NeMo-Curator/pull/509.

### DIFF
--- a/nemo_curator/modules/fuzzy_dedup/fuzzyduplicates.py
+++ b/nemo_curator/modules/fuzzy_dedup/fuzzyduplicates.py
@@ -17,7 +17,6 @@ from __future__ import annotations
 import logging
 import os
 import time
-import warnings
 from typing import Optional, Union
 
 from nemo_curator.datasets import DocumentDataset
@@ -273,7 +272,7 @@ class FuzzyDuplicates(BaseModule):
         """
 
         if not duplicates_to_remove:
-            warnings.warn("No fuzzy duplicates to remove, returning original dataset")
+            print("No fuzzy duplicates to remove, returning original dataset")
             return dataset
 
         result = remove_duplicates(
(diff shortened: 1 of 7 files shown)


nemo_curator/modules/fuzzy_dedup/fuzzyduplicates.py
```diff
@@ -257,19 +258,24 @@ def identify_duplicates(
...
+          The dataset containing IDs of the fuzzy duplicates to remove
         Returns
         -------
         DocumentDataset containing only non-duplicate documents
         """
+
         if not duplicates_to_remove:
-            return None
+            warnings.warn("No fuzzy duplicates to remove, returning original dataset")
+            return dataset
```
sarahyurick, line 277: The main thing I wanted to change with this PR is to return the original dataset when there are no duplicates, instead of returning None.

praateekmahajan: I also prefer removing the Dataset, but I think @ayushdg might have some thoughts
https://github.com/NVIDIA/NeMo-Curator/pull/326#discussion_r1842311388

sarahyurick: I think this comment is a bit different than the change I am making here. This does not return an empty dataset, it returns the original dataset (AKA a dataset with no duplicates removed).

ayushdg: So it sounds like `identify` will still retain the behavior of returning None but `remove` might default to returning the original dataset.

sarahyurick: Yes that is correct.

praateekmahajan (approved):
LGTM!

PR #578, opened 2025-02-28
Add `TextFixerConfig` to `UnicodeReformatter`

sarahyurick:
Closes https://github.com/NVIDIA/NeMo-Curator/issues/476.

### DIFF
--- a/nemo_curator/modifiers/unicode_reformatter.py
+++ b/nemo_curator/modifiers/unicode_reformatter.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2024, NVIDIA CORPORATION.  All rights reserved.
+# Copyright (c) 2025, NVIDIA CORPORATION.  All rights reserved.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


nemo_curator/modifiers/unicode_reformatter.py
```diff
@@ -12,14 +12,138 @@
...
+        fix_c1_controls: bool = True,
+        fix_latin_ligatures: bool = False,
+        fix_character_width: bool = False,
+        uncurl_quotes: bool = False,
+        fix_line_breaks: bool = False,
+        fix_surrogates: bool = True,
+        remove_control_chars: bool = True,
+        normalization: Literal["NFC", "NFD", "NFKC", "NFKD"] | None = None,
+        max_decode_length: int = 1000000,
+        explain: bool = True,
```
sarahyurick, line 42: These are the default values set by DataTrove: https://github.com/huggingface/datatrove/pull/319.

tests/test_cleaning.py
```diff
@@ -58,7 +58,7 @@ def test_reformatting(self):
         ]
         expected_results.sort()
 
-        modifier = Modify(UnicodeReformatter())
+        modifier = Modify(UnicodeReformatter(uncurl_quotes=True))
```
sarahyurick, line 61: DataTrove uses a default value `uncurl_quotes=False`, with their reasoning being "while [`uncurl_quotes=False`] makes sense, models then [do not] recognize curly quotes at all".

See here: https://github.com/huggingface/datatrove/blob/main/src/datatrove/pipeline/formatters/ftfy.py#L32

cc @ryantwolf if you have an opinion here?

nemo_curator/modifiers/unicode_reformatter.py
ryantwolf: Summary of changes from ftfy defaults vs datatrove defaults.
| Parameter | `ftfy` default | `datatrove` default |
| ------ | ------------- | -------------------- |
| `fix_latin_ligatures` | `True` | `False` |
| `fix_character_width` | `True` | `False` |
| `uncurl_quotes` |  `True` | `False` |
| `fix_line_breaks` | `True` | `False` |
| `normalization` | `"NFC"` | `None` |

tests/test_cleaning.py
ryantwolf: What you have done here is good.

nemo_curator/modifiers/unicode_reformatter.py
sarahyurick: Thanks!

ryantwolf (approved):

PR #581, opened 2025-03-07
Add batched pairwise similarity method for Semantic Dedup

praateekmahajan:
## Description
Resolves https://github.com/NVIDIA/NeMo-Curator/issues/520

Currently if a single cluster is large enough, we'll likely OOM since `M @ M.T` requires `N**2` storage. A batched version breaks it up into smaller batches B and performs `M @ B.T`. The only thing we need to be careful is how to zero out the diagonals and get the upper triangular matrix.

## other nits
1. `get_cluster_reps -> get_normalized_cluster_reps` : now **always** l2 normalize the embedding vector so that `M @ M.T` results in an absolute max value of 1
1. renamed `_semdedup` to `pairwise_similarity`
3. added tests for existing function + the batched approach.
4. default now is batched implementation

## Usage

```python
```
## Checklist

### DIFF
--- a/nemo_curator/modules/semantic_dedup/semanticclusterleveldedup.py
+++ b/nemo_curator/modules/semantic_dedup/semanticclusterleveldedup.py
@@ -65,8 +65,8 @@ class SemanticClusterLevelDedup:
             embedding_column (str): The column name that stores the embeddings.
                 Default is "embeddings".
             batched_cosine_similarity (int): Whether to use batched cosine similarity (has less memory usage).
-                Default is 1024. When >0, batching is used and memory requirements are O(N*B) where N is the number of items in the cluster and B is the batch size.
-                When <=0, no batching is used and memory requirements are O(N^2) where N is the number of items in the cluster.
+                Default is 1024. When greater than 0, batching is used and memory requirements are O(N*B) where N is the number of items in the cluster and B is the batch size.
+                When less than or equal to 0, no batching is used and memory requirements are O(N^2) where N is the number of items in the cluster.
             logger (Union[logging.Logger, str]): Existing logger to log to, or a path to a log directory.
                 Default is "./".
             profile_dir (Optional[str]): If specified, directory to write Dask profile.
--- a/nemo_curator/modules/semantic_dedup/semanticclusterleveldedup.py
+++ b/nemo_curator/modules/semantic_dedup/semanticclusterleveldedup.py
@@ -23,7 +23,6 @@ import dask.bag as db
 
 from nemo_curator.datasets import DocumentDataset
 from nemo_curator.log import create_logger
-from nemo_curator.modules.config import SemDedupConfig
 from nemo_curator.utils.distributed_utils import performance_report_if_with_ts_suffix
 from nemo_curator.utils.file_utils import expand_outdir_and_mkdir
 from nemo_curator.utils.semdedup_utils import (
@@ -43,6 +42,7 @@ class SemanticClusterLevelDedup:
         which_to_keep: str = "hard",
         output_dir: str = "./clustering_results",
         embedding_column: str = "embeddings",
+        batched_cosine_similarity: int = 1024,
         logger: Union[logging.Logger, str] = "./",
         profile_dir: Optional[str] = None,
     ) -> None:
@@ -64,6 +64,9 @@ class SemanticClusterLevelDedup:
                 Default is "./clustering_results".
             embedding_column (str): The column name that stores the embeddings.
                 Default is "embeddings".
+            batched_cosine_similarity (int): Whether to use batched cosine similarity (has less memory usage).
+                Default is 1024. When >0, batching is used and memory requirements are O(N*B) where N is the number of items in the cluster and B is the batch size.
+                When <=0, no batching is used and memory requirements are O(N^2) where N is the number of items in the cluster.
             logger (Union[logging.Logger, str]): Existing logger to log to, or a path to a log directory.
                 Default is "./".
             profile_dir (Optional[str]): If specified, directory to write Dask profile.
@@ -82,6 +85,7 @@ class SemanticClusterLevelDedup:
         )
         self.computed_semantic_match_dfs = False
         self.embedding_column = embedding_column
+        self.batched_cosine_similarity = batched_cosine_similarity
         self.logger = self._setup_logger(logger)
         self.profile_dir = profile_dir
 
@@ -144,6 +148,7 @@ class SemanticClusterLevelDedup:
                     output_dir=self.semdedup_pruning_tables_dir,
                     embedding_col=self.embedding_column,
                     which_to_keep=self.which_to_keep,
+                    batched_cosine_similarity=self.batched_cosine_similarity,
                 )
             )
             tasks.compute()
--- a/nemo_curator/utils/semdedup_utils.py
+++ b/nemo_curator/utils/semdedup_utils.py
@@ -18,7 +18,7 @@ import os
 import random
 import shutil
 import time
-from typing import List, Literal, Optional, Tuple, Union
+from typing import List, Literal, Optional, Tuple
 
 import cudf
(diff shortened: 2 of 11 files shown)


nemo_curator/modules/semantic_dedup/semanticclusterleveldedup.py
```diff
@@ -64,6 +64,9 @@ def __init__(
                 Default is "./clustering_results".
             embedding_column (str): The column name that stores the embeddings.
                 Default is "embeddings".
+            batched_cosine_similarity (int): Whether to use batched cosine similarity (has less memory usage).
+                Default is 1024. When >0, batching is used and memory requirements are O(N*B) where N is the number of items in the cluster and B is the batch size.
```
sarahyurick, line 68:
```suggestion
                Default is 1024. When greater than 0, batching is used and memory requirements are O(N*B) where N is the number of items in the cluster and B is the batch size.
```

nemo_curator/modules/semantic_dedup/semanticclusterleveldedup.py
```diff
@@ -64,6 +64,9 @@ def __init__(
...
+                When <=0, no batching is used and memory requirements are O(N^2) where N is the number of items in the cluster.
```
sarahyurick, line 69:
```suggestion
                When less than or equal to 0, no batching is used and memory requirements are O(N^2) where N is the number of items in the cluster.
```

nemo_curator/utils/semdedup_utils.py
```diff
@@ -179,22 +179,62 @@ def rank_within_cluster(
...
+    """
+    Compute pairwise cosine similarity between cluster items,
+    then replace to diagonal with zeros to ignore self similarity
+    """
+    # Move to device
+    cluster_reps = cluster_reps.to(device)
+    # Normalize embeddings
+    cluster_reps = cluster_reps / cluster_reps.norm(dim=1, keepdim=True)
+    # Compute pairwise cosine similarity
+    pairwise_sim_matrix = cluster_reps @ (cluster_reps.T)
```
sarahyurick, line 195: Haha I don't think I've used `@` before. Would there be any advantage to using `torch.mm`, `torch.matmul`, etc.?

tests/test_semdedup.py
```diff
@@ -233,3 +238,62 @@ def get_reference_embeddings(
...
+                [[1, 2, 3], [4, 5, 6], [7, 8, 9], [10, 11, 12], [1, 2, 3]],
+            ),
+            dtype=torch.float32,
+        )
+        self.expected_similarity = torch.tensor(
+            [0.0000, 0.974631, 0.998190, 0.999618, 1.0000]
+        )
+        self.expected_indices = [0, 0, 1, 2, 0]
+
+    # paramterize device
```
sarahyurick, line 256:
```suggestion
```

tutorials/image-curation/image-curation.ipynb
```diff
@@ -681,6 +681,7 @@
     "    id_column_type=\"str\",\n",
     "    embedding_col=\"image_embedding\",\n",
     "    which_to_keep=\"hard\",\n",
+    "    batched_cosine_similarity=1024,\n",
```
sarahyurick, line 684: Has this been tested?

sarahyurick (commented):
Nice PR, added a couple minor comments.

tutorials/image-curation/image-curation.ipynb
praateekmahajan: Nope, do we always manually run these notebooks for such PRs? That'll be a time sink but I'm okay to do it that's the practice

sarahyurick: If it is expected to produce the same results as before, it is okay with me. Sometimes I leave notebooks unchanged (or add changes to have it keep the previous default) if the output is expected to change, so that the user won't be confused when their cell outputs are different than the ones on GitHub.

sarahyurick: But it sounds like that isn't the case here?

nemo_curator/utils/semdedup_utils.py
```diff
@@ -179,22 +179,62 @@ def rank_within_cluster(
...
-    cluster_reps.to(device)
-    pair_w_sim_matrix = cluster_reps @ (cluster_reps.T)
+    """
+    Compute pairwise cosine similarity between cluster items,
+    then replace to diagonal with zeros to ignore self similarity
+    """
+    # Move to device
+    cluster_reps = cluster_reps.to(device)
+    # Normalize embeddings
+    cluster_reps = cluster_reps / cluster_reps.norm(dim=1, keepdim=True)
```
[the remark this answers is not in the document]
VibhuJawa: Minor nit, but can we move the normalization logic to inside the `get_cluster_reps ` function, so that we dont repeat it in both `pairwise_cosine_similarity_batched ` and `pairwise_cosine_similarity `

tutorials/image-curation/image-curation.ipynb
praateekmahajan: It'll be the same result as before, and the notebook output should also not change

nemo_curator/utils/semdedup_utils.py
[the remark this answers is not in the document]
praateekmahajan: @ryantwolf It can be configurable, however I'm not sure if we need more configs in `SemDedupConfig` 😢 

I think given the time consumption of other steps this a minor drop in the bucket that I won't expect to move the needle much if not at all.

tests/test_semdedup.py
```diff
@@ -233,3 +238,61 @@ def get_reference_embeddings(
...
+        self, device: Literal["cpu", "cuda"], batch_size: int
+    ):
+        N = 1024
+        D = 512
+        rand_arr = torch.randn(N, D, device=device)
+        max_similarity, max_indices = pairwise_cosine_similarity(rand_arr, device)
+        max_similarity_batched, max_indices_batched = (
+            pairwise_cosine_similarity_batched(rand_arr, device, batch_size=batch_size)
+        )
+        assert torch.allclose(max_similarity, max_similarity_batched, atol=1e-3)
```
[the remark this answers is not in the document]
praateekmahajan: Not sure if this is still relevant, as I've removed the lower tolerance of only 1e-3 and even though it's still `assert_close` the default tolerance is 1e-6 for fp32

VibhuJawa (approved):
LGTM to me. 

Thanks for pushing on this

tests/test_semdedup.py
```diff
@@ -233,3 +245,57 @@ def get_reference_embeddings(
...
+        self, device: Literal["cpu", "cuda"], batch_size: int
+    ):
+        N = 1024
+        D = 512
+        rand_arr = torch.randn(N, D, device=device)
+        max_similarity, max_indices = pairwise_cosine_similarity(rand_arr, device)
+        max_similarity_batched, max_indices_batched = (
+            pairwise_cosine_similarity_batched(rand_arr, device, batch_size=batch_size)
+        )
+        torch.testing.assert_close(max_similarity, max_similarity_batched)
```
praateekmahajan, line 300: @ryantwolf even for this where we thought assert equality should be fine, even then it wasn't. Long live floating imprecisions

E       Mismatched elements: 1 / 1024 (0.1%)
E       Greatest absolute difference: 1.33514404296875e-05 at index (3,) (up to 1e-05 allowed)
E       Greatest relative difference: 6.5386520873289555e-06 at index (3,) (up to 1.3e-06 allowed)

nemo_curator/utils/semdedup_utils.py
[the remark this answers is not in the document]
ryantwolf: I was also concerned that maybe the algorithm could be done without normalization entirely for some reason, but rereading the paper since they normalized the embeddings I suppose it's fine.

PR #594, opened 2025-03-17
Add more tests to test_dataset

sarahyurick:


tests/test_dataset.py
```diff
@@ -1,10 +1,98 @@
...
+def test_to_pickle(tmpdir):
+    original_df = pd.DataFrame({"first_col": [1, 2, 3], "second_col": ["a", "b", "c"]})
+    dataset = DocumentDataset.from_pandas(original_df)
+
+    output_file = str(tmpdir / "output.pkl")
+    with pytest.raises(NotImplementedError):
+        dataset.to_pickle(output_file)
+
+
+def test_read_json_or_parquet(tmpdir):
```
ryantwolf, line 68: Why do you test this function instead of `DocumentDataset.read_json` or `DocumentDataset.read_parquet` directly?

sarahyurick: Good point, I think either way would be okay. I did `_read_json_or_parquet` since I was leafing through uncovered code in https://app.codecov.io/gh/NVIDIA/NeMo-Curator/blob/main/nemo_curator%2Fdatasets%2Fdoc_dataset.py, it is a more direct test.

ryantwolf (approved):

PR #602, opened 2025-03-21
Automatically run gpuCI on PR updates

sarahyurick:
Currently, a user wishing to run gpuCI on a PR must remove and re-add the `gpuci` label in order to rerun it after pushing new commits. This PR allows us to autorun gpuCI when new commits are pushed, as long as the PR has the `gpuci` label on it.

### DIFF
--- a/.github/workflows/gpuci.yml
+++ b/.github/workflows/gpuci.yml
@@ -11,8 +11,7 @@ on:
       - "[rv][0-9].[0-9].[0-9]"
       - "[rv][0-9].[0-9].[0-9]rc[0-9]"
     # PR has to be labeled with "gpuCI" label
-    # If new commits are added, the "gpuCI" label has to be removed and re-added to rerun gpuCI
-    types: [labeled]
+    types: [labeled, synchronize]
 
 concurrency:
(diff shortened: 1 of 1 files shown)


praateekmahajan (approved):

ryantwolf (changes requested):
Can @ko3n1g give his input? I'm worried this is a security concern.

.github/workflows/gpuci.yml
```diff
@@ -11,8 +11,7 @@ on:
       - "[rv][0-9].[0-9].[0-9]"
       - "[rv][0-9].[0-9].[0-9]rc[0-9]"
     # PR has to be labeled with "gpuCI" label
-    # If new commits are added, the "gpuCI" label has to be removed and re-added to rerun gpuCI
-    types: [labeled]
+    types: [labeled, synchronize]
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
```
ko3n1g, line 17: It's unrelated to this PR: But if we add another part like `${{ github.event.label.name }}` of the group we can prevent that we're accidentally restarting a running CI when an (unrelated) label is added. I'd strongly recommend this change in this or another PR:)

PR #618, opened 2025-04-02
Semantic Dedup bug fix to support string id types

praateekmahajan:
## Description
`cupy` doesn't support string arrays, therefore `.values` won't work as intended. 

Originally I was going to use `.to_arrow()` to solve this problem, but then i figured we could just let it be a cudf.Series

## Usage

```python
```
## Checklist


VibhuJawa (approved):
LGTM

VibhuJawa:
The `test_classifiers` is failing right now. I think we should merge this one in while we fix CI.

PR #606, opened 2025-03-23
Add miscellaneous unit tests

ryantwolf:
## Description

Adds miscellaneous unit tests to test

- `nemo_curator/services`
- `nemo_curator/nemo_run`
- `nemo_curator/utils/distributed_utils.py`
- `nemo_curator/utils/file_utils.py`
- reactivated aegis tests in `tests/test_classifiers.py`

## Usage

N/A
## Checklist

- [x] I am familiar with the [Contributing Guide](https://github.com/NVIDIA/NeMo-Curator/blob/main/CONTRIBUTING.md).
- [x] New or Existing tests cover these changes.
- [x] The documentation is up to date with these changes.

PR #620, opened 2025-04-02
Fix gpuci errors around huggingface timeouts

ayushdg:
## Description

Prior to #606, gpuci would skip tests around the aegis classifiers. 
The updates allow those tests to work using a HF token if CI is running on main or a branch with the repo, but for security reasons ci running on branches of forks don't have access to secrets and would fail. 

This PR just updates the conditional so that tests are skipped with the HF_TOKEN env variable is not set or set to an empty string (the value propagated via GitHub Actions if the token value evaluates to null), skipping these tests when run from a fork. 

We can explore options in a followup that use the `pull_request_target` event or a different approach to allow pr's by forks to run the given tests. 

## Usage

N/A
## Checklist


VibhuJawa (approved):
LGTM

PR #626, opened 2025-04-04
Support writing gzip compressed jsonl files

ayushdg:
## Description

Adds support for writing gzip compressed jsonl files which was always supported in the pandas backend but also added in recent releases for the cuDF backend as well.

## Usage

```python
doc_dataset.to_json(..., compression="gzip")
```
## Checklist

- [X] I am familiar with the [Contributing Guide](https://github.com/NVIDIA/NeMo-Curator/blob/main/CONTRIBUTING.md).
- [X] New or Existing tests cover these changes.
- [X] The documentation is up to date with these changes.

PR #631, opened 2025-04-07
Fix bug when removing extensions from a filename

ayushdg:
## Description

Fixes a bug introduced in #626 which ends up truncating filenames of files like `abc.xyz.def.file_ext` style names. 
The current solution is a stopgap but I suspect there's something more robust that can be done here by utilizing fsspec or other filesystem utils and refactoring the IO sections.

## Usage

```python
```
## Checklist


sarahyurick (approved):
LGTM!

PR #635, opened 2025-04-07
text modality doc updates for 24.04

lbliii:
## Description

- New article for stop words
- Added mention of gzip support for jsonl
- Added mention of trafilatura support

Please review the stopwords article, as it's net new.


docs/user-guide/documentdataset.rst
```diff
@@ -79,6 +78,32 @@ Let's walk through this code line by line.
...
+
+.. code-block:: python
+
+    # Read gzip compressed JSONL files
+    compressed_books = DocumentDataset.read_json("books_dataset/*.jsonl.gz")
+
+    # You can also mix compressed and uncompressed files
+    all_books = DocumentDataset.read_json(["books_dataset/books_00.jsonl.gz", 
+                                         "books_dataset/books_01.jsonl"])
+
```
ayushdg, line 103: Would recommend avoid this and even mentioning something like avoiding mixing compressed and uncompressed files as it's undefined behavior and may break in certain cases/with certain options.

docs/user-guide/personalidentifiableinformationidentificationandremoval.rst
```diff
@@ -241,7 +268,38 @@ For example:
...
+
+When working with non-English text or when you want to customize how the LLM identifies PII entities, you can provide a custom system prompt. However, ensure that the JSON schema is included exactly as shown in the default system prompt.
+
+For reference, the default system prompt is:
+
+.. code-block:: text
+
+    "You are an expert redactor. The user is going to provide you with some text. 
+    Please find all personally identifying information from this text. 
+    Return results according to this JSON schema: {JSON_SCHEMA}
```
sarahyurick, line 296: I think the most important part here is to specify:
```
JSON_SCHEMA = {
    "type": "array",
    "items": {
        "type": "object",
        "required": ["entity_type", "entity_text"],
        "properties": {
            "entity_type": {"type": "string"},
            "entity_text": {"type": "string"},
        },
    },
}
```

docs/user-guide/personalidentifiableinformationidentificationandremoval.rst
```diff
@@ -241,7 +268,38 @@ For example:
...
+    Only return results for entities which actually appear in the text. 
+    It is very important that you return the entity_text by copying it exactly from the input. 
+    Do not perform any modification or normalization of the text. 
+    The entity_type should be one of these: {PII_LABELS}"
```
sarahyurick, line 300: Not as important, but maybe specify that `PII_LABELS` is a list of strings, e.g., `name`, `ip_addres`, etc.

docs/user-guide/stopwords.rst
```diff
@@ -0,0 +1,179 @@
...
+
+Why Stop Words Matter in NeMo Curator
+------------------------------------
+
+In NeMo Curator, stop words play several important roles:
+
+* **Text Extraction**: The text extraction process (especially for Common Crawl data) uses stop word density as a key metric to identify meaningful content
+* **Language Detection**: Stop words help in language detection and processing
+* **Efficient Processing**: Filtering stop words reduces the amount of data that needs to be processed
+* **Boilerplate Removal**: Stop word density helps differentiate between main content and boilerplate in web pages
```
sarahyurick, line 21: The **Text Extraction** and **Boilerplate Removal** sections go hand in hand and mean almost the same thing semantically, consider combining them into a single bullet point.

docs/user-guide/stopwords.rst
```diff
@@ -0,0 +1,179 @@
...
+   * - Chinese
+     - ``zh_stopwords.py``
+     - ~800
+
+These stop word lists are implemented as Python frozen sets for efficient lookup and immutability.
+
+Thai Stop Words
+~~~~~~~~~~~~~~
+
+Thai stop words are available in two files: ``thai_stopwords.py`` and ``th_stopwords.py``. Both contain around 120 common Thai words like "กล่าว" (to say), "การ" (the), and "ของ" (of).
```
sarahyurick, line 50: There is only `th_stopwords.py` (which was originally called `thai_stopwords.py` but the original name does not need to be mentioned in documentation).

docs/user-guide/stopwords.rst
```diff
@@ -0,0 +1,179 @@
...
+
+* **Text Extraction**: The text extraction process (especially for Common Crawl data) uses stop word density as a key metric to identify meaningful content
+* **Language Detection**: Stop words help in language detection and processing
+* **Efficient Processing**: Filtering stop words reduces the amount of data that needs to be processed
+* **Boilerplate Removal**: Stop word density helps differentiate between main content and boilerplate in web pages
+
+Available Stop Word Lists
+------------------------
+
+NeMo Curator includes built-in stop word lists for the following languages:
```
sarahyurick, line 26: This table is not accurate - we support all the languages accessible by `justext.get_stoplist`, which are listed here: https://github.com/miso-belica/jusText/tree/main/justext/stoplists

Since jusText does not have stoplists for Thai, Japanese, or Chinese, we have the lists in Curator.

sarahyurick (changes requested):
Very nice, thank you! Added some suggestions.

docs/user-guide/stopwords.rst
```diff
@@ -0,0 +1,179 @@
...
+
+Stop words are a critical component in NeMo Curator's text extraction algorithms. Here's how they're used in different extractors:
+
+JusText Extractor
+~~~~~~~~~~~~~~~~
+
+The JusText algorithm uses stop word density to classify text blocks as main content or boilerplate:
+
+1. **Context-Free Classification**: Text blocks with a high density of stop words are classified as "good" (likely main content)
+2. **Parameter Customization**: You can customize the stop word density thresholds via ``stopwords_low`` and ``stopwords_high`` parameters
```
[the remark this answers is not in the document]
sarahyurick: For example, in the docstrings we have "Studies on stopword lists and their distribution in various text corpora often suggest that around 30-40% of a typical English text consists of stopwords."

docs/user-guide/personalidentifiableinformationidentificationandremoval.rst
```diff
@@ -241,7 +268,147 @@ For example:
...
+This will raise the following error:
+
+.. code-block:: text
+
+    ValueError: No matching recognizers were found to serve the request.
+
+Solution
+~~~~~~~~~~~~~~~~~~~~~~
+
+To enable multilingual PII redaction, modify lines ``99-102`` in ``algorithm.py`` by passing an additional argument ``models`` to ``CustomNlpEngine``. This argument must be a list of dictionaries, where each dictionary contains 2 key-value pairs: ``lang_code`` and ``model_name``.
```
sarahyurick, line 369: I didn't realize this solution involves editing source code. I don't think we should include this, especially since line numbers can change with commits. A better approach would be to hold off on this section for now until we can edit the source code to implicitly handle multilingual PII. What do you think?

lbliii: sure, that sounds good to me

sarahyurick: Opened https://github.com/NVIDIA/NeMo-Curator/issues/661 for now. Thanks!

sarahyurick (approved):

sarahyurick:
Please fix the failing pre-commit CI, then it should be ready to merge.

PR #650, opened 2025-04-09
Ruff - Base PR

praateekmahajan:
Closes https://github.com/NVIDIA/NeMo-Curator/issues/599.


pyproject.toml
```diff
@@ -188,3 +185,42 @@ version = { attr = "nemo_curator.package_info.__version__" }
 [tool.setuptools.packages.find]
 include = ["*"]
 exclude = ["tests", "tests.*"]
+
+[tool.ruff]
+line-length = 119
```
sarahyurick, line 190: Should line length be 119 or 88 like it was before?

praateekmahajan: I changed it to follow what we have in https://github.com/NVIDIA/NeMo/blob/4609f97f476e66d8c833bf5aa9134b4d827c9da5/pyproject.toml#L86 based on @ayushdg 's comment

sarahyurick (approved):
LGTM! It looks like the only GPU test that failed was an Aegis classifier one, due to connection issues. I think that is okay to merge.

PR #653, opened 2025-04-10
Ruff formatting for `nemo_curator/tasks` and `nemo_curator/modules`

sarahyurick:
Closes https://github.com/NVIDIA/NeMo-Curator/issues/645.


nemo_curator/modules/dataset_ops.py
```diff
@@ -136,23 +136,23 @@ def blend_datasets(
...
     weight_sum = sum(sampling_weights)
     sampling_weights = [weight / weight_sum for weight in sampling_weights]
-    num_documents_per_dataset = [
-        math.ceil(weight * target_size) for weight in sampling_weights
-    ]
+    num_documents_per_dataset = [math.ceil(weight * target_size) for weight in sampling_weights]
 
     blend_components = []
-    for dataset, num_documents in zip(datasets, num_documents_per_dataset):
+    for dataset, num_documents in zip(datasets, num_documents_per_dataset, strict=False):
```
praateekmahajan, line 149: Shouldn't this be `strict=True`?

sarahyurick: By default Python 3.10 sets `strict=False`, so the above should maintain the same behavior as before.

nemo_curator/modules/fuzzy_dedup/jaccardsimilarity.py
```diff
@@ -166,34 +151,29 @@ def _compute_jaccard_pair(self, docs_df, anchor_df):
...
         nbytes = df[text_field].str.byte_count().sum()
         # Number of exmploded bytes
         exploded_bytes = nbytes * 5 * 2
         max_chars_allowed = 2_147_483_647
         byte_ratio = int(exploded_bytes) // max_chars_allowed
-        if byte_ratio > 1:
-            nrows_at_once = len(df) // byte_ratio
-        else:
-            nrows_at_once = len(df)
+        nrows_at_once = len(df) // byte_ratio if byte_ratio > 1 else len(df)
```
praateekmahajan, line 177: should this be covered in paranthesis? maybe ruff removes it even if you add them
```suggestion
        nrows_at_once = len(df) // (byte_ratio if byte_ratio > 1 else len(df))
```

sarahyurick: I think the current way is already correct. With your way, it would divide by `byte_ratio` if it is greater than 1, else divide by `len(df)`, which is incorrect.

praateekmahajan: Oh I see, it was the confusion I had about the behavior when we have variable = `a OPERATOR b if CONDITION else c`

nemo_curator/modules/semantic_dedup/embeddings.py
```diff
@@ -17,14 +17,13 @@
...
 import time
 from dataclasses import dataclass
-from typing import Optional, Union
 
 import dask_cudf
 import torch
-import torch.nn as nn
 from crossfit import op
 from crossfit.backend.torch.hf.model import HFModel
-from torch.nn import functional as F
```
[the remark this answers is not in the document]
sarahyurick: Yeah... we already use it as `torch.nn.functional` a couple places in our codebase, so I don't mind either way.

praateekmahajan (approved):
LGTM! Let's merge to ruff

PR #654, opened 2025-04-10
Ruff :  `examples` and `nemo_curator/scripts`

praateekmahajan:
## Description
Closes https://github.com/NVIDIA/NeMo-Curator/issues/652
## Usage

```python
```
## Checklist

### DIFF
--- a/nemo_curator/scripts/filter_documents.py
+++ b/nemo_curator/scripts/filter_documents.py
@@ -15,7 +15,10 @@
 import argparse
 import os
 
+import dask.dataframe as dd
+
 import nemo_curator
+from nemo_curator import Sequential
 from nemo_curator.datasets import DocumentDataset
 from nemo_curator.utils.config_utils import build_filter_pipeline
 from nemo_curator.utils.distributed_utils import get_client, read_data, write_to_disk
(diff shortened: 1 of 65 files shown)


nemo_curator/scripts/verify_classification_results.py
sarahyurick, line 1: TBH, I would be fine with deleting this file altogether. It was originally created to help verify correct results for the domain classifier, but I don't think it is used for anything else now.

nemo_curator/scripts/find_matching_ngrams.py
```diff
@@ -22,21 +22,19 @@
 from nemo_curator.utils.script_utils import ArgumentHelper
 
 
-def main(args):
-    client = get_client(**ArgumentHelper.parse_client_args(args))
+def main(args: argparse.Namespace) -> None:
+    get_client(**ArgumentHelper.parse_client_args(args))
```
sarahyurick, line 26: I like this a lot actually. Do you think we could do `get_client(...)` instead of `client = get_client(...)` everywhere that complains about F841?

nemo_curator/scripts/semdedup/clustering.py
```diff
@@ -46,15 +45,11 @@ def main(args):
     )
 
     client = get_client(**ArgumentHelper.parse_client_args(args))
-    dt1 = datetime.now()
+    dt1 = datetime.now()  # noqa: DTZ005
```
sarahyurick, line 48: Instead of `datetime.now()` it could be `time.perf_counter()` like elsewhere in this PR?

nemo_curator/scripts/semdedup/clustering.py
```diff
@@ -74,7 +69,7 @@ def main(args):
 
     clustered_embeddings = clustering_model(embedding_dataset)
     clustered_embeddings.df.head(10)
-    dt2 = datetime.now()
+    dt2 = datetime.now()  # noqa: DTZ005
```
sarahyurick, line 72: Instead of `datetime.now()` it could be `time.perf_counter()` like elsewhere in this PR?

nemo_curator/scripts/verify_classification_results.py
praateekmahajan: Let's do that as a followup PR so that we're not confused in future why a ruff PR deleted a file

nemo_curator/scripts/filter_documents.py
```diff
@@ -15,39 +15,38 @@
...
-def get_score_fields(pipeline):
+def get_score_fields(pipeline: Sequential) -> list[str]:
     score_fields = []
     for nc_module in pipeline.modules:
-        if isinstance(nc_module, nemo_curator.Score) or isinstance(
-            nc_module, nemo_curator.ScoreFilter
-        ):
-            if nc_module.score_field:
-                score_fields.append(nc_module.score_field)
+        if (isinstance(nc_module, (nemo_curator.Score, nemo_curator.ScoreFilter))) and nc_module.score_field:
```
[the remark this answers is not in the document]
praateekmahajan: Yup
```python
In [1]: isinstance(2, (int, str))
Out[1]: True

In [2]: isinstance("string", (int, str))
Out[2]: True

In [3]: isinstance(3.5, (int, str))
Out[3]: False
```

nemo_curator/scripts/find_matching_ngrams.py
praateekmahajan: Haha only in this PR I had added back `client = ??` after Ruff had removed it after chatting with Vibhu about it and seems like this was missed

(See this commit https://github.com/NVIDIA/NeMo-Curator/pull/654/commits/4a80251f874e9f2e3c2dda3c8a3eeab0f892bb2c)

The rationale for adding it back was, our customers might copy code over from examples / script and we want the variable to exist in case they want to see the dashboard url etc. It's a minor reason, but for that reason we let it be.

sarahyurick (approved):
LGTM!

PR #662, opened 2025-04-11
Ruff formatting for `nemo_curator/synthetic` and `nemo_curator/download`

sarahyurick:
Closes https://github.com/NVIDIA/NeMo-Curator/issues/646.


nemo_curator/download/arxiv.py
```diff
@@ -39,13 +40,12 @@
 
 
 class ArxivDownloader(DocumentDownloader):
-
-    def __init__(self, download_dir, verbose=False):
+    def __init__(self, download_dir: str, verbose: bool = False):
```
praateekmahajan, line 43: Good catch

nemo_curator/download/commoncrawl.py
```diff
@@ -548,8 +530,14 @@ def iterate(self, file_path):
 
 
 class CommonCrawlWARCExtractor(DocumentExtractor):
+    def __init__(
+        self,
+        algorithm: JusTextExtractor | ResiliparseExtractor | TrafilaturaExtractor | None = None,
```
praateekmahajan, line 535: All of them extend `HTMLExtractorAlgorithm` so maybe we can just use that?

nemo_curator/download/zh_stopwords.py
```diff
@@ -761,38 +761,38 @@
...
-        "＞",
-        "？",
-        "＠",
-        "［",
-        "］",
-        "｛",
-        "｜",
-        "｝",
-        "～",
+        "！",  # noqa: RUF001
```
praateekmahajan, line 764: This works but also for ease in future we can just add it to `tool.ruff.lint.per-file-ignores`

praateekmahajan (approved):
LGTM except one minor type hint, feel free to ignore it. Can be merged to ruff

PR #668, opened 2025-04-14
Ruff formatting for remaining tests

sarahyurick:
Closes https://github.com/NVIDIA/NeMo-Curator/issues/660.


tests/test_download.py
```diff
@@ -30,7 +34,7 @@
     CommonCrawlWARCExtractor,
     CommonCrawlWARCIterator,
     JusTextExtractor,
-    ResiliparseExtractor,
+    ResiliparseExtractor,  # noqa: F811
```
praateekmahajan, line 37: Why is this `ResiliparseExtractor`? I believe it can be removed (we don't need the line neither noqa)

tests/test_download.py
```diff
@@ -105,96 +109,28 @@ def html_string():
...
-        from nemo_curator.download import (
-            JusTextExtractor,
-            ResiliparseExtractor,
-            TrafilaturaExtractor,
-            download_arxiv,
-            download_common_crawl,
-            download_wikipedia,
-        )
-
+    def test_imports(self) -> None:
```
praateekmahajan, line 115: Ruff seems to think they're not needed. But that's what the test is. Let's add these back along with noqa so that they're not removed in a subsequent `ruff check --fix`

tests/test_download.py
sarahyurick: Ah, didn't realize the imports were a bit wonky here. Will fix in a few minutes, just want to wait for tests to run first.

tests/test_exact_dedup.py
```diff
@@ -24,7 +24,7 @@
 
 
 @pytest.fixture(params=[False, pytest.param(True, marks=pytest.mark.gpu)], ids=["no gpu", "gpu"])
-def exact_dedup_data(request: pytest.SubRequest) -> DocumentDataset:
+def exact_dedup_data(request) -> DocumentDataset:  # noqa: ANN001
```
praateekmahajan, line 27: In my tests I figured the datatype of this is `from pytest import FixtureRequest`

praateekmahajan (approved):
LGTM, small nit on typehint for fixture (feel free to ignore for now)

Let's merge to ruff

PR #670, opened 2025-04-15
Fix broken Aegis Permissive test

sarahyurick:


tests/test_classifiers.py
```diff
@@ -116,11 +115,7 @@ def test_aegis_classifier(gpu_client, aegis_variant: str) -> None:  # noqa: ANN0
     result_dataset = classifier(dataset=input_dataset)
     result_pred = result_dataset.df.compute()["aegis_pred"]
 
-    if "Defensive" in aegis_variant:
-        expected_pred = cudf.Series(["safe", "O3", "O4", "O13", "O3"])
-    else:
-        # Permissive
-        expected_pred = cudf.Series(["safe", "O3", "safe", "O13", "O3"])
+    expected_pred = cudf.Series(["safe", "O3", "O13", "O3"])
```
sarahyurick, line 118: My best guess is that this is caused by the latest Transformers release: https://pypi.org/project/transformers/#history.

I tried reverting all Aegis and classifier files back to how they were (pre Ruff) and that did not resolve the issue. Furthermore, the expected result for the third prompt is flaky, between `O4` and `unknown` (when it was originally labeled `safe`). To avoid flakiness I removed the third prompt altogether.

ryantwolf: Strange. If it's flakey we can definitely remove it. It would be nice to have a prompt that is different between the variants, but this will do for now.

ryantwolf (approved):

PR #681, opened 2025-04-22
Edit download examples and README

sarahyurick:


ayushdg (approved):
Generally lgtm! I'm guessing these changes are needed post #541 which made it easier to materialize the dataset. Curious if we want to do a to_json here instead of `head` but defer to @ryantwolf in case he has thoughts.

ryantwolf (approved):
Looks good!

PR #684, opened 2025-04-22
Add classifier CLI script tests

sarahyurick:
Loosely modeled after the NeMo setup:

- https://github.com/NVIDIA/NeMo/tree/1aa704e83ac3585a7e5f2f6451462b58a95b8914/tests/functional_tests
- https://github.com/NVIDIA/NeMo/blob/main/.github/workflows/cicd-main-e2e-tests.yml

### DIFF
--- a/tests/cli_scripts_tests/run_cli_scripts.sh
+++ b/tests/cli_scripts_tests/run_cli_scripts.sh
@@ -1,0 +1,43 @@
+#!/bin/bash
+set -euo pipefail
+
+LOG_FILE="/tmp/run_cli_scripts.log"
+FAILURES=()
+
+# Generate test data
+echo "🛠️ Generating test data..."
+python3 /tests/scripts_tests/generate_input_data.py
+DOMAIN_DATA_DIR=/tmp/domain_dataset
+
+# Domain classifier inference
+echo "▶️ Running domain_classifier_inference..."
+OUTPUT_DIR=/tmp/domain_classifier_inference_output
+mkdir -p $OUTPUT_DIR
+LOG_FILE="/tmp/domain_classifier_inference.log"
+domain_classifier_inference \
+  --input-data-dir "$DOMAIN_DATA_DIR" \
+  --input-file-type "jsonl" \
+  --input-text-field "text" \
+  --output-data-dir "$OUTPUT_DIR" \
+  > "$LOG_FILE" 2>&1
+if [ $? -ne 0 ]; then
+    echo "❌ domain_classifier_inference failed! Here's the log output:"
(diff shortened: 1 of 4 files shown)


tests/cli_scripts_tests/run_cli_scripts.sh
```diff
@@ -0,0 +1,74 @@
...
+    echo ""
+done
+
+if [ -z "${HF_TOKEN:-}" ]; then
+    echo ""
+    echo "⚠️ WARNING: HF_TOKEN is not set. The following CLI scripts were skipped:"
+    echo "➖ run_aegis_classifier_inference"
+    echo "➖ run_instruction_data_guard_classifier_inference"
+    echo ""
+fi
```
ryantwolf, line 60: Any reason in particular you are putting this warning here in addition to the warnings already in the corresponding scripts?

ryantwolf (commented):
Just a few comments. I think this looks good overall though, the structure makes sense.

tests/cli_scripts_tests/run_cli_scripts.sh
```diff
@@ -0,0 +1,80 @@
...
+    echo "🚨 Some CLI scripts failed:"
+    for f in "${FAILURES[@]}"; do
+        echo "❌ $f"
+    done
+    exit 1
+else
+    echo ""
+    echo "🎉 All CLI scripts passed."
+    exit 0
+fi
```
praateekmahajan, line 80: Is there a way log the summary even for succesfull runs in the end. it'll help identify the root cause in case our github action starts taking a lot of time

```
P succeeded in X seconds ✅
Q failed in Y seconds ❌
```

sarahyurick: The for loop prints the time to run for each script, regardless of whether it passed or failed. If the job takes longer than 30 minutes (the max time I have given it for now), then the script will force exit without displaying the summary text at the end.

IMO the for loop outputs plus the GitHub Actions report (which includes total runtime for the job) should be sufficient. What do you think?

praateekmahajan: The issue with the for loop is that its too much text (because of the cat `"$LOG_FILE"`) which makes it hard to read why.. But its fine if you think we don't need it, my preference would be to show success + failure both along with some message that says `$num_tests_succeeded/$num_total_tests tests succeeded in TOTAL_TIME` and then the individual breakdown

sarahyurick:
Closing for now, since I am not sure the `scripts/` directory will be relevant after the Ray refactor.

PR #691, opened 2025-04-29
ci: Add codecov rules

ko3n1g:
## Description

This PR updates `codecov.yml` so that it adds a failing status check to a PR if coverage drops below 80%. Coverage will only be uploaded if GPU tests ran. 

## Usage

```python
```
## Checklist


ryantwolf:
There's a weird edgecase I want to understand with this. Consider this scenario.

1. User opens a PR that only needs CPU unit tests, but they don't add more unit tests for their code.
2. CPU Unit tests pass.
3. Code coverage remains at 80% from previous GPU Unit test run.
4. PR gets approved and merge.
5. Another user opens a PR that needs GPU unit tests and they do add all needed unit tests.
6. The code coverage gets run with their PR, but their PR gets blocked due to low code coverage from the previous PR.

Is this scenario possible? What can we do to prevent it?

.github/workflows/test.yml
```diff
@@ -54,6 +55,7 @@ jobs:
 
       - name: Upload coverage reports to Codecov
         uses: codecov/codecov-action@v5
+        if: github.event.label.name == 'gpuci'
```
sarahyurick, line 58: What about something like:
```
contains(github.event.pull_request.labels.*.name, 'gpuci') || (github.event.label.name == 'gpuci') || (github.ref == 'refs/heads/main')
```
?

ko3n1g:
> There's a weird edgecase I want to understand with this. Consider this scenario.
> 
> 1. User opens a PR that only needs CPU unit tests, but they don't add more unit tests for their code.
> 2. CPU Unit tests pass.
> 3. Code coverage remains at 80% from previous GPU Unit test run.
> 4. PR gets approved and merge.
> 5. Another user opens a PR that needs GPU unit tests and they do add all needed unit tests.
> 6. The code coverage gets run with their PR, but their PR gets blocked due to low code coverage from the previous PR.
> 
> Is this scenario possible? What can we do to prevent it?

I guess this is where codecov's separation between project and patch coverage could be interesting. From this perspective, we probably would want PRs to be checked against patch coverage (each measured PR's diff must be covered at least by 80%). But if a PR doesn't go through codecov, we'd still want to have 80% on main. In that case we'd want to get a notification so that we can submit a fixing PR that lifts project coverage up again.

ko3n1g:
@ryantwolf: So the discussed issue indeed works: Codecov can be configured to react to patch vs project changes. Given that the presubmit-CI might be based on partial results, I think this makes sense. 

However, this PR would not yet solve the alerting if the project coverage drops below 80.

If you generally agree with this path, I can take this as the next step of this PR.

chtruong814:
@ko3n1g thanks for this.  I was literally about to jump on this.

ryantwolf:
This makes sense and I am supportive of this path. Could you update the CONTRIBUTING.md file explaining this new requirement and how to satisfy it so people aren't confused?

.github/workflows/test.yml
```diff
@@ -54,6 +55,7 @@ jobs:
 
       - name: Upload coverage reports to Codecov
         uses: codecov/codecov-action@v5
+        if: contains(github.event.pull_request.labels.*.name, 'gpuci') || (github.event.label.name == 'gpuci') || (github.ref == 'refs/heads/main')
```
sarahyurick, line 58: Does this have to be updated now that we use `/ok to test`?

ko3n1g: Good catch, removed it

ryantwolf (approved):
Thank you!

PR #694, opened 2025-05-04
ci: Migrate to using Nvidia Github Runners

chtruong814:
## Description
* Migrate to using Nvidia Github Runners
* Use cpu runner for building the docker image
* The Nvidia Github Runner will log into Azure to download and push images as needed
* Uses an environment for the build and test steps so that this can be configured on the Azure side to allow it to assume the necessary role

## Usage

```python
```
## Checklist


.github/workflows/gpuci.yml
```diff
@@ -4,80 +4,82 @@ on:
...
-    #    then subsequent pushes to the PR will not autorun gpuCI
-    #    unless the "gpuCI" label is removed and re-added again.
-    # 3. PR is merged to main.
-    if: >-
-      (
-        contains(github.event.pull_request.labels.*.name, 'gpuci') &&
-        contains(
-          '["ayushdg", "ko3n1g", "praateekmahajan", "ryantwolf", "sarahyurick", "VibhuJawa"]',
-          github.event.pull_request.user.login
-        )
```
sarahyurick, line 37: Can we re-add this? For known maintainers (NeMo Curator team members), I would like not to need to re-add the gpuCI label every time. This check was added to allow that.

chtruong814: So with this migration to nv-gh-runners, the mechanics has changed a little.  We aren't relying on the gpuci label anymore to run this, you'd add a comment like `/ok to test` and it'd trigger the job.  That will be requried at least for any external user PRs that happen.  One of the vetters from this list will need to review and approve CI for those PRs:

https://github.com/NVIDIA/NeMo-Curator/blob/main/.github/copy-pr-bot.yaml

We can have it so that anyone in the approved list will just automatically have CI run on every change event in their PR.  Maybe we limit it to non-draft?  Just looking to be more resposible for using the resource.  What do you think?  

@ko3n1g any thoughts on your side as well?

ko3n1g: I think we could have both? After `/ok to test`, we refresh CI with each synchronize event. But expensive tests will only run if commit-author is core contributor. In other cases it's either a no-op or some cheap steps like linting/CPU tests

sarahyurick: Ah I see, thanks! I like this idea and limited to non-drafts seems reasonable to me. I agree that linting is cheap enough to run on all commits. Up to you about CPU tests automatically running or not.

.github/workflows/gpuci.yml
```diff
@@ -48,36 +27,43 @@ jobs:
...
 
   # Then, we run our PyTests in the container we just built
   run-gpu-tests:
-    needs: build-container
-    # This is the tag on our Azure runner found in Actions -> Runners -> Self-hosted runners
-    # It has 2 A100 GPUs
-    runs-on: self-hosted-azure
+    needs: [build-container]
+    runs-on: linux-amd64-gpu-rtxa6000-latest-1
+    environment: nemo-ci
```
ko3n1g, line 42: If I understood the conversation with @sarahyurick correctly, we want to guard this job behind the label gpuci.

The workflow should detach the label after one full pass _and_ the github.actor is not a core-member.

```yaml
     # pseudo-code:
     - name: Remove label if not cancelled
        if: success or failure and !contains([...], github.actor)
        env:
          GH_TOKEN: ${{ github.token }}
          PR_NUMBER: ${{ github.event.number }}
        run: gh pr edit "$PR_NUMBER" --remove-label "gpuci"
```

ko3n1g (changes requested):
Overall great change, I like the cleanup of the dockerfile! I think only proper guarding of gpu-jobs is missing, but then we're ready to go 🚀

.github/workflows/gpuci.yml
chtruong814: I thought @sarahyurick 's concern was mainly around needing to constantly re-add the label?  So if our copy-pr-bot config allows Sarah and other Nvidia devs to have this always run on their commits for non-draft PRs, I don't think we need to worry about the label?  

Otherwise, I think we just fully embrace the `/ok to test` without the label check.  

Sarah, what do you think?

sarahyurick: I think if we are removing the `gpuci` label altogether in favor of `/ok to test`, then I am okay with automatically running the GPU workflow for users in the copy-pr-bot per every commit in a non-draft PR, and for all other users requiring `/ok to test` for every commit.

Does this mean `\ok to test` will always trigger both CPU and GPU workflows?

chtruong814: The CPU workflows will always run.  Only the GPU workflow is gated.  

I updated the PR so that it only runs the GPU tests on certain file changes.  Please take another look when you get a chance.

ko3n1g (approved):

thomasdhc (approved):
Thanks for the update

.github/workflows/gpuci.yml
```diff
@@ -4,41 +4,61 @@ on:
...
+      - name: Get changed files
+        id: changed-files
+        uses: step-security/changed-files@v45.0.1
+        with:
+          files: |
+            nemo_curator/**
+            config/**
+            .github/**
+            pyproject.toml
+            Dockerfile
```
sarahyurick, line 54: I was wondering if `tests/**` could be added here too?

sarahyurick (commented):
LGTM, added a minor request.

PR #695, opened 2025-05-05
Curriculum Dataprep tutorial

sarahyurick:


tutorials/curriculum-dataprep/main.py
```diff
@@ -0,0 +1,483 @@
...
+        default=False,
+        help="""
+        If True, the datasets will be interleaved globally, i.e., row by row.
+        If False, the datasets will be interleaved approximately, i.e., partition by partition.
+        Default is False.
+        """,
+    )
+
+    parser.add_argument(
+        "--input",
```
Maghoumi, line 449: So it appears that we expect users to provide an input directory, under which we'll search for all JSONL files.

However, the chances are most folks don't want to process the whole dataset. Is it possible to add another input arg (called something like `filename_filter`) where users pass `["math_v1.1, "code"]`. Then when we list all the files, we only keep the ones that match the user's filter. If no filter provided, then all files are used for curation.

Maghoumi (commented):
Some first impression comments after going through the code in the first pass

tutorials/curriculum-dataprep/main.py
```diff
@@ -0,0 +1,483 @@
...
+        help="Path to the input JSONL file or directory containing JSONL files.",
+    )
+    parser.add_argument(
+        "--tokenizer",
+        type=str,
+        default="meta-llama/Llama-3.1-8B-Instruct",
+        help="Hugging Face tokenizer",
+    )
+    parser.add_argument(
+        "--model_path",
```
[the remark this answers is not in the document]
sarahyurick: I think I would prefer not to download the model via Python, which could introduce some overhead about where the user expects the model to be downloaded to (or unnecessarily download the model when it already exists somewhere), and instead assume the user already has it downloaded somewhere.

I already include the link in the README, although I can add a snippet about the `wget` command to download the model. What do you think?

tutorials/curriculum-dataprep/main.py
sarahyurick: Added.

tutorials/curriculum-dataprep/main.py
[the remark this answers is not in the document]
Maghoumi: Adding `wget` instructions sounds reasonable to me.

tutorials/curriculum-dataprep/main.py
```diff
@@ -455,8 +468,10 @@ def attach_args() -> argparse.ArgumentParser:
         """,
     )
 
+    # TODO: --generate-statistics
+
     parser.add_argument(
-        "--input",
+        "--input-dir",
         type=str,
         help="Path to the input JSONL file or directory containing JSONL files.",
```
Maghoumi, line 498: The help string might need updating

tutorials/curriculum-dataprep/README.md
```diff
@@ -2,38 +2,78 @@
...
+
 - Only take samples used for Nemotron Nano training
 - Remove empty and malformed samples
 - Remove non-English samples
 - Remove samples longer than 8k tokens (with chat template applied via the tokenizer). For faster performance, the user can specify the boolean flag `--skip-tokenize`, which skips the tokenization step and returns the text length (number of characters) instead.
+- Remove any columns specified by the `--remove-columns` parameter
 
 After filtering, it sorts all samples by completion length, then "interleaves" thinking ON/thinking OFF for curriculum learning. For large datasets, we recommend setting `--device "gpu"` and using an approximate interleaving approach which interleaves the data by Dask partition rather than row by row. If you prefer to interleave globally (per row), then you can specify the boolean flag `--global-interleave` from the script command.
+
+## Debugging Out of Memory Errors
```
Maghoumi, line 77: Thanks for making this section and adding more details. I think it provides a great overview of the issues we frequently faced and how to go about solving them.

tutorials/curriculum-dataprep/README.md
```diff
@@ -2,38 +2,78 @@
...
+## Usage
+
+This tutorial can be run with:
 
 ```bash
 python main.py \
-    --input "./data" \
+    --input-dir "/path/to/input/data" \
     --filename-filter "code" "math" \
+    --remove-columns "used_in_training" "license" \
```
Maghoumi, line 54: Please also add `"version", "category", "generator"` by default.

tutorials/curriculum-dataprep/main.py
```diff
@@ -0,0 +1,653 @@
...
+        )
+
+    @batched
+    def keep_document(self, scores: pd.Series) -> pd.Series:
+        return scores
+
+
+# TODO: Add this to NeMo Curator modules
+# Tokenize system_prompt, input, and output and filter out samples with too many tokens
+class TokenCountFilter(DocumentFilter):
```
sarahyurick, line 209: Hi @Maghoumi can you please verify the correctness of this implementation?

sarahyurick (commented):
Hi @Maghoumi @ruchaa-apte this should be ready for another round of reviews. Thanks!

tutorials/curriculum-dataprep/main.py
```diff
@@ -0,0 +1,653 @@
...
+        text = self.tokenizer.apply_chat_template(
+            [
+                {"role": "system", "content": system},
+                *inpt,
+                {"role": "assistant", "content": outpt},
+            ],
+            tokenize=False,
+            add_generation_prompt=False,
+        )
+        return str(text).replace("\n", " ").strip()
```
Maghoumi, line 235: We shouldn't do this. It should just return the resulting string without modification.

It was fine to do it above for the language filter because fast text would complain if we don't do it.

tutorials/curriculum-dataprep/main.py
Maghoumi: I think it looks good (left a comment about a minor issue).

If you wanna get super fancy with it, and make the code more compact you could define a super class `class AbstractTokenFilter(DocumentFilter)` which has `def apply_chat_template()`. It could also have the logic for `load_object_on_worker` to load the tokenizer as needed.

Then english filter, token count filter, completion filter can all inherit from that class and just call `apply_chat_template` and the `load` functions as needed. 

We don't have to do it now though, just a suggestion to avoid code duplication.

tutorials/curriculum-dataprep/main.py
```diff
@@ -0,0 +1,653 @@
...
+    prompt = tokenizer.apply_chat_template(
+        [
+            {"role": "system", "content": system_prompt},
+            *inpt,
+        ],
+        tokenize=False,
+        add_generation_prompt=True,
+    )
+
+    completion = prompt_and_completion[len(prompt) :]
```
Maghoumi, line 335: Please add a short comment explaining that we are extracting the completion part with string manipulation.

tutorials/curriculum-dataprep/main.py
```diff
@@ -0,0 +1,653 @@
...
+        add_generation_prompt=False,
+    )
+
+    prompt = tokenizer.apply_chat_template(
+        [
+            {"role": "system", "content": system_prompt},
+            *inpt,
+        ],
+        tokenize=False,
+        add_generation_prompt=True,
```
Maghoumi, line 332: Add a comment about `add_generation_prompt=True`, something like "we expect the model to start predicting tokens after it sees the start of the `assistant` response turn".

tutorials/curriculum-dataprep/main.py
```diff
@@ -0,0 +1,653 @@
...
+                score_type=int,
+            ),
+        ]
+    )
+    dataset_df = filter_steps(dataset).df
+
+    if args.remove_columns:
+        dataset_df = dataset_df.drop(columns=args.remove_columns, axis=1)
+
+    # TODO: A lot of re-used logic around apply_chat_template...
```
Maghoumi, line 499: Ah, I see you had already thought about this :)

tutorials/curriculum-dataprep/main.py
```diff
@@ -0,0 +1,653 @@
...
+    parser.add_argument(
+        "--lang-id-model-path",
+        type=str,
+        help="Path to the FastText model",
+        required=True,
+    )
+    parser.add_argument(
+        "--max-token-count",
+        type=int,
+        default=16_000,
```
Maghoumi, line 620: [nit] 16384

Maghoumi (commented):
I reviewed, I think it looks great now. Will now start testing and let you know how it goes.

Thanks again for putting all this together.

tutorials/curriculum-dataprep/main.py
sarahyurick: I think that is a great idea. For now I am going to keep this PR as is unless there are any other suggestions. As a follow-up to this PR I have created Issue https://github.com/NVIDIA/NeMo-Curator/issues/706 for adding some of the classes from this tutorial to NeMo Curator's modules. I think your suggestion from above would be better suited for when I am working on that.

For now, let's prioritize getting this tutorial merged. Thanks for all the feedback so far!

ruchaa-apte: LGTM @sarahyurick - Thank you so much :D

tutorials/curriculum-dataprep/README.md
```diff
@@ -1,6 +1,21 @@
-# Curriculum Data Preparation Tutorial
+# Curating the Llama Nemotron Reasoning Dataset with NVIDIA NeMo Curator
```
Maghoumi, line 1: I like this new title.

Could you also rename the tutorial's folder from `curriculum-dataprep` to something that resonates with this title?

tutorials/curriculum-dataprep/README.md
```diff
@@ -1,6 +1,21 @@
...
+The [Llama Nemotron Post-Training Dataset](https://huggingface.co/datasets/nvidia/Llama-Nemotron-Post-Training-Dataset) is a curated collection of approximately 30 million high-quality synthetic samples, designed to enhance the reasoning capabilities of large language models.
+Organized into distinct subsets for supervised fine-tuning (SFT) or reinforcement learning (RL), it encompasses samples from various problem domains.
+All samples are in JSON lines (JSONL) format and contain metadata such as license type, source model, as well as the [Llama Nemotron](https://www.nvidia.com/en-us/ai-data-science/foundation-models/llama-nemotron/) model(s) trained with that sample.
+
+Each sample consists of a prompt, along with an expected reponse with detailed chain-of-thought (CoT) reasoning traces followed by responses (i.e., "reasoning on"), as well as samples with direct responses (i.e., "reasoning off").
+More concretely, the relevant attributes for this tutorial are:
+
```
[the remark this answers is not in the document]
Maghoumi: Feel free to put this sample wherever it makes the most sense

Maghoumi (approved):
Thanks for making all the changes. Looks great now. Approved.

Maghoumi:
@sarahyurick Would you kindly rename the PR's title from "curriculum..." to reference llama nemotron explicitly?
This is along the lines of what we did yesterday :)

praateekmahajan (approved):

ruchaa-apte (approved):

PR #698, opened 2025-05-07
Simplifying clustering logic

praateekmahajan:
# TLDR
Existing implementation is both convoluted and can be very slow [unless a user explicitly does `to_backend("cudf")` (read point 1 below).]
A simpler implementation performs similar even when user doesn't do the `to_backend(..)`. 

The code has been testing on 37mn embedding dataset (80gb worth of data) on 7 GPUs and much smaller dataset of 2gb on both Rapids 25.02 and Rapids 25.06.

Here is the GPU utilization for 90 gb of embeddings where we see that during KMeans.fit it spikes to 2x
image

Here are the results which show that our existing implementation times out when we don't do `to_backend('cudf')` 

image

# Longer Version
Our existing codes flow was as follows
```python
# Step 1) read df
embeddings_df = DocumentDataset.read_parquet(...., backend="cudf", blocksize or fpp).df
# Step 2) persist to pandas
embeddings_df = embeddings_df.to_backend("pandas").persist()
# Step 3) Assert for len > n_clusters
if len(embeddings_df) > kmeans_n_clusters: raise error
# Step 4) persist to cudf
embeddings_df = embeddings_df.to_backend("cudf")
# Step 5) get embeddings array
embeddings_arr = embeddings_df.map_partitions(get_array)
# Step 6) compute_chunk_sizes
embeddings_arr.compute_chunk_sizes()
# Step 7) train kmeans
KMeans().fit(embeddings_arr)
```

The few takeaways in above code are
1. If between step 1 and step 2 we didn't do `embeddings_df.to_backend("cudf")` the performance at scale was very slow (no idea why this would be the case) **this is the main reason for this PR**
2. Step 3, while a good user experience is super expensive (since the user might've read using FPP in which case all data is read). We expect an error to anyway be thrown at `KMeans().fit`
3. Step 5 started failing atleast from `dask 25.02` due to `Cannot fuse tasks with multiple outputs` (fixed in 25.06 atleast, not sure about 25.04)
2. Step 6 wasn't needed since 25.02 atleast
4. Step 7 persists `embeddings_arr` anyway within the code

The final version of code now looks like this
```python 
# Step 1) read df (this calls `dask_cudf.read_parquet(...)`)
embeddings_df = DocumentDataset.read_parquet(...., backend="cudf", blocksize or fpp).df 
# Step 2) get embeddings array
embeddings_arr = embeddings_df.map_partitions(get_array)
# Step 3) Persist the array
embeddings_arr.persist()
# Step 4) `compute_chunk_sizes` so that we can calculate length
embeddings_arr.compute_chunk_sizes()

# Step 5) Assert for len > n_clusters
if len(embeddings_df) > kmeans_n_clusters: raise error
# Step 6) 
KMeans().fit(embeddings_arr)
```

If we're okay to get rid of the assert `len > n_clusters` the code could be even simpler (however this is prone to runtime errors that the user might not be able to understand)

```python 
# Step 1) read df (this calls `dask_cudf.read_parquet(...)`)
embeddings_df = DocumentDataset.read_parquet(...., backend="cudf", blocksize or fpp).df 
# Step 2) get embeddings array
embeddings_arr = embeddings_df.map_partitions(get_array)
# Step 3) 
KMeans().fit(embeddings_arr)
```

## Usage

```python
```
## Checklist

### DIFF
--- a/nemo_curator/modules/semantic_dedup/clusteringmodel.py
+++ b/nemo_curator/modules/semantic_dedup/clusteringmodel.py
@@ -116,32 +116,26 @@ class ClusteringModel:
             raise ValueError(msg)
 
         with performance_report_if_with_ts_suffix(self.profile_dir, "clustering-model"):
+            t0 = time.perf_counter()
             if not self.keep_all_columns:
                 embeddings_df = embeddings_df[[self.id_col, self.embedding_column]]
 
             if self.clustering_input_partition_size is not None:
+                print(
+                    f"Repartitioning embeddings with partition size {self.clustering_input_partition_size}.",
+                    flush=True,
+                )
                 embeddings_df = embeddings_df.repartition(partition_size=self.clustering_input_partition_size)
 
-            try:
-                embeddings_df = embeddings_df.to_backend("pandas").persist()
-                embeddings_length = embeddings_df.shape[0].compute()
-
-                if embeddings_length < self.n_clusters:
-                    msg = (
-                        "Number of clusters is greater than the number of documents in your dataset: "
-                        f"dataset length is {embeddings_length} while n_clusters is set to {self.n_clusters}. "
-                        f"Please reduce n_clusters to be less than or equal to {embeddings_length}."
-                    )
-                    raise ValueError(msg)
-            except IndexError as e:
-                msg = (
-                    f'Original error message: "{e}". '
-                    "This could be due to empty partitions in your DocumentDataset. "
-                    "Please check your dataset for empty partitions and remove them if necessary."
-                )
-                raise IndexError(msg) from e
-
-            embeddings_df = embeddings_df.to_backend("cudf")
+            print(
+                f"Calling kmeans on embeddings with npartitions={embeddings_df.npartitions} and optimized partitions {embeddings_df.optimize().npartitions}.",
+                flush=True,
+            )
+            # We need this optimize call since the map_partitions(get_array_from_df) seems to be returning different
+            # number of partitions
+            embeddings_df = embeddings_df.optimize()
+            print(f"After optimization, embeddings with npartitions={embeddings_df.npartitions}.", flush=True)
+            t1 = time.perf_counter()
             # Normalize embeddings before clustering
             embeddings_df = embeddings_df.map_partitions(
                 normalize_embeddings_col_in_df,
@@ -151,8 +145,11 @@ class ClusteringModel:
             cupy_normalized_darr = embeddings_df.map_partitions(
                 get_array_from_df, self.embedding_column, meta=cp.ndarray([1, 1])
             )
-            cupy_normalized_darr.compute_chunk_sizes()
-
+            t2 = time.perf_counter()
+            print(
+                f"Took {(t2 - t1):.2f} seconds to normalize + get_array_from_df (npartitions={cupy_normalized_darr.npartitions}) + compute_chunk_sizes.",
+                flush=True,
+            )
             # Perform KMeans clustering (KMeans.fit)
             t0 = time.time()
             kmeans = KMeans(
@@ -163,6 +160,8 @@ class ClusteringModel:
             )
             self.logger.info("KMeans starting fit")
             kmeans.fit(cupy_normalized_darr)
+            t3 = time.perf_counter()
+            print(f"Took {(t3 - t2):.2f} seconds to call KMeans Fit.", flush=True)
             self.logger.info("KMeans fit complete")
             self.logger.info(f"Time taken for KMeans fit: {time.time() - t0}")
             # Compute nearest centroids using kmeans.predict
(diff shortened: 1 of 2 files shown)


nemo_curator/modules/semantic_dedup/clusteringmodel.py
```diff
@@ -122,36 +122,39 @@ def __call__(self, embeddings_dataset: DocumentDataset) -> DocumentDataset:  # n
...
             )
+            # It's okay to add persist here since KMeansMG will persist the array too
             cupy_normalized_darr = embeddings_df.map_partitions(
                 get_array_from_df, self.embedding_column, meta=cp.ndarray([1, 1])
             )
-            cupy_normalized_darr.compute_chunk_sizes()
+            cupy_normalized_darr.persist()
+
+            try:
+                cupy_normalized_darr.compute_chunk_sizes()
```
praateekmahajan, line 141: I'm not using a constant from `_compat.py` to check for versions because I'm not sure which package causes this error `dask / cudf / cupy / dask_cudf / dask_cuda`. What I do know is in nightly it works just fine, while in stable we need the task fusion logic

nemo_curator/modules/semantic_dedup/clusteringmodel.py
```diff
@@ -122,36 +122,39 @@ def __call__(self, embeddings_dataset: DocumentDataset) -> DocumentDataset:  # n
...
                 normalize_embeddings_col_in_df,
                 embedding_col=self.embedding_column,
                 meta=embeddings_df._meta.copy(),  # noqa: SLF001
             )
+            # It's okay to add persist here since KMeansMG will persist the array too
             cupy_normalized_darr = embeddings_df.map_partitions(
                 get_array_from_df, self.embedding_column, meta=cp.ndarray([1, 1])
             )
-            cupy_normalized_darr.compute_chunk_sizes()
+            cupy_normalized_darr.persist()
```
VibhuJawa, line 138:
```suggestion
            cupy_normalized_darr = cupy_normalized_darr.persist()
```

nemo_curator/modules/semantic_dedup/clusteringmodel.py
```diff
@@ -122,36 +122,39 @@ def __call__(self, embeddings_dataset: DocumentDataset) -> DocumentDataset:  # n
...
+                    cupy_normalized_darr.compute_chunk_sizes()
+
+            # TODO: Remove once https://github.com/rapidsai/cuml/issues/6643 is fixed
+            if len(cupy_normalized_darr) < self.n_clusters:
+                msg = (
+                    f"Number of clusters is greater than the number of documents in your dataset: "
+                    f"dataset length is {len(cupy_normalized_darr)} while n_clusters is set to {self.n_clusters}. "
+                    f"Please reduce n_clusters to be less than or equal to {len(cupy_normalized_darr)}."
+                )
+                raise ValueError(msg)
```
VibhuJawa, line 157: Can we do something like below to log this error/warnings 

```python3

except Exception as exc:
   warnings.warn(
        "compute_chunk_sizes failed (%s); retrying with task‑fusion disabled", exc
    )

    # ❷ Second attempt: disable task fusion (work‑around for RAPIDS 25.02/25.04)
    try:
        with dask.config.set({"optimization.fuse.active": False}):
            cupy_normalized_darr.compute_chunk_sizes()
    except Exception as fallback_exc:
        # ❸ Give up early with a clear diagnostic
        raise RuntimeError(
            "Unable to compute chunk sizes for the input array – even with task‑fusion "
            "disabled. Verify that your RAPIDS stack is ≥25.06 or apply the fix linked "
            "in https://github.com/rapidsai/cuml/issues/6643."
        ) from fallback_exc 
 ```

**Review by VibhuJawa — DISMISSED** [review]
Some nits around persist and data handling . I think we will end up with some memory overhead but i think that is probably fine given speedup you are seeing

File: nemo_curator/modules/semantic_dedup/clusteringmodel.py [hunk 4]
```diff
@@ -122,36 +122,39 @@ def __call__(self, embeddings_dataset: DocumentDataset) -> DocumentDataset:  # n
...
-                    f'Original error message: "{e}". '
-                    "This could be due to empty partitions in your DocumentDataset. "
-                    "Please check your dataset for empty partitions and remove them if necessary."
-                )
-                raise IndexError(msg) from e
-
-            embeddings_df = embeddings_df.to_backend("cudf")
+            # We need this optimize call since the map_partitions(get_array_from_df) seems to be returning different
+            # number of partitions
+            embeddings_df = embeddings_df.optimize()
```
sarahyurick, line 127: I was wondering if you could explain more why this is needed?

sarahyurick (commented):
PR looks good to me, just had a question.

nemo_curator/modules/semantic_dedup/clusteringmodel.py
praateekmahajan: Added a comment re the same, if its not clear lmk

praateekmahajan:
@VibhuJawa as discussed offline, the `arr.persist()` doesn't result in anymore memory overhead than what we had. The only issue was I/O re-read if we don't persist embeddings df. Now I have added that as well in addition, which makes arr.persist() in that workflow a no-op.

PR #728, opened 2025-06-16
c: Update to use nemo runners

chtruong814:
## Description
Update to use nemo runners and remove unnecessary test code.

## Usage

```python
```
## Checklist

### DIFF
--- a/nemo_curator/log.py
+++ b/nemo_curator/log.py
@@ -26,6 +26,7 @@ def create_logger(
     log_level: int = logging.INFO,
     stdout: bool = False,
 ) -> logging.Logger:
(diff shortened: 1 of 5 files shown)


nemo_curator/log.py
```diff
@@ -26,6 +26,7 @@ def create_logger(
     log_level: int = logging.INFO,
     stdout: bool = False,
 ) -> logging.Logger:
+    # Minor change
```
ayushdg, line 29: Debug/test comment left behind?

chtruong814: Oof yeah.  Let me get rid of that.

PR #730, opened 2025-06-16
[Ray] Add Jsonl / Parquet Writer Stage

praateekmahajan:
## Usage

```python
```
## Checklist


ray-curator/ray_curator/stages/io/writer/dataframe.py
```diff
@@ -0,0 +1,113 @@
...
+
+        Args:
+            task (DocumentBatch): DocumentBatch containing data to write
+
+        Returns:
+            FileGroupTask: Task containing paths to written files
+        """
+        # Get source files from metadata for deterministic naming
+        if source_files := task._metadata.get("source_files"):  # noqa: SLF001
+            filename = writer_utils.get_deterministic_hash(source_files, task.task_id)
```
praateekmahajan: I am using Task._metadata to decide the filename, from the ReadStage (which will create a  source_files key in metadata).
However this is error prone as
User might read + intermediate transformation + write, where intermediate transformation might have
1. done a fan-out, in which case multiple tasks will have the same source_files  and hence all of them will produce the same hash(source_files) which will lead to them writing to the same file
2. not carried over metadata from ReadStage (which means write doesn't have access to source_files)
3. overwritten the key source_files (tho solvable, if we use a dict that doesn't allow keys to be overwritten)
For WriteStage, hence, I believe we shouldn't allow this checkpointing-ish behavior. Or even if we do, hide it behind a flag, or something like key_for_filename

cc @abhinavg4 @ayushdg @sarahyurick

ray-curator/ray_curator/stages/io/writer/jsonl.py
```diff
@@ -0,0 +1,12 @@
+from .dataframe import DataFrameWriter
+
+
+class JsonlWriter(DataFrameWriter):
```
ayushdg: What are your thoughts on following this pattern and having all the writing logic live in the parent dataframeWriter class vs having a base class for all the common stuff and handling concrete arguments in the `JsonlWriter` classes.

I see a lot of common code supporting the argument for a base class. However I can see a lot of writer specific options that can diverge or vary a ton between formats. With the `kwargs` approach we allow passing those in but give the user lesser visibility on what those options might be vs concretely supporting format specific args?

ray-curator/ray_curator/stages/io/writer/utils.py
```diff
@@ -0,0 +1,7 @@
+import hashlib
+
+
+def get_deterministic_hash(inputs: list[str], seed: str = "") -> str:
```
ayushdg, line 4: Nit: I typically expect `seed` to be an integer. I'm not too familiar with hashing so maybe string seeds are common enough as well?

ray-curator/ray_curator/stages/io/writer/dataframe.py
```diff
@@ -0,0 +1,113 @@
...
+        return FileGroupTask(
+            task_id=f"{filename}_written",  # TODO : Discuss how task_id should be handled
+            dataset_name=task.dataset_name,
+            data=[file_path],
+            _metadata={
+                **task._metadata,  # noqa: SLF001
+                "output_dir": self.output_dir,
+                "format": self.format,
+            },
+            _stage_perf=task._stage_perf,  # noqa: SLF001
```
ayushdg: Is it worth adding `SLF001` to the ignore list if this is going to be a common pattern throughout tasks?

ayushdg (commented):
Some minor nits and question around having a branched `DataFrameWriter` vs having the core implementation in `JsonlWriter` `ParquetWriter` etc.

ray-curator/ray_curator/stages/io/writer/utils.py
praateekmahajan: It essentially adds seed as a suffix to the string to get a hash.

ray-curator/ray_curator/stages/io/writer/jsonl.py
praateekmahajan: I agree, since we've gotten burnt with our current dasks one single monolithy function of `read_data`, I'm happy to move most of the implementation out to JsonlWriter and ParquetWriter (or rather remove any base class for now, and if in future we see another writer then add that)

ray-curator/pyproject.toml
```diff
@@ -74,5 +75,9 @@ fixable = ["ALL"]
 ]
 "tests/**/*.py" = [
     "S101", # asserts allowed in tests...
+    "ANN201", # Allow methods to not return something
     "PLR2004", # Magic value used in comparison, ...
 ]
+
+[tool.pytest.ini_options]
+testpaths = ["tests"]
```
praateekmahajan, line 83: This allows us to not use conftest from `NeMo-Curator/tests`

ray-curator/ray_curator/stages/io/writer/dataframe.py
```diff
@@ -0,0 +1,90 @@
...
+        else:
+            logger.warning("The task does not have source_files in metadata, using UUID for base filename")
+            filename = uuid.uuid4().hex
+
+        # Create output directory
+        self.fs.makedirs(self.output_dir, exist_ok=True)
+
+        # Generate filename with appropriate extension
+        file_extension = self.get_file_extension()
+        file_path = os.path.join(self.output_dir, f"{filename}.{file_extension}")
```
ayushdg: Nit: Should we make this more generic with `fs.sep.join` in favor of `os.path.join`. 
The only catch here is I'm not sure if fs.sep.join works with `Path` object or only with Strings.

ray-curator/ray_curator/stages/io/writer/dataframe.py
```diff
@@ -0,0 +1,90 @@
...
+            filename = writer_utils.get_deterministic_hash(source_files, task.task_id)
+        else:
+            logger.warning("The task does not have source_files in metadata, using UUID for base filename")
+            filename = uuid.uuid4().hex
+
+        # Create output directory
+        self.fs.makedirs(self.output_dir, exist_ok=True)
+
+        # Generate filename with appropriate extension
+        file_extension = self.get_file_extension()
```
ayushdg, line 68: We're not doing much with file extension logic apart from appending which is fine. But wanted to note that the extension handling can get complex pretty quickly so we should avoid parsing/handling them unless necessary. 
eg: `file.jsonl.gz` type extensions etc.

ray-curator/ray_curator/stages/io/writer/parquet.py
```diff
@@ -0,0 +1,34 @@
...
+    file_extension: str = "parquet"
+
+    @property
+    def name(self) -> str:
+        return "parquet_writer"
+
+    def write_data(self, task: DocumentBatch, file_path: str) -> None:
+        """Write data to Parquet file using pandas DataFrame.to_parquet."""
+        df = task.to_pandas()  # Convert to pandas DataFrame if needed
+
```
ayushdg, line 24: Do you prefer this over the direct pa.Table -> file codepath? 
Maybe doesn't need to be done now but benchmarked and addressed in a followup?

ray-curator/tests/stages/io/writer/conftest.py
```diff
@@ -0,0 +1,71 @@
...
+        data=df,
+        _metadata={"dummy_key": "dummy_value"},
+        _stage_perf=[
+            StagePerfStats(
+                stage_name="test_stage",
+                process_time=0.1,  # 100ms = 0.1s
+                actor_idle_time=0.0,
+                input_data_size_mb=0.001,  # Small test data
+                num_items_processed=3,  # 3 rows in the DataFrame
+            )
```
ayushdg, line 33: Do we need to populate this with dummy stats?

ray-curator/ray_curator/stages/io/writer/parquet.py
praateekmahajan: Had to chose one of them, so chose pandas, I don't mind chosing either

ray-curator/tests/stages/io/writer/conftest.py
praateekmahajan: WDYM? or just keep empty?

ray-curator/ray_curator/stages/io/writer/dataframe.py
praateekmahajan: Yup, i ended up adding `file_extension` as an init param because of that reason, in case we want to end some "compression" `jsonl_kwargs` and then write as `.jsonl.gz`

ray-curator/ray_curator/stages/io/writer/parquet.py
ayushdg: I think pandas is fine to begin with.

ray-curator/tests/stages/io/writer/conftest.py
ayushdg: Does every fixture/test stage need to add dummy stats like the example here? Mainly thinking if we can avoid needing to add this for most tests/fixtures manually

praateekmahajan: I had a similar thought not for perfstats but `pandas_document_batch / pyarrow_document_batch` as I believe that's a relatively common fixture that might live across multiple tests, when we get there we could abstract this out too?

ayushdg (approved):

PR #579, opened 2025-03-03
Re-add Common Crawl tests with skips

sarahyurick:
Follow up for https://github.com/NVIDIA/NeMo-Curator/pull/577.

Additional information:

- https://github.com/NVIDIA/NeMo-Curator/pull/541
- https://github.com/NVIDIA/NeMo-Curator/issues/236

Issue #236, opened 2024-09-09
Re-add `test_uneven_common_crawl_range` PyTest


sarahyurick:
PR https://github.com/NVIDIA/NeMo-Curator/pull/235 skips `test_uneven_common_crawl_range` because of how flaky it is. In the future, we may want to debug and re-add it.

```
def test_uneven_common_crawl_range(self):
    start_snapshot = "2021-03"
    end_snapshot = "2021-11"
    urls = get_common_crawl_urls(start_snapshot, end_snapshot)

    assert (
        urls[0]
        == "https://data.commoncrawl.org/crawl-data/CC-MAIN-2021-10/segments/1614178347293.1/warc/CC-MAIN-20210224165708-20210224195708-00000.warc.gz"
    )
    assert (
        urls[-1]
        == "https://data.commoncrawl.org/crawl-data/CC-MAIN-2021-04/segments/1610704847953.98/warc/CC-MAIN-20210128134124-20210128164124-00799.warc.gz"
    )
    assert len(urls) == 143840
```

sarahyurick:
More flaky tests from `test_download` have been skipped in https://github.com/NVIDIA/NeMo-Curator/pull/454.

sarahyurick:
See https://github.com/NVIDIA/NeMo-Curator/pull/541.

PR #738, opened 2025-06-17
[Ray] Add Common Crawl Download Stage

praateekmahajan:
## Usage

```python
```
## Checklist


ray-curator/ray_curator/examples/common_crawl_example.py
```diff
@@ -0,0 +1,195 @@
...
+        from ray_curator.backends.experimental.ray_data import RayDataExecutor
+
+        executor = RayDataExecutor()
+    elif args.executor == "ray_actors":
+        from ray_curator.backends.experimental.ray_actors import RayActorExecutor
+
+        executor = RayActorExecutor()
+    else:
+        msg = f"Invalid executor type: {args.executor}"
+        raise ValueError(msg)
```
ayushdg: If this is going to be a common pattern let's just create a simple utility that returns the default variant of these executors instead of the logic shared across examples?

One motivation for keeping it here is only to show how users can choose a specific executor.

ray-curator/ray_curator/examples/common_crawl_example.py
```diff
@@ -0,0 +1,195 @@
...
+    # Add Common Crawl pipeline stage
+    # The CommonCrawlPipeline is a CompositeStage that will be decomposed
+    # into its constituent stages during pipeline building
+    pipeline.add_stage(
+        CommonCrawl(
+            start_snapshot=start_snapshot,
+            end_snapshot=end_snapshot,
+            download_dir=str(download_dir),
+            crawl_type=crawl_type,
+            html_extraction=html_extraction_algorithm,
```
ayushdg: Is it worth showing the use case of passing in a HTMLExtractor as well instead of string only?

ray-curator/ray_curator/stages/download/text/common_crawl/html_extractor.py
```diff
@@ -9,20 +9,37 @@
...
+        algorithm_kwargs = algorithm_kwargs or {}
         if algorithm is None:
             logger.warning("No algorithm provided, using justext with default parameters")
             algorithm = JusTextExtractor()
+        elif isinstance(algorithm, str):
+            if algorithm == "justext":
+                algorithm = JusTextExtractor(**algorithm_kwargs)
+            elif algorithm == "resiliparse":
+                algorithm = ResiliparseExtractor(**algorithm_kwargs)
+            elif algorithm == "trafilatura":
```
ayushdg: Nit: Also raise here if it's another value.

ray-curator/ray_curator/stages/download/text/common_crawl/stage.py
```diff
@@ -0,0 +1,80 @@
...
+
+        # HTML extraction stage
+        html_stage = CommonCrawlHTMLExtractor(
+            algorithm=self.html_extraction,
+            algorithm_kwargs=self.html_extraction_kwargs,
+            stop_lists=self.stop_lists,
+        )
+
+        # Set up the pipeline
+        self.stages = [
```
ayushdg: Question (for my own curiosity): Any reason you prefer this over having this logic in decompose?

ray-curator/ray_curator/stages/download/text/common_crawl/url_generation.py
```diff
@@ -0,0 +1,255 @@
...
+        end_date = end_dt_obj.date()
+
+        if start_date > end_date:
+            msg = f"Start snapshot '{self.start_snapshot_str}' is after end snapshot '{self.end_snapshot_str}'"
+            raise ValueError(msg)
+
+        today_utc_date = datetime.now(tz=timezone.utc).date()
+        if end_date > today_utc_date:
+            logger.warning(
+                f"Requested end date {end_date} is in the future. Adjusting end date to today's date ({today_utc_date})."
```
ayushdg, line 56: Neat!

ray-curator/ray_curator/examples/common_crawl_example.py
```diff
@@ -0,0 +1,195 @@
...
+
+def create_common_crawl_pipeline(  # noqa: PLR0913
+    download_dir: Path,
+    output_dir: Path,
+    output_format: Literal["parquet", "jsonl"],
+    crawl_type: Literal["main", "news"],
+    start_snapshot: str,
+    end_snapshot: str,
+    html_extraction_algorithm: str = "justext",
+    aws: bool = False,
```
ayushdg: My understanding of this param is that it'll either use s5cmd to download from the s3://commoncrawl bucket or use wget to pull that url (which might still be hosted on s3). 

I'm not sure what the motivation for either option is but should be rename this to something else or update the doctstring to set the expectation here?

ray-curator/tests/stages/download/text/common_crawl/test_download.py
```diff
@@ -0,0 +1,111 @@
+import os
```
ayushdg: Nit: Do you prefer these being split into separate files or a single file `test_common_crawl` with separate classes for each part, download, extract

ray-curator/tests/stages/download/text/common_crawl/test_html_extractor.py
```diff
@@ -0,0 +1,117 @@
+import pandas as pd
```
ayushdg: Filename for this and the next one test_html_extractors is a bit confusing.

ray-curator/tests/stages/download/text/common_crawl/test_stage.py
```diff
@@ -0,0 +1,93 @@
...
+        """Test different HTML extraction algorithms initialization."""
+        download_dir = str(tmp_path / "downloads")
+
+        # Test with string algorithm
+        stage_justext = CommonCrawl(
+            start_snapshot="2021-04", end_snapshot="2021-10", download_dir=download_dir, html_extraction="justext"
+        )
+
+        # Get the HTML extractor stage (4th stage)
+        html_stage = stage_justext.decompose()[3]
```
ayushdg: Historically these tests can sometimes be flaky when hitting the external common crawl buckets: #236, #579 etc. 

Does it make sense to patch the url/downloader to return some predetermined paths. If the files are small enough we can include them in the data folder. 

We should still keep the tests which hit the cc service around but maybe agree to skip in CI or run at a different cadence.

ayushdg (commented):
Minor comments/questions. Looks good overall.

ray-curator/ray_curator/stages/download/text/common_crawl/stage.py
praateekmahajan: Intuitively made more sense to have all stages be initialized in `_post_init` (or `init` if not a dataclass) then inside a method.. Happy to move it around

ray-curator/tests/stages/download/text/common_crawl/test_download.py
praateekmahajan: I personally prefer that tests also follow a similar directory structure as the original modules, so that there is a close enough 1 to 1 mapping. 
Having a single file with separate classes, leads you to have 1000+ line files, which makes harder to work on... 
tldr : separation is preffered

ray-curator/tests/stages/download/text/common_crawl/test_stage.py
praateekmahajan: Good callout, I realised I never added them. Now created a new file called test_url_generation, where I've mocked the content of the files

ray-curator/tests/stages/download/text/common_crawl/test_html_extractor.py
praateekmahajan: I agree, do you have any recommendations, not only for the tests but also modules.. 
We have a module called `html_extractors` which has `justext` `resiliparse` `trafilatura`.. And then we have the stage inside the `html_extractor.py` which implements the stage and one of the algorithms from the module

ray-curator/ray_curator/examples/common_crawl_example.py
praateekmahajan: I can't imagine doing that atleast through the argparse setup, without bloating it a lot... Given each HtmlExtractor has a bunch of args too.. maybe that works when we have pipeline.yml

ray-curator/ray_curator/examples/common_crawl_example.py
praateekmahajan: I think we can split it out.. Related question is, are we going to have one example per module? That seems like an overkill..

ray-curator/ray_curator/examples/common_crawl_example.py
praateekmahajan: Good callout, I've renamed `aws` to `use_aws_to_download` (could be renamed to `use_s3_to_download`)

On the other hand, when I started testing it, I realised I don't have the right credentials to download from.. are you or @sarahyurick able to download? 

Can you test `aws s3 ls --region us-east-1  s3://commoncrawl/crawl-data/CC-MAIN-2018-17/segments/1524125937193.1/warc/` 
For me I'm getting `An error occurred (AccessDenied) when calling the ListObjectsV2 operation: Access Denied` but that might be because my s3 profile might be different

ray-curator/ray_curator/stages/download/text/common_crawl/warc_reader.py
```diff
@@ -0,0 +1,83 @@
...
+
+import pandas as pd
+from loguru import logger
+from warcio.archiveiterator import ArchiveIterator  # TODO: consider using fastwarc
+
+from ray_curator.stages.base import ProcessingStage
+from ray_curator.tasks import DocumentBatch, FileGroupTask
+
+
+class WarcReader(ProcessingStage[FileGroupTask, DocumentBatch]):
```
sarahyurick: Curious why you renamed this from `CommonCrawlWARCIterator`?

ray-curator/ray_curator/stages/download/text/common_crawl/html_extractor.py
```diff
@@ -0,0 +1,111 @@
...
+
+from .html_extractors import HTMLExtractorAlgorithm
+from .html_extractors.justext import JusTextExtractor
+from .html_extractors.resiliparse import ResiliparseExtractor
+from .html_extractors.trafilatura import TrafilaturaExtractor
+from .utils import get_stop_list_dict
+
+
+@dataclass
+class CommonCrawlHTMLExtractor(ProcessingStage[DocumentBatch, DocumentBatch]):
```
sarahyurick: Same here, you would prefer not to call it `CommonCrawlWARCExtractor` anymore?

ray-curator/ray_curator/examples/common_crawl_example.py
```diff
@@ -0,0 +1,195 @@
...
+        Pipeline: Configured pipeline
+    """
+    # Define pipeline
+    pipeline = Pipeline(name="common_crawl_processing", description="Download and process Common Crawl data")
+
+    # Add Common Crawl pipeline stage
+    # The CommonCrawlPipeline is a CompositeStage that will be decomposed
+    # into its constituent stages during pipeline building
+    pipeline.add_stage(
+        CommonCrawl(
```
sarahyurick: Would it make more sense to call this `DownloadCommonCrawl` (to match the old name `download_common_crawl`)?

ray-curator/ray_curator/stages/download/text/common_crawl/html_extractors/__init__.py
sarahyurick, line 1: Nitpicky, but since the jusText, Resiliparse, and Trafilatura extractors are not used anywhere else in our codebase, maybe it would be okay to combine all this into the `html_extractor.py` file for easier navigation?

ray-curator/ray_curator/stages/download/text/common_crawl/stage.py
```diff
@@ -0,0 +1,82 @@
...
+
+from .download import CommonCrawlWARCDownloader
+from .html_extractor import CommonCrawlHTMLExtractor
+from .html_extractors import HTMLExtractorAlgorithm
+from .url_generation import MainCommonCrawlUrlStage, NewsCommonCrawlUrlStage
+from .warc_reader import WarcReader
+
+
+@dataclass
+class CommonCrawl(CompositeStage[_EmptyTask, DocumentBatch]):
```
sarahyurick: I was wondering what you think about having this return a `FileGroupTask` (or maybe `DocumentBatch | FileGroupTask`) instead of a `DocumentBatch` (by adding a final write stage to this class)? The reasoning behind this being that (usually) a user would immediately want to save the results of a large extraction job and have a separate script for additional Curator pipelines.

One pro IMO is that if a download job is interrupted, then some of the batches will have automatically already written the extracted outputs, so it would be easier to create a resuming script without having to start all over.

An argument against adding the write stage here is that the user could just add it to the end of the pipeline in their own script, which is more customizable.

ray-curator/ray_curator/stages/download/text/common_crawl/download.py
```diff
@@ -0,0 +1,107 @@
...
+import subprocess
+from urllib.parse import urlparse
+
+from loguru import logger
+
+from ray_curator.stages.base import ProcessingStage
+from ray_curator.tasks import FileGroupTask
+
+
+class CommonCrawlWARCDownloader(ProcessingStage[FileGroupTask, FileGroupTask]):
```
sarahyurick: I was thinking about the pros and cons of keeping the parent `DocumentDownloader`, etc. classes from https://github.com/NVIDIA-NeMo/Curator/blob/main/nemo_curator/download/doc_builder.py.

These parent classes really have no functional value, but they do interlink the ArXiv, Common Crawl, and Wikipedia implementations nicely, to make it clear that all of our download modules follow the same 3 steps.

They are also used in several of our tutorials:

- https://github.com/NVIDIA-NeMo/Curator/blob/main/tutorials/bitext_cleaning/docbuilder.py
- https://github.com/NVIDIA-NeMo/Curator/blob/main/tutorials/dapt-curation/code/docbuilder.py
- https://github.com/NVIDIA-NeMo/Curator/blob/main/tutorials/peft-curation/docbuilder.py
- https://github.com/NVIDIA-NeMo/Curator/blob/main/tutorials/peft-curation-with-sdg/docbuilder.py
- https://github.com/NVIDIA-NeMo/Curator/blob/main/tutorials/tinystories/docbuilder.py

Which are really nice implementations that demonstrate how users can take this 3 step pipeline and customize it to their own data that they want to bring.

If we really wanted to be ambitious, we could support custom download-iterate-extract pipelines with a composite `DownloadCustom` stage that expects those 3 parent classes.

Just my thoughts on why it may be nice to keep these parent classes.

sarahyurick (commented):
Very nice, left some high level comments.

ray-curator/tests/stages/download/text/common_crawl/test_html_extractor.py
sarahyurick: I was thinking all of the extractor logic could be combined into a single file, unless you are worried about bloat?

ray-curator/ray_curator/stages/download/text/common_crawl/html_extractor.py
praateekmahajan: Technically WarcReader or WarcIterator extracts content from WARC, while this one does HTML parsing, hence I believe this is more correctly named

ray-curator/ray_curator/stages/download/text/common_crawl/warc_reader.py
praateekmahajan: Renamed it to `CommonCrawlWarcIterator`

ray-curator/ray_curator/examples/common_crawl_example.py
praateekmahajan: Renamed it to `CommonCrawlDownloadExtractStage` since `DownloadCommonCrawl` is incomplete as we not only download but also extract contents.. WDYT?

ray-curator/ray_curator/stages/download/text/common_crawl/html_extractors/__init__.py
praateekmahajan: Would prefer keeping it separate to avoid file bloat as all of them are roughly ~100 lines and have their own set of imports. We have one single pytest file for them because we go over fixtures

ray-curator/ray_curator/stages/download/text/common_crawl/stage.py
praateekmahajan: > An argument against adding the write stage here is that the user could just add it to the end of the pipeline in their own script, which is more customizable.

I think this basically. We want to be able to support custom pipelines where they download / run classifiers etc etc in one go.

ray-curator/ray_curator/stages/download/text/common_crawl/download.py
praateekmahajan: This was a great idea and I've now implemented it. Thank you so much for calling this out. Please take a look at `stages/download/text/base`

ray-curator/ray_curator/examples/common_crawl_example.py
```diff
@@ -0,0 +1,203 @@
...
+    if args.executor == "xenna":
+        executor = XennaExecutor()
+    elif args.executor == "ray_data":
+        from ray_curator.backends.experimental.ray_data import RayDataExecutor
+
+        executor = RayDataExecutor()
+    elif args.executor == "ray_actors":
+        from ray_curator.backends.experimental.ray_actors import RayActorExecutor
+
+        executor = RayActorExecutor()
```
sarahyurick: Remove this.

ray-curator/ray_curator/examples/common_crawl_example.py
```diff
@@ -0,0 +1,203 @@
...
+        Pipeline: Configured pipeline
+    """
+    # Define pipeline
+    pipeline = Pipeline(name="common_crawl_processing", description="Download and process Common Crawl data")
+
+    # Add Common Crawl pipeline stage
+    # The CommonCrawlPipeline is a CompositeStage that will be decomposed
+    # into its constituent stages during pipeline building
+    pipeline.add_stage(
+        CommonCrawlDownloadExtractStage(
```
sarahyurick: Wish we didn't have to create a pipeline just to add 1 stage... oh well.

Btw, does this write the extracted results?

ray-curator/ray_curator/examples/common_crawl_example.py
```diff
@@ -0,0 +1,203 @@
...
+    parser.add_argument("--url_limit", type=int, default=5, help="Limit the number of WARC files to process")
+    parser.add_argument("--record_limit", type=int, default=None, help="Limit the number of records to process")
+    parser.add_argument("--add_filename_column", action="store_true", help="Add filename column to output")
+
+    # Executor configuration
+    parser.add_argument(
+        "--executor",
+        type=str,
+        default="xenna",
+        choices=["xenna", "ray_data", "ray_actors"],
```
sarahyurick:
```suggestion
        choices=["xenna", "ray_data"],
```

ray-curator/ray_curator/stages/download/text/README.md
```diff
@@ -0,0 +1,125 @@
...
+├── download.py        # Download implementation
+├── iterator.py        # File iteration logic
+└── extract.py         # Data extraction logic
+```
+
+### 2. Implement Core Components
+
+**URL Generator** (`url_generation.py`):
+```python
+from ray_curator.stages.download.text import URLGenerator
```
sarahyurick, line 31: Any chance you can imagine these modules being used to download other modalities?

ray-curator/ray_curator/stages/download/text/README.md
```diff
@@ -0,0 +1,125 @@
...
+)
+
+pipeline = Pipeline("your_pipeline")
+pipeline.add_stage(stage)
+results = pipeline.run(executor)
+```
+
+## 💡 Key Notes
+
+- **Extractor is optional** - omit if iteration produces final format
```
sarahyurick, line 114: Should `URLGenerator` be optional too? What if a user just wants to pass a list of URLs into the downloader?

ray-curator/ray_curator/stages/download/text/base/download.py
```diff
@@ -0,0 +1,147 @@
...
+            if self._verbose:
+                logger.info(f"File: {output_file} exists. Not downloading")
+            return output_file
+
+        # Download to temporary file
+        success, error_message = self._download_to_path(url, temp_file)
+
+        if success:
+            # Download successful, atomically move temp file to final location
+            os.rename(temp_file, output_file)
```
sarahyurick, line 79: Could you add a more explanatory comment about why this is done?

ray-curator/ray_curator/stages/download/text/common_crawl/extract.py
```diff
@@ -0,0 +1,93 @@
...
+        # Content from WARC records is bytes, even though type annotation suggests str
+        html = decode_html(html_content)
+
+        if html is not None:
+            # Language detection and HTML extraction
+            lang = lang_detect(html)
+
+            text = None
+            # TODO: Understand more on why we need to check for stop_lists here and why only
+            # few of the records make it
```
sarahyurick: Happy to discuss this portion of the codebase if there is anything unclear here.

ray-curator/ray_curator/stages/download/text/common_crawl/html_extractors/base.py
sarahyurick, line 1: Okay, in this case can we move the HTML extractors outside of Common Crawl? Mostly for visibility purposes, so users know they are generic enough to be used outside of Common Crawl.

Would love to encourage other users to consider them when implementing their own data downloaders :)

ray-curator/ray_curator/stages/download/text/common_crawl/url_generation.py
```diff
@@ -0,0 +1,240 @@
...
+            start_date = date(2013, 1, 1)
+        snapshot_data = self._snapshot_index
+        warc_gz_urls = []
+        crawl_data_main_prefix = urljoin(self.data_prefix, "crawl-data/CC-MAIN")
+
+        for snapshot_info in snapshot_data:
+            snapshot_id = snapshot_info.get("id")
+            # 2008-2010 are old snapshots and not supported by this script
+            if not snapshot_id or snapshot_id in {"CC-MAIN-2009-2010", "CC-MAIN-2008-2009"}:
+                continue
```
sarahyurick: Should we raise a warning or error here?

ray-curator/ray_curator/stages/download/text/common_crawl/utils/__init__.py
sarahyurick, line 1: Same here, would like to move this out of Common Crawl since it is generic enough to be used for other modules.

ray-curator/ray_curator/stages/download/text/README.md
praateekmahajan: I do, in which case we can move it out, but for now I think we can keep it here, as that's the only place its used.

ray-curator/ray_curator/stages/download/text/README.md
praateekmahajan: > list of URLs
For now we don't allow any primitive data types to be passed when you want to start a pipeline and it has to be some Task. If you meant `list[FileGroupTask]`, you're right that a user could pass in a `list[FileGroupTask]`.

Currently that is somehow possible where the user can explicitly create a pipeline, instead of using our composite stage pipeline as below
```python
Pipeline(
  stages=[
    DownloadStage(...), 
    IterateStage(), 
    ExtractStage(), 
    WriteStage()
  ]
).run(executor, initial_tasks=list[FileGroupTask])
```
I think supporting in our CompositeStage will be tricky since Stages go from X to Y, and when URLGenerator is not present then X is FileGroupTask otherwise its EmptyTask. The cleaner way to do would be have a new CompositeStage called `DownloadAndExtractWithCustomUrls` or something. 

But I don't know if we should encourage that, since URLGenerator is a rather trivial stage.

ray-curator/ray_curator/stages/download/text/common_crawl/html_extractors/base.py
praateekmahajan: Good idea!

ray-curator/ray_curator/stages/download/text/common_crawl/url_generation.py
```diff
@@ -0,0 +1,241 @@
...
+            start_date = date(2013, 1, 1)
+        snapshot_data = self._snapshot_index
+        warc_gz_urls = []
+        crawl_data_main_prefix = urljoin(self.data_prefix, "crawl-data/CC-MAIN")
+
+        for snapshot_info in snapshot_data:
+            snapshot_id = snapshot_info.get("id")
+            # 2008-2010 are old snapshots and not supported by this script
+            if not snapshot_id or snapshot_id in {"CC-MAIN-2009-2010", "CC-MAIN-2008-2009"}:
+                logger.warning(f"Skipping old snapshot for common crawl: {snapshot_id}")
```
sarahyurick:
```suggestion
                logger.warning(f"Skipping old snapshot for Common Crawl: {snapshot_id}")
```
Nit.

ray-curator/ray_curator/stages/download/text/base/iterator.py
```diff
@@ -0,0 +1,103 @@
...
+
+        for file_path in task.data:
+            try:
+                record_count = 0
+                iterator_result = self.iterator.iterate(file_path)
+                if iterator_result is not None:
+                    for record_dict in iterator_result:
+                        if self.record_limit and record_count >= self.record_limit:
+                            break
+                        # TODO: this should work with cloud storage
```
sarahyurick: Should we open issues for the remaining TODOs?

ray-curator/ray_curator/stages/download/text/common_crawl/url_generation.py
```diff
@@ -0,0 +1,241 @@
...
+            gz_path_urls = self.generate_path_urls()
+        else:
+            gz_path_urls = [path_urls] if isinstance(path_urls, str) else path_urls
+
+        all_individual_warc_urls = []
+
+        if not gz_path_urls:
+            return []
+
+        for gz_path_url in gz_path_urls:  # TODO: check this
```
sarahyurick: See above comment.

ray-curator/ray_curator/stages/download/text/common_crawl/warc_iterator.py
```diff
@@ -0,0 +1,40 @@
+from collections.abc import Iterator
+from pathlib import Path
+from typing import Any
+
+from loguru import logger
+from warcio.archiveiterator import ArchiveIterator  # TODO: consider using fastwarc
```
sarahyurick: See above comment.

ray-curator/pyproject.toml
```diff
@@ -22,6 +22,13 @@ dependencies = [
     "transformers>=4.48.0",
     "pyarrow",
     "absl-py>=2.0.0,<3.0.0",
+    # Text Download / Extract
```
sarahyurick:
```suggestion
```
I keep adding nits, sorry. Let's remove this comment.

ray-curator/ray_curator/stages/download/text/base/extract.py
```diff
@@ -0,0 +1,102 @@
...
+        # Convert to DataFrame
+        df = pd.DataFrame(extracted_records)
+
+        return DocumentBatch(
+            task_id=task.task_id,
+            dataset_name=task.dataset_name,
+            data=df,
+            _metadata={
+                **task._metadata,
+            },
```
sarahyurick, line 95: Can't it just be `_metadata=task._metadata`?

ray-curator/ray_curator/stages/download/text/base/iterator.py
```diff
@@ -0,0 +1,103 @@
...
+        # Convert to DataFrame
+        df = pd.DataFrame(records)
+
+        return DocumentBatch(
+            task_id=task.task_id,
+            dataset_name=task.dataset_name,
+            data=df,
+            _metadata={
+                **task._metadata,
+            },
```
sarahyurick, line 96: Same comment.

ray-curator/pyproject.toml
```diff
@@ -22,6 +22,13 @@ dependencies = [
...
+    "justext",
```
abhinavg4, line 37: Can we put this inside a different text dependency? I think as we add more dependencies, the core might get very large.

ray-curator/ray_curator/stages/download/text/README.md
```diff
@@ -0,0 +1,125 @@
+# Adding Custom Download Pipelines
```
abhinavg4, line 1: The nice document really helps to follow along.

ray-curator/ray_curator/stages/download/text/base/download.py
```diff
@@ -0,0 +1,148 @@
...
+    def __init__(self, download_dir: str, verbose: bool = False):
+        """Initialize the downloader.
+
+        Args:
+            download_dir: Directory to store downloaded files
+            verbose: If True, logs detailed download information
+        """
+        self._download_dir = download_dir
+        self._verbose = verbose
+        os.makedirs(download_dir, exist_ok=True)
```
abhinavg4, line 24: Can we use fsspec instead of os ? Not sure if we can support remote download extract but a good step towards it.

ray-curator/ray_curator/stages/download/text/base/download.py
```diff
@@ -0,0 +1,148 @@
...
+        Checks for existing file to avoid re-downloading. Supports resumable downloads.
+        Args:
+            url: URL to download
+
+        Returns:
+            Path to downloaded file, or None if download failed
+        """
+        # Generate output filename
+        output_name = self._get_output_filename(url)
+        output_file = os.path.join(self._download_dir, output_name)
```
abhinavg4, line 65: Same fsspec comment. Are we assuming that the code and data lives on the same machine. Was wondering how much effort is it to support remote data? If it's just fsspec, we should do it.

Can use a lot of utils from here:

https://github.com/marin-community/marin/blob/main/marin/utils.py

ray-curator/ray_curator/stages/download/text/base/download.py
```diff
@@ -0,0 +1,148 @@
...
+        temp_file = output_file + ".tmp"
```
abhinavg4, line 66: Can we use 

`with tempfile.NamedTemporaryFile(mode='w+', delete=False) as tmp_file:` 

and 

`with tempfile.TemporaryDirectory() as tmp_dir:`

It has several benefits, including permission issues, automatic cleanup, symlink attack security etc.

ray-curator/ray_curator/stages/download/text/base/extract.py
```diff
@@ -0,0 +1,102 @@
...
+
+    @property
+    def name(self) -> str:
+        """Return stage name."""
+        extractor_name = self.extractor.__class__.__name__
+        return f"extract_{extractor_name.lower()}"
+
+    def inputs(self) -> tuple[list[str], list[str]]:
+        """Define input requirements - expects DocumentBatch with dict records."""
+        return (["data"], self.extractor.input_columns() + ([self.filename_col] if self.add_filename_column else []))
```
abhinavg4: Unrelated, but I think we should get rid of inputs and outputs first return type. Will it always be data?

ray-curator/ray_curator/stages/download/text/common_crawl/download.py
```diff
@@ -0,0 +1,86 @@
...
+        if self._verbose:
+            logger.info(f"Downloading {url_to_download} to {path}")
+
+        # Download with either wget or s5cmd (aws) to temporary file
+        if self.use_aws_to_download:
+            cmd = ["s5cmd", "cp", url_to_download, path]
+        else:
+            # We don't use -c (for continue resume) because we want to download file to temp path using -O
+            # but -c and -O don't work well together
+            cmd = ["wget", url_to_download, "-O", path]
```
abhinavg4, line 68: Btw, in my previous experience, this is super unreliable, and the HTTP endpoints for CC always seem to error out before downloading the whole thing. For context, I was downloading Huge amounts of data in parallel (About 10% of the whole CC for 2020 or something)

ray-curator/ray_curator/stages/download/text/common_crawl/download.py
```diff
@@ -0,0 +1,86 @@
...
+        urlpath = urlparse(url).path[1:]
+
+        url_to_download = os.path.join("s3://commoncrawl/", urlpath) if self.use_aws_to_download else url
+
+        if self._verbose:
+            logger.info(f"Downloading {url_to_download} to {path}")
+
+        # Download with either wget or s5cmd (aws) to temporary file
+        if self.use_aws_to_download:
+            cmd = ["s5cmd", "cp", url_to_download, path]
```
abhinavg4, line 64: Incase you are curious another way of doing this is via s3_fs:

https://github.com/marin-community/marin/blob/92d91981be2e0bf66e7d247d0f98a1b498901765/marin/crawl/fineweb_edu/convert_fineweb_edu_to_html.py#L107

ray-curator/ray_curator/stages/download/text/common_crawl/warc_iterator.py
```diff
@@ -0,0 +1,43 @@
...
+
+class CommonCrawlWarcIterator(DocumentIterator):
+    """Processes downloaded WARC files."""
+
+    def iterate(self, file_path: str) -> Iterator[dict[str, Any]]:
+        """Process a task containing WARC files and extract their contents."""
+        filename = file_path.name if isinstance(file_path, Path) else file_path.split("/")[-1]
+
+        num_records = 0
+        # TODO: Support cloud storage https://github.com/NVIDIA-NeMo/Curator/issues/779
```
abhinavg4, line 21: I think I mentioned this above but this supports cloud storage directly:

https://github.com/marin-community/marin/blob/92d91981be2e0bf66e7d247d0f98a1b498901765/marin/crawl/fineweb_edu/convert_fineweb_edu_to_html.py#L107

ray-curator/ray_curator/stages/download/text/base/download.py
```diff
@@ -0,0 +1,148 @@
...
+
+        # If final file exists and is non-empty, assume it's complete
+        if os.path.exists(output_file) and os.path.getsize(output_file) > 0:
+            if self._verbose:
+                logger.info(f"File: {output_file} exists. Not downloading")
+            return output_file
+
+        # If temp file exists delete it
+        if os.path.exists(temp_file):
+            os.remove(temp_file)
```
abhinavg4: This prevents resumability inside _download_to_path ?

praateekmahajan: Good catch! Had forgotten to remove it.

ray-curator/ray_curator/stages/download/text/base/extract.py
praateekmahajan: I agree, hopefully once the video prs come in we'll see a better consistent pattern.

ray-curator/ray_curator/stages/download/text/base/download.py
praateekmahajan: Left it for later https://github.com/NVIDIA-NeMo/Curator/issues/779

ray-curator/pyproject.toml
praateekmahajan: Good point!

ray-curator/ray_curator/stages/download/text/base/extract.py
praateekmahajan: I think it can, I've made it intentionally a copy because.. I observed something weird with fanout stages (e.g. URLGeneration / FilePartitioning) where if we use the same as `task._perf_stats` then they all end up mutating the same object. Which is why I intentionally kept it a copy until we figured that out

ray-curator/ray_curator/stages/download/text/base/download.py
praateekmahajan: I don't think we can use tmp file if we want resumability as we want consistent name each time. If we can have consistent name and no-delete with NamedTemp file that'll be cool.

sarahyurick (approved):
LGTM! It looks like the PyTests are failing because of the `[text]` dependencies, once that is updated it should be good to merge.

PR #740, opened 2025-06-18
[Ray] Add Ray Data as an experimental backend

praateekmahajan:
## Description
**Add RayDataStageExecutor**
Takes in a `list[Task]` and creates a dataset using `ds = ray.data.from_items`. 
Each subsequent `RayDataStageAdapter` transforms the data using `ds.map_batches(RayDataStageAdapter(..).func)`

**Add RayDataStageAdapter**
This wraps `stage`s `process_batch` function and passes it to process_batch. 

**Changes to `ProcessingStage `**
1. Added `ray_stage_spec` 
    - The setting for `is_fanout_stage` (default is False) allows us to
   
**Changes to `FilePartitioningStage `**
1. Change `Resource` to `0.5` since its just a ls stage (also enables Ray Data to not fuse)

 
## Learnings
1. Originally I was converting each task to a dict (and had a Task.to_dict / Task.from_dict). 
    - The issue with that is RayData can serialize a pd.DataFrame differently based on some internal serialization works 
    - Repro here https://gist.github.com/praateekmahajan/aaae567b6a573e7b75915e37979fb169
2. If we have a fanout stage, and we don't repartition, then all subsequent stages are considered to be in the same block and hence executed serially. This makes us repartition data and hence the `is_fanout_stage`.
    - Repro here https://gist.github.com/praateekmahajan/1f803f4f36f15f7b73c1c7b13bbc1ab5
3. Ray Data Task Fusion
   - If we have multiple stages which have the same resources, Ray will fuse them together.
   - This means if one stage is I/O intensive and other is Compute intensive with vastly different runtimes, they'll get fused.
   - To circumvent fusion and hence enable multi-stage-atusocaling, we need to make sure resources are different. (e.g.  FilePartitioningStage)
4. Ray Data `repartition(target_num_rows_per_block=1)` leads to an error unclear when
   - Gist here https://gist.github.com/praateekmahajan/85c3bdc1b2fc48905d70c69af768a74d

## TODO for future
1. Implement setup / setup_on_node functionality
6. Tasks vs Actors - investigate. If to `map_batches` we pass `Class.fn` vs pass `Class.__call__` is there a difference?

## Usage

```python
```
## Checklist


ray-curator/ray_curator/backends/experimental/ray_data/executor.py
```diff
@@ -0,0 +1,114 @@
...
+    3. Leverages Ray Data's automatic parallelization and resource management
+    4. Returns final results as a list of tasks
+    """
+
+    def __init__(self, config: dict[str, Any] | None = None):
+        super().__init__(config)
+
+        # Initialize Ray if not already initialized
+        if not ray.is_initialized():
+            ray.init()
```
sarahyurick: Hmm should we allow this?

ray-curator/ray_curator/backends/experimental/ray_data/executor.py
```diff
@@ -0,0 +1,114 @@
...
+        """Convert Ray Data dataset back to list of tasks.
+
+        Args:
+            dataset: Ray Data dataset containing task dictionaries
+
+        Returns:
+            List of Task objects
+        """
+        # Get all items from dataset
+        task_dicts = dataset.take_all()
```
sarahyurick: Maybe I am reading this incorrectly, is this on the whole dataset or just a batch? Couldn't this OOM?

ray-curator/ray_curator/backends/experimental/__init__.py
sarahyurick, line 1: In addition to being in the `experimental` folder, we should log/print to the user that this is experimental as well.

sarahyurick (commented):
Tested with https://github.com/NVIDIA-NeMo/Curator/pull/746 (no changes) and it worked for me.

ray-curator/ray_curator/backends/experimental/ray_data/executor.py
praateekmahajan: Removed it for now

ray-curator/ray_curator/backends/experimental/ray_data/executor.py
praateekmahajan: If you see where `_dataset_to_tasks` is called there is a TODO that says 
`            # TODO: add pipeline configuration to check if user wants to return last stages output to driver`

I believe Xenna has configuration that says "return_to_driver" and that is what we should mimic. However have left that for now

ray-curator/ray_curator/examples/simple_read_example.py
```diff
@@ -107,6 +107,14 @@ def main(args: argparse.Namespace) -> None:
...
     if args.executor == "xenna":
         executor = XennaExecutor()
+    elif args.executor == "ray_data":
+        from ray_curator.backends.experimental.ray_data import RayDataExecutor
+
+        executor = RayDataExecutor()
+    elif args.executor == "ray_actors":
+        from ray_curator.backends.experimental.ray_actors import RayActorExecutor
+
+        executor = RayActorExecutor()
```
sarahyurick: Since this is still very experimental, maybe we omit from our basic example script for now?

sarahyurick: Plus, there is no `RayActorExecutor` yet, right?

ray-curator/ray_curator/backends/utils.py
```diff
@@ -0,0 +1,9 @@
+import ray
+
+from ray_curator.backends.base import NodeInfo, WorkerMetadata
+
+
+def get_worker_metadata_and_node_id() -> tuple[NodeInfo, WorkerMetadata]:
```
abhinavg4: Can we move these utils inside ray_data? My thinking is that anything that is not related to xenna lives in respective folders to avoid confusion.

ray-curator/ray_curator/examples/simple_read_example.py
abhinavg4: I agree with Sarah. Anything that is experimental should probably go under examples/experimental or somehting

ray-curator/ray_curator/stages/base.py
```diff
@@ -221,6 +221,15 @@ def xenna_stage_spec(self) -> dict[str, Any]:
         """
         return {}
 
+    @property
+    def ray_stage_spec(self) -> dict[str, Any]:
```
abhinavg4: Keeping with the above theme, would it be better to move this inside backends/experimental?

One way might be to have mappings from attributes to specific classes inside the adapter.

In case we are keeping it here. Best to add in the doc string that this is only used in Ray Data and put it at the end of the class definition. I think it might confuse first time users as to what is the use of this function.

ray-curator/ray_curator/utils/performance_utils.py
```diff
@@ -20,6 +20,7 @@
 class StagePerfStats:
     """Statistics for tracking stage performance metrics.
     Attributes:
+        stage_name: Name of the processing stage.
```
abhinavg4, line 23: Thanks for catching this.

abhinavg4:
I think it would be good to move all Ray data related functions inside backends/experimental just to hide it from the end-users

abhinavg4:
Also it might be better add some tests for code coverage right away else we might have lower coverage later.

ray-curator/ray_curator/backends/experimental/ray_data/setup_utils.py
```diff
@@ -0,0 +1,31 @@
...
+    setup_on_node_fn: Callable[[NodeInfo | None, WorkerMetadata | None], None],
+) -> None:
+    """Setup the stage with coordination.
+
+    Args:
+        stage (ProcessingStage): Processing stage to setup
+        setup_fn (Callable): Function to call for per-worker setup
+        setup_on_node_fn (Callable): Function to call for per-node setup
+    """
+    logger.warning("Ray Data setup / setup_on_node_fn is likely ineffecient.")
```
abhinavg4: Do you think we can make this more informative and actionable, or convert this to a TODO?

ray-curator/ray_curator/backends/experimental/ray_data/setup_utils.py
```diff
@@ -0,0 +1,31 @@
...
+from ray_curator.backends.utils import get_worker_metadata_and_node_id
+from ray_curator.stages.base import ProcessingStage
+
+
+def setup_stage_with_coordination(
+    stage: ProcessingStage,  # noqa: ARG001
+    setup_fn: Callable[[WorkerMetadata], None],
+    setup_on_node_fn: Callable[[NodeInfo | None, WorkerMetadata | None], None],
+) -> None:
+    """Setup the stage with coordination.
```
abhinavg4: What do you mean by "coordination"?

Can we explain this a bit more here ? Also possibly change function name to something more informative.

ray-curator/ray_curator/backends/experimental/ray_data/adapter.py
```diff
@@ -0,0 +1,127 @@
...
+    @property
+    def batch_size(self) -> int | None:
+        """Get the batch size for this stage."""
+        return self._batch_size
+
+    def _setup_if_needed(self) -> None:
+        """Setup the stage if it hasn't been setup yet.
+
+        This method ensures setup happens exactly once per Ray Data worker,
+        and setup_on_node happens exactly once per node.
```
abhinavg4: How are we ensuring that setup_on_node is done just once? `_setup_done` is per stage right?

Is it because there is just one instance of each stage for Ray Data, so setup and setup_on_node are essentially the same functions?

ray-curator/ray_curator/backends/experimental/ray_data/adapter.py
```diff
@@ -0,0 +1,127 @@
...
+            dataset (Dataset): Ray Data dataset containing Task objects
+
+        Returns:
+            Dataset: Processed Ray Data dataset
+        """
+        processed_dataset = dataset.map_batches(
+            create_named_ray_data_stage_adapter(self.stage).map_batch_fn,
+            batch_size=self.batch_size,
+            num_cpus=self.stage.resources.cpus,
+            num_gpus=self.stage.resources.gpus,
```
abhinavg4, line 87: Can we add a todo here for the NVENCs and NVDECs? Ray can ideally support any custom resources, and in case we want, we can add NVENCs and NVDECs later to Ray Data.


Also can we check:

1. What happens if someone provides self.stage.resources.gpus > actual num gpus on the cluster ? From what I know Ray might wait for such resource to become available instead of crashing.

2. Can we add a check if the user provides nvencs and nvdecs, we just raise an error saying we cannot spport those stages with Ray data rn.

ray-curator/ray_curator/backends/experimental/ray_data/executor.py
```diff
@@ -0,0 +1,111 @@
...
+                current_dataset = adapter.process_dataset(current_dataset)
+
+                # Log progress
+                logger.info(f"  Stage {i + 1} completed")
+
+        except Exception as e:
+            logger.error(f"Error during pipeline execution: {e}")
+            raise
+
+        else:
```
abhinavg4, line 74: Can we remove this else since we are raising the error? I feel like this try/except/else construct is not widely used/understood.

abhinavg4 (commented):
Added review related to just ray data executor. Before one was for other files

ray-curator/ray_curator/backends/experimental/ray_data/executor.py
praateekmahajan: Learnt from ruff. The link shows why its good practice too https://docs.astral.sh/ruff/rules/try-consider-else/ 
I encourage we keep it

ray-curator/ray_curator/backends/utils.py
praateekmahajan: Good callout

ray-curator/ray_curator/stages/base.py
praateekmahajan: We'll need to keep this inside the base stage spec so that concrete implementations can override it. This is later used inside RayData's adapter.
Moved it to the end, and added a comment that this experimental

ray-curator/ray_curator/backends/experimental/ray_data/setup_utils.py
praateekmahajan: Added a TODO in code, and also its tracked as a backlog item in our project board

ray-curator/ray_curator/backends/experimental/ray_data/adapter.py
praateekmahajan: This is a bad comment, good catch!
`setup_on_node` isn't correctly functional.

ray-curator/ray_curator/backends/experimental/ray_data/setup_utils.py
praateekmahajan: Renamed it to `setup_stage` and added a comment

ray-curator/ray_curator/examples/simple_read_example.py
praateekmahajan: I'm not sure if we need so many examples, just to show how to change the executor. Having a warning should be enough in my opinion for one of the examples. Can remove RayActor though

ray-curator/ray_curator/backends/experimental/ray_data/adapter.py
praateekmahajan: Good callout, we now raise an error and added a todo. 

I checked when `stage.resources.gpus > num_gpus` then RayData hangs but does repetitively tell you in warning
```
(autoscaler +3s) Error: No available node types can fulfill resource request {'CPU': 1.0, 'GPU': 6.0}. Add suitable node types to this cluster to resolve this issue.                                                                                                                                                               
```

That I believe is good enough for now

ray-curator/ray_curator/examples/simple_read_example.py
sarahyurick: It looks like `RayActorExecutor` still has not been removed here.

ray-curator/ray_curator/backends/experimental/ray_data/adapter.py
abhinavg4: Why do we do the num GPU <= 0 check? For multi GPUs the user anyways gets all the GPUs, nvencs and decs.

It's conceptually ok but I think confusing when reading the code.

ray-curator/ray_curator/backends/experimental/ray_data/setup_utils.py
abhinavg4: Sorry can't see the TODO. Probably I'm missing it somehow. Can you please point me to it.

Also I think we should still change the warning. The current warning is not very useful. Since it has ambiguity.

Can we do something like

Currently ray data setup is inefficient due to......consider putting your code logic inside process when using ray Data as backend

Something like this

ray-curator/ray_curator/backends/experimental/ray_data/executor.py
```diff
@@ -0,0 +1,111 @@
...
+        """Convert list of tasks to Ray Data dataset.
+
+        Args:
+            tasks: List of Task objects
+
+        Returns:
+            Ray Data dataset containing Task objects directly
+        """
+        # Create Ray Data dataset directly from Task objects
+        return ray.data.from_items(tasks)
```
ayushdg, line 93: Calling out a nuance here which shouldn't be an issue in most cases: `from_items` will materialize all `tasks` into the object store before creating a ray data object from it. So everything unto this point will eagerly be computed.

ray-curator/ray_curator/backends/experimental/ray_data/executor.py
```diff
@@ -0,0 +1,111 @@
...
+        """Convert Ray Data dataset back to list of tasks.
+
+        Args:
+            dataset: Ray Data dataset containing Task objects
+
+        Returns:
+            List of Task objects
+        """
+        # Get all items from dataset
+        items = dataset.take_all()
```
ayushdg, line 105: Similar caveat. Take all will first materialize this whole dataset in object store memory and then possibly bring it to the driver process (similar to `compute` in dask)

ray-curator/ray_curator/tasks/__init__.py
```diff
@@ -1,6 +1,14 @@
...
 
-__all__ = ["DocumentBatch", "EmptyTask", "FileGroupTask", "ImageBatch", "ImageObject", "Task", "_EmptyTask"]
+__all__ = [
+    "DocumentBatch",
+    "EmptyTask",
+    "FileGroupTask",
+    "ImageBatch",
+    "ImageObject",
+    "Task",
+    "_EmptyTask",
```
ayushdg, line 13: Why do we need to expose _EmptyTask directly?

ayushdg (commented):
Looks good overall! Does it make sense to move some of the `learnings` to a  file within the `experimental ray executor namespace?

ray-curator/ray_curator/backends/experimental/ray_data/executor.py
praateekmahajan: This is the start of the pipeline, so at any point you'll need to seed your pipeline with some tasks which already exist in memory i.e. are materialized. So no-op here.

ray-curator/ray_curator/backends/experimental/ray_data/executor.py
praateekmahajan: This is true! @sarahyurick also pointed this out, I have this as a TODO to fix later
> If you see where _dataset_to_tasks is called there is a TODO that says
`# TODO: add pipeline configuration to check if user wants to return last stages output to driver`
I believe Xenna has configuration that says "return_to_driver" and that is what we should mimic. However have left that for now

ray-curator/ray_curator/tasks/__init__.py
praateekmahajan: It was always there. `_EmptyTask` is the actual class so we need it for type hints while `EmptyTask` is the object. maybe the naming should be swapped

PR #749, opened 2025-06-27
ci: Add placeholder codecov check for when tests do not run

chtruong814:
## Description
Add placeholder codecov check for when tests do not run

If no test coverage is uploaded, then the `codecov/patch` status will never exist. So we add a placeholder status check using the same name for when tests do not run.

Here is an example when I forced the tests to be skipped:
Screenshot 2025-06-26 at 8 21 39 PM

* Lastly, added a `gpu-ci-result` job so that it always runs even if gpu job is skipped if it's a docs only case. This would be good to have to enable a status check on it.
* Updated cpu tests to run all PR events not just labeld and synchronized

## Usage

```python
```
## Checklist

### DIFF
--- a/.github/workflows/gpuci.yml
+++ b/.github/workflows/gpuci.yml
@@ -217,7 +217,7 @@ jobs:
           if [ "${{ needs.pre-flight.outputs.allowed_to_run }}" == "false" ]; then
             echo "No tests were required to run, marking as success"
             exit 0
-          elif [ "${{ needs.run-gpu-tests.conclusion }}" == "success" ]; then
+          elif [ "${{ needs.run-gpu-tests.result }}" == "success" ]; then
             echo "GPU tests passed successfully"
             exit 0
           else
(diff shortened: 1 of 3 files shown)


sarahyurick (commented):
Lgtm, thanks! Was wondering if we should wait until outside RAPIDS issues are solved before merging this PR?

.github/workflows/gpuci.yml
```diff
@@ -206,3 +206,31 @@ jobs:
...
+        run: |
+          if [ "${{ needs.pre-flight.outputs.allowed_to_run }}" == "false" ]; then
+            echo "No tests were required to run, marking as success"
+            exit 0
+          elif [ "${{ needs.run-gpu-tests.conclusion }}" == "success" ]; then
+            echo "GPU tests passed successfully"
+            exit 0
+          else
+            echo "GPU tests failed"
+            exit 1
```
sarahyurick, line 225: Looks like the CI went here, even though the GPU tests passed...

chtruong814: Ah yeah let me take a look

sarahyurick (approved):

PR #753, opened 2025-06-27
[Ray] Classifiers

sarahyurick:
Remaining tasks:

- [x] Address TODOs
- [x] Benchmark
- [x] Add tests

Follow-up issues and PRs:

- https://github.com/NVIDIA-NeMo/Curator/issues/838
- https://github.com/rapidsai/crossfit/issues/130
- https://github.com/NVIDIA-NeMo/Curator/issues/877
- https://github.com/NVIDIA-NeMo/Curator/issues/878
- https://github.com/NVIDIA-NeMo/Curator/pull/881


ray-curator/ray_curator/stages/classifiers/quality.py
```diff
@@ -101,32 +106,83 @@ def __init__(  # noqa: PLR0913
...
-            batch_size=self.batch_size,
+            model_batch_size=self.model_batch_size,
             label_col=self.pred_column,
             text_field=self.text_field,
             prob_col=self.prob_column,
         )
-        return DocumentDataset(df)
+
+
+class StreamingQualityClassifier(StreamingDataClassifier):
```
sarahyurick: Composite stage example.

ray-curator/ray_curator/stages/classifiers/base.py
```diff
@@ -0,0 +1,246 @@
...
+
+        msg = "filter_by must be a string or list type"
+        raise TypeError(msg)
+
+    def get_labels(self) -> list[str]:
+        return self.labels
+
+
+@dataclass(kw_only=True)
+class StreamingDataClassifier(CompositeStage[DocumentBatch, DocumentBatch]):
```
sarahyurick: Composite stage implementation. Playing around with just importing the `Tokenizer` step from CrossFit versus a CPU-only tokenizer step (see `Tokenizer` class in this file).

ray-curator/ray_curator/stages/classifiers/base.py
```diff
@@ -0,0 +1,479 @@
...
+        self.model = AutoModel.from_pretrained(config["base_model"])
+        self.dropout = nn.Dropout(config["fc_dropout"])
+        self.fc = nn.Linear(self.model.config.hidden_size, len(config["id2label"]))
+
+    @property
+    def device(self) -> torch.device:
+        return next(self.parameters()).device
+
+    def _forward(self, batch: dict[str, torch.Tensor]) -> torch.Tensor:
+        with torch.no_grad():
```
sarahyurick: TODO: Check everywhere that uses `torch.no_grad()` for the inference step.

Also look into using `cleanup_torch_cache` from CrossFit.

ray-curator/ray_curator/stages/classifiers/aegis.py
```diff
@@ -0,0 +1,555 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import os
+
+os.environ["RAPIDS_NO_INITIALIZE"] = "1"
```
praateekmahajan: Do we always need this? Especially that "stages" get serialized to processes with "gpu" and only then they cudf's cuda context must be initialized? 
cc @ayushdg

ray-curator/ray_curator/stages/classifiers/aegis.py
```diff
@@ -0,0 +1,555 @@
...
+        super().__init__()
+        base_model = AutoModelForCausalLM.from_pretrained(
+            pretrained_model_name_or_path, torch_dtype=dtype, token=token
+        )
+        # Importing PeftModel here to prevent cuda context issues
+        # that seem to happen on Transformers 4.48.3
+        # See related: https://github.com/rapidsai/crossfit/pull/113
+        from peft import PeftModel
+
+        self.model = PeftModel.from_pretrained(base_model, peft_model_name_or_path)
```
praateekmahajan: shouldn't this be in setup?

praateekmahajan: oh nvm this is a model while the stage is below

ray-curator/ray_curator/stages/classifiers/aegis.py
```diff
@@ -0,0 +1,555 @@
...
+        return df_cpu
+
+    def process(self, batch: DocumentBatch) -> DocumentBatch:
+        processed_outputs = []
+        df_cpu = batch.to_pandas()
+
+        for model_input_batch in self.yield_next_batch(df_cpu):
+            # Forward pass
+            with torch.no_grad():
+                outputs = self.model(model_input_batch)
```
praateekmahajan: Given that self.model's def _forward already has a no_grad wrapper do we need one more?

ray-curator/ray_curator/stages/classifiers/aegis.py
```diff
@@ -0,0 +1,555 @@
...
+
+        if self.add_instruction_data_guard:
+            df_cpu[self.prob_column] = collected_output["preds"].tolist()
+            df_cpu[self.pred_column] = (collected_output["preds"] >= 0.5).tolist()  # noqa: PLR2004
+        else:
+            df_cpu[self.pred_column] = collected_output["preds"].tolist()
+
+        return df_cpu
+
+    def process(self, batch: DocumentBatch) -> DocumentBatch:
```
praateekmahajan: The `process` seems to be generic enough, can/should this be abstracted? Also can you create a dumb README.md that shows how to add a new classifier? I always (even in dask) world get confused which classes we had to extend

ray-curator/ray_curator/stages/classifiers/aegis.py
```diff
@@ -0,0 +1,555 @@
...
+    def inputs(self) -> tuple[list[str], list[str]]:
+        return ["data"], [self.text_field]
+
+    def outputs(self) -> tuple[list[str], list[str]]:
+        return ["data"], ["_hidden_text"]
+
+    def _wrap_in_prompt(self, df: pd.DataFrame) -> pd.DataFrame:
+        documents = df[self.text_field].tolist()
+        prompts = [format_aegis(doc[: self.max_chars]) for doc in documents]
+        df["_hidden_text"] = prompts
```
praateekmahajan: instead of `_hiddent_text` what if we use `_curator_hidden_text` or something so that we don't override a column the user has coincidentally.. i understand unlikely but just to avoid namespace colision

ray-curator/ray_curator/stages/classifiers/aegis.py
```diff
@@ -0,0 +1,555 @@
...
+
+        # Collect all outputs
+        collected_output = self.collect_outputs(processed_outputs)
+
+        # Create output Pandas DataFrame
+        df_cpu = self.create_output_dataframe(df_cpu, collected_output)
+
+        # Sort by seq_order to preserve original order from tokenizer
+        if self.has_seq_order:
+            df_cpu = df_cpu.sort_values(by="_curator_seq_order", ignore_index=True).drop(
```
praateekmahajan: maybe moving _curator_seq_order to some constant somewhere 🤷

ray-curator/ray_curator/stages/classifiers/aegis.py
```diff
@@ -0,0 +1,555 @@
...
+    """
+    PostProcessResponsesStage is a stage that post-processes the responses from the AEGIS model.
+    """
+
+    pred_column: str
+    raw_pred_column: str
+    keep_raw_pred: bool
+    _name = "postprocess_responses"
+
+    def inputs(self) -> tuple[list[str], list[str]]:
```
praateekmahajan: I believe this should be an ActorStage ie. in RayData can you tag this as `IS_ACTOR_STAGE` given that it has a setup.. and similarly should the `setup_on_node` to ensure the tokenizers are downloaded before we start a stage.


Also I saw  this https://github.com/nvidia-cosmos/cosmos-xenna/blob/c62aa91db62d14dcbf20cba69cfe3247b95016a9/examples/simple_vlm_inference.py#L129-L155 

What I really liked was `local_files_only=True` during setup, this guarantees that we don't waste time during "processing"

praateekmahajan: ^ true for all classifiers

ray-curator/ray_curator/stages/classifiers/aegis.py
```diff
@@ -0,0 +1,555 @@
...
+        return ["data"], [self.pred_column] + ([self.raw_pred_column] if self.keep_raw_pred else [])
+
+    def filter_by_category(self, value: str) -> bool:
+        return value in self.filter_by
+
+    def decompose(self) -> list[ProcessingStage]:
+        stages = [
+            WrapInPromptStage(
+                text_field=self.text_field,
+                max_chars=6000,
```
praateekmahajan: configurable? and what about max_seq_len below?

ray-curator/ray_curator/stages/classifiers/fineweb_edu.py
```diff
@@ -0,0 +1,371 @@
...
+            micro_batch_size=micro_batch_size,
+            padding_side="right",
+        )
+
+        self.float_score_column = float_score_column
+        self.int_score_column = int_score_column
+        self.autocast = autocast
+
+    def inputs(self) -> tuple[list[str], list[str]]:
+        return ["data"], ["input_ids", "attention_mask"] + (["_curator_seq_order"] if self.has_seq_order else [])
```
praateekmahajan: Seems generic enough, can be abstracted or no?

ray-curator/tests/stages/test_classifiers.py
```diff
@@ -0,0 +1,595 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import pandas as pd
+import pytest
+
+from ray_curator.stages.classifiers import (
```
praateekmahajan: Let's try to maintain the same directory structure as the original code. If that means breaking up into different tests for different files lets do that.

ray-curator/ray_curator/stages/classifiers/aegis.py
```diff
@@ -0,0 +1,555 @@
+# Copyright (c) 2025, NVIDIA CORPORATION.  All rights reserved.
```
praateekmahajan, line 1: naming nit.. should these be `stages/text_models/base.py` + stages/text_models/classifiers/{aegis, domain}.py` + `stages/text_models/embedding_model.py`?

Given even SemDedup's embedding generation looks very similar to this (and it need not live inside semdedup but rather here)..

ray-curator/ray_curator/stages/classifiers/base.py
```diff
@@ -0,0 +1,479 @@
+# Copyright (c) 2025, NVIDIA CORPORATION.  All rights reserved.
```
praateekmahajan: naming nit.. should these be stages/text_models/base.py + stages/text_models/classifiers/{aegis, domain}.py + stages/text_models/embedding_model.py?

Given even SemDedup's embedding generation looks very similar to this (and it need not live inside semdedup but rather here)..

cc @ayushdg
cc @abhinavg4 @suiyoubi (since a similar situation) likely to happen in video too

i don't have a strong opinion

praateekmahajan: Irresepctive of what we decide can you seee if the current code will also be able to support the embedding generation model?

ray-curator/ray_curator/stages/classifiers/aegis.py
sarahyurick: I can double check to see if the model can be moved elsewhere. Main thing to avoid here are the CUDA issues.

ray-curator/ray_curator/stages/classifiers/aegis.py
sarahyurick: Sequence length shouldn't be configurable IMO, but I think exposing `max_chars` is reasonable.

ray-curator/tests/stages/test_classifiers.py
sarahyurick: I prefer having a single test file for the classifiers, that it also how we have it for Dask. Besides, we only have 1 test function per classifier.

ray-curator/ray_curator/stages/classifiers/base.py
sarahyurick: Sure, I can try to split up into a more generic structure.

ray-curator/tests/stages/test_classifiers.py
praateekmahajan: TBH I am not a fan of the current tests world either where we lump all the classifiers or downloaders into a single test. From a separation-of-concerns perspective, that gets messy fast. Once your test file hits 700 lines and covers logic spread across 10 different classes, it gets tough to maintain—never mind the merge conflicts and cognitive overhead for anyone touching those tests.

I do think the following pattern is much better for the reasons mentioned below
```
src/
  module1/
    foo.py
    bar.py
  module2/
    foo2.py
    bar2.py

tests/
  module1/
    test_foo.py
    test_bar.py
  module2/
    test_foo2.py
    test_bar2.py

```

1. **Maintainability**: If I need to update or refactor module1/foo.py, I instantly know where the tests live—tests/module1/test_foo.py. The 1:1 mapping cuts down on hunting, makes code reviews easier, and keeps things tidy if files are renamed or deleted.
2. **Encapsulation** :  test file is focused, covering a single module or class. That keeps tests easy to reason about, especially when debugging or reviewing changes.
1. **Scalability** :  As our library grows, we'll have more files and classes to test. One single `test_classifiers.py` file can become enormous, catch-all and hard to navigate as more features get added. Reading through it is just more cognitive overload to jump through bunch of unrelated code.

If someone has actual pros to the “all-in-one” test file approach, To zoom out a little, pandas also organizes similarly [pandas/io](https://github.com/pandas-dev/pandas/tree/main/pandas/io)  and [pandas/tests/io](https://github.com/pandas-dev/pandas/tree/main/pandas/tests/io)  instead of having one `test_io.py`

ray-curator/ray_curator/stages/classifiers/aegis.py
```diff
@@ -0,0 +1,555 @@
...
+            task_id=batch.task_id,
+            dataset_name=batch.dataset_name,
+            data=df,
+            _metadata=batch._metadata,
+            _stage_perf=batch._stage_perf,
+        )
+
+
+@dataclass
+class PostProcessResponsesStage(ProcessingStage[DocumentBatch, DocumentBatch]):
```
praateekmahajan: Should we name this something else if it's Aegis specific? Since PostProcessResponse seems generic and will be registered to our Stage Registry https://github.com/NVIDIA-NeMo/Curator/blob/efcb5002587e966b0a1c985e43e98106e59d2157/ray-curator/ray_curator/stages/base.py#L19

ray-curator/tests/stages/test_classifiers.py
sarahyurick: Idk, not really a fan of having 5-10 test files less than 100-200 lines each (and 1-3 functions each)... especially since their functionalities are so tightly coupled. All classifier implementations live in the same directory, so the mapping to a test file is still there to an extent. I/O testing is very different than testing classifiers (e.g., we are regularly updating or tweaking I/O things but we haven't added a new classifier in 6 months, and our current Dask classifier test file really only ever changes if we are updating `HF_TOKEN` related things for Aegis). The classifier tests are just about testing unchanging functionality on a small scale at this point (i.e., as long as the Hugging Face files do not change, then we do not expect these tests to change).

Just seems like a way for my name to be attributed to more files/lines of code than necessary haha. When I think about potentially needing to edit a test in the future, it seems more direct to me to be able to go to a single file for editing (again, especially since all of the classifiers share multiple underlying classes).

sarahyurick: Discussed on Slack.

ray-curator/ray_curator/stages/classifiers/aegis.py
sarahyurick: My understanding is that its useful for preventing UCX-related errors. Although this may only be relevant for Dask.

ray-curator/ray_curator/stages/classifiers/aegis.py
sarahyurick: I think I had originally edited it to be different for Aegis, but then reverted back to the underlying `process` function. I can verify that it is okay to fall back on the parent implementation.

ray-curator/ray_curator/stages/classifiers/aegis.py
sarahyurick: Sure, can rename the pre and post processing stages.

ray-curator/ray_curator/stages/classifiers/aegis.py
sarahyurick: Seems like having the extra wrapper helps, without it I get an OOM with my tests.

ray-curator/ray_curator/stages/classifiers/base.py
```diff
@@ -0,0 +1,508 @@
...
+                padding="max_length",
+                return_tensors="np",
+                truncation=True,
+                add_special_tokens=True,
+                return_token_type_ids=False,
+            )
+
+        output = df.copy()
+        output[INPUT_ID_COLUMN] = tokens.input_ids.tolist()
+        output[ATTENTION_MASK_COLUMN] = tokens.attention_mask.tolist()
```
VibhuJawa: Minor feedback here is to maintain numpy arrays inside the dataframe 


```suggestion
        output[INPUT_ID_COLUMN] = [id_ar for id_ar in tokens.input_ids]
        output[ATTENTION_MASK_COLUMN] = [mask_ar for mask_ar in tokens.attention_mask]
```

ray-curator/ray_curator/stages/classifiers/fineweb_edu.py
```diff
@@ -0,0 +1,357 @@
...
+class HFFineWebModelStage(HFModel):
+    """
+    Stage for Hugging Face model inference.
+
+    Args:
+        model_identifier: The identifier of the Hugging Face model.
+        pred_column: The name of the prediction column.
+        float_score_column: The name of the float score column.
+        int_score_column: The name of the integer score column.
+        micro_batch_size: The size of the micro-batch. Defaults to 256.
```
VibhuJawa: Why did we call this `micro_batch_size` and just not `batch_size` , what is a `micro-batch` ?

VibhuJawa (changes requested):
Mostly looks good, minor nits. Thanks for this PR, learnt a lot.

ray-curator/ray_curator/stages/classifiers/quality.py
```diff
@@ -0,0 +1,72 @@
...
+        pred_column: str = "quality_pred",
+        prob_column: str | None = None,
+        text_field: str = "text",
+        filter_by: list[str] | None = None,
+        max_chars: int = 6000,
+        sort_by_length: bool = True,
+        micro_batch_size: int = 256,
+        autocast: bool = True,
+    ):
+        self._name = QUALITY_CLASSIFIER_MODEL_IDENTIFIER.split("/")[-1].replace("-", "_").lower() + "_classifier"
```
VibhuJawa: Should  we just centralize this logic into a util ?

ray-curator/ray_curator/stages/classifiers/fineweb_edu.py
sarahyurick: `batch_size` is already a property for each `ProcessingStage` (representing the number of `DocumentBatch` objects to pass into a stage at a time), so it was a naming issue.

Although maybe `model_batch_size` or `inference_batch_size` may be clearer. I can update.

VibhuJawa: Yeah, something like `model_inference_batch_size `  i think is a better idea .   

Like, `micro_batch_size` in training workflows refers to batch size per gpu when doing distributed training, i got confused about this.

ray-curator/ray_curator/stages/classifiers/base.py
```diff
@@ -0,0 +1,508 @@
...
+
+        for model_input_batch in self.yield_next_batch(df_cpu):
+            # Forward pass
+            with torch.no_grad():
+                if self.unpack_inference_batch:
+                    outputs = self.model(**model_input_batch)
+                else:
+                    outputs = self.model(model_input_batch)
+
+            del model_input_batch
```
VibhuJawa: We should use `dataloader` here with to allow pre fetching using dataloaders. This will allow that our next batches are loaded into memory async while we infer the batch.

VibhuJawa (commented):
We should explore using a pytorch dataloader which will allow us to load the batch async into memory.

ray-curator/ray_curator/stages/classifiers/base.py
sarahyurick: Good point, thank you. I can try to explore, and depending on how it goes I may try to include it in this PR or as a follow-up. Plus I think automatic batch size tuning from https://github.com/NVIDIA-NeMo/Curator/issues/838 will come into play with this as well.

sarahyurick: Opened https://github.com/NVIDIA-NeMo/Curator/issues/877 for now, would love to get this initial PR in more quickly.

sarahyurick:
TODO:

- [x] Re-organize file structure
- [x] Maintain list versus NumPy arrays inside the resulting DataFrame
- [x] Check `setup` and `setup_on_node` for Aegis
- [x] Follow-up issue/PR for adding tests

ray-curator/ray_curator/stages/classifiers/base.py
sarahyurick: This works, but I am seeing a warning:
```
UserWarning: Creating a tensor from a list of numpy.ndarrays is extremely slow. Please consider converting the list to a single numpy.ndarray with numpy.array() before converting to a tensor. (Triggered internally at /pytorch/torch/csrc/utils/tensor_new.cpp:254.)
```

sarahyurick: I see more slowness too, for example with `tolist()` the content type classifier was 861 seconds, on the same dataset with NumPy arrays it was 967 seconds.

ray-curator/ray_curator/stages/text/models/tokenizer.py
sarahyurick, line 1: Not sure that `tokenizer.py` living under `models/` is the best way to do it. Happy to accept other suggestions for this directory name.

ray-curator/ray_curator/stages/classifiers/base.py
sarahyurick: Updated directory structure so that the base `TokenizerStage` and `ModelStage` classes live separately from the classifiers that use them.

ray-curator/ray_curator/stages/classifiers/base.py
VibhuJawa: Very interesting.   Would have expected it in the other direction . Looks good as is then. No need to change. Thanks for the test

ray-curator/ray_curator/stages/text/models/tokenizer.py
VibhuJawa: I think this is fine, models and tokenizers are pretty tied in together.

PR #764, opened 2025-07-02
Adding with_ for options in base class

abhinavg4:
## Description
This PR adds with_ to the Ray functionalities.

Tests are majorly by Cursor and can be trimmed down if needed

Certain things to take care in this implementation:
1. We are using class variables and instance variables interchangeably for BaseStage, this can create an issue if someone tries to directly edit the private variable.
Eg:-
```python
class Concrete1(ProcessingStage[EmptyTask, EmptyTask]):
    _name = "ConcreteProcessingStage"
    _batch_size = 2
    def process(self, task: EmptyTask) -> EmptyTask:
        return task

concrete1 = Concrete1()
print("concrete inst variable", concrete1.resources)
ProcessingStage._resources = Resources(cpus=2)
print("concrete inst variable", concrete1.resources)

```

would produce
```python
concrete inst variable Resources(cpus=1.0, gpu_memory_gb=0.0, nvdecs=0, nvencs=0, entire_gpu=False, gpus=0.0)
concrete inst variable Resources(cpus=2, gpu_memory_gb=0.0, nvdecs=0, nvencs=0, entire_gpu=False, gpus=0.0)
```
instead of expected
```python
concrete inst variable Resources(cpus=1.0, gpu_memory_gb=0.0, nvdecs=0, nvencs=0, entire_gpu=False, gpus=0.0)
concrete inst variable Resources(cpus=1.0, gpu_memory_gb=0.0, nvdecs=0, nvencs=0, entire_gpu=False, gpus=0.0)
```

2. We need to ensure we are doing super() inside any CompositeStage to initialize `_with_operations` 
## Checklist


ray-curator/ray_curator/stages/base.py
```diff
@@ -3,7 +3,7 @@
 from __future__ import annotations
 
 from abc import ABC, ABCMeta, abstractmethod
-from typing import TYPE_CHECKING, Any, Generic, TypeVar
+from typing import TYPE_CHECKING, Any, Generic, Self, TypeVar
```
praateekmahajan: `from typing import Self`  is 3.11+; I believe our project is 3.10+

ray-curator/tests/stages/test_base.py
```diff
@@ -0,0 +1,102 @@
...
+    @property
+    def num_items(self) -> int:
+        return len(self.data)
+
+    def validate(self) -> bool:
+        return True
+
+
+@dataclass
+class ConcreteProcessingStage(ProcessingStage[MockTask, MockTask]):
```
praateekmahajan: Your Concrete  in test case is a `dataclass` (this is related class variable vs instance, when you decorate with @dataclass those are overridden as instance variables, vs when you remove @dataclass it still works, because you have overridden the class variable)

ray-curator/tests/stages/test_base.py
```diff
@@ -0,0 +1,102 @@
...
+class TestProcessingStageWith:
+    """Test the with_ method for ProcessingStage."""
+
+    def test_all(self):
+        stage = ConcreteProcessingStage()
+        assert stage.resources.cpus == 2.0
+
+        # Test the simplified with_ method that modifies the instance directly
+        stage.with_(resources=Resources(cpus=4.0))
+        assert stage.resources.cpus == 4.0
```
praateekmahajan: For Resources equality check why not check for `Resources(cpus=4) == Resources(cpus=4)` instead of checking for attributes of it (in this case cpu). 
If equality isn't true, let's implement __eq__ on `Resources`

ray-curator/tests/stages/test_base.py
```diff
@@ -0,0 +1,102 @@
...
+        stage.with_(name=None, resources=None, batch_size=None)
+
+        # Values should remain unchanged
+        assert stage.name == original_name
+        assert stage.resources == original_resources
+        assert stage.batch_size == original_batch_size
+
+    def test_chained_with_calls(self):
+        """Test that with_ can be chained and returns self."""
+        stage = ConcreteProcessingStage()
```
praateekmahajan: Awesome that you added chaining.

ray-curator/ray_curator/stages/base.py
```diff
@@ -264,6 +277,11 @@ def decompose(self) -> list[ProcessingStage]:
             List of execution stages that will actually run
         """
 
+    def with_(self, stage_with_dict: dict[str, Any]) -> Self:
+        """We don't support with_() for composite stages."""
+        msg = "with_() is not supported for composite stages."
+        raise NotImplementedError(msg)
```
praateekmahajan: I noticed you had composite stage in the previous version, why did you remove that? Can we make implement in Composite too

praateekmahajan (changes requested):
I think this works but I have two primary concerns with it
1. We seem to be interchangeably using object variables vs instance variables. This can be buggy when the state gets mutated within different threads. We should avoid using class variables for per-instance state, especially for attributes that are expected to differ between instances or be mutated independently
3. Mutability : `with_` seems to mutate self? That's probably less ideal I'd prefer it returns a copy.

Simple repro showing how it can be buggy in a multi-threaded context.
```python
import threading
import time

class Resources:
    def __init__(self, cpus=1.0, gpus=0.0):
        self.cpus = cpus
        self.gpus = gpus

    def __repr__(self):
        return f"Resources(cpus={self.cpus}, gpus={self.gpus})"

class ProcessingStage:
    _name = "ProcessingStage"
    _resources = Resources(cpus=1.0)
    _batch_size = 1

    def __init__(self):
        pass

    @property
    def name(self):
        return self._name

    @property
    def resources(self):
        return self._resources

    @property
    def batch_size(self):
        return self._batch_size

    def with_(self, name=None, resources=None, batch_size=None):
        if name is not None:
            self._name = name
        if resources is not None:
            self._resources = resources
        if batch_size is not None:
            self._batch_size = batch_size
        return self

class Concrete1(ProcessingStage):
    _name = "ConcreteProcessingStage"
    _resources = Resources(cpus=2.0)
    _batch_size = 2

def worker(stage, worker_id):
    print(f"Worker {worker_id} sees stage name: {stage.name}")
    if worker_id == 0:
        print(f"Worker {worker_id} mutating stage name to 'MutatedByWorker0'")
        stage.with_(name="MutatedByWorker0")
    time.sleep(0.1)
    print(f"Worker {worker_id} sees stage name after mutation: {stage.name}")

stage = Concrete1()

threads = []
for i in range(2):
    t = threading.Thread(target=worker, args=(stage, i))
    t.start()
    threads.append(t)

for t in threads:
    t.join()
```

I haven't looked at the approach closely but https://stackoverflow.com/a/71868253 seems interesting where they have `AbstractClassVar` however I'm not sure if that solves our problem for non-abstract-class-vars like Resources(..) or Batch Size

abhinavg4:
Nice comment about mutation. I think both the points you mentioned can be solved by copying instead inplace substitution ?

```
import threading
import time
import copy

class Resources:
    def __init__(self, cpus=1.0, gpus=0.0):
        self.cpus = cpus
        self.gpus = gpus

    def __repr__(self):
        return f"Resources(cpus={self.cpus}, gpus={self.gpus})"

class ProcessingStage:
    _name = "ProcessingStage"
    _resources = Resources(cpus=1.0)
    _batch_size = 1

    def __init__(self):
        pass

    @property
    def name(self):
        return self._name

    @property
    def resources(self):
        return self._resources

    @property
    def batch_size(self):
        return self._batch_size

    def with_(self, name=None, resources=None, batch_size=None):
        # Create a copy of the current instance
        new_instance = copy.deepcopy(self)
        
        # Set instance-level attributes to override class-level ones
        if name is not None:
            new_instance._name = name
        if resources is not None:
            new_instance._resources = resources
        if batch_size is not None:
            new_instance._batch_size = batch_size
        
        return new_instance

class Concrete1(ProcessingStage):
    _name = "ConcreteProcessingStage"
    _resources = Resources(cpus=2.0)
    _batch_size = 2

def worker(stage, worker_id):
    print(f"Worker {worker_id} sees stage name: {stage.name}")
    if worker_id == 0:
        print(f"Worker {worker_id} mutating stage name to 'MutatedByWorker0'")
        stage = stage.with_(name="MutatedByWorker0")
    time.sleep(0.1)
    print(f"Worker {worker_id} sees stage name after mutation: {stage.name}")

stage = Concrete1()

threads = []
for i in range(2):
    t = threading.Thread(target=worker, args=(stage, i))
    t.start()
    threads.append(t)

for t in threads:
    t.join()
```

ray-curator/ray_curator/stages/base.py
abhinavg4: For the composite stage, it's non-trivial, since decomposing can initialise the stages.

So we need to ensure we store all the calls to with_ and then apply them after decompose but decompose is abstract method for us.

praateekmahajan:
TODO -
1. Add `num_workers` also as an init property
2. Add `xenna_stage_spec` and `ray_stage_spec` as init properties
3. Consider switching  `_name` <> `name` (i.e. create a property called `_name` since users will less interact with that and likely interact more with creating their own stages)
4. @abhinavg4 to expand on with an example what he means by "We need to ensure we are doing `super()` inside any `CompositeStage` to initialize _with_operations"
5. Consider having a variable called `.stages` inside `CompositeStages` that essentially is returned as-is during `decompose`. This will make our `with_` implementation in `CompositeStages` much simpler to reason about.

PR #770, opened 2025-07-04
[Ray] Add integration test to test backends for a specified pipeline

praateekmahajan:
## Description
1. Creates a conftest within `tests/backends` to reuse the same cluster as per suggestion from https://docs.ray.io/en/latest/ray-contribute/testing-tips.html
2. Creates a pipeline with Read (Composite) -> AddLength -> SplitIntoRows (fanout) -> Write stages
3. Create a single file for now called `test_integration` which implements a class `TestBackendIntegrations` which loops over the backend configurations specified at a class level. Then the class implements each different tests which parse the output tasks / output files or stdout for each of the backend.

If we want custom test unique to a particular backend, we have one example where a RayData only tests specifically parses out execution plan which exists in stdout, and ensures we don't have task fusion.

## Usage

```python
```
## Checklist


ray-curator/tests/backends/test_integration.py
```diff
@@ -0,0 +1,148 @@
...
+                assert perf_stats.process_time > 0, "Process time should be non-zero for all stages"
+            assert task._stage_perf[1].num_items_processed == task._stage_perf[2].num_items_processed, (
+                "Mismatch in number of items processed by add_length and split_into_rows"
+            )
+            # Because we split df into a single row each, the number of items processed by jsonl_writer should be only 1 i.e 1 row
+            assert task._stage_perf[3].num_items_processed == 1, (
+                "Mismatch in number of items processed by jsonl_writer"
+            )
+
+    def test_ray_data_execution_plan(self):
```
sarahyurick, line 126: Is there no equivalent test for this with `XennaExecutor`?

ray-curator/tests/backends/utils.py
```diff
@@ -0,0 +1,195 @@
...
+
+def create_test_pipeline(input_dir: Path, output_dir: Path) -> Pipeline:
+    """Create a test pipeline for integration testing."""
+    pipeline = Pipeline(
+        name="integration_test_pipeline", description="Integration test pipeline for backend comparison"
+    )
+
+    # Add JsonlReader stage
+    pipeline.add_stage(
+        JsonlReader(
```
sarahyurick, line 134: Would it be worth it to test `ParquetReader` here as well?

ray-curator/tests/backends/utils.py
```diff
@@ -0,0 +1,195 @@
...
+            file_paths=str(input_dir),
+            files_per_partition=2,  # Create multiple tasks for better testing
+            reader="pandas",
```
sarahyurick, line 137: What about PyArrow?

ray-curator/tests/backends/test_integration.py
praateekmahajan: In Xenna the logical plan is the same as execution plan. I'm sure we can add something later as we learn more about Xenna, things like which resource will a stage T be scheduled at vs Stage T+1 and such

ray-curator/tests/backends/utils.py
praateekmahajan: This being an integration test for the backends, I don't see the need to test Jsonl & Parquet. Similarly for pandas and pyarrow. We can and should have separate tests for our `Readers` where we test that. But here the goal is an integration test that does a pipeline work with our backends, and the said pipeline should cover interesting situations such as fanout etc. Another "interesting" thing we can add later is one of the stages can be a gpu stage. But that I'll leave for a subsequent PR.

ray-curator/tests/backends/utils.py
praateekmahajan: See comment above https://github.com/NVIDIA-NeMo/Curator/pull/770/files#r2190792380

sarahyurick (approved):

PR #775, opened 2025-07-09
Aot/ray video reader

suiyoubi:
## Usage

```python
```
## Checklist

### DIFF
--- a/ray-curator/pyproject.toml
+++ b/ray-curator/pyproject.toml
@@ -22,6 +22,7 @@ dependencies = [
     "transformers>=4.48.0",
     "pyarrow",
     "absl-py>=2.0.0,<3.0.0",
+    "pynvml==11.5.3"
 ]
 
 [project.optional-dependencies]
@@ -39,6 +40,11 @@ dev = [
     "ruff==0.11.4",
     "debugpy",
     "pytest"
+]
+
+video = [
+    "av==13.1.0",
+    "opencv-python",
 ]
 
 [tool.setuptools]
--- a/ray-curator/ray_curator/examples/video/video_read_example.py
+++ b/ray-curator/ray_curator/examples/video/video_read_example.py
@@ -1,0 +1,45 @@
+import argparse
+
+from ray_curator.backends.xenna import XennaExecutor
+from ray_curator.pipeline import Pipeline
+from ray_curator.stages.video.io.video_download import VideoDownloadStage
+
+
+def create_video_reading_pipeline(args: argparse.Namespace) -> Pipeline:
+
+    # Define pipeline
+    pipeline = Pipeline(name="video_reading", description="Read videos from a folder and extract metadata on video level.")
+
+    # Add stages
+    pipeline.add_stage(VideoDownloadStage(folder_path=args.video_folder, debug=args.debug))
+
+    # TODO: Add Writer stage in the following PR
+
+    return pipeline
+
+
+def main(args: argparse.Namespace) -> None:
+
+    pipeline = create_video_reading_pipeline(args)
+
+    # Print pipeline description
+    print(pipeline.describe())
+    print("\n" + "=" * 50 + "\n")
+
+    # Create executor
+    executor = XennaExecutor()
+
+    # Execute pipeline
+    print("Starting pipeline execution...")
+    pipeline.run(executor)
+
+    # Print results
+    print("\nPipeline completed!")
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
+    # General arguments
+    parser.add_argument("--debug", action="store_true", default=False, help="Run in debug mode")
+    parser.add_argument("--video-folder", type=str, required=True, help="Path to the video folder")
+    args = parser.parse_args()
+    main(args)
--- a/ray-curator/ray_curator/stages/video/io/video_download.py
+++ b/ray-curator/ray_curator/stages/video/io/video_download.py
@@ -1,0 +1,186 @@
+import os
+import pathlib
+from dataclasses import dataclass
+
+from loguru import logger
+
+from ray_curator.stages.base import ProcessingStage
+from ray_curator.tasks import Video, VideoTask, _EmptyTask
+from ray_curator.utils.file_utils import get_all_files_paths_under
+
+
+@dataclass
+class VideoDownloadStage(ProcessingStage[_EmptyTask, VideoTask]):
+    """Stage that downloads video files from storage and extracts metadata.
+
+    This class processes video files through a series of steps including downloading,
+    extracting metadata, and storing the results in the task.
+    """
+    folder_path: str = None
+    debug: bool = False
+
+    @property
+    def name(self) -> str:
+        return "video_download"
+
+    def inputs(self) -> tuple[list[str], list[str]]:
+        return [], []
+
+    def outputs(self) -> tuple[list[str], list[str]]:
+        return ["data"], []
+
+    def process(self, _: _EmptyTask) -> list[VideoTask]:
+        """Process a single group of video files.
+
+        Args:
+            task (FileGroupTask): FileGroupTask containing file paths and configuration
+
+        Returns:
+            VideoTask: VideoTask with the data from these files
+        """
+        # Get list of files
+        files = self._get_file_list()
+        logger.info(f"Found {len(files)} files")
+
+        video_tasks = []
+        for fp in files:
+            if not os.path.exists(fp):
+                msg = f"Video file {fp} does not exist"
+                raise FileNotFoundError(msg)
+
+            file_path =fp
+            if isinstance(file_path, str):
+                file_path = pathlib.Path(file_path)
+
+            video = Video(input_video=file_path)
+            video_task = VideoTask(
+                task_id=f"{file_path}_processed",
+                dataset_name=self.folder_path,
+                data=video,
+            )
+
+            # Download video bytes
+            if not self._download_video_bytes(video):
+                continue
+
+            # Extract metadata and validate video properties
+            if not self._extract_and_validate_metadata(video):
+                continue
+
+            # Log video information
+            self._log_video_info(video)
+
+            video_tasks.append(video_task)
+        return video_tasks
+
+    def _download_video_bytes(self, video: Video) -> bool:
+        """Download video bytes from storage.
+
+        Args:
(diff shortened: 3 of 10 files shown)


ray-curator/pyproject.toml
```diff
@@ -22,6 +22,7 @@ dependencies = [
     "transformers>=4.48.0",
     "pyarrow",
     "absl-py>=2.0.0,<3.0.0",
+    "pynvml==11.5.3"
```
abhinavg4: Thanks for taking care of this. Probably we need to add ray and Xenna too at some points but I think we can leave it for now

ray-curator/ray_curator/examples/video/video_read_example.py
```diff
@@ -0,0 +1,45 @@
...
+    pipeline.add_stage(VideoDownloadStage(folder_path=args.video_folder, debug=args.debug))
+
+    # TODO: Add Writer stage in the following PR
+
+    return pipeline
+
+
+def main(args: argparse.Namespace) -> None:
+
+    pipeline = create_video_reading_pipeline(args)
```
abhinavg4, line 28: NIT: Can we just pass the path and debug flag to this function instead of whole args?

ray-curator/ray_curator/stages/video/io/video_download.py
```diff
@@ -0,0 +1,186 @@
...
+    debug: bool = False
+
+    @property
+    def name(self) -> str:
+        return "video_download"
+
+    def inputs(self) -> tuple[list[str], list[str]]:
+        return [], []
+
+    def outputs(self) -> tuple[list[str], list[str]]:
```
abhinavg4, line 23: Can you please put correct second level output keys.

So basically, data is the first level and the second level would be keys inside the data dataclass.

Currently, we don't use these, but as you can imagine, we can use these guys later for optimizations. E.g.,- remove a field in case no downstream uses it.

ray-curator/ray_curator/stages/video/io/video_download.py
```diff
@@ -0,0 +1,186 @@
...
+        """Get the list of files to process."""
+        if self.folder_path is None:
+            msg = "folder_path is not set"
+            raise ValueError(msg)
+        files = get_all_files_paths_under(
+            self.folder_path,
+            recurse_subdirectories=True,
+            keep_extensions=[".mp4", ".mov", ".avi", ".mkv", ".webm"],
+        )
+        if self.debug:
```
abhinavg4: Actually, can we use num_files instead of debug? This can help run experiments at different scales.

ray-curator/ray_curator/stages/video/io/video_download.py
```diff
@@ -0,0 +1,186 @@
...
+        return [], []
+
+    def outputs(self) -> tuple[list[str], list[str]]:
+        return ["data"], []
+
+    def process(self, _: _EmptyTask) -> list[VideoTask]:
+        """Process a single group of video files.
+
+        Args:
+            task (FileGroupTask): FileGroupTask containing file paths and configuration
```
abhinavg4: This should be EmptyTask?

ray-curator/ray_curator/stages/video/io/video_download.py
```diff
@@ -0,0 +1,186 @@
...
+        Returns:
+            VideoTask: VideoTask with the data from these files
+        """
+        # Get list of files
+        files = self._get_file_list()
+        logger.info(f"Found {len(files)} files")
+
+        video_tasks = []
+        for fp in files:
+            if not os.path.exists(fp):
```
abhinavg4: This won't work with s3? Also, for local will this condition ever be true?

ray-curator/ray_curator/stages/video/io/video_download.py
```diff
@@ -0,0 +1,186 @@
...
+
+            video = Video(input_video=file_path)
+            video_task = VideoTask(
+                task_id=f"{file_path}_processed",
+                dataset_name=self.folder_path,
+                data=video,
+            )
+
+            # Download video bytes
+            if not self._download_video_bytes(video):
```
abhinavg4: This should be in a separate stage. The Video Download stage will have only a single instance since it has just one input task (EmptyTask). So, essentially, you are reading/downloading every video sequentially. 

Ideally, we should have a very lightweight stage that reads the file paths (ls stage) and then for each path, we download/read from the disk.

ray-curator/ray_curator/stages/video/io/video_download.py
```diff
@@ -0,0 +1,186 @@
...
+                continue
+
+            # Extract metadata and validate video properties
+            if not self._extract_and_validate_metadata(video):
+                continue
+
+            # Log video information
+            self._log_video_info(video)
```
abhinavg4, line 46: Should this be only if we have debug as True? Also, can this get pretty noisy? Do we need this logging? Might not be necessary? Also, this should move to the download stage.

ray-curator/ray_curator/tasks/video.py
```diff
@@ -0,0 +1,394 @@
+import os
```
abhinavg4, line 1: Didn't review this file assuming it's coming from cosmos curate

ray-curator/ray_curator/tasks/video.py
```diff
@@ -0,0 +1,394 @@
...
+
+        Returns:
+            Input path of the video.
+
+        """
+        # TODO: Support StorageClient type input
+        return self.input_video.as_posix()
+
+@dataclass
+class VideoTask(Task[Video]):
```
abhinavg4, line 343: This looks good

ray-curator/ray_curator/tasks/video.py
```diff
@@ -0,0 +1,394 @@
...
+        return 1
+
+class SplitPipeTask(Task[Video]):
+    """
+    Task for splitting a video into multiple clips.
+    """
+    data: Video = field(default_factory=Video)
+
+    @property
+    def fraction(self) -> float:
```
abhinavg4, line 368: Why do we need these functions? Can't we just do

task.data.fraction... wherever required?

ray-curator/ray_curator/utils/decoder_utils.py
```diff
@@ -0,0 +1,639 @@
+import enum
```
abhinavg4, line 1: Didn't review assuming coming from curate

ray-curator/ray_curator/utils/operation_utils.py
```diff
@@ -0,0 +1,110 @@
+"""Provide utilities for getting operation info."""
```
abhinavg4, line 1: make_pipeline_named_temporary_file

abhinavg4 (changes requested):
I think we majorly need to split this into 2 stages:

1. Get file list from local/s3. (potentially decides which files are already processed, or it can be a seperate stage, preferred)
2. Download that file and decode it (This stage is generally very intensive and where xenna shines, We will use nvencs and nvdecs for this later)

ray-curator/ray_curator/tasks/video.py
suiyoubi: we could, this is just to ease the layer of abstraction.

ray-curator/ray_curator/stages/video/io/video_download.py
suiyoubi: we need major refactor once we introduce S3.

ray-curator/ray_curator/stages/video/io/video_download.py
suiyoubi: I now use `verbose` for logging this

ray-curator/ray_curator/examples/video/video_read_example.py
suiyoubi: I think its easier to leave it like this as in the future I am going to add more stages and that would make the argument list super long

ray-curator/ray_curator/utils/operation_utils.py
suiyoubi: could you elaborate a bit

ray-curator/ray_curator/tasks/video.py
abhinavg4: Wait all this is wrong. I think it should be `self.data.fraction` and not `self.video.fraction` ?

ray-curator/ray_curator/stages/video/io/video_download.py
```diff
@@ -0,0 +1,148 @@
...
+    extracting metadata, and storing the results in the task.
+    """
+    verbose: bool = False
+
+    @property
+    def name(self) -> str:
+        return "video_download"
+
+    def inputs(self) -> tuple[list[str], list[str]]:
+        return [], []
```
abhinavg4: this should be

["data"], ["input_video"] ?

ray-curator/ray_curator/stages/video/io/video_download.py
```diff
@@ -0,0 +1,145 @@
...
+            VideoTask: VideoTask with the processed video data
+        """
+        video = task.data
+        # Download video bytes
+        if not self._download_video_bytes(video):
+            return task
+
+        # Extract metadata and validate video properties
+        if not self._extract_and_validate_metadata(video):
+            return task
```
sarahyurick, line 42: Hi, I was wondering why this is not split into 2 stages?

suiyoubi: Yeah, we could split this into a separate stage, but extracting the metadata is very fast. I think the overhead of creating a new stage might actually be larger, which could end up making the performance worse.

ray-curator/ray_curator/stages/video/io/video_reader.py
```diff
@@ -0,0 +1,55 @@
...
+    video_limit: int = -1
+    _name: str = "video_reader"
+
+    def inputs(self) -> tuple[list[str], list[str]]:
+        return [], []
+
+    def outputs(self) -> tuple[list[str], list[str]]:
+        return ["data"], ["input_video"]
+
+    def process(self, _: _EmptyTask) -> list[VideoTask]:
```
praateekmahajan, line 24: Can you add `ray_stage_spec` with the property of IS_FANOUT_STAGE?

PR #783, opened 2025-07-10
Add video splitting pipeline with fixed stride extraction and transco…

suiyoubi:
…ding stages

Note this requires https://github.com/NVIDIA-NeMo/Curator/pull/775 to be merged first (currently the base is set to `aot/;ray-video-reader` instead of `ray-api`)
- Introduced `video_split_clip_example.py` to demonstrate video splitting functionality.
- Added `ClipTranscodingStage` and `FixedStrideExtractorStage` for processing video clips.
- Implemented command-line arguments for configuring video processing parameters.
- Created utility functions for grouping iterables in `grouping.py`.
- Added unit tests for the new stages in `test_clip_transcoding_stage.py` and `test_fixed_stride_extractor_stage.py`.

## Usage

```python

python Curator/ray-curator/ray_curator/examples/video/video_split_clip_example.py --debug --verbose
```
## Checklist

### DIFF
--- a/ray-curator/ray_curator/examples/video/video_split_clip_example.py
+++ b/ray-curator/ray_curator/examples/video/video_split_clip_example.py
@@ -1,0 +1,264 @@
+import argparse
+
+from ray_curator.backends.xenna import XennaExecutor
+from ray_curator.pipeline import Pipeline
+from ray_curator.stages.video.clipping.clip_extraction_stages import ClipTranscodingStage, FixedStrideExtractorStage
+from ray_curator.stages.video.io.video_download import VideoDownloadStage
+
+
+def create_video_splitting_pipeline(args: argparse.Namespace) -> Pipeline:
+
+    # Define pipeline
+    pipeline = Pipeline(name="video_splitting", description="Split videos into clips")
+
+    # Add stages
+    pipeline.add_stage(VideoDownloadStage(folder_path=args.video_folder, debug=args.debug))
+
+    if args.splitting_algorithm == "fixed_stride":
+        pipeline.add_stage(
+            FixedStrideExtractorStage(
+                clip_len_s=args.fixed_stride_split_duration,
+                clip_stride_s=args.fixed_stride_split_duration,
+                min_clip_length_s=args.fixed_stride_min_clip_length_s,
+                limit_clips=args.limit_clips,
+            )
+        )
+    else:
+        msg = f"Splitting algorithm {args.splitting_algorithm} not supported"
+        raise ValueError(msg)
+
+    pipeline.add_stage(ClipTranscodingStage(
+        num_cpus_per_worker=args.transcode_cpus_per_worker,
+        encoder=args.transcode_encoder,
+        encoder_threads=args.transcode_encoder_threads,
+        encode_batch_size=args.transcode_ffmpeg_batch_size,
+        use_hwaccel=args.transcode_use_hwaccel,
+        use_input_bit_rate=args.transcode_use_input_video_bit_rate,
+        num_clips_per_chunk=args.clip_re_chunk_size,
+        verbose=args.verbose,
+        debug=args.debug,
+    ))
+
+    return pipeline
+
+
+def main(args: argparse.Namespace) -> None:
+
+    pipeline = create_video_splitting_pipeline(args)
+
+    # Print pipeline description
+    print(pipeline.describe())
+    print("\n" + "=" * 50 + "\n")
+
+    # Create executor
+    executor = XennaExecutor()
+
+    # Execute pipeline
+    print("Starting pipeline execution...")
+    pipeline.run(executor)
+
+    # Print results
+    print("\nPipeline completed!")
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
+    # General arguments
+    parser.add_argument("--debug", action="store_true", default=False, help="Run in debug mode")
+    parser.add_argument("--video-folder", type=str, default="/home/aot/Videos")
+    parser.add_argument("--verbose", action="store_true", default=False)
+    parser.add_argument("--output-clip-path", type=str, default="/mnt/mint/output")
+    parser.add_argument(
+        "--no-upload-clips",
+        dest="upload_clips",
+        action="store_false",
+        default=True,
+        help="Whether to upload clips to output path.",
+    )
+    parser.add_argument(
+        "--dry-run",
+        action="store_true",
+        default=False,
+        help="If set only write minimum metadata",
+    )
+
+    # Splitting parameters
+    parser.add_argument(
+        "--splitting-algorithm",
+        type=str,
+        default="fixed_stride",
+        choices=["fixed_stride", "transnetv2"],
+        help="Splitting algorithm to use",
+    )
+    parser.add_argument(
+        "--fixed-stride-split-duration",
+        type=float,
+        default=10.0,
+        help="Duration of clips (in seconds) generated from the fixed stride splitting stage.",
+    )
+    parser.add_argument(
+        "--fixed-stride-min-clip-length-s",
+        type=float,
+        default=2.0,
+        help="Minimum length of clips (in seconds) for fixed stride splitting stage.",
+    )
+    parser.add_argument(
+        "--limit-clips",
+        type=int,
+        default=0,
+        help="limit number of clips from each input video to process. 0 means no limit.",
+    )
+    parser.add_argument(
+        "--transnetv2-frame-decoder-mode",
+        type=str,
+        default="pynvc",
+        choices=["pynvc", "ffmpeg_gpu", "ffmpeg_cpu"],
+        help="Choose between ffmpeg on CPU or GPU or PyNvVideoCodec for video decode.",
+    )
+
+    # Transcoding arguments
+    parser.add_argument(
+        "--transcode-cpus-per-worker",
+        type=float,
+        default=6.0,
+        help="Number of CPU threads per worker. The stage uses a batched ffmpeg "
+        "commandline with batch_size (-transcode-ffmpeg-batch-size) of ~64 and per-batch thread count of 1.",
+    )
+    parser.add_argument(
+        "--transcode-encoder",
+        type=str,
+        default="libopenh264",
+        choices=["libopenh264", "h264_nvenc", "libx264"],
+        help="Codec for transcoding clips; None to skip transocding.",
+    )
+    parser.add_argument(
+        "--transcode-encoder-threads",
+        type=int,
+        default=1,
+        help="Number of threads per ffmpeg encoding sub-command for transcoding clips.",
+    )
+    parser.add_argument(
+        "--transcode-ffmpeg-batch-size",
+        type=int,
+        default=16,
+        help="FFMPEG batchsize for transcoding clips. Each clip/sub-command in "
+        "the batch uses --transcode-encoder-threads number of CPU threads",
+    )
+    parser.add_argument(
+        "--transcode-use-hwaccel",
+        action="store_true",
+        default=False,
+        help="Whether to use cuda acceleration for decoding in transcoding stage.",
+    )
+    parser.add_argument(
+        "--transcode-use-input-video-bit-rate",
+        action="store_true",
+        default=False,
+        help="Whether to use input video's bit rate for encoding clips.",
+    )
+    parser.add_argument(
+        "--clip-re-chunk-size",
+        type=int,
+        default=32,
+        help="Number of clips per chunk after transcoding stage.",
+    )
+
+    # Motion vector decoding arguments
+    parser.add_argument(
+        "--motion-filter",
+        choices=["disable", "enable", "score-only"],
+        default="disable",
+        help=(
+            "Control motion filtering behavior:\n"
+            "  - disable: No filtering or scoring.\n"
+            "  - enable: Automatically filter clips based on motion thresholds.\n"
+            "      (controlled by --motion-global-mean-threshold and --motion-per-patch-min-256-threshold).\n"
+            "  - score-only: Calculate motion scores without filtering clips."
+        ),
+    )
+    parser.add_argument(
+        "--motion-global-mean-threshold",
+        type=float,
+        default=0.00098,
+        help=(
+            "Threshold for global average motion magnitude. "
+            "Clips with global motion below this value may be flagged as low-motion. "
+            "Only applies when --motion-filter is set to 'enable' or 'score-only'."
+        ),
+    )
+    parser.add_argument(
+        "--motion-per-patch-min-256-threshold",
+        type=float,
+        default=0.000001,
+        help=(
+            "Threshold for minimal average motion magnitude in any 256x256-pixel patch. "
+            "Clips containing patches below this threshold may be flagged as low-motion. "
+            "Only applies when --motion-filter is set to 'enable' or 'score-only'."
+        ),
+    )
+    parser.add_argument(
+        "--motion-decode-target-fps",
+        type=float,
+        default=2.0,
+        help="Target frames per second to sample for motion vector decoding.",
+    )
+    parser.add_argument(
+        "--motion-decode-target-duration-ratio",
+        type=float,
+        default=0.5,
+        help="Target ratio of video duration to sample for motion vector decoding (0.5 = 50%%).",
+    )
+    parser.add_argument(
+        "--motion-decode-cpus-per-worker",
+        type=float,
+        default=4.0,
+        help="Number of CPUs per worker allocated to motion vector decoding.",
+    )
+    parser.add_argument(
+        "--motion-score-batch-size",
+        type=int,
+        default=64,
+        help="Batch size for motion score computation.",
+    )
+    parser.add_argument(
+        "--motion-score-gpus-per-worker",
+        type=float,
+        default=0.5,
+        help="Number of GPUs per worker allocated to motion score computation. Set to 0 to use CPU instead of GPU.",
+    )
+    parser.add_argument(
+        "--clip-extraction-target-res",
+        type=int,
+        default=-1,
+        help="Target resolution for clip extraction as (height, width). A value of -1 implies disables resize",
+    )
+
+    # Aesthetic arguments
+    parser.add_argument(
+        "--aesthetic-threshold",
+        type=float,
+        default=None,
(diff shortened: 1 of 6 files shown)


ray-curator/ray_curator/examples/video/video_split_clip_example.py
```diff
@@ -0,0 +1,265 @@
...
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
+    # General arguments
+    parser.add_argument("--video-folder", type=str, default="/home/aot/Videos")
+    parser.add_argument("--video-limit", type=int, default=-1, help="Limit the number of videos to read")
+    parser.add_argument("--verbose", action="store_true", default=False)
+    parser.add_argument("--output-clip-path", type=str, default="/mnt/mint/output")
+    parser.add_argument(
+        "--no-upload-clips",
```
abhinavg4, line 74: Is this argument used anywhere ?

ray-curator/ray_curator/examples/video/video_split_clip_example.py
```diff
@@ -0,0 +1,265 @@
...
+        dest="upload_clips",
+        action="store_false",
+        default=True,
+        help="Whether to upload clips to output path.",
+    )
+    parser.add_argument(
```
abhinavg4, line 80: What about this argument ? Is this being used anywhere ?

ray-curator/ray_curator/stages/video/clipping/clip_extraction_stages.py
```diff
@@ -0,0 +1,355 @@
...
+        if self.encoder not in {"libopenh264", "libx264", "h264_nvenc"}:
+            error_msg = f"Expected encoder of `libopenh264`, `libx264`, or `h264_nvenc`. Got {self.encoder}"
+            raise ValueError(error_msg)
+
+    @property
+    def resources(self) -> Resources:
+        """Resource requirements for this stage."""
+        if self.encoder == "h264_nvenc" or self.use_hwaccel:
+            # TODO: support partial GPU usage
+            return Resources(entire_gpu=True)
```
abhinavg4: 1. Can you please add your name to the TODO so that we know it's our TODO and not of cosmos curate.
2. I think it should not be entire_gpu instead Resouces(gpus=1). The diff is of nvencs and nvdecs. entire GPU gives you nvencs and nvdecs as well whereas gpus=1 does not

abhinavg4: Also just wondering if there are any blockers for making this partial GPU ?

ray-curator/ray_curator/stages/video/clipping/clip_extraction_stages.py
```diff
@@ -0,0 +1,355 @@
...
+                uuid=uuid.uuid5(
+                    uuid.NAMESPACE_URL,
+                    f"{file}_{start_event}_{end_event}",
+                ),
+                source_video=str(file),
+                span=span,
+            )
+            video.clips.append(clip)
+
+        logger.info(f"Extracted {len(task.data.clips)} clips from {task.data.input_video}")
```
abhinavg4, line 375: Maybe this should be logger.debug ?

abhinavg4: Or put under verbose

ray-curator/ray_curator/stages/video/clipping/clip_extraction_stages.py
```diff
@@ -0,0 +1,355 @@
...
+                dataset_name=task.dataset_name,
+                data=Video(
+                    input_video=video.input_video,
+                    metadata=video.metadata,
+                    clips=clip_chunks[idx],
+                    num_total_clips=len(video.clips),
+                    num_clip_chunks=len(clip_chunks),
+                    clip_chunk_index=idx,
+                    errors=copy.deepcopy(video.errors),
+                ),
```
abhinavg4, line 151: I think you need to also copy the stats. This function has the same error as we did in quickstart.py

ray-curator/ray_curator/stages/video/clipping/clip_extraction_stages.py
```diff
@@ -0,0 +1,355 @@
...
+    software (libx264, libopenh264) and hardware (NVENC) encoding with configurable parameters.
+    """
+    num_cpus_per_worker: float = 6.0
+    encoder: str = "libx264"
+    encoder_threads: int = 1
+    encode_batch_size: int = 16
+    nb_streams_per_gpu: int = 3
+    use_hwaccel: bool = False
+    use_input_bit_rate: bool = False
+    num_clips_per_chunk: int = 32
```
abhinavg4, line 46: Would it be possible to add some docstrings around these variables? They look like a mystery right now. And also maybe a bit better docstring for the function as well?

abhinavg4: Can we mention in the docstrong that 

This stage takes in the source bytes for the whole video and actually chunks the video into corresponding clips. It can also additionally, regroup those clips into smaller tasks.

Eg:- If a video is too long (120 clips) it can make 120/32 tasks from it....Something like this

ray-curator/ray_curator/examples/video/video_split_clip_example.py
suiyoubi: this is used when we have S3 client support, I just have it here in advance

ray-curator/ray_curator/examples/video/video_split_clip_example.py
suiyoubi: dry run is used in the clip writter that comes later

ray-curator/ray_curator/stages/video/clipping/clip_extraction_stages.py
suiyoubi: we need to use encoder and decoder for hw accelerator encoding

suiyoubi: Modify this to use the partial GPU now

ray-curator/ray_curator/stages/video/clipping/clip_extraction_stages.py
suiyoubi: used with `verbose`

abhinavg4: Is this still pending? I don't see verbose here?

ray-curator/ray_curator/utils/grouping.py
```diff
@@ -0,0 +1,88 @@
...
+"""
+
+import itertools
+import typing
+from collections.abc import Generator, Iterable
+
+T = typing.TypeVar("T")
+
+
+def split_by_chunk_size(
```
praateekmahajan, line 22: Can you add tests for these?

ray-curator/ray_curator/examples/video/video_split_clip_example.py
```diff
@@ -0,0 +1,267 @@
+import argparse
```
praateekmahajan, line 1: I think we don't want to commit these? let's check with arham and team what is the final thing? one e.g. per module is gonna be some serious bloat.. we can have an integration test if we want to "test it"

suiyoubi: If you mean this file, I actually directly use this file to integrate the rest of the modules.

suiyoubi:
/ok to test [9f39885](https://github.com/NVIDIA-NeMo/Curator/pull/783/commits/9f398851431996f39e389e0472596475b45a80c9)

suiyoubi:
\ok to test [8472134](https://github.com/NVIDIA-NeMo/Curator/pull/783/commits/8472134740be42a2d4a4a8589ebf09d0a75c0e34)

suiyoubi:
/ok to test [8472134](https://github.com/NVIDIA-NeMo/Curator/pull/783/commits/8472134740be42a2d4a4a8589ebf09d0a75c0e34)

ray-curator/ray_curator/examples/video/video_split_clip_example.py
praateekmahajan: nit: i would say as general practice lets add things when the time comes, otherwise the diff becomes hard to reason about today (and in future). SImilarly in the `Video` dataclass you have more fields than what the "currently" merged stage need. This makes it hard to prune out what's needed where.

ray-curator/ray_curator/stages/video/clipping/clip_extraction_stages.py
```diff
@@ -0,0 +1,386 @@
...
+    This stage handles the conversion of video clips using FFmpeg, supporting both
+    software (libx264, libopenh264) and hardware (NVENC) encoding with configurable parameters.
+
+    Args:
+        num_cpus_per_worker: Number of CPUs per worker.
+        encoder: Video encoder to use.
+        encoder_threads: Number of threads per encoder.
+        encode_batch_size: Number of clips to encode in parallel.
+        nb_streams_per_gpu: Number of streams per GPU.
+        use_hwaccel: Whether to use hardware acceleration.
```
praateekmahajan, line 33: It seems like even when `use_hwaccel` is False then we use `resources(gpus=1)` (at line 76). Can you elaborate what happens there?

ray-curator/ray_curator/stages/video/clipping/clip_extraction_stages.py
```diff
@@ -0,0 +1,386 @@
...
+            if self.nb_streams_per_gpu > 0:
+                # Assume that we have same type of GPUs
+                gpu_info = _get_local_gpu_info()[0]
+                nvencs = _make_gpu_resources_from_gpu_name(gpu_info.name).num_nvencs
+                gpu_memory_gb = _get_gpu_memory_gb()
+                return Resources(nvencs=nvencs // self.nb_streams_per_gpu, gpu_memory_gb=gpu_memory_gb // self.nb_streams_per_gpu)
+            else:
+                return Resources(gpus=1)
+
+        return Resources(cpus=self.num_cpus_per_worker)
```
praateekmahajan: Let's not override properties but in __post_init__ override the `self._resources`. If you override properties we'll end up with a weird case if someone does `ClipTranscodingStage.with(resources=???)`

ray-curator/ray_curator/stages/video/clipping/clip_extraction_stages.py
```diff
@@ -0,0 +1,386 @@
...
+        return {
+            RayStageSpecKeys.IS_ACTOR_STAGE: True,
+            RayStageSpecKeys.IS_FANOUT_STAGE: True,
+        }
+
+    def process(self, task: VideoTask) -> VideoTask:
+        video = task.data
+        if video.source_bytes is None:
+            msg = "Video source bytes are not available"
+            raise ValueError(msg)
```
praateekmahajan: Is this fro IDE to be happy? Because in theory the validate(..) of the stage should check against the inputs source_bytes, right? If not then we've gone somewhere wrong in our validate implementation

ray-curator/ray_curator/stages/video/clipping/clip_extraction_stages.py
```diff
@@ -0,0 +1,386 @@
...
+        # we are done with source_bytes
+        video.source_bytes = None
+
+        # TODO: log_stats
+
+        # Consider craking into smaller chunks of clips
+        output_tasks = []
+        clip_durations = [clip.duration for clip in video.clips]
+        if len(clip_durations) > 0:
+            logger.info(
```
praateekmahajan, line 125: inside verbose?

ray-curator/ray_curator/stages/video/clipping/clip_extraction_stages.py
```diff
@@ -0,0 +1,386 @@
...
+    def inputs(self) -> tuple[list[str], list[str]]:
+        return ["data"], ["source_bytes"]
+
+    def outputs(self) -> tuple[list[str], list[str]]:
+        return ["data"], []
+
+    def ray_stage_spec(self) -> dict[str, Any]:
+        """Ray stage specification for this stage."""
+        return {
+            RayStageSpecKeys.IS_ACTOR_STAGE: True,
```
praateekmahajan: fanout makes sense since we go from `X -> list[X]` but do we need this to be Actor? For general context let's use Actors when we want to maintain state or our `setup(..)` loads a model or something i.e. our init time is expensive.. If it's a simpler map style operation then we don't need it to be an Actor and you can remove that value and RayData executor will autodecide if it should be one.

FWIW RayData won't work with this stage because it has nvencs.

ray-curator/tests/utils/test_grouping.py
```diff
@@ -0,0 +1,233 @@
...
+        chunks = list(split_by_chunk_size(data, 3, drop_incomplete_chunk=True))
+        expected = [[1, 2, 3]]
+        assert chunks == expected
+
+    def test_drop_incomplete_chunk_false_explicit(self):
+        """Test keeping incomplete chunks when drop_incomplete_chunk=False."""
+        data = [1, 2, 3, 4, 5]
+        chunks = list(split_by_chunk_size(data, 3, drop_incomplete_chunk=False))
+        expected = [[1, 2, 3], [4, 5]]
+        assert chunks == expected
```
praateekmahajan, line 37: this could be combined with line 18 test..

ray-curator/tests/utils/test_grouping.py
```diff
@@ -0,0 +1,233 @@
...
+
+
+class TestSplitIntoNChunks:
+    """Test cases for split_into_n_chunks function."""
+
+    def test_basic_functionality(self):
+        """Test basic splitting into n chunks."""
+        data = [1, 2, 3, 4, 5, 6, 7, 8, 9]
+        chunks = list(split_into_n_chunks(data, 3))
```
praateekmahajan, line 107: how does this differ from functionality  from `list(split_by_chunk_size(data, 3))`)

ray-curator/ray_curator/examples/video/video_split_clip_example.py
praateekmahajan: I mean you can keep it locally but need not push.

suiyoubi: I use `video_split_clip_example.py` for the rest of all modules/stages integration.

ray-curator/ray_curator/stages/video/clipping/clip_extraction_stages.py
suiyoubi:
```    @property
    def resources(self) -> Resources:
        """Resource requirements for this stage."""
        if self.encoder == "h264_nvenc" or self.use_hwaccel:
            if self.nb_streams_per_gpu > 0:
                # Assume that we have same type of GPUs
                gpu_info = _get_local_gpu_info()[0]
                nvencs = _make_gpu_resources_from_gpu_name(gpu_info.name).num_nvencs
                gpu_memory_gb = _get_gpu_memory_gb()
                return Resources(nvencs=nvencs // self.nb_streams_per_gpu, gpu_memory_gb=gpu_memory_gb // self.nb_streams_per_gpu)
            else:
                return Resources(gpus=1)

        return Resources(cpus=self.num_cpus_per_worker)
```
if `use_hwaccel = False`, we will use CPU instead

suiyoubi: The else branch for L76 is for `if self.nb_streams_per_gpu > 0:` condition

ray-curator/ray_curator/stages/video/clipping/clip_extraction_stages.py
suiyoubi: No its just when I start implementing this I am not aware of the `validate()` stage. These could be safely removed

ray-curator/ray_curator/stages/video/clipping/clip_extraction_stages.py
suiyoubi: I honestly think we should log this without `verbose`. Since we are always processing based on clips, it is good to log the number of clips coming out from each video in a nutshell

ray-curator/ray_curator/stages/video/clipping/clip_extraction_stages.py
suiyoubi: sounds good, removed

ray-curator/tests/utils/test_grouping.py
suiyoubi: I actually don't think so. `test_incomplete_chunk_kept_by_default` test the default behviour of the function, and `test_drop_incomplete_chunk_false_explicit` test when we explictily set `drop_incomplete_chunk=False`.

I don't think they are equivalent test as if we change the default value someday in the future, we need some test to capture this.

ray-curator/tests/utils/test_grouping.py
suiyoubi: for `split_by_chunk_size`, you specify the SIZE of each chunk

for `split_into_n_chunks`, you specify the NUMBER of chunks you want

ray-curator/ray_curator/stages/video/clipping/clip_extraction_stages.py
suiyoubi: sounds good. I'll remove both `def resource` and `def name`

suiyoubi:
/ok to test [22f2ffe](https://github.com/NVIDIA-NeMo/Curator/pull/783/commits/22f2ffe8fcb36ec19a8bd0617833f478374edc4c)

suiyoubi:
/ok to test [99dd29f](https://github.com/NVIDIA-NeMo/Curator/pull/783/commits/99dd29f79547fa712fb3fbb472d6353e4f467c02)

ray-curator/ray_curator/examples/video/video_split_clip_example.py
sarahyurick: +1 I think if we are not planning to support by the 25.09 release, then it should not be included yet.

suiyoubi:
/ok to test [2291a8f](https://github.com/NVIDIA-NeMo/Curator/pull/783/commits/2291a8f8503026be7265b932724e1cb693918a74)

PR #792, opened 2025-07-10
[Ray] Use Ray Actors where viable

praateekmahajan:
## Description
1. Ray Data supports `setup_on_node`
2. Added a conftest that creates a 11 cpu cluster (2 / 3 / 6) CPUs (We only need 6.5 cpus, but Xenna / Ray Data hang if it's not ~9-10. Created a ticket here https://github.com/orgs/NVIDIA-NeMo/projects/15/views/13?sliceBy%5Bvalue%5D=Core%2FBase&pane=issue&itemId=120026772 )
3. Ray Data supports `Actor` using `IS_ACTOR_STAGE` spec.
4. For `StreamingRepartition` we rely on a Ray >= 2.47.1 because of https://github.com/ray-project/ray/pull/54469
 

## Usage

```python
```
## Checklist


ray-curator/ray_curator/backends/experimental/ray_data/adapter.py
```diff
@@ -80,53 +72,81 @@ def process_dataset(self, dataset: Dataset) -> Dataset:
             msg = "Ray Data does not support nvdecs / nvencs. Please use gpus instead."
             raise ValueError(msg)
 
-        processed_dataset = dataset.map_batches(
-            create_named_ray_data_stage_adapter(self.stage).map_batch_fn,
-            batch_size=self.batch_size,
-            num_cpus=self.stage.resources.cpus,
-            num_gpus=self.stage.resources.gpus,
-        )
+        is_actor_stage_ = self.stage.ray_stage_spec().get(RayStageSpecKeys.IS_ACTOR_STAGE, is_actor_stage(self.stage))
```
ayushdg, line 75: Nit: I like the concept but find it a bit verbose/difficult to follow. I wonder if it's easy enough for cases like these to have something like `self.stage.ray_stage_spec().is_actor_stage()` or if that leads to too much complexity.

ray-curator/ray_curator/backends/experimental/ray_data/utils.py
```diff
@@ -0,0 +1,115 @@
...
+class RayStageSpecKeys(str, Enum):
+    """String enum of different flags that define keys inside ray_stage_spec."""
+
+    IS_ACTOR_STAGE = "is_actor_stage"
+    IS_FANOUT_STAGE = "is_fanout_stage"
+
+
+def get_available_cpu_gpu_resources() -> tuple[int, int]:
+    """Get available CPU and GPU resources from Ray."""
+    available_resources = ray.available_resources()
```
ayushdg, line 23: Note: Ray is sometimes slow to update value if a previous usage just ended. Not sure if it's valid for this particular use case where the func is called at when a pipeline is executed, but if the utility is used elsewhere it can sometimes print out of date result. 
I've worked around this in the past by adding a really small sleep but it's not best solution.

ray-curator/ray_curator/backends/experimental/ray_data/utils.py
```diff
@@ -0,0 +1,115 @@
...
+
+    # Get available resources from Ray
+    available_cpus, available_gpus = get_available_cpu_gpu_resources()
+    # Calculate based on CPU and GPU requirements
+    max_cpu_actors = float("inf")
+    max_gpu_actors = float("inf")
+
+    # CPU constraint
+    if stage.resources.cpus > 0:
+        max_cpu_actors = int(available_cpus / stage.resources.cpus)
```
ayushdg: Nit: `available_cpus // stage.resources.cpus` though if you prefer the current version for readability can keep as is.

ray-curator/ray_curator/backends/experimental/ray_data/utils.py
```diff
@@ -0,0 +1,115 @@
...
+
+    # Take the minimum of CPU and GPU constraints
+    max_actors = min(max_cpu_actors, max_gpu_actors)
+    return (1, int(max_actors))
+
+
+def is_actor_stage(stage: ProcessingStage) -> bool:
+    """Check if the stage is an actor stage."""
+    overridden_setup = type(stage).setup is not ProcessingStage.setup
+    has_gpu_and_cpu = (stage.resources.gpus > 0) and (stage.resources.cpus > 0)
```
ayushdg, line 63: Nit/Question: ideally cpu=1 and gpu>1 is not allowed. Do we need the has cpu_and_gpu check?

ray-curator/ray_curator/backends/experimental/ray_data/utils.py
```diff
@@ -0,0 +1,115 @@
...
+    from ray.util.scheduling_strategies import NodeAffinitySchedulingStrategy
+
+    ray_tasks = []
+    for node in ray.nodes():
+        node_id = node["NodeID"]
+        logger.info(f"Executing setup on node {node_id} for {len(stages)} stages")
+        for stage in stages:
+            # Create NodeInfo and WorkerMetadata for this node
+            node_info = NodeInfo(node_id=node_id)
+            worker_metadata = WorkerMetadata(worker_id="", allocation=None)
```
ayushdg: Can move this to the outer for loop

ray-curator/ray_curator/backends/experimental/ray_data/utils.py
```diff
@@ -0,0 +1,115 @@
...
+
+            ray_tasks.append(
+                _setup_stage_on_node.options(
+                    num_cpus=1,
+                    scheduling_strategy=NodeAffinitySchedulingStrategy(node_id=node_id, soft=False),
+                ).remote(stage, node_info, worker_metadata)
+            )
+    ray.get(ray_tasks)
```
ayushdg, line 115: I don't recall if xenna prints logs on setup completed for node. If it does, it might be worth printing here.

ayushdg (changes requested):
This is neat! Left a small set of questions/nits.

ray-curator/ray_curator/backends/experimental/ray_data/adapter.py
praateekmahajan: Both `xenna_stage_spec` and `ray_stage_spec` are `dict[str, Any]` and I think theres value in that since in future we can make them `with`-able (i.e be able to derive a new stage w/o overriding the whole class and just overriding that value).

I'm happy to refactor this in another way but I would love to avoid "backend" logic inside the `PorcessingStage` class. Just a dict, and then the backend itself should know how to proceed.

ray-curator/ray_curator/backends/experimental/ray_data/utils.py
praateekmahajan: Good point thanks for lmk. This will happen only during pipeline's start of execute, i.e. no resources would be used at this point, so this should be _upto date hopefully_

ray-curator/ray_curator/backends/experimental/ray_data/utils.py
praateekmahajan: Maybe there is misunderstanding here, this is because map_batches(TASK) doesn't like both cpus and gpus being specified from here https://docs.ray.io/en/latest/data/api/doc/ray.data.Dataset.map_batches.html (what we had discussed on call)


image

ray-curator/ray_curator/backends/experimental/ray_data/utils.py
praateekmahajan: Added the log

ray-curator/ray_curator/backends/experimental/ray_data/adapter.py
```diff
@@ -17,8 +19,9 @@ class RayDataStageAdapter(BaseStageAdapter):
     This adapter converts stages to work with Ray Data datasets by:
     1. Working directly with Task objects (no dictionary conversion)
     2. Using Ray Data's map_batches for parallel processing
-    3. Handling single and batch processing modes
-    4. TODO: Supporting setup() and setup_on_node() calls like other backends
+        a. If stage has both gpus and cpus specified, then we use actors
+        b. If stage.setup is overridden, then we use actors
+        c. Else we use tasks
```
sarahyurick, line 24: Will we want make this overridable (e.g., with a parameter) in the future? Any chance it could depend on the stage's modality?

sarahyurick (commented):
In general PR makes a lot of sense to me. Please LMK if there are any specific portions I can help review more closely.

ray-curator/ray_curator/backends/experimental/ray_data/adapter.py
praateekmahajan: Yes, have a todo for that here https://github.com/orgs/NVIDIA-NeMo/projects/15/views/13?itemId=120026128&pane=issue

PR #797, opened 2025-07-14
Add motion filtering stages to video splitting pipeline

suiyoubi:
- Integrated `MotionVectorDecodeStage` and `MotionFilterStage` into the video splitting pipeline for enhanced motion analysis.
- Updated command-line arguments to configure motion filtering options, including GPU memory allocation and thresholds.
- Modified `Clip` class to include a type for decoded motion data.

## Usage

```python
```
## Checklist

### DIFF
--- a/ray-curator/ray_curator/examples/video/video_split_clip_example.py
+++ b/ray-curator/ray_curator/examples/video/video_split_clip_example.py
@@ -3,6 +3,7 @@ import argparse
 from ray_curator.backends.xenna import XennaExecutor
 from ray_curator.pipeline import Pipeline
 from ray_curator.stages.video.clipping.clip_extraction_stages import ClipTranscodingStage, FixedStrideExtractorStage
+from ray_curator.stages.video.filtering.motion_filter import MotionFilterStage, MotionVectorDecodeStage
 from ray_curator.stages.video.io.clip_writer import ClipWriterStage
 from ray_curator.stages.video.io.video_download import VideoDownloadStage
 from ray_curator.stages.video.io.video_reader import VideoReaderStage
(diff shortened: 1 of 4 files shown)


ray-curator/ray_curator/examples/video/video_split_clip_example.py
```diff
@@ -238,10 +255,10 @@ def main(args: argparse.Namespace) -> None:
         help="Batch size for motion score computation.",
     )
     parser.add_argument(
-        "--motion-score-gpus-per-worker",
```
abhinavg4: Can we keep this parameter the same and internally we can figure out the amount of memory. This is for the users for cosmos curate moving to Nemo Curator.

abhinavg4: Also 0 for GPU still works. So we can keep the line Set to 0 to use CPU instead of GPU__

ray-curator/ray_curator/stages/video/filtering/motion_filter.py
```diff
@@ -0,0 +1,164 @@
...
+                    )
+        video.clips = passing_clips
+
+        logger.info(
+            f"Video {video.input_video} chunk-{video.clip_chunk_index} has "
+            f"{len(video.clips)}/{len(video.filtered_clips) + len(video.clips)} clips "
+            "passed/filtered",
+        )
+
+        # TODO: free memory periodically
```
abhinavg4: Can we add your name to the todo so that it;s clear? Also why can't we just put the empty cache ?

abhinavg4 (changes requested):
Can you please change the args back to original ?

ray-curator/ray_curator/stages/video/filtering/motion_filter.py
suiyoubi: We could invoke `torch.cuda.empty_cache` after each `process` call , but it would be too excessive. I think we need some mechanism from a more base level that call this less frequent .

suiyoubi:
/ok to test [ecf1d82](https://github.com/NVIDIA-NeMo/Curator/pull/797/commits/ecf1d82bc53152b9321ebde297395f8caaa84930)

abhinavg4 (approved):
Finally got the rebase correct hahah. Looks good

PR #803, opened 2025-07-15
Add frame extraction stage to video splitting pipeline

suiyoubi:
- Integrated `ClipFrameExtractionStage` into the video splitting pipeline to support frame extraction at specified rates based on user-defined aesthetics and embeddings.
- Updated command-line argument handling to accommodate new frame extraction configurations.
- Added logic to determine target frame rates based on the presence of aesthetics and embeddings.

## Usage

```python
```
## Checklist


ray-curator/ray_curator/examples/video/video_split_clip_example.py
```diff
@@ -3,10 +3,12 @@
 from ray_curator.backends.xenna import XennaExecutor
 from ray_curator.pipeline import Pipeline
 from ray_curator.stages.video.clipping.clip_extraction_stages import ClipTranscodingStage, FixedStrideExtractorStage
+from ray_curator.stages.video.clipping.clip_frame_extraction import ClipFrameExtractionStage
 from ray_curator.stages.video.filtering.motion_filter import MotionFilterStage, MotionVectorDecodeStage
 from ray_curator.stages.video.io.clip_writer import ClipWriterStage
 from ray_curator.stages.video.io.video_download import VideoDownloadStage
 from ray_curator.stages.video.io.video_reader import VideoReaderStage
+from ray_curator.utils.decoder_utils import FrameExtractionPolicy
```
abhinavg4: decoder_utils should be part of this PR?

ray-curator/ray_curator/examples/video/video_split_clip_example.py
```diff
@@ -58,6 +60,28 @@ def create_video_splitting_pipeline(args: argparse.Namespace) -> Pipeline:
...
         ))
 
+
+    has_embeddings = args.generate_embeddings
+    has_aesthetics = args.aesthetic_threshold is not None
+    # If both aesthetics AND embeddings are needed: [1, 2] - extract frames at both 1 FPS and 2 FPS
+    # If only aesthetics is needed: [1] - extract frames at 1 FPS
+    # If only embeddings is needed: [2] - extract frames at 2 FPS
+    target_fps: list[float | int] = (
+        [1, 2] if has_aesthetics and has_embeddings else [1] if has_aesthetics else [2] if has_embeddings else []
```
abhinavg4: This seems like it should be a dataclass or an ENUM that dictates what the FPS should be for each model?

Currently, it seems like it would be super hard to implement a new embedding model for the video curator. Can we make this process a bit smoother for the user?

ray-curator/ray_curator/examples/video/video_split_clip_example.py
suiyoubi: decoder_utils is already included in previous PR

ray-curator/ray_curator/examples/video/video_split_clip_example.py
suiyoubi: I am not sure if I understand you correctly, why do you think it is "super hard" to implemnet a new embedding model ?

sarahyurick: Yeah I think it is easy enough to understand now, but I agree a `dataclass` or `Enum` would be nice to add here. Or some type of logic that makes setting `target_fps` less hardcoded than this current implementation. Like:
```
class FramePurpose(Enum):
    AESTHETICS = 1
    EMBEDDINGS = 2

purposes = []
if args.aesthetic_threshold is not None:
    purposes.append(FramePurpose.AESTHETICS)
if args.generate_embeddings:
    purposes.append(FramePurpose.EMBEDDINGS)

target_fps = [p.value for p in purposes]
```

Also, in addition to letting the user set `target_fps` directly, there could be another parameter called `extract_purpose: Literal["aesthetics", "embeddings"] | list[str]` that bakes the above logic into `ClipFrameExtractionStage`?

ray-curator/ray_curator/stages/video/clipping/clip_frame_extraction.py
```diff
@@ -0,0 +1,121 @@
...
+    def resources(self) -> Resources:
+        return Resources(cpus=self.num_cpus)
+
+    def lcm_multiple(self, fps: list[float | int]) -> float | int:
+        """Compute LCM of a list of fps targets."""
+
+        def lcm(a: float, b: float) -> float | int:
+            return abs(a * b) // math.gcd(int(a), int(b))
+
+        return reduce(lcm, fps)
```
sarahyurick: Can this be replaced with `math.lcm(*fps)`? Maybe with an additional check to convert it to a list of integers as needed?

Or is a float like 2.5 valid here? Wouldn't we always want to round it first anyway when computing this?

abhinavg4 (approved):
Looks good modulo ENUM stuff

ray-curator/ray_curator/examples/video/video_split_clip_example.py
suiyoubi: Thanks I've changed to use Enum instead

ray-curator/ray_curator/stages/video/clipping/clip_frame_extraction.py
suiyoubi: sure I can use math.lcm and round it before passing into that

suiyoubi:
/ok to test [65a70c7](https://github.com/NVIDIA-NeMo/Curator/pull/803/commits/65a70c72146c1610a5f7bea322756f7beea2dec6)

PR #834, opened 2025-07-20
Adding get client feature

abhinavg4:
## Description
This PR adds a get client function.

It sets up prometheus and graphana for the user and also allows customizing almost any parameter for Ray cluster.
## Usage
`get_client()`
```python
```
## Checklist


ray-curator/ray_curator/core/client.py
```diff
@@ -0,0 +1,365 @@
...
+    DEFAULT_NEMO_CURATOR_METRICS_PATH,
+    DEFAULT_PROMETHEUS_WEB_PORT,
+    DEFAULT_RAY_DASHBOARD_HOST,
+    DEFAULT_RAY_DASHBOARD_PORT,
+    DEFAULT_RAY_METRICS_PORT,
+    DEFAULT_RAY_PORT,
+    DEFAULT_RAY_TEMP_DIR,
+    GRAPHANA_DASHBOARD_YAML_TEMPLATE,
+    GRAPHANA_DATASOURCE_YAML_TEMPLATE,
+    GRAPHANA_INI_TEMPLATE,
```
sarahyurick: Is it Graphana or Grafana?

ray-curator/ray_curator/core/client.py
```diff
@@ -0,0 +1,365 @@
...
+    ray_port: int = DEFAULT_RAY_PORT,
+    ray_dashboard_port: int = DEFAULT_RAY_DASHBOARD_PORT,
+    ray_temp_dir: str = DEFAULT_RAY_TEMP_DIR,
+    prometheus_web_port: int = DEFAULT_PROMETHEUS_WEB_PORT,
+    grafana_web_port: int = DEFAULT_GRAFANA_WEB_PORT,
+    ray_metrics_port: int = DEFAULT_RAY_METRICS_PORT,
+    ray_dashboard_host: str = DEFAULT_RAY_DASHBOARD_HOST,
+    num_gpus: int | None = None,
+    num_cpus: int | None = None,
+    enable_object_spilling: bool = False,
```
ayushdg: Do we want to expose kwargs here to also pass in kwargs to ray init?

ray-curator/ray_curator/core/client.py
```diff
@@ -0,0 +1,365 @@
...
+        prometheus_web_port = get_next_free_port(prometheus_web_port)
+        grafana_web_port = get_next_free_port(grafana_web_port)
+
+        return_dict["prometheus_web_port"] = prometheus_web_port
+        return_dict["grafana_web_port"] = grafana_web_port
+
+        downloaded, file_name = download_prometheus()
+        if not downloaded:
+            logger.error("Failed to download Prometheus.")
+            sys.exit(1)
```
ayushdg: Should Curator be responsible for downloading prometheus? Maybe it's a requirement for users to do instead?

ray-curator/ray_curator/core/client.py
```diff
@@ -0,0 +1,365 @@
...
+                    service_discovery_path=os.path.join(ray_temp_dir, "prom_metrics_service_discovery.json")
+                )
+            )
+
+        prometheus_cmd = [
+            f"{prometheus_dir}/prometheus",
+            "--config.file",
+            str(prometheus_config_path),
+            "--web.enable-lifecycle",
+            f"--web.listen-address=:{prometheus_web_port}",
```
ayushdg: might be cleaner to write a separate method responsible for this logic. Similar to `download_prometheus()` something like `start_prometheus()`?

ray-curator/ray_curator/core/client.py
```diff
@@ -0,0 +1,365 @@
...
+        # Grafana setup and launch
+        # -----------------------------
+        try:
+            # Determine download URL based on architecture
+            arch = platform.machine()
+            if arch not in ("x86_64", "amd64"):
+                logger.warning(
+                    "Automatic Grafana installation is only tested on x86_64/amd64 architectures. "
+                    "Please install Grafana manually if the following steps fail."
+                )
```
ayushdg: Similar comment as above

ray-curator/ray_curator/core/client.py
```diff
@@ -0,0 +1,365 @@
...
+
+            grafana_version = "12.0.2"
```
ayushdg: Should this go to constants?

ray-curator/ray_curator/core/client.py
```diff
@@ -0,0 +1,365 @@
...
+            logger.error(f"Failed to setup or start Grafana: {error}")
+
+        # Depending upon the executor, we need to start the ray dashboard.
+
+    if num_gpus is not None:
+        if executor != "Xenna":
+            msg = "num_gpus is only supported for Xenna. For other executors, use CUDA_VISIBLE_DEVICES"
+            raise ValueError(
+                msg,
+            )
```
ayushdg: Can you elaborate?

ray-curator/ray_curator/core/client.py
```diff
@@ -0,0 +1,365 @@
...
+            num_gpus,
+            num_gpus,
+        )
+
+    if os.environ.get("RAY_ADDRESS"):
+        logger.info("Ray is already running. Skipping the setup.")
+    else:
+        ray_dashboard_port = get_next_free_port(ray_dashboard_port)
+        ray_metrics_port = get_next_free_port(ray_metrics_port)
+        ray_port = get_next_free_port(ray_port)
```
ayushdg: This should reuse the args provided from the spec/function args

ray-curator/ray_curator/core/utils.py
```diff
@@ -0,0 +1,134 @@
...
+
+        # We need to set this env var to avoid ray from setting CUDA_VISIBLE_DEVICES.
+        # We set these manually in Xenna because we allocate the gpus manually instead of relying on ray's mechanisms.
+        # This will *only* get picked up from here if the cluster is started from this script. In the case of previously
+        # existing clusters, this needs to be set in the processes that set up the cluster.
+        os.environ["RAY_EXPERIMENTAL_NOSET_CUDA_VISIBLE_DEVICES"] = "0"
+        # These need to be set to allow listing debug info about more than 10k actors.
+        os.environ["RAY_MAX_LIMIT_FROM_API_SERVER"] = str(API_LIMIT)
+        os.environ["RAY_MAX_LIMIT_FROM_DATA_SOURCE"] = str(API_LIMIT)
+        os.environ["XENNA_RAY_METRICS_PORT"] = str(ray_metrics_port)
```
ayushdg: There may be a possibility of moving some of these args like `RAY_EXPERIMENTAL_NOSET_CUDA_VISIBLE_DEVICES` into actor/stage runtime_env args in the xenna wrapper or executor so that the `get_client` call is executor agnostic.

ray-curator/ray_curator/core/utils.py
```diff
@@ -0,0 +1,134 @@
...
+    ray_command.extend(["--metrics-export-port", str(ray_metrics_port)])
+    ray_command.extend(["--dashboard-host", ray_dashboard_host])
+    ray_command.extend(["--dashboard-port", str(ray_dashboard_port)])
+    ray_command.extend(["--temp-dir", ray_temp_dir])
+    ray_command.extend(["--disable-usage-stats"])
+    if enable_object_spilling:
+        ray_command.extend(
+            [
+                "--system-config",
+                '{"local_fs_capacity_threshold": 0.95, "object_spilling_config": "{ "type": "filesystem", "params": {"directory_path": "/tmp/ray_spill", "buffer_size": 1000000 } }"}',
```
ayushdg: Can we not use `ray.init()` instead of starting via CLI?

ray-curator/ray_curator/core/client.py
abhinavg4: We need to use ray start so there is no 1-1 mapping of the python and the command line arguments. I can use something like onto_slurm for making RayConfig and we can add stuff to it if required .

ray-curator/ray_curator/core/client.py
abhinavg4: Hmmm, we can have a parameter, but I think in the spirit of making things e2e, I would prefer we do download

ray-curator/ray_curator/core/client.py
abhinavg4: 100%

ray-curator/ray_curator/core/client.py
abhinavg4: Yes

ray-curator/ray_curator/core/client.py
abhinavg4: Yup

ray-curator/ray_curator/core/client.py
abhinavg4: This will change

ray-curator/ray_curator/core/utils.py
abhinavg4: I initially had ray init only but I think it does not support some of the features we need. Same was observed with cosmos curate:

https://github.com/nvidia-cosmos/cosmos-curate/blob/21a1f6b7414a7f3ab3ab7b9369689543e47bb85e/cosmos_curate/scripts/onto_slurm.py#L23

ray-curator/tests/test_get_ray_client.py
```diff
@@ -0,0 +1,106 @@
...
+        prometheus_port = client_info["prometheus_web_port"] or 9090  # Fallback if already running
+
+        session_name = _latest_session_name(ray_tmp)
+        assert _wait_for_metric(session_name, prometheus_port), (
+            f"Metric for session {session_name} not found in Prometheus on port {prometheus_port}."
+        )
+
+    # TODO: This will kill all the ray processes. Find a better way to do this.
+    # Cleanup : stop any Ray processes we started
+    subprocess.run(["ray", "stop", "--force"], check=False, capture_output=True)  # noqa: S603, S607
```
praateekmahajan: yup this todo might be important.. would hate for it to kill my existing job if i end up running it on my cluster while i have the job.. i had to hack it in conftest  in this PR too https://github.com/NVIDIA-NeMo/Curator/pull/844

praateekmahajan (changes requested):
Took a high level look it mostly looks good.. here are some nits
1. `get_client` by default could only get client. and we grafana prometheus behind a boolean which is false.
2. let's refactor it such that grafana prometheus are behind functions so `get_client` is super trivial to follow.. at the end of the day get_client should get client and everything else is nice to have
3. instead of core move it to core/client; similar for test let's follow the same nested structure instead of parent level
4. I love your test for `test_two_clusters_metrics_visible`. If we can add similarly more tests that'll be great too!

ray-curator/ray_curator/core/client.py
abhinavg4: For now I will make a ticket for adding config. Similar to https://github.com/nvidia-cosmos/cosmos-curate/blob/21a1f6b7414a7f3ab3ab7b9369689543e47bb85e/cosmos_curate/scripts/onto_slurm.py#L172

abhinavg4: https://github.com/orgs/NVIDIA-NeMo/projects/15/views/13?sliceBy%5Bvalue%5D=Core%2FBase&pane=issue&itemId=123675311

ray-curator/ray_curator/core/client.py
abhinavg4: Added a param which is default to True. Open to making it False as default

ray-curator/ray_curator/core/client.py
abhinavg4: Can you please elaborate on this?

ray-curator/tests/test_get_ray_client.py
abhinavg4: Hmm, any good solution for this? In the case of getting a client, the ray init happens in a subprocess, so it's not straightforward to do a shutdown on it?

abhinavg4:
@praateekmahajan 

1. I added a boolean, it's True by default right now but open to making it False.
2. Yeah, refactored it that way.
3. Hmmm, it's already inside a client file. Do you mean put the 2 files inside a folder? I was hoping that Constant.py and util.py will be same as we grow core more. As for test, yes let me nest it.
4. Any tests that come to mind?

ray-curator/ray_curator/core/client.py
```diff
@@ -0,0 +1,156 @@
...
+    ray_metrics_port: int = DEFAULT_RAY_METRICS_PORT,
+    ray_dashboard_host: str = DEFAULT_RAY_DASHBOARD_HOST,
+    num_gpus: int | None = None,
+    num_cpus: int | None = None,
+    enable_object_spilling: bool = False,
+) -> dict:
+    """
+    This function is used to setup the ray cluster and the metrics dashboard.
+    It does this by first checking if the prometheus and grafana are running.
+    If they are not running, it downloads the prometheus and grafana and starts them.
```
sarahyurick:
```suggestion
    It does this by first checking if prometheus and grafana are running.
    If they are not running, it downloads prometheus and grafana and starts them.
```

ray-curator/ray_curator/core/client.py
```diff
@@ -0,0 +1,156 @@
...
+
+            downloaded, file_name = download_prometheus()
+            if not downloaded:
+                error_msg = "Failed to download Prometheus."
+                logger.error(error_msg)
+                raise RuntimeError(error_msg)
+
+            file_path = move_and_extract_prometheus(file_name)
+
+            prometheus_dir = file_path.rstrip(".tar.gz")  # noqa: B005
```
sarahyurick: Why wouldn't we have the user download Prometheus and Grafana themselves? This function is a bit overloaded IMO.

sarahyurick: From my understanding, Prometheus and Grafana should not be strict requirements here... we should have more flexibility by instructing the user to download it themselves.

abhinavg4: Hmmm, I just wanted to make sure we are supporting end-to-end use case. 

> 
This function is a bit overloaded IMO.

I agree. Especially since this will get more overloaded overtime as we add support for multinode. I have added a flag to disable the Prometheus and Grafana setup. Can I make that False by default?  I can factor out the full Prometheus and Grafana setup login out and just add a function too, but would wanna keep the loginc e2e here.

ray-curator/tests/test_get_ray_client.py
praateekmahajan: You can see how I try to do in conftest. Here is the issue I opened, some users mentioned some dirty tricks, in case you want to explore those https://github.com/ray-project/ray/issues/54989

ray-curator/ray_curator/core/client.py
sarahyurick: I meant why is the current code path for the dashboard like:
```
if is_prometheus_running() or is_grafana_running():
    # Skip setup, etc.
else:
   download_prometheus()
   # Further down
   download_grafana()
```
? Is there not a way we can verify if they are downloaded already, and if not print an error/warning with the appropriate instructions to download?

ray-curator/scripts/xenna_grafana_dashboard.json
VibhuJawa, line 1: Would move it to something which is not scripts. 

```
ray-curator/monitoring/grafana/xenna_dashboard.json
```

ray-curator/ray_curator/core/client.py
VibhuJawa: +1, Do not think we can download softwares on the fly. This will probably raise a lot of security concerns.   The minimum requirement is you have to prompt the user to confirm download. 

We will also need to do a bunch of  stuff like checksum verification, GPG signature checks, stricter _safe_extract etc.  Strong suggestion of moving this out. 

While I agree with the sentiment of this of allowing folks to launch prometheus seemlessly, do not think we can go with this approach as is.  Another approach will be  to start a docker's python sdk and launching using that.  I think this requires more design thought and do not think will be able to make it in this release .

ray-curator/ray_curator/core/client.py
```diff
@@ -0,0 +1,156 @@
...
+                logger.error(error_msg)
+                raise
+
+            # -----------------------------
+            # Grafana setup and launch
+            # -----------------------------
+            try:
+                grafana_dir = download_grafana()
+                grafana_ini_path = write_grafana_configs(grafana_web_port, prometheus_web_port)
+                launch_grafana(grafana_dir, grafana_ini_path)
```
VibhuJawa: Same here. Do not think we can download this on the fly .

ray-curator/ray_curator/core/client.py
```diff
@@ -0,0 +1,156 @@
...
+                time.sleep(2)
+            except Exception as error:
+                error_msg = f"Failed to setup or start Grafana: {error}"
+                logger.error(error_msg)
+                raise
+
+    if os.environ.get("RAY_ADDRESS"):
+        logger.info("Ray is already running. Skipping the setup.")
+    else:
+        ray_dashboard_port = get_next_free_port(ray_dashboard_port)
```
VibhuJawa: If default port then we should get the next free port but if the user provided the port this should error out.

ray-curator/ray_curator/core/utils.py
VibhuJawa: Can we file and link an issue to ensure  to ray , we can work with them on adding the missing features on ray.init please.

ray-curator/ray_curator/core/utils.py
```diff
@@ -0,0 +1,378 @@
...
+
+
+def _logger_custom_deserializer(
+    _: None,
+) -> "loguru.Logger":
+    # Initialize a default logger
+    return logger
+
+
+def init_or_connect_to_cluster(  # noqa: PLR0913
```
VibhuJawa: Could you point me to where we  connect to the cluster here, seems like we only start it using CLI ?

ray-curator/ray_curator/core/client.py
```diff
@@ -0,0 +1,144 @@
...
+        include_dashboard: Whether to include dashboard integration. If true, adds Ray metrics service discovery.
+        ray_metrics_port: The port number of the Ray metrics.
+        ray_dashboard_host: The host of the Ray dashboard.
+        num_gpus: The number of GPUs to use.
+        num_cpus: The number of CPUs to use.
+        enable_object_spilling: Whether to enable object spilling.
+
+    Note:
+        To start monitoring services (Prometheus and Grafana), use the standalone
+        start_prometheus_grafana.py script separately.
```
sarahyurick, line 62: Does this script exist yet?

ray-curator/ray_curator/core/client.py
```diff
@@ -0,0 +1,144 @@
...
+
+    def stop(self) -> None:
+        if self.ray_process:
+            self.ray_process.kill()
+            self.ray_process.wait()
+            # Current there is no good way of stopping a particular ray cluster. https://github.com/ray-project/ray/issues/54989
+            # We kill the ray gcs process to stop the cluster. but still we have some ray processes running.
+            msg = "Nemo Curator has stopped the ray cluster it started by killing the ray gcs process. "
+            msg += "It is advised to wait for a few seconds before running any ray commands to ensure ray can clenup other processes."
+            msg += "If you are seeing any ray commands like `ray status` failing, please ensure /tmp/ray/ray_current_cluster has correct information."
```
sarahyurick:
```suggestion
            # Currently there is no good way of stopping a particular Ray cluster. https://github.com/ray-project/ray/issues/54989
            # We kill the Ray GCS process to stop the cluster, but still we have some Ray processes running.
            msg = "NeMo Curator has stopped the Ray cluster it started by killing the Ray GCS process. "
            msg += "It is advised to wait for a few seconds before running any Ray commands to ensure Ray can cleanup other processes."
            msg += "If you are seeing any Ray commands like `ray status` failing, please ensure /tmp/ray/ray_current_cluster has correct information."
```

ray-curator/ray_curator/core/client.py
abhinavg4: Yes. I have it. Will make a PR.

ray-curator/ray_curator/core/client.py
```diff
@@ -0,0 +1,144 @@
...
+            )
+            # Register atexit handler only when we have a ray process
+            atexit.register(self.stop)
+
+    def stop(self) -> None:
+        if self.ray_process:
+            self.ray_process.kill()
+            self.ray_process.wait()
+            # Current there is no good way of stopping a particular ray cluster. https://github.com/ray-project/ray/issues/54989
+            # We kill the ray gcs process to stop the cluster. but still we have some ray processes running.
```
sarahyurick: Yeah as I was testing https://github.com/NVIDIA-NeMo/Curator/pull/920 I noticed that there were still a lot of lingering Ray processes. I do not think this should block us merging, but maybe we should open a ticket to eventually find a better way?

ray-curator/ray_curator/core/utils.py
```diff
@@ -0,0 +1,119 @@
...
+    ray_port: int,
+    ray_temp_dir: str,
+    ray_dashboard_port: int,
+    ray_metrics_port: int,
+    ray_client_server_port: int,  # noqa: ARG001
+    ray_dashboard_host: str,
+    num_gpus: int | None = None,
+    num_cpus: int | None = None,
+    enable_object_spilling: bool = False,
+    block: bool = True,
```
VibhuJawa, line 77: I dont think we should support non blocking mode today, lets remove this option for now please as if non blocking we can not kill it

ray-curator/tests/core/test_get_ray_client.py
```diff
@@ -0,0 +1,63 @@
...
+    with tempfile.TemporaryDirectory(prefix="ray_test_single_") as ray_tmp:
+        # Clear the environment variable RAY_ADDRESS
+        os.environ.pop("RAY_ADDRESS", None)
+        client = RayClient(ray_temp_dir=ray_tmp)
+        client.start()
+        time.sleep(10)  # Wait for ray to start.
+        with open(os.path.join(ray_tmp, "ray_current_cluster")) as f:
+            assert f.read() == f"127.0.0.1:{client.ray_port}"
+        client.stop()
+
```
VibhuJawa, line 49: Can you add a asset here to verify stop actually worked ?

VibhuJawa (changes requested):
Mostly looks good, requested some minor changes

ray-curator/tests/core/test_get_ray_client.py
abhinavg4: Hmm there is no easy way to verify that the stop actually worked.

ray-curator/ray_curator/core/utils.py
abhinavg4: I think once ray allows to stop a sepcific server this might be used. I can add an assert that block is always true.

ray-curator/ray_curator/core/client.py
abhinavg4: HI yes, if you wait they should eventually get killed.

abhinavg4: I'll make a ticket of this.

ray-curator/ray_curator/core/utils.py
abhinavg4: Ok we are not passing `block` as a param from get_client class (which means it's not an option). But incase the user wants to use this util without `block`, I don't see a reason why they should not be able to ?

VibhuJawa (approved):
LGTM. Thanks for all the changes

abhinavg4:
/ok to test [2bdd0b0](https://github.com/NVIDIA-NeMo/Curator/pull/834/commits/2bdd0b04ecd7980bb4b7a350b0c49d00a6a6c067)

PR #835, opened 2025-07-20
Adding function decorator for very simple functions to be converted into stages

abhinavg4:
## Description
Adding function decorator for very simple functions to be converted into stages

## Usage
See the example

## Checklist


ray-curator/ray_curator/stages/function_decorators.py
```diff
@@ -0,0 +1,120 @@
...
+from ray_curator.stages.resources import Resources
+from ray_curator.stages.function_decorators import processing_stage
+
+
+@processing_stage(name="WordCountStage", resources=Resources(cpus=1.0), batch_size=1)
+def word_count(task: SampleTask) -> SampleTask:
+    # Add a *word_count* column to the task's dataframe
+    task.data["word_count"] = task.data["sentence"].str.split().str.len()
+    return task
+```
```
sarahyurick: In the case of a `DocumentBatch`, I was wondering if we could also support a `pd.DataFrame -> pd.DataFrame` or `pa.Table -> pa.Table` approach as well? I was thinking something like:

```
@processing_stage(...)
def do_a_bunch_of_pd_ops(df: pd.DataFrame, other_param: ...) -> pd.DataFrame:
   df = df.sort_values(...)
   df["new_column"] = ...
   return df
```
this would really allow the user to think in a DataFrame mindset without having to know anything about Curator's data types. WDYT?

Also, for a global operation like sorting, do we have any plans for how to enable it across entire batches instead of just within batches?

ray-curator/ray_curator/stages/function_decorators.py
```diff
@@ -0,0 +1,120 @@
...
+
+The variable ``word_count`` now holds an *instance* of a concrete
+``ProcessingStage`` subclass that can be added directly to a
+:class:`ray_curator.pipeline.Pipeline`.
+"""
```
praateekmahajan, line 53: Can you add an e.g. here in the comment too for how that pipeline looks like

ray-curator/ray_curator/stages/function_decorators.py
```diff
@@ -0,0 +1,120 @@
...
+    name: str,
+    resources: Resources | None = None,
+    batch_size: int | None = None,
+) -> Callable[[Callable[[TIn], TOut | list[TOut]]], ProcessingStage[TIn, TOut]]: ...
+
+
+def processing_stage(
+    *,
+    name: str,
+    resources: Resources | None = None,
```
praateekmahajan: let's also allow `dict[str, float]` so that users don't have to import resources, and they can just do 
```
@processing_stage(name="WordCountStage", resources={"cpus" : 1}, batch_size=1)
def word_count(...)
```

ray-curator/ray_curator/stages/function_decorators.py
```diff
@@ -0,0 +1,120 @@
...
+                return cast("TOut | list[TOut]", self._fn(task))
+
+            # The user requested to “not worry about inputs/outputs”, so we leave
+            # them as the base-class defaults (empty lists).
+
+        # Give the dynamically-created class a *nice* __name__ so that logs and
+        # error messages are meaningful.  We purposefully use the *stage* name
+        # instead of the function name to avoid confusion.
+        _FunctionProcessingStage.__name__ = name
+        _FunctionProcessingStage.__qualname__ = name
```
praateekmahajan, line 154: Awesome!

ray-curator/ray_curator/stages/function_decorators.py
praateekmahajan: Also let's support `ray_stage_spec` and `xenna_stage_spec`. Or to support that do we need to do this https://github.com/orgs/NVIDIA-NeMo/projects/15/views/13?sliceBy%5Bvalue%5D=Core%2FBase&pane=issue&itemId=120026128

praateekmahajan: Similarly can we provide the user the flexibility to add inputs outputs too?

ray-curator/ray_curator/stages/function_decorators.py
```diff
@@ -0,0 +1,120 @@
...
+            _resources: Resources = resources  # type: ignore[assignment]
+            _batch_size: int | None = batch_size  # type: ignore[assignment]
+
+            # Keep a reference to the original function for introspection /
+            # debugging.
+            _fn: Callable[[TIn], TOut | list[TOut]] = staticmethod(func)  # type: ignore[assignment]
+
+            def process(self, task: TIn) -> TOut | list[TOut]:  # type: ignore[override]
+                # Delegate to the wrapped function.
+                return cast("TOut | list[TOut]", self._fn(task))
```
praateekmahajan: Thinking out loud, I haven't fully debugged the root cause of it but in fan out stages we must copy the _perf_stats. See https://github.com/NVIDIA-NeMo/Curator/pull/738#discussion_r2205687315

Should we leave it to the user to do that or can we do that here
```
out = cast("TOut | list[TOut]", self._fn(task))
out._metadata = out._metadata.copy()
out._perf_stats = out._perf_stats.copy()
```

ray-curator/tests/stages/test_function_decorators.py
```diff
@@ -0,0 +1,98 @@
...
+# -----------------------------------------------------------------------------
+# Helper stages created via the decorator
+# -----------------------------------------------------------------------------
+
+# A stage that increments the task's integer payload
+resources_inc = Resources(cpus=1.5)
+
+
+@processing_stage(name="IncrementStage", resources=resources_inc, batch_size=4)
+def _increment(task: MockTask) -> MockTask:
```
praateekmahajan: super nit, can you call these `increment_stage`, `duplicate_stage` and just use like that in subsequent tests

ray-curator/tests/stages/test_function_decorators.py
```diff
@@ -0,0 +1,98 @@
...
+
+        stage = _duplicate
+        task = MockTask(value=42)
+        result = stage.process(task)
+        assert isinstance(result, list)
+        assert len(result) == 2
+        # All returned objects should be MockTask instances pointing at the same task
+        assert all(isinstance(t, MockTask) for t in result)
+        assert all(t is task for t in result)
+
```
praateekmahajan, line 119: Let's add a test to see that even inherited `process_batch` works as expected.
Secondly, let's add / replace an existing stage with one decorated processing stage to our integration test too, which checks it works across backends in a pipeline.

praateekmahajan (changes requested):
Great work on this, suggested some enhanacements.

ray-curator/tests/stages/test_function_decorators.py
```diff
@@ -0,0 +1,98 @@
...
+    def test_invalid_signature_raises(self) -> None:
+        """Functions with an invalid signature should raise *ValueError* when
+        decorated.
+        """
+
+        with pytest.raises(ValueError):  # noqa: PT011
+
+            @processing_stage(name="BadStage")
+            def _bad(task: MockTask, _: int):  # type: ignore[valid-type]  # noqa: ANN202
+                return task
```
praateekmahajan, line 129: Also lets add a test that we can see our new ProcessingStage inside `STAGE_RESGISTRY` which might help us in future when we want to parse yamls and create stages

https://github.com/NVIDIA-NeMo/Curator/blob/efcb5002587e966b0a1c985e43e98106e59d2157/ray-curator/ray_curator/stages/base.py#L19

ray-curator/ray_curator/stages/function_decorators.py
sarahyurick: I don't think this fully works... it just names them `_FunctionProcessingStage` in the stage registry.

ray-curator/ray_curator/stages/function_decorators.py
sarahyurick: I like doing it automatically for the user, unless there is some reason to think/expect the user to do it themselves.

I will add your suggested code snippet for now, and if sometime down the road we want to give flexibility to the user, then we can adjust accordingly.

ray-curator/tests/stages/test_function_decorators.py
```diff
@@ -0,0 +1,117 @@
...
+# limitations under the License.
+
+import pytest
+
+from ray_curator.stages.base import ProcessingStage, get_stage_class
+from ray_curator.stages.function_decorators import processing_stage
+from ray_curator.stages.resources import Resources
+from ray_curator.tasks import Task
+
+# TODO: Add tests for ray pipelines too
```
sarahyurick: TODO...

ray-curator/ray_curator/stages/function_decorators.py
sarahyurick: I like the suggestion for resources, will add.

IMO `ray_stage_spec`, `xenna_stage_spec`, `inputs`, and `outputs` is a bit overkill. If the user wants control over those things, they can implement a `ProcessingStage` directly instead of using the decorator. Does this sound reasonable?

ray-curator/ray_curator/stages/function_decorators.py
sarahyurick: We can do this as a future task. I would rather focus on getting this PR in ASAP.

ray-curator/tests/stages/test_function_decorators.py
sarahyurick: Not really sure what you are asking for with the second request. Adding `process_batch` tests.

praateekmahajan (approved):
LGTM! Thanks for accommodating the changes

PR #836, opened 2025-07-21
Update Authors

thomasdhc:
## Usage

```python
```
## Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -21,7 +21,7 @@ name = "nemo_curator"
 description = "Scalable Data Preprocessing Tool for Training Large Language Models"
 readme = { file = "README.md", content-type = "text/markdown" }
 authors = [
-    { name = "Ryan Wolf", email = "rywolf@nvidia.com" },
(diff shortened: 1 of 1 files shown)


abhinavg4 (approved):
Thanks. I think We should still keep Ryan. But maybe there email is not valid anymore. Can we just add Ryan's name again ?

pyproject.toml
```diff
@@ -21,7 +21,7 @@ name = "nemo_curator"
 description = "Scalable Data Preprocessing Tool for Training Large Language Models"
 readme = { file = "README.md", content-type = "text/markdown" }
 authors = [
-    { name = "Ryan Wolf", email = "rywolf@nvidia.com" },
+    { name = "Abhinav Garg", email = "abhgarg@nvidia.com" },
```
ayushdg, line 24: Could we add @praateekmahajan as well.

thomasdhc: Updated @ayushdg please review

PR #837, opened 2025-07-21
Initial Minhash implementation on Ray

ayushdg:
## Description
Implements the Minhash logic using Ray actors

Goals:
1. Support reading in jsonl/parquet data & adding our own ID independent of any ID column in the data.
    1. Motivation for this is to ensure that we do not have a string->int id conversion step for connected components by control and setting int id's at the beginning.
    2. Allows for more efficient removal by mapping source files to an ID range.

Benchmarking:

| Dataset           | Resources    | Dask (old) | Xenna    | Ray Data |
|-------------------|--------------|------------|----------|----------|
| 1.4TB CC snapshot | 4x A100 GPUs | 550-590s   | 470-500s | 480-500s |
|                   |              |            |          |          |
|                   |              |            |          |          |

Benchmarking Script: https://gist.github.com/ayushdg/7ace675b72e70b5f6abc4b9d47aeaaf6
Notes:
Main delta in perf b/w old and new versions the old version was writing string ID's which was slower in to_parquet and produced larger files.

## Usage

```python
# TODO: Need to add
```
## Checklist


ray-curator/ray_curator/stages/deduplication/io_utils.py
```diff
@@ -0,0 +1,57 @@
...
+    def read_parquet(self, filepath: str | list[str], assign_id: bool = False, **kwargs) -> "cudf.DataFrame":
+        read_kwargs = kwargs.copy()
+        read_kwargs["allow_mismatched_pq_schemas"] = True
+        df = cudf.read_parquet(filepath, **read_kwargs)
+        return self.assign_id(filepath, df) if assign_id and self.id_generator else df
+
+    def write_parquet(self, df: "cudf.DataFrame", filepath: str, **kwargs) -> None:
+        fs = get_fs(filepath, storage_options=kwargs.get("storage_options", {}))
+        # TODO: Add overwrite behavior here
+        fs.makedirs(fs._parent(filepath), exist_ok=True)
```
ayushdg, line 38: Previous behavior was to overwrite if directory already existed. Might want to keep the same here.

ray-curator/ray_curator/stages/deduplication/io_utils.py
```diff
@@ -0,0 +1,57 @@
...
+        df.to_parquet(filepath, **kwargs)
+
+    def custom_read(
```
ayushdg, line 41: Still leaving this as is though it's unused for the time being.

ray-curator/ray_curator/stages/deduplication/fuzzy/minhash.py
```diff
@@ -0,0 +1,326 @@
...
+from ray_curator.stages.deduplication.io_utils import DeduplicationIO
+from ray_curator.stages.resources import Resources
+from ray_curator.tasks import FileGroupTask
+from ray_curator.utils.file_utils import get_fs
+
+if TYPE_CHECKING:
+    from ray_curator.backends.base import WorkerMetadata
+
+
+class MinHash(ABC):
```
ayushdg, line 23: Not sure if I should have the base minhash classes in the same module as the stage or a different one.

ray-curator/ray_curator/stages/deduplication/fuzzy/minhash.py
```diff
@@ -0,0 +1,326 @@
...
+        if not isinstance(text_series, cudf.Series):
+            msg = "Expected data of type cudf.Series"
+            raise TypeError(msg)
+
+        # Compute minhashes
+        minhash_method = self.minhash64 if self.use_64bit_hash else self.minhash32
+        return minhash_method(text_series)
+
+
+class GPUMinHashStage(ProcessingStage[FileGroupTask, FileGroupTask], DeduplicationIO):
```
sarahyurick: Nit: We can call it `MinHashStage`?

ray-curator/ray_curator/stages/deduplication/fuzzy/minhash.py
```diff
@@ -0,0 +1,326 @@
...
+            self,
+            seed=seed,
+            num_hashes=num_hashes,
+            char_ngrams=char_ngrams,
+            use_64bit_hash=use_64bit_hash,
+        )
+
+        # Initialize memory pool for cuDF
+        if pool:
+            rmm.reinitialize(pool_allocator=pool)
```
sarahyurick, line 83: Curious about what is happening here. All the tests are `pool=False` while the default in the minhash stage is `pool=True`. Do we still need to worry about memory pools if we are controlling 1 GPU for this stage and each `GPUMinHashStage` actor is independent of the others?

ray-curator/ray_curator/stages/deduplication/fuzzy/minhash.py
```diff
@@ -0,0 +1,326 @@
...
+        read_kwargs["storage_options"] = task._metadata.get("storage_options", {})
+
+        # Read input file based on format
+        if self.read_format == "jsonl":
+            df = self.read_jsonl(filepath=task.data, columns=[self.text_column], assign_id=True, **read_kwargs)
+        elif self.read_format == "parquet":
+            df = self.read_parquet(filepath=task.data, columns=[self.text_column], assign_id=True, **read_kwargs)
+        else:
+            msg = f"Unsupported read format: {self.read_format}"
+            raise ValueError(msg)
```
sarahyurick, line 305: This is okay with me, but at some point will we want to accept a `DocumentBatch` input in addition to a `FileGroupTask` input (i.e., `process(self, task: FileGroupTask | DocumentBatch)`)?

ray-curator/ray_curator/stages/deduplication/fuzzy/minhash.py
```diff
@@ -0,0 +1,326 @@
...
+
+        result_df = df[[CURATOR_DEDUP_ID_STR]]
+        result_df[self.minhash_column] = self.minhash_processor.compute_minhashes(df[self.text_column])
+
+        # Write output file
+        self.write_parquet(df=result_df, filepath=output_file, **self.write_kwargs)
```
sarahyurick, line 311: Same here, it is more beneficial/negligible impact to write here instead of as a separate stage?

ray-curator/ray_curator/stages/deduplication/fuzzy/minhash.py
ayushdg: Fair. I don't think we'll have a CPU MinHashStage so we should be good to rename

ray-curator/ray_curator/stages/deduplication/fuzzy/minhash.py
ayushdg: Calling `rmm.reinitialize` multiple times in the same process was leading to segfaults. I think rmm is a singleton that wants to be initialized only once.
In practice since these are called from actors/processes with Ray it's not an issue but the tests don't create new actors, just run the `setup` directly on the driver/client process which leads to segfaults unless pool is `False`

ray-curator/ray_curator/stages/deduplication/fuzzy/minhash.py
ayushdg: Longer term there might be a plan to accept DocumentBatch as well for streaming use cases until MinHash but for now we're targeting dedup to be it's own pipeline run with actors outside of ray-data/xenna so this paradigm should be fine.

ray-curator/pyproject.toml
```diff
@@ -43,14 +43,17 @@ text = [
     "sentencepiece",
     "mwparserfromhell==0.6.5"
 ]
-cuda12x = [
+
+# Installs CPU + GPU text curation modules
+deduplication_cuda12x = [
```
praateekmahajan, line 50: It's a todo on my mind, feel free to add it here that [gpu] might also be needed which installs `xenna[gpu]`

ray-curator/ray_curator/stages/deduplication/fuzzy/minhash.py
ayushdg: I haven't explored this avenue but longer term we do want to explore splitting out the IO components and experimenting with fusion at the pipeline level if separate stages are slower.

ray-curator/ray_curator/stages/deduplication/fuzzy/minhash.py
praateekmahajan: what's the need for this? is there a world we extend MinHash more?

ray-curator/ray_curator/stages/deduplication/fuzzy/minhash.py
```diff
@@ -0,0 +1,326 @@
...
+        ).remote()
+
+        # Initialize the GPU minhash processor
+        self.minhash_processor = GPUMinHash(
+            seed=self.seed,
+            num_hashes=self.num_hashes,
+            char_ngrams=self.char_ngrams,
+            use_64bit_hash=self.use_64bit_hash,
+            pool=self.pool,
+        )
```
praateekmahajan, line 268: Out of curioisity, wondering why this is as a design choice of splitting stage and processor separately rather than keeping everything inside a `MinhashStage`?

ray-curator/ray_curator/stages/deduplication/fuzzy/minhash.py
```diff
@@ -0,0 +1,326 @@
...
+
+        if self.minhash_processor is None or self.id_generator is None:
+            msg = "MinHash processor or ID generator not initialized. Call setup() first."
+            raise RuntimeError(msg)
+
+        fs = get_fs(self.output_dir, self.write_kwargs.get("storage_options", {}))
+        output_file = fs.sep.join([self.output_dir, self._name, f"{task.task_id}.parquet"])
+
+        read_kwargs = self.read_kwargs.copy()
+        read_kwargs["storage_options"] = task._metadata.get("storage_options", {})
```
praateekmahajan: I'm sus about passing in storage_options from `task_metadata`. I think it bloats, likely leaks keys etc. I know we have it there, but I'm curious why not lead any reader just accept storage_option as a key?

ray-curator/ray_curator/stages/deduplication/fuzzy/minhash.py
praateekmahajan: @sarahyurick the reason for this per my understanding is that df is cudf.DataFrame so passing it as DocumentBatch means it gets serialized to CPU and then transferred over network to the next stage which will write using CPU. Might as well write it here while we're at it

praateekmahajan:
Mostly looks good, can you share the example file that you've been running to test this too

ray-curator/ray_curator/stages/deduplication/fuzzy/minhash.py
ayushdg: MinHash is a feature that's being used externally as well and we may add ngram based minhashing. That combined with some possibility of supporting APIs independent of our the Stage Framework was one of the reasons.

ray-curator/pyproject.toml
ayushdg: > [gpu] might also be needed which installs xenna[gpu]

Could you elaborate. I'm not sure I understood.

ray-curator/ray_curator/stages/deduplication/fuzzy/minhash.py
ayushdg: Probably the same as this: https://github.com/NVIDIA-NeMo/Curator/pull/837#discussion_r2261207607

ayushdg:
> Mostly looks good, can you share the example file that you've been running to test this too

Updated the description with the link.

ray-curator/ray_curator/stages/deduplication/fuzzy/minhash.py
ayushdg: I'm okay with that approach as well. The main motivation was this:
For general readers (like jsonl) which are compositeStages you provide the `storage_options` once and that'll be used for both FilePartitioningStage as well as JsonlReaderStage.

Since there's no composite here the user experience would be
```
FilePartitioningStage(storage_options)
MinHashStage(same_storage_options)
```
I'm fine with that if reusing from the prev task can lead to unexpected behavior.

ray-curator/tests/stages/deduplication/fuzzy/test_minhash.py
```diff
@@ -0,0 +1,185 @@
...
+                "A different test string",
+                "A different object",
+                "The quick brown fox jumps over the lazy dog",
+                "The quick black cat jumps over the lazy dog",
+            ],
+        }
+    )
+
+
+@pytest.mark.gpu
```
ayushdg, line 50: Ci tests seem to be failing because of this not being recognized. Do I need to add the pytest config as well?

sarahyurick (approved):
LGTM! Implementation design and decisions make sense to me.

sarahyurick:
Conflict is because the directory structure is now `stages/text/io/reader`. Sorry for the inconvenience.

ray-curator/pyproject.toml
ayushdg: Opened: https://github.com/orgs/NVIDIA-NeMo/projects/15/views/13?sliceBy%5Bvalue%5D=Core%2FBase&pane=issue&itemId=123856146 to followup with this

ray-curator/ray_curator/stages/deduplication/fuzzy/minhash.py
ayushdg: Updated to use `read_kwargs` directly which could contain `storage_options` in it. 
We can use a followup to decide if `storage_options` should be a separate explicit arg in addition to read_kwargs.

praateekmahajan (approved):
LGTM

PR #739, opened 2025-06-17
Update CI testing workflow for ray branch

ayushdg:
## Description

Updates the CI testing workflow to build the ray-curator package for the ray branch.

## Usage

```python
# Add snippet demonstrating usage
```
## Checklist

- [ ] I am familiar with the [Contributing Guide](https://github.com/NVIDIA/NeMo-Curator/blob/main/CONTRIBUTING.md).
- [ ] New or Existing tests cover these changes.
- [ ] The documentation is up to date with these changes.

PR #840, opened 2025-07-22
Re-enable CI/CD for Ray API branch

sarahyurick:
Since test.yml was deleted on main, we need to re-enable Ray testing with the new setup.

Guided by https://github.com/NVIDIA-NeMo/Curator/pull/739.


thomasdhc:
@sarahyurick Instead of deleting what is currently not used shall we comment out gpu-tests, and needs gpu-test from nemo-cicd-test?

This branch will eventually require both. I guess this branch is in rapid development, but since it is the new main would we want to protect it and require some checks to pass?

Decision is ultimately up to you

PR #844, opened 2025-07-22
Initialize and shutdown ray session in each executor

praateekmahajan:
## Description

This PRs goal is to ensure we are able to run different executor pipeline in the same python process e.g. Xenna pipeline followed by Ray Data pipeline (and vice versa). It is currently not possible to do so without shutdown because ray preserve envvars from the previous sessions. 

If https://github.com/nvidia-cosmos/cosmos-xenna/pull/6/files does not merge, even then this PR works as expected as Xenna sets `os.environ[RAY_EXPERIMENTAL_NOSET_CUDA_VISIBLE_DEVICES]` which means if Ray Data runs after Xenna then Ray Data will not set `CUDA_VISIBLE_DEVICES` i.e. use `gpu_id=0`.  We solve that by forcefully setting `RAY_EXPERIMENTAL_NOSET_CUDA_VISIBLE_DEVICES=""` inside envvars while doing ray.init inside Ray Data.

## Workarounds needed
However because we introduce `ray.shutdown()` now 
1. running two subsequent Ray Data pipelines will fail due to https://github.com/ray-project/ray/issues/54841
2. Whenever we create a ray cluster inside the process, it'll be killed by the first pipeline. To avoid that we can use subprocess.run to create our cluster
 
See the following for clearer understanding

1. You start a Ray Cluster outside the process and run pipelines in the same process
```python
ray start --head --port 1234

os.environ["RAY_ADDRESS"] = "localhost:1234"
ray.init(ignore_reinit=True) # this will connect to 1234
run_pipeline("xenna")
ray.shutdown()

ray.init(ignore_reinit=True) # this will connect to 1234
run_pipeline("ray data")
ray.shutdown()
```

2. You start a ray cluster inside the process and run pipelines in the same process
```python
ray.init(ignore_reinit=True) # this will start a new cluster
run_pipeline("xenna")
ray.shutdown() # this will kill the ray cluster


ray.init(ignore_reinit=True) # this will start a new cluster
run_pipeline("xenna")
ray.shutdown()
```

3. You start a ray cluster explicitly inside the new process
```python
ray.init(ignore_reinit=True) # this will start a new cluster A

ray.init(ignore_reinit=True) # this will connect to cluster A
run_pipeline("xenna")
ray.shutdown() # this will kill the cluster A


ray.init(ignore_reinit=True) # this will start cluster B
run_pipeline("xenna")
ray.shutdown()
```
4. You use subprocess to start a ray cluster
```python
subprocess.run(["ray", "start", "--head", "--port", "1234"]) # this will start a new cluster A

ray.init(ignore_reinit=True) # this will connect to cluster A @ 1234
run_pipeline("xenna")
ray.shutdown() # this will exit session

ray.init(ignore_reinit=True) # this will connect to cluster A @ 1234
run_pipeline("xenna")
ray.shutdown()
```

## Usage

```python
```
## Checklist

### DIFF
--- a/ray-curator/tests/backends/conftest.py
+++ b/ray-curator/tests/backends/conftest.py
@@ -7,7 +7,6 @@ proper isolation through Ray's actor/task lifecycle management.
 
 import os
 import subprocess
-from pathlib import Path
 
 import pytest
 import ray
@@ -23,7 +22,7 @@ def find_free_port():
 
 
 @pytest.fixture(scope="session", autouse=True)
-def shared_ray_cluster():
+def shared_ray_cluster(tmp_path_factory: pytest.TempPathFactory):
     """Set up a shared Ray cluster for all tests in the session.
 
     This fixture automatically sets up Ray at the beginning of the test session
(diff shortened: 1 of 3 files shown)


ray-curator/ray_curator/backends/experimental/ray_data/executor.py
```diff
@@ -44,17 +44,20 @@ def execute(self, stages: list["ProcessingStage"], initial_tasks: list[Task] | N
...
         register_loguru_serializer()
-        execute_setup_on_node(stages)
-
-        logger.info(f"Setup on node complete for all stages. Starting Ray Data pipeline with {len(stages)} stages")
         # Initialize with initial tasks if provided, otherwise start with EmptyTask
         tasks: list[Task] = initial_tasks if initial_tasks else [EmptyTask]
+        output_tasks: list[Task] = []
+        try:
+            # Initialize ray
+            ray.init(ignore_reinit_error=True)
```
abhinavg4: NIT: Can we please add a comment here? Why are we doing this again here? Since getting the client also might have ray init/ray start.

ray-curator/ray_curator/backends/experimental/ray_data/executor.py
```diff
@@ -69,15 +72,14 @@ def execute(self, stages: list["ProcessingStage"], initial_tasks: list[Task] | N
...
             # TODO: add pipeline configuration to check if user wants to return last stages output to driver
-            logger.info(f"Converting final dataset to tasks -> {current_dataset=}")
-            final_tasks = self._dataset_to_tasks(current_dataset)
-            logger.info(f"Pipeline completed. Final results: {len(final_tasks)} tasks")
-
-            return final_tasks
+            output_tasks = self._dataset_to_tasks(current_dataset)
+            logger.info(f"Pipeline completed. Final results: {len(output_tasks)} tasks")
+        finally:
+            ray.shutdown()
```
abhinavg4, line 86: Same as above. A comment here would be useful. If you have made a issue out of our findings, we can just link that here.

ray-curator/ray_curator/backends/experimental/ray_data/executor.py
abhinavg4: Also this should have the loguru serialiser?

ray-curator/ray_curator/backends/xenna/executor.py
```diff
@@ -113,13 +114,14 @@ def execute(self, stages: list[ProcessingStage], initial_tasks: list[Task] | Non
         logger.info(f"Execution mode: {exec_mode.name}")
 
         try:
-            # Run the pipeline
+            # Run the pipeline (this will initialize ray)
```
abhinavg4: My recommendation would be to add ray. init here as well, along with Ray Loguru serializer, so that we are not dependent on Xenna. And the call is exactly the same across executors. 
If we add `ray.init` here, the ray init inside Xenna will effectively be useless, and I think that is what we want?

ray-curator/tests/backends/conftest.py
```diff
@@ -6,12 +6,20 @@
...
+import subprocess
 from pathlib import Path
 
 import pytest
 import ray
 from loguru import logger
-from ray.cluster_utils import Cluster
+
+
+def find_free_port():
```
abhinavg4, line 16: I think this function might be present in the actual code too. Can we add a TODO to use that maybe and remove this ?

ray-curator/tests/backends/conftest.py
```diff
@@ -36,20 +40,32 @@ def shared_ray_cluster():
...
+            str(ray_port),
+            "--num-cpus",
+            "11",
+            "--num-gpus",
+            "0",
+            "--object-store-memory",
+            str(2 * ONE_GB),
+            "--block",
+        ],
+        env={**os.environ, "RAY_MAX_LIMIT_FROM_API_SERVER": "40000", "RAY_MAX_LIMIT_FROM_DATA_SOURCE": "40000"},
```
abhinavg4: Add a TODO to use get_client here in the future?

ray-curator/ray_curator/backends/experimental/ray_data/executor.py
praateekmahajan: This does have a loguru few lines above. Added a comment too

ray-curator/ray_curator/backends/xenna/executor.py
praateekmahajan: Yup made that change, and in fact that helps us solve this PR without Xenna changes too (along with one more change)

ray-curator/tests/backends/conftest.py
praateekmahajan: the reason we find the port ourselves is so that subsequently we can connect to that port using `ray.init` (instead of address="auto" to avoid connecting to another cluster in case multiple are running)..

ray-curator/tests/backends/conftest.py
praateekmahajan: Added, i hope we can. the only nuance is that we need to know the pid of the process that was started by ray start so that we can kill it without doing `ray stop` which might kill all ray processes

Issue #801, opened 2025-07-14
[Ray] Explore fusion options for `ScoreFilter` modules


sarahyurick:
As a follow up to https://github.com/NVIDIA-NeMo/Curator/pull/746, we should explore fusing multiple `DocumentFilter` operations into a single `ScoreFilter` stage and/or general techniques for combining filters together.

The goal here would be to reduce communication from the Ray object store.

sarahyurick:
Closed by https://github.com/NVIDIA-NeMo/Curator/pull/850.

PR #850, opened 2025-07-23
Allow users to fuse multiple `DocumentFilter` objects into a single `ScoreFilter` stage

sarahyurick:
Closes https://github.com/NVIDIA-NeMo/Curator/issues/801.


ray-curator/ray_curator/stages/text/modules/score_filter.py
```diff
@@ -33,42 +34,74 @@ class Score(ProcessingStage[DocumentBatch, DocumentBatch]):
...
+            if isinstance(self.text_field, list) and len(self.text_field) > 1:
+                msg = f"More text fields than filters provided: {self.text_field}"
+                raise ValueError(msg)
+
+            self.score_fn = [self.score_fn]
+            self.score_field = [self.score_field]
+            self.text_field = [self.text_field]
+
+        elif isinstance(self.score_fn, list):
+            self._name = "score_fn_chain"
```
VibhuJawa: Minor nit to have name based on filters applied.

ray-curator/ray_curator/stages/text/modules/score_filter.py
```diff
@@ -33,42 +34,74 @@ class Score(ProcessingStage[DocumentBatch, DocumentBatch]):
...
-        if isinstance(self.score_fn, DocumentFilter):
-            if hasattr(self.score_fn, "load_model"):
-                self.score_fn.load_model()
-            elif hasattr(self.score_fn, "load_tokenizer"):
-                self.score_fn.load_tokenizer()
+        for score_fn in self.score_fn:
+            if isinstance(score_fn, DocumentFilter):
+                if hasattr(score_fn, "load_model"):
+                    score_fn.load_model()
+                elif hasattr(score_fn, "load_tokenizer"):
```
VibhuJawa:
```suggestion
                if hasattr(score_fn, "load_tokenizer"):
                    score_fn.load_tokenizer()
```

ray-curator/ray_curator/stages/text/modules/score_filter.py
```diff
@@ -82,10 +115,13 @@ def process(self, batch: DocumentBatch) -> DocumentBatch | None:
...
-        score_fn = self.score_fn.score_document if isinstance(self.score_fn, DocumentFilter) else self.score_fn
-
         df = batch.to_pandas()
-        df[self.score_field] = df[self.text_field].apply(score_fn)
+
+        for score_fn_i, text_field_i, score_field_i in zip(
+            self.score_fn, self.text_field, self.score_field, strict=True
+        ):
+            inner_score_fn = score_fn_i.score_document if isinstance(score_fn_i, DocumentFilter) else score_fn_i
+            df[score_field_i] = df[text_field_i].apply(inner_score_fn)
```
VibhuJawa, line 100: As a follow up, we may want to just apply a function on text series instead of strings for controlling stuff like batching, etc. 

This is true for both this PR and my PR https://github.com/NVIDIA-NeMo/Curator/pull/898/files which is based on a lot of your work anyways. 

Nothing to do here, just keeping it mind for a follow up PR/Issue of having that option in the base class.

ray-curator/ray_curator/stages/text/modules/score_filter.py
```diff
@@ -106,46 +142,81 @@ class Filter(ProcessingStage[DocumentBatch, DocumentBatch]):
...
         if isinstance(self.filter_fn, DocumentFilter):
             self._name = self.filter_fn.name
 
+        if isinstance(self.filter_fn, (DocumentFilter, Callable)):
+            if isinstance(self.filter_field, list) and len(self.filter_field) > 1:
+                msg = f"More filter fields than filters provided: {self.filter_field}"
+                raise ValueError(msg)
+            if isinstance(self.invert, list) and len(self.invert) > 1:
+                msg = f"More invert flags than filters provided: {self.invert}"
+                raise ValueError(msg)
```
VibhuJawa: Probably we can move to a function to prevent code replication

ray-curator/ray_curator/stages/text/modules/score_filter.py
```diff
@@ -162,18 +233,20 @@ def process(self, batch: DocumentBatch) -> DocumentBatch | None:
...
-        result_df = df[bool_mask]
 
-        if len(result_df) == 0:
-            print(f"All documents filtered out for batch {batch.task_id}")
+        for filter_fn_i, filter_field_i, invert_i in zip(self.filter_fn, self.filter_field, self.invert, strict=True):
+            bool_mask = self.compute_filter_mask(df, filter_fn_i, filter_field_i, invert_i)
+            df = df[bool_mask]
+
+        if len(df) == 0:
+            logger.info(f"All documents filtered out for batch {batch.task_id}")
```
VibhuJawa, line 193: Like, if its None, does our downstream pipeline logic handle it or will it fail ? 

Like, should we create a empty data object here ?

ray-curator/ray_curator/stages/text/modules/score_filter.py
```diff
@@ -182,70 +255,120 @@ def process(self, batch: DocumentBatch) -> DocumentBatch | None:
...
+    filter_obj: DocumentFilter | list[DocumentFilter]
+    text_field: str | list[str] = "text"
+    score_field: str | list[str] | None = None
+    invert: bool | list[bool] = False
     _name: str = "score_filter"
 
     def __post_init__(self):
-        self._name = self.filter_obj.name
+        if isinstance(self.filter_obj, list):
+            self._name = "score_filter_chain"
```
VibhuJawa: Same note about name here

ray-curator/ray_curator/stages/text/modules/score_filter.py
```diff
@@ -182,70 +255,120 @@ def process(self, batch: DocumentBatch) -> DocumentBatch | None:
...
+
+            num_filters = len(self.filter_obj)
+
+            # Okay to assume same text field and invert flag for all filters
+            if isinstance(self.text_field, str):
```
VibhuJawa: Probably use a function to validate to help with reducing code deplication

ray-curator/ray_curator/stages/text/modules/score_filter.py
```diff
@@ -182,70 +255,120 @@ def process(self, batch: DocumentBatch) -> DocumentBatch | None:
...
+            self.score_field = [self.score_field]
+            self.invert = [self.invert]
 
     def inputs(self) -> tuple[list[str], list[str]]:
-        return ["data"], [self.text_field]
+        return ["data"], self.text_field
 
     def outputs(self) -> tuple[list[str], list[str]]:
-        return ["data"], [self.score_field] if self.score_field else []
+        return ["data"], self.score_field if self.score_field is not None else []
```
VibhuJawa: Should we add validation that text field is not dropped here too ? 
```suggestion
        return ["data"], [self.text_field] + (self.score_field if self.score_field is not None else [])
```

ray-curator/ray_curator/stages/text/modules/score_filter.py
```diff
@@ -262,18 +385,22 @@ def process(self, batch: DocumentBatch) -> DocumentBatch | None:
...
-        if len(result_df) == 0:
-            print(f"All documents filtered out for batch {batch.task_id}")
+        for filter_obj_i, text_field_i, score_field_i, invert_i in zip(
+            self.filter_obj, self.text_field, self.score_field, self.invert, strict=True
+        ):
+            bool_mask = self.compute_filter_mask(df, filter_obj_i, text_field_i, score_field_i, invert_i)
+            df = df[bool_mask]
+
+        if len(df) == 0:
+            logger.info(f"All documents filtered out for batch {batch.task_id}")
```
VibhuJawa, line 311: Same question here.

ray-curator/tests/stages/text/test_filters.py
```diff
@@ -124,7 +124,6 @@ def test_score_filter(self, letter_count_data: DocumentBatch) -> None:
         letter_filter = LetterCountFilter()
         filter_step = ScoreFilter(letter_filter, text_field="documents")
 
-        filter_step.setup()
```
VibhuJawa, line 127: Why did we remove this, just curious ?

VibhuJawa (changes requested):
Mostly looks good to me, just added some nits. Thanks a lot for this PR, i mostly used this PR to build my work here: https://github.com/NVIDIA-NeMo/Curator/pull/898

Thanks again for this work. Made my life a lot easier.

ray-curator/tests/stages/text/test_filters.py
sarahyurick: The `ScoreFilter` setup function really does nothing unless it is a FastText filter or the `TokenCountFilter`. But I may rewrite to do what you did for modifiers: have a helper function that calls setup and process instead of writing it in for each test.

ray-curator/ray_curator/stages/text/modules/score_filter.py
sarahyurick: Let me check actually. I never ran into any issues with my large or small scale tests, but maybe it was luck. I will check specifically for this and maybe add a test as well.

ray-curator/ray_curator/stages/text/modules/score_filter.py
sarahyurick: Oh interesting, I am curious if this would make a difference in performance? I thought for Pandas it will do it sequentially either way.

ray-curator/ray_curator/stages/text/modules/score_filter.py
sarahyurick: Good catch, thanks.

ray-curator/ray_curator/stages/text/modules/score_filter.py
sarahyurick: Yeah, it looks like the rest of the pipeline fails if I try to set it to `None` or `EmptyTask`. For now, I will just have it return the empty DataFrame.

ray-curator/ray_curator/stages/text/modules/score_filter.py
VibhuJawa: I think eventually we need to enable batching and not just run at once based on filter. 

Related, just realized pandas 2.1 also added numba support  so we can even explore parallel here. 

https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.apply.html

Image

We should just file an issue to follow up on this

ray-curator/ray_curator/stages/text/modules/score_filter.py
VibhuJawa: I think that is fine, returning empty frames is expected behavior and lines it up with what i would expect

PR #847, opened 2025-07-23
Add aesthetic filtering capabilities to video processing pipeline

suiyoubi:
- Introduced ClipAestheticFilterStage to filter video clips based on aesthetic scores using a new AestheticScorer model.
- Updated video_split_clip_example.py to include aesthetic threshold and reduction method arguments.
- Added new models for aesthetic scoring and CLIP image embeddings, enhancing the video processing framework.
- Implemented comprehensive unit tests for the new aesthetic scoring models and the filtering stage, ensuring robust functionality and integration.

## Usage

```python
# Add snippet demonstrating usage
```
## Checklist

- [ ] I am familiar with the [Contributing Guide](https://github.com/NVIDIA/NeMo-Curator/blob/main/CONTRIBUTING.md).
- [ ] New or Existing tests cover these changes.
- [ ] The documentation is up to date with these changes.

PR #864, opened 2025-07-28
add DataEntry, AudioReaderStage

karpnv:
## Description
Add speech processing

## Usage

```python
```
## Checklist

### DIFF
--- a/ray-curator/ray_curator/stages/audio/io/audio_reader.py
+++ b/ray-curator/ray_curator/stages/audio/io/audio_reader.py
@@ -1,0 +1,56 @@
+import pathlib
+from dataclasses import dataclass
+
+from loguru import logger
+
+from ray_curator.stages.base import ProcessingStage
+from ray_curator.tasks import Video, DataEntry, _EmptyTask
+from ray_curator.utils.file_utils import get_all_files_paths_under
+
+
+@dataclass
+class AudioReaderStage(ProcessingStage[_EmptyTask, DataEntry]):
+    """Stage that reads video files from storage and extracts metadata."""
+    input_audio_path: str
+    audio_limit: int = -1
+    filepath_key: str = "audio_filepath"
+    _name: str = "audio_reader"
+
+    def inputs(self) -> tuple[list[str], list[str]]:
+        return [], []
+
+    def outputs(self) -> tuple[list[str], list[str]]:
+        return ["data"], ["input_audio"]
+
+    def process(self, _: _EmptyTask) -> list[DataEntry]:
+        """Process a single group of audio files."""
+        if self.input_audio_path is None:
+            msg = "input_audio_path is not set"
+            raise ValueError(msg)
+        files = get_all_files_paths_under(
+            self.input_audio_path,
+            recurse_subdirectories=True,
+            keep_extensions=[".wav", ".mp3", ".flac", ".ogg", ".opus"],
+        )
+        logger.info(f"Found {len(files)} files under {self.input_audio_path}")
+
+        if self.audio_limit > 0:
+            files = files[:self.audio_limit]
+            logger.info(f"Using first {len(files)} files under {self.input_audio_path} since audio_limit is set to {self.audio_limit}")
+
+        audio_tasks = []
+        for fp in files:
+
+            file_path = fp
+            if isinstance(file_path, str):
+                file_path = pathlib.Path(file_path)
+
+            audio = {self.filepath_key: file_path}
+            audio_task = DataEntry(
+                task_id=f"{file_path}_processed",
+                dataset_name=self.input_audio_path,
+                data=audio,
+            )
+            audio_tasks.append(audio_task)
+
+        return audio_tasks
--- a/ray-curator/ray_curator/tasks/__init__.py
+++ b/ray-curator/ray_curator/tasks/__init__.py
@@ -3,6 +3,7 @@ from .file_group import FileGroupTask
 from .image import ImageBatch, ImageObject
 from .tasks import EmptyTask, Task, _EmptyTask
 from .video import Clip, ClipStats, SplitPipeTask, Video, VideoMetadata, VideoTask
+from .audio import DataEntry
 
 __all__ = [
     "Clip",
@@ -17,5 +18,6 @@ __all__ = [
     "Video",
     "VideoMetadata",
     "VideoTask",
+    "DataEntry",
     "_EmptyTask",
 ]
--- a/ray-curator/ray_curator/tasks/audio.py
+++ b/ray-curator/ray_curator/tasks/audio.py
@@ -1,0 +1,27 @@
+import os
+from dataclasses import dataclass, field
+from typing import Any, Dict
+
+from .tasks import Task
+
+@dataclass
+class DataEntry(Task[list]):
+    """A wrapper for data entry + any additional metrics."""
+
+    def __init__(self, data: Dict = None, metrics: Any = None, dataset_name: str = "", task_id: int = 0, **kwargs):
+        self.data = data  # data can be None to drop the entry
+        self.metrics = metrics
+        self.filepath_key = "audio_filepath"
+        super().__init__(data=data, task_id=task_id, dataset_name=dataset_name, **kwargs)
+
+    @property
+    def num_items(self) -> int:
+        return 1
+
+    def validate(self) -> bool:
+        """Validate the task data."""
+        if not os.path.exists(self.data[self.filepath_key]):
+            print(f"Video {self.data[self.filepath_key]} does not exist")
+            return False
+        return True
+    


ray-curator/ray_curator/stages/audio/io/audio_reader.py
```diff
@@ -0,0 +1,58 @@
...
+from ray_curator.tasks import DataEntry, _EmptyTask
+from ray_curator.utils.file_utils import get_all_files_paths_under
+
+
+@dataclass
+class AudioReaderStage(ProcessingStage[_EmptyTask, DataEntry]):
+    """Stage that reads video files from storage and extracts metadata."""
+
+    input_audio_path: str
+    audio_limit: int = -1
```
sarahyurick:
```suggestion
    audio_file_limit: int = -1
```
or `file_limit`?

ray-curator/ray_curator/tasks/audio.py
```diff
@@ -0,0 +1,28 @@
+import os
+from dataclasses import dataclass
+
+from .tasks import Task
+
+
+@dataclass
+class DataEntry(Task[list]):
```
sarahyurick:
```suggestion
class AudioClip(Task[list]):
```
or something similar? I was wondering what this Task will eventually encapsulate.

praateekmahajan: +1 can you share what data will reside in this task..

ray-curator/ray_curator/stages/audio/io/audio_reader.py
karpnv: The name is inspired by VideoReader class: 
https://github.com/NVIDIA-NeMo/Curator/blob/ray-api/ray-curator/ray_curator/stages/video/io/video_reader.py#L235
Should it be consistent?

ray-curator/ray_curator/tasks/audio.py
karpnv: Usually, we have a dict with several fields like this: 
`self.data = {"audio_filepath": "/home/user1/a.wav", "text": "the can sat on a mat", "lang": "en"} `

karpnv: It is pretty universal

ray-curator/ray_curator/tasks/__init__.py
```diff
@@ -9,6 +10,8 @@
     "FileGroupTask",
     "ImageBatch",
     "ImageObject",
+    "SpeechDataEntry",
+    "SplitPipeTask",
```
huvunvidia: I guess we're missing SplitPipeTask file?

ray-curator/ray_curator/tasks/audio.py
```diff
@@ -0,0 +1,28 @@
+import os
+from dataclasses import dataclass
+
+from .tasks import Task
+
+
+@dataclass
+class SpeechDataEntry(Task[list]):
```
huvunvidia: As I understand, the self.data will have dict type.
Hence, should this be: "class SpeechDataEntry(Task[dict]):"?

ray-curator/ray_curator/stages/audio/io/audio_reader.py
```diff
@@ -0,0 +1,59 @@
+import pathlib
```
huvunvidia: Can you add an simple example of how to this stage?
You can learn from the example in video: ray-curator/ray_curator/examples/video/video_read_example.py

ray-curator/ray_curator/tasks/audio.py
karpnv: Done

ray-curator/ray_curator/tasks/__init__.py
karpnv: Removed

ray-curator/ray_curator/stages/audio/io/asr_inference.py
```diff
@@ -0,0 +1,140 @@
...
+
+from ray_curator.stages.base import CompositeStage, ProcessingStage
+from ray_curator.stages.io.reader.file_partitioning import FilePartitioningStage
+from ray_curator.tasks import SpeechEntry, _EmptyTask
+from ray_curator.tasks.file_group import FileGroupTask
+
+
+@dataclass
+class AsrNemoInferenceStage(ProcessingStage[FileGroupTask, SpeechEntry]):
+    """Stage that reads video files from storage and extracts metadata."""
```
sarahyurick: Nit: a couple mentions of video instead of audio in this file.

ray-curator/ray_curator/tasks/audio.py
```diff
@@ -0,0 +1,28 @@
...
+        super().__init__(data=data, task_id=task_id, dataset_name=dataset_name, **kwargs)
+
+    @property
+    def num_items(self) -> int:
+        return 1
+
+    def validate(self) -> bool:
+        """Validate the task data."""
+        if not os.path.exists(self.data[self.filepath_key]):
+            print(f"Video {self.data[self.filepath_key]} does not exist")
```
sarahyurick: Nit: should be audio.

ray-curator/ray_curator/stages/audio/io/asr_inference.py
```diff
@@ -0,0 +1,140 @@
...
+            text = outputs[i]
+            file_path = files[i]
+
+            entry = {self.filepath_key: file_path, self.filepath_key: text}
+
+            audio_task = SpeechEntry(
+                task_id=f"{file_path}_task_id",
+                dataset_name=f"{self.model_name}_inference",
+                data=entry,
+            )
```
sarahyurick: I think other params like `_stage_perf` and `_metadata` could be initialized here too.

ray-curator/ray_curator/stages/audio/io/asr_inference.py
```diff
@@ -0,0 +1,140 @@
...
+    """Stage that reads video files from storage and extracts metadata."""
+
+    model_name: str
+    filepath_key: str = "audio_filepath"
+    text_key: str = "text"
+    _name: str = "audio_inference"
+
+    def setup(self) -> None:
+        """Initialise heavy object self.asr_model: nemo_asr.models.ASRModel"""
+        self.asr_model = nemo_asr.models.ASRModel.from_pretrained(model_name=self.model_name)
```
sarahyurick: Is this a CPU-only stage?

ray-curator/ray_curator/stages/audio/io/asr_inference.py
```diff
@@ -0,0 +1,140 @@
...
+            audio_task = SpeechEntry(
+                task_id=f"{file_path}_task_id",
+                dataset_name=f"{self.model_name}_inference",
+                data=entry,
+            )
+            audio_tasks.append(audio_task)
+        return audio_tasks
+
+    def process(self, task: FileGroupTask) -> list[SpeechEntry]:
+        pass
```
sarahyurick: This can be removed since you are overwriting `process_batch`.

sarahyurick (commented):
Nice PR so far, added some comments.

ray-curator/ray_curator/tasks/audio.py
praateekmahajan: Video..? And let's use loguru.logger with the debug

sarahyurick:
Looks like we should add `nemo_toolkit["asr"]` as a dependency in the pyproject file.

ray-curator/ray_curator/stages/audio/io/asr_inference.py
karpnv: we can't, since it is an abstract method

ray-curator/ray_curator/stages/audio/io/asr_inference.py
karpnv: It works both for CPU and GPU

ray-curator/ray_curator/tasks/audio.py
karpnv: Done

ray-curator/ray_curator/stages/audio/io/asr_inference.py
karpnv: Yes, we initialize them here too since SpeechObject inherited from Task class

ray-curator/ray_curator/tasks/audio.py
karpnv: done

ray-curator/ray_curator/stages/audio/io/asr_inference.py
karpnv: done

ray-curator/ray_curator/tasks/audio.py
karpnv: renamed to SpeechObject, inspired by ImageObject

ray-curator/ray_curator/stages/audio/io/asr_inference.py
sarahyurick: Ah ok, thanks. The Curator team should update this at some point. Good to leave as is for now.

ray-curator/ray_curator/stages/audio/io/asr_inference.py
sarahyurick: Got it, okay. How can the user configure this?

One way could be to add a `use_gpu` boolean parameter. Then, in a post init function, set the `Resources` accordingly?

ray-curator/ray_curator/stages/audio/io/asr_inference.py
sarahyurick: Got it, keeping as default for now.

ray-curator/pyproject.toml
```diff
@@ -28,6 +28,7 @@ dependencies = [
     "torch",
     "transformers>=4.48.0",
     "pynvml==11.5.3",
+    "nemo_toolkit[asr]",
```
sarahyurick: Below under `project.optional-dependencies` we have `text`, `cuda12x`, `dev`, and `video` dependencies. Could you create a dependency for `audio` and move `nemo_toolkit[asr]` there?

ray-curator/ray_curator/stages/audio/io/asr_inference.py
```diff
@@ -0,0 +1,143 @@
...
+class AsrNemoInference(CompositeStage[_EmptyTask, SpeechObject]):
+    """Composite stage that read audio files and do speech recognition inference using NeMo model.
+
+    This stage combines FilePartitioningStage and AsrNemoInferenceStage into a single
+    high-level operation for reading audiop files from a directory and processing
+    them.
+
+    Args:
+        input_audio_path: Path to the directory containing audio files
+        model_name: name of the speech recognition NeMo model
```
sarahyurick: I was wondering if there is a list of models supported here? Could you add a link?

ray-curator/ray_curator/stages/audio/io/asr_inference.py
```diff
@@ -0,0 +1,143 @@
...
+        audio_limit: Maximum number of audios to process (None for unlimited)
```
sarahyurick:
```suggestion
        audio_limit: Maximum number of audio files to process (None for unlimited)
```
?

ray-curator/ray_curator/stages/audio/io/asr_inference.py
```diff
@@ -0,0 +1,143 @@
...
+@dataclass
+class AsrNemoInferenceStage(ProcessingStage[FileGroupTask, SpeechObject]):
+    """Stage that do speech recognition inference using NeMo model."""
+
+    model_name: str
+    filepath_key: str = "audio_filepath"
+    text_key: str = "text"
+    _name: str = "audio_inference"
+
+    def setup(self) -> None:
```
sarahyurick: Could you also implement `setup_on_node` to ensure that the model is already downloaded from Hugging Face? For example for text models we do: https://github.com/NVIDIA-NeMo/Curator/blob/bd176474788d1f5b3c3d3e2de15cfccae2294bdb/ray-curator/ray_curator/stages/classifiers/base.py#L95-L100

ray-curator/pyproject.toml
karpnv: done

ray-curator/ray_curator/stages/audio/io/asr_inference.py
karpnv: no need, because ASRModel.from_pretrained() under the hood use try_to_load_from_cache from huggingface_hub

sarahyurick: Yes, but `setup_on_node` ensures that if the model isn't already downloaded, then we only attempt to download once.

ray-curator/ray_curator/stages/audio/io/asr_inference.py
karpnv: added map_location

ray-curator/ray_curator/stages/audio/io/asr_inference.py
karpnv: added to AsrNemoInferenceStage()

ray-curator/ray_curator/stages/audio/io/asr_inference.py
karpnv: -1 is not working. Looks like a bug

ray-curator/ray_curator/stages/audio/io/asr_inference.py
karpnv: done

ray-curator/ray_curator/stages/audio/io/asr_inference.py
karpnv: comment fixed

ray-curator/ray_curator/stages/audio/io/asr_inference.py
```diff
@@ -0,0 +1,185 @@
+import time
```
huvunvidia: I wonder why do move this file to under /audio/io?
Usually, /io/ is used for reading or writing stage, while AsrNemoInferenceStage is a regular data processing stage.

ray-curator/ray_curator/stages/audio/io/asr_inference.py
```diff
@@ -0,0 +1,185 @@
...
+from dataclasses import dataclass, field
```
huvunvidia: Can you add an example script (e.g. ray-curator/ray_curator/examples/audio/audio_curation_example.py) of the full pipeline?
You can learn from the example in for images at: https://github.com/NVIDIA-NeMo/Curator/blob/huvu/image_curation_2/ray-curator/ray_curator/examples/image/image_curation_example.py

ray-curator/ray_curator/stages/audio/io/asr_inference.py
```diff
@@ -0,0 +1,185 @@
...
+        pass
+
+    def finalize(self) -> None:
+        elapsed = time.time() - self._start_time
+        logger.info(f"Stage {self.name} completed in {elapsed:.2f} seconds.")
+        self._metrics["elapsed"] = elapsed
+
+
+@dataclass
+class AsrNemoInference(CompositeStage[_EmptyTask, SpeechObject]):
```
huvunvidia: I'm not sure if it makes sense to have a composite Stage combining reading step and processing step.
 
I think these steps should be separated. Because reading step might have multiple smaller steps (e.g., download, reader), and processing step also would have multiple smaller steps (e.g., asr inference, filtering incorrect transcription, etc.)

From other examples that use CompositeStage:
- ray-curator/ray_curator/stages/io/reader/jsonl.py
- ray-curator/ray_curator/stages/video/io/video_reader_download.py

CompositeStage is only used to combine reading steps (download + reader, or FilePartitioningStage + JsonlReaderStage).

In my opinion, we don't necessarily need CompositeStage in this scenario. Having separate stages for downloading, reading, asrInference would be enough.

huvunvidia (commented):
I have several comments regarding

ray-curator/ray_curator/stages/audio/io/asr_inference.py
huvunvidia: @sarahyurick @abhinavg4 would love to hear what you think

sarahyurick: Agree, although I would like to avoid having the user interact directly with the `FilePartitioningStage`, so besides `filepath_key` I am curious what else would be stored in a basic `SpeechObject` entry.

Also, do we expect the user to do ASR inference immediately after the `FilePartitioningStage` for every audio pipeline supported by Curator? If yes, then keeping the `CompositeStage` is okay with me. If no, then I agree we should separate the stages.

karpnv: Ok, I thought the composite Stage is just an example of how we can apply Stages one after another.

huvunvidia: We don't need Composite Stage for that. We just need to to define pipeline as a list of stages.
You can find an example here: https://github.com/NVIDIA-NeMo/Curator/blob/huvu/image_curation_2/ray-curator/ray_curator/examples/image/image_curation_example.py

sarahyurick:
Hi, audio dependencies need to be added to CI here: https://github.com/NVIDIA-NeMo/Curator/blob/ray-api/.github/workflows/cicd-main.yml#L149-L156

ray-curator/ray_curator/stages/audio/io/asr_inference.py
praateekmahajan: +1. I
I'm not too experienced with audio but its interesting to me that the first thing we do after deciding the list of files is run a DL model on top i.e. inference

ray-curator/ray_curator/examples/audio/fleurs_and_wer_example.py
```diff
@@ -0,0 +1,71 @@
...
+    # Add the composite stage that combines reading and downloading
+    pipeline.add_stage(
+        CreateInitialManifestFleursStage(
+            lang=args.lang,
+            split=args.split,
+            raw_data_dir=args.raw_data_dir,
+        )
+    )
+    pipeline.add_stage(
+        AsrNemoInferenceStage(model_name=args.model_name, batch_size=16, cuda="cuda", _resources=Resources(gpus=1.0))
```
abhinavg4: A good idea might be to do this instad:

`AsrNemoInferenceStage(model_name=args.model_name, cuda="cuda").with_(batch_size=16 resources=Resources(gpus=1.0)))`

ray-curator/ray_curator/stages/audio/io/asr_inference.py
```diff
@@ -0,0 +1,199 @@
...
+    """
+
+    model_name: str
+    filepath_key: str = "audio_filepath"
+    text_key: str = "text"
+    cuda: str = ""
+    name: str = "ASR_inference"
+    _start_time = time.time()
+    _metrics: ClassVar[dict] = {}
+    batch_size: int = 16
```
abhinavg4: batch_size should be `_batch_size`

ray-curator/ray_curator/stages/audio/io/asr_inference.py
```diff
@@ -0,0 +1,199 @@
...
+         Args:
+            files: list of audio file paths.
+
+        Returns:
+            list of predicted texts.
+        """
+        outputs = self.asr_model.transcribe(files)
+        return [output.text for output in outputs]
+
+    def process_batch(self, tasks: list[FileGroupTask | DocumentBatch | SpeechObject]) -> list[SpeechObject]:
```
sarahyurick: Hi, wanted to ask why it is `list[FileGroupTask | DocumentBatch | SpeechObject]` here?

karpnv: Because it can work with all of them:
FileGroupTask in case we seek files by extension
DocumentBatch in case we get a list of files from jsonl manifest file
SpeechObject in the current case

ray-curator/ray_curator/examples/audio/fleurs_and_wer_example.py
karpnv: done

ray-curator/ray_curator/stages/audio/io/asr_inference.py
karpnv: done

ray-curator/ray_curator/examples/audio/fleurs_and_wer_example.py
```diff
@@ -0,0 +1,71 @@
+from dataclasses import dataclass
+
+from loguru import logger
+from nemo.core.config import hydra_runner
+from omegaconf import OmegaConf
+
+from ray_curator.backends.xenna import XennaExecutor
+from ray_curator.pipeline import Pipeline
+from ray_curator.stages.audio.datasets.fleurs.create_initial_manifest import CreateInitialManifestFleursStage
```
sarahyurick: Hi, it looks like this file might be missing?

ray-curator/ray_curator/stages/audio/io/asr_inference.py
```diff
@@ -0,0 +1,199 @@
...
+
+from ray_curator.backends.base import NodeInfo, WorkerMetadata
+from ray_curator.stages.base import CompositeStage, ProcessingStage
+from ray_curator.stages.io.reader.file_partitioning import FilePartitioningStage
+from ray_curator.stages.resources import Resources
+from ray_curator.tasks import DocumentBatch, FileGroupTask, SpeechObject, _EmptyTask
+
+
+@dataclass
+class InferenceAsrNemoStage(ProcessingStage[FileGroupTask | DocumentBatch | SpeechObject, SpeechObject]):
```
huvunvidia: As mentioned in previous comment (https://github.com/NVIDIA-NeMo/Curator/pull/864#discussion_r2257483662), I think the data reading (io) and data processing should be separated. Currently, InferenceAsrNemoStage both reads and run asr inference.

You can look through image and video PRs for consistency:
- Image Curation: https://github.com/NVIDIA-NeMo/Curator/pull/880
- Video Curation: https://github.com/NVIDIA-NeMo/Curator/pull/847

ray-curator/ray_curator/examples/audio/fleurs_and_wer_example.py
```diff
@@ -0,0 +1,78 @@
+from dataclasses import dataclass
+
+from loguru import logger
+from nemo.core.config import hydra_runner
+from omegaconf import OmegaConf
```
sarahyurick: Does this mean we should add these as dependencies?

ray-curator/ray_curator/stages/audio/inference/asr_nemo.py
```diff
@@ -0,0 +1,145 @@
...
+
+@dataclass
+class InferenceAsrNemoStage(ProcessingStage[FileGroupTask | DocumentBatch | SpeechObject, SpeechObject]):
+    """Stage that do speech recognition inference using NeMo model.
+
+    Args:
+        model_name (str): name of the speech recognition NeMo model. See full list at https://docs.nvidia.com/nemo-framework/user-guide/latest/nemotoolkit/asr/all_chkpt.html
+        filepath_key (str): which key of the data object should be used to find the path to audiofile. Defaults to “audio_filepath”
+        pred_text_key (str): key is used to identify the field containing the predicted transcription associated with a particular audio sample. Defaults to “pred_text”
+        cuda (str): device to run inference on it. Could be cpu, cuda, ipu, xpu, mkldnn, opengl, opencl, ideep, hip, ve, fpga, maia, xla, lazy, vulkan, mps, meta, hpu, mtia or cuda number (digit). Defaults to “” (empty string)
```
sarahyurick:
```suggestion
        cuda (str): device to run inference on it. Could be cpu or cuda.
```
Would prefer not to allow the user to set a GPU number, etc. Also, I think having the `cuda` parameter and the `.with(Resources(...))` is unnecessary, just 1 should be needed. How about using `with`, and if GPUs > 0, then set the `map_location` accordingly?

ray-curator/ray_curator/stages/audio/inference/asr_nemo.py
```diff
@@ -0,0 +1,145 @@
...
+        name (str): Stage name. Defaults to "ASR_inference"
+    """
+
+    model_name: str
+    filepath_key: str = "audio_filepath"
+    pred_text_key: str = "pred_text"
+    cuda: str = ""
+    name: str = "ASR_inference"
```
sarahyurick:
```suggestion
    _name: str = "ASR_inference"
```
?

ray-curator/ray_curator/stages/audio/inference/asr_nemo.py
```diff
@@ -0,0 +1,145 @@
...
+    _start_time = time.time()
+    _metrics: ClassVar[dict] = {}
```
sarahyurick: These can be removed in favor of storing it in `_stage_perf`, right?

ray-curator/ray_curator/stages/audio/inference/asr_nemo.py
```diff
@@ -0,0 +1,145 @@
...
+            List of SpeechObject with self.filepath_key .
+            If errors occur, the task is returned with error information stored.
+        """
+
+        if isinstance(tasks[0], FileGroupTask):
+            files = [task.data[0] for task in tasks]
+        elif isinstance(tasks[0], DocumentBatch):
+            files = []
+            for task in tasks:
+                files.extend(list(task.data[self.filepath_key]))
```
sarahyurick: I think we can remove the `DocumentBatch` case. Otherwise it is misleading to use the text modality here.

ray-curator/ray_curator/stages/audio/inference/asr_nemo.py
```diff
@@ -0,0 +1,145 @@
...
+
+        Args:
+            tasks: List of FileGroupTask containing a path to audop file for inference.
+
+        Returns:
+            List of SpeechObject with self.filepath_key .
+            If errors occur, the task is returned with error information stored.
+        """
+
+        if isinstance(tasks[0], FileGroupTask):
```
sarahyurick: In addition to checking the first item of the list, can you add a check that all list items are the same type?

ray-curator/ray_curator/stages/audio/inference/asr_nemo.py
```diff
@@ -0,0 +1,145 @@
...
+        self.finalize()
+        return audio_tasks
+
+    def process(self, task: FileGroupTask) -> list[SpeechObject]:
+        pass
+
+    def finalize(self) -> None:
+        elapsed = time.time() - self._start_time
+        logger.info(f"Stage {self.name} completed in {elapsed:.2f} seconds.")
+        self._metrics["elapsed"] = elapsed
```
sarahyurick: Don't think we need this. Can be consolidated with `_stage_perf` instead.

ray-curator/ray_curator/stages/audio/io/common.py
```diff
@@ -0,0 +1,107 @@
+from abc import abstractmethod
+from dataclasses import dataclass
+from operator import eq, ge, gt, le, lt, ne
+
+import soundfile
```
sarahyurick: Do we need to add this dependency?

ray-curator/ray_curator/stages/audio/metrics/get_wer.py
```diff
@@ -0,0 +1,58 @@
+from dataclasses import dataclass
+
+import editdistance
```
sarahyurick: New dependency? Could we use `nltk` or another existing dependency instead?

ray-curator/ray_curator/stages/audio/datasets/fleurs/create_initial_manifest.py
```diff
@@ -0,0 +1,134 @@
...
+        return speech_tasks
+
+    def download_extract_files(self, dst_folder: str) -> None:
+        """downloading and extracting files"""
+
+        os.makedirs(dst_folder, exist_ok=True)
+
+        # downloading all files
+        for file_url in get_fleurs_url_list(self.lang, self.split):
+            download_file(file_url, str(dst_folder))
```
sarahyurick: A bit surprised that you didn't run into some type of concurrency issue here. What if you made downloading its own stage?

cc @praateekmahajan

ray-curator/ray_curator/stages/audio/io/common.py
sarahyurick, line 1: Maybe this shouldn't be in `io/`, the directory can be called something else (maybe `filters/`)?

ray-curator/ray_curator/examples/audio/fleurs_and_wer_example.py
karpnv: it is in nemo_toolkit[asr]'s dependencies

ray-curator/ray_curator/stages/audio/io/audio_reader.py
karpnv: Done

ray-curator/ray_curator/stages/audio/io/asr_inference.py
karpnv: done

ray-curator/ray_curator/stages/audio/io/asr_inference.py
karpnv: done

ray-curator/ray_curator/stages/audio/io/asr_inference.py
karpnv: removed

ray-curator/ray_curator/stages/audio/io/asr_inference.py
karpnv: I think it is impossible with NeMo

ray-curator/ray_curator/stages/audio/inference/asr_nemo.py
karpnv: done

ray-curator/ray_curator/stages/audio/inference/asr_nemo.py
karpnv: removed

ray-curator/ray_curator/stages/audio/inference/asr_nemo.py
karpnv: removed

ray-curator/ray_curator/stages/audio/io/common.py
karpnv: it is in nemo_toolkit[asr]'s dependencies

ray-curator/ray_curator/stages/audio/metrics/get_wer.py
karpnv: it is in nemo_toolkit[asr]'s dependencies

ray-curator/ray_curator/stages/audio/io/common.py
karpnv: moved out from io

ray-curator/ray_curator/stages/audio/inference/asr_nemo.py
karpnv: Our audio metadata is usually in JSONL format. Jsonl reader and jsonl writer both work with DocumentBatch. 
So, I believe we will use DocumentBatch in the newly created audio stages, while keeping SpeechObject for stages that came from SDP

ray-curator/ray_curator/stages/audio/inference/asr_nemo.py
karpnv: done

ray-curator/ray_curator/stages/audio/inference/asr_nemo.py
karpnv: renamed SpeechObject to DocumentObject

ray-curator/ray_curator/examples/audio/fleurs_and_wer.py
```diff
@@ -0,0 +1,80 @@
...
+    )
+    pipeline.add_stage(GetPairwiseWerStage(text_key="text", pred_text_key="pred_text", wer_key="wer"))
+    pipeline.add_stage(GetAudioDurationStage(audio_filepath_key="audio_filepath", duration_key="duration"))
+    pipeline.add_stage(PreserveByValueStage(input_value_key="wer", target_value=args.wer_threshold, operator="le"))
+    if args.output_manifest_file is not None:
+        pipeline.add_stage(WriteJsonlStage(output_manifest_file=args.output_manifest_file))
+    return pipeline
+
+
+@hydra_runner(config_name="TranscriptionConfig", schema=TranscriptionConfig)
```
huvunvidia: Is there a specific reason to use hydra_runner and TranscriptionConfig here, instead of regular argparse.ArgumentParser?
All other examples use argparse.ArgumentParser, so should keep it consistent across repo.

ray-curator/ray_curator/tasks/audio.py
```diff
@@ -0,0 +1,36 @@
+import os
+from dataclasses import dataclass
+
+from loguru import logger
+
+from .tasks import Task
+
+
+@dataclass
+class DataObject(Task[dict]):
```
huvunvidia: Can we change this to audio-specific (for example, AudioDataObject)?
Because the class is defined in /tasks/audio and specifically only used is audio pipeline, the name "DataObject" (i.e. from tasks import DataObject) might cause confusion.

ray-curator/ray_curator/stages/audio/datasets/fleurs/create_initial_manifest.py
huvunvidia: +1.
It seems like when this stage is run parallelly with multiple actors, they will be downloading and processing the same data and potentially overwrite each other.

You can either divide the dataset into chunks so actors can download parallelly. 
Or as Sarah suggested, make downloading its own stage and run it with only one actor. @abhinavg4 I remember you can set a stage to be run with only one actor, right. Can you share how? I believe we can set stage's method num_workers() to return 1. Is it correct?

ray-curator/ray_curator/stages/audio/inference/asr_nemo.py
```diff
@@ -0,0 +1,124 @@
...
+         Args:
+            files: list of audio file paths.
+
+        Returns:
+            list of predicted texts.
+        """
+        outputs = self.asr_model.transcribe(files)
+        return [output.text for output in outputs]
+
+    def process_batch(self, tasks: list[FileGroupTask | DocumentBatch | DataObject]) -> list[DataObject]:
```
huvunvidia: When overriding process_batch, we need to still keep self.validate_input() method.
https://github.com/NVIDIA-NeMo/Curator/blob/ca0b3b1b06be9cd48e9ff019a08d9674000a3c59/ray-curator/ray_curator/stages/base.py#L165

ray-curator/ray_curator/stages/audio/io/write_jsonl.py
```diff
@@ -0,0 +1,51 @@
...
+from ray_curator.tasks import DataObject, DocumentBatch
+
+
+class ObjectToBatchStage(ProcessingStage[DataObject, DocumentBatch]):
+    """
+    Stage to conver DocumentObject to DocumentBatch
+
+    """
+
+    def process_batch(self, tasks: list[DataObject]) -> list[DocumentBatch]:
```
huvunvidia: I wonder where this stage ObjectToBatchStage is used in the script.
If we keep it, let's make sure it also has the self.validate_input method when overriding process_batch().

ray-curator/ray_curator/tasks/audio.py
karpnv: It is not really audio-specific, just single data entry specific. I moved it to data_object.py

ray-curator/ray_curator/stages/audio/inference/asr_nemo.py
karpnv: added

ray-curator/ray_curator/stages/audio/io/write_jsonl.py
karpnv: added

ray-curator/ray_curator/stages/audio/datasets/fleurs/create_initial_manifest.py
karpnv: I can also download and extract it in setup_on_node(), right?

sarahyurick: Unfortunately no, because in a case like a multi-node Slurm job it would still attempt to download several times. I see 2 options for handling this:

1. Make download a stage as described above by Huy, or
2. Break out download into its own Python function for now, and just have it run as a separate thing before the pipeline is set up and run in the `examples/audio/fleurs_and_wer.py` script.

ray-curator/ray_curator/stages/audio/datasets/fleurs/create_initial_manifest.py
```diff
@@ -0,0 +1,139 @@
...
+            shutil.move(file_path, dest_file_path)
+            logger.info(f"Moved {file_path} to {dest_file_path}")
+        else:
+            os.remove(file_path)
+            logger.info(f"File {file_name} already exists in {target_folder}, deleted from source.")
+
+    def setup_on_node(
+        self, _node_info: NodeInfo | None = None, _worker_metadata: WorkerMetadata | None = None
+    ) -> None:
+        self.download_extract_files(self.raw_data_dir)
```
sarahyurick: Will need to be removed either way.

ray-curator/ray_curator/stages/audio/inference/asr_nemo.py
```diff
@@ -0,0 +1,127 @@
...
+    name: str = "ASR_inference"
+    _batch_size: int = 16
+    _resources: Resources = field(default_factory=lambda: Resources(cpus=1.0))
+
+    def check_cuda(self) -> torch.device:
+        return torch.device("cuda") if self.resources.gpus > 0 else torch.device("cpu")
+
+    def setup_on_node(self, _node_info: NodeInfo | None = None, _worker_metadata: WorkerMetadata = None) -> None:
+        # TODO: load asr_model file only once per node
+        pass
```
sarahyurick: Ideally this should be a no-op if the model is already downloaded. Curious why `huggingface_hub` does not work for you here?

ray-curator/ray_curator/stages/audio/inference/asr_nemo.py
```diff
@@ -0,0 +1,127 @@
...
+                msg = f"Task {task!s} failed validation for stage {self}"
+                raise ValueError(msg)
+            if isinstance(task, FileGroupTask):
+                files.append(task.data[0])
+            elif isinstance(task, DocumentBatch):
+                files.extend(list(task.data[self.filepath_key]))
+            elif isinstance(task, DataObject):
+                files.append(task.data[self.filepath_key])
+            else:
+                raise TypeError(str(task))
```
sarahyurick: I think we should verify that all tasks are the same type.

ray-curator/ray_curator/stages/audio/io/write_jsonl.py
sarahyurick: +1 not sure this is being used anywhere.

ray-curator/ray_curator/stages/audio/io/write_jsonl.py
```diff
@@ -0,0 +1,58 @@
...
+                dataset_name="ObjectToBatch",
+            )
+        ]
+
+    def process(self, _: DataObject) -> None:
+        pass
+
+
+@dataclass
+class WriteJsonlStage(ProcessingStage[DataObject, DataObject]):
```
sarahyurick: I am worried about concurrency here too. For each task it writes out a single file, or it writes all tasks as a single file?

ray-curator/ray_curator/tasks/data_object.py
```diff
@@ -0,0 +1,27 @@
+import os
+from dataclasses import dataclass
+
+from loguru import logger
+
+from .tasks import Task
+
+
+@dataclass
+class DataObject(Task[dict]):
```
sarahyurick: I can see why why this is generic enough to call it a `DataObject` instead of audio, but would still prefer to have a child of `Task` that is explicitly called `AudioBatch` or `AudioTask`. I would really like every modality (text, image, audio, video) to each have its own task type.

ray-curator/ray_curator/stages/audio/datasets/fleurs/create_initial_manifest.py
karpnv: Only downloading step needs to be moved to another stage or extracting too?

ray-curator/ray_curator/stages/audio/io/write_jsonl.py
karpnv: Any other way to save it into one JSONL file?

ray-curator/ray_curator/tasks/data_object.py
karpnv: Even if AudioBatch will be exactly the same as DocumentBatch? Currently, this is going to happen

karpnv:
added num_workers()  return 1  to CreateInitialManifestFleursStage

ray-curator/ray_curator/stages/audio/datasets/fleurs/create_initial_manifest.py
```diff
@@ -0,0 +1,132 @@
...
+    lang: str
+    split: str
+    raw_data_dir: str
+    filepath_key: str = "audio_filepath"
+    text_key: str = "text"
+    _name: str = "CreateInitialManifestFleurs"
+    _batch_size: int = 1
+
+    def num_workers(self) -> int:
+        return 1
```
sarahyurick: Would like to check that this works as expected.

ray-curator/ray_curator/stages/audio/inference/asr_nemo.py
sarahyurick: Bumping this.

ray-curator/ray_curator/stages/audio/datasets/fleurs/create_initial_manifest.py
```diff
@@ -0,0 +1,132 @@
...
+                parts = line.strip().split("\t")
+                if len(parts) < min_num_parts:
+                    continue
+
+                file_name, transcript_text = parts[1], parts[2]
+                wav_file = os.path.join(root, file_name)
+
+                entries.append({self.filepath_key: os.path.abspath(wav_file), self.text_key: transcript_text})
+                count += 1
+                if count == self._batch_size:
```
sarahyurick: Can this be accessed with `self.batch_size`?

ray-curator/ray_curator/stages/audio/inference/asr_nemo.py
sarahyurick: Bumping this.

ray-curator/ray_curator/stages/audio/io/convert.py
```diff
@@ -0,0 +1,20 @@
...
+    Stage to conver DocumentObject to DocumentBatch
+
+    """
+
+    def process(self, task: AudioBatch) -> list[DocumentBatch]:
+        return [
+            DocumentBatch(
+                data=pd.DataFrame(task.data),
+                task_id="",
+                dataset_name="ObjectToBatch",
```
sarahyurick: Nit: Call it `AudioToDocument` or something similar?

karpnv: done

ray-curator/ray_curator/stages/audio/datasets/fleurs/create_initial_manifest.py
karpnv: yes

ray-curator/ray_curator/stages/audio/datasets/fleurs/create_initial_manifest.py
karpnv: removed

huvunvidia (approved):
LGTM!
Please address the failed tests.
Thank you.

ray-curator/ray_curator/examples/audio/fleurs/pipeline.py
```diff
@@ -0,0 +1,80 @@
...
+    pipeline.add_stage(GetPairwiseWerStage(text_key="text", pred_text_key="pred_text", wer_key="wer"))
+    pipeline.add_stage(GetAudioDurationStage(audio_filepath_key="audio_filepath", duration_key="duration"))
+    pipeline.add_stage(PreserveByValueStage(input_value_key="wer", target_value=args.wer_threshold, operator="le"))
+    pipeline.add_stage(AudioToDocumentStage().with_(batch_size=1))
+    result_dir = os.path.join(args.raw_data_dir, "result")
+    if os.path.isdir(result_dir):
+        shutil.rmtree(result_dir)  # clean up resulting folder
+    pipeline.add_stage(
+        JsonlWriter(
+            output_dir=result_dir,
```
sarahyurick:
```suggestion
            path=result_dir,
```
Code needs to be updated here.

ray-curator/ray_curator/examples/audio/fleurs/pipeline.yaml
```diff
@@ -0,0 +1,68 @@
...
+
+  - _target_: ray_curator.stages.audio.common.PreserveByValueStage
+    input_value_key: "wer"
+    target_value: 75.
+    operator: "le"
+
+  - _target_: ray_curator.stages.audio.io.convert.AudioToDocumentStage
+
+  - _target_: ray_curator.stages.text.io.writer.JsonlWriter
+    output_dir: ${output_dir}
```
sarahyurick: Same here, this needs to be updated.

karpnv: done

ray-curator/ray_curator/examples/audio/fleurs/pipeline.py
karpnv: doen

karpnv:
/ok to test [ae09a7a](https://github.com/NVIDIA-NeMo/Curator/pull/864/commits/ae09a7a09f9f5299c1a0cd9fc484548198688b2e)

karpnv:
/ok to test [fb2080c8](https://github.com/NVIDIA-NeMo/Curator/pull/864/commits/fb2080c8b6ddc252a820af316cee74e209320c34)

karpnv:
/ok to test [fb2080c](https://github.com/NVIDIA-NeMo/Curator/pull/864/commits/fb2080c8b6ddc252a820af316cee74e209320c34)

sarahyurick (approved):

PR #876, opened 2025-08-04
[RAY] Add ID Module

VibhuJawa:
## Description
This PR adds a initial implementation of add id. 

### TODO:

- [x] Add tests
- [x] Test robustness at multi GB scale
- [x] Add task scaling results

## Dataset Benchmark

| Dataset Size | Machine Setup | Dask (seconds) | Xenna (seconds) | Ray Data (Tasks) (seconds) |
|--------------|----------------|----------------|------------------|-----------------------------|
| Small        | 64 CPUs         | 10.92          | 11.168           | 19.182                      |
| Medium       | 64 CPUs         | 1846.003       | 465.714          | 632.264                     |


ray-curator/ray_curator/tasks/tasks.py
```diff
@@ -26,6 +27,7 @@ class Task(ABC, Generic[T]):
     data: T
     _stage_perf: list[StagePerfStats] = field(default_factory=list)
     _metadata: dict[str, Any] = field(default_factory=dict)
+    _uuid: str = field(init=False, default_factory=lambda: str(uuid.uuid4()))
```
praateekmahajan, line 30: that's a good idea, maybe in future we can just nuke `task_id`.. Can you add a simple test to test_tasks.py (might not exist) which ensures that in a simple fanout stage where 

```python
def process(self, task : X) -> list[X]):
    return [x] * 10
```

all the newly generated tasks have distinct uuids. This ensures that users have no way to shoot themselves in the food (because if the above isn't distinct then we don't have much benefits over task_id iiuc)

praateekmahajan (commented):
thanks this is great! 
can you share what the pipeline is? I'm surprised to see the delta at large scale between  
1. dask vs (ray data / xenna).. 
    - its a simple map style operation, maybe dask should've been similar to ray data (tasks) because they'll operate the same way i.e. fuse all stages and execute them for each task in parallel.. the only advantage i can think ray would have is scheduler might be nicer for many small tasks, but still delta is huge
3. ray data vs xenna.. 
    - primarily because ray data should've fused read / add_id / write (assuming you had a three stage pipeline) so the i/o overhead should've been minimal and hence faster

ray-curator/ray_curator/stages/modules/add_id.py
```diff
@@ -0,0 +1,51 @@
...
+    The module responsible for adding unique IDs to each document record.
+
+    This stage adds a unique identifier to each document in the batch by combining
+    the task UUID with a sequential index.
+
+    Args:
+        id_field (str): The field where the generated ID will be stored.
+    """
+
+    id_field: str
```
ayushdg, line 23: Could we also support adding a user specified prefix to the ID. It's fairly common for folks to prefix with dataset-name or some other naming scheme. The id could be in the format `{user-prefix}_{uuid}_{i}` for each row

ray-curator/ray_curator/tasks/tasks.py
ayushdg: ➕ 1 on this. Would be good to possibly replace `task_id` with this attribute instead.

ray-curator/ray_curator/stages/modules/add_id.py
```diff
@@ -0,0 +1,51 @@
...
+
+        Args:
+            batch (DocumentBatch): The batch to add IDs to
+
+        Returns:
+            DocumentBatch: A batch with unique IDs added to each document
+        """
+        df = batch.to_pandas()
+        prefix = str(batch._uuid)
+        df[self.id_field] = [f"{prefix}_{i}" for i in range(len(df))]
```
ayushdg, line 60: Nit: Should we check if this column already exists and raise (unless users want to overwrite) or warn?

ayushdg (changes requested):
Generally looks good! Minor comments inline

ray-curator/ray_curator/tasks/tasks.py
sarahyurick: Would it be beneficial to propagate the `task_id` through the entire pipeline for any reason (as we are doing now)? If not then I agree it would be nice to get rid of it. Plus, we already have `dataset_name` too.

It may be easier to remove `task_id` in a follow-up PR though.

Nit: The docstring should be updated to reflect the `_metadata` and `_uuid` attributes.

ray-curator/ray_curator/stages/modules/add_id.py
VibhuJawa: Makes sense. Let me add that.

ray-curator/ray_curator/stages/modules/add_id.py
VibhuJawa: Let me add this

VibhuJawa:
> can you share what the pipeline is? I'm surprised to see the delta at large scale between


Dask: https://gist.github.com/VibhuJawa/e193230b25437d11872e78670f6a61e2#file-add_id_dask-py

Ray: https://gist.github.com/VibhuJawa/e193230b25437d11872e78670f6a61e2#file-add_id_ray-py

Yeah, i dont understand ray enough to know we perform poorly with xenna against ray data.

ray-curator/ray_curator/tasks/tasks.py
VibhuJawa: Happy to remove `task_id` in a follow up PR.

VibhuJawa: > 
> def process(self, task : X) -> list[X]):
>     return [x] * 10
> all the newly generated tasks have distinct uuids. This ensures that users have no way to shoot themselves in the > food (because if the above isn't distinct then we don't have much benefits over task_id iiuc)


In the current setup, if we split out a task using `[task for i in range(self.times)]`, each copy shares the same id. However, if we instantiate a new object (e.g., using replace()), each copy gets a unique id.

There are two questions we should consider here:

1. Should we enforce a contract that all tasks are unique?

In that case, fan-out scenarios like the above (repeating the same task) may be considered unrealistic. I’d appreciate your input on whether that assumption makes sense given how the codebase is structured.

2. Should we generate a new unique id in the add_id stage on the fly?

This would ensure that any task going through add_id ends up with a unique id — even if it started as a shallow copy. Note: _uuid would remain the same, but the data itself might still differ.

Both options are straightforward to implement, but `Option 1` might be more meaningful — since it guarantees deterministic output for a given task.  `Option 2` ensures it works independently of the `task_id`  

My only motivation for switching from task_id to a UUID is to minimize user-generated ID bugs — having it generated internally reduces the chances of foot guns (and i actually ran into this in my testing so thats a data point of 1)  

I am gonna rely on your all understanding of the user contracts we want to make here to choose b/w options. 


Example code for us to consider a valid stage or not. 

```python3
from dataclasses import dataclass, replace

from ray_curator.stages.base import ProcessingStage
from ray_curator.tasks import Task


@dataclass
class Repeat(ProcessingStage[Task, Task]):
    """
    Dummy stage that returns *n* identical (shallow-copied) versions of the
    incoming task.  Useful for testing fan-out behaviour.

    Args:
        times (int): How many copies to emit.  Defaults to 10.
    """

    times: int = 10
    _name: str = "repeat"

        return [], []

        return [], []

    def process(self, task: Task) -> list[Task]:
        """Return ``times`` copies of *task*."""
        return [task. for i in range(self.times)]
        #return [DocumentBatch(task_id=task.task_id,
        #                      data=task.to_pandas(),
        #                      dataset_name=task.dataset_name) for i in range(self.times)]
        
        
def _sample_batch() -> DocumentBatch:
    df = pd.DataFrame({"text": ["a", "b", "c"]})
    return DocumentBatch(task_id="b1", dataset_name="dummy", data=df)


batch = _sample_batch()
stage = Repeat(times=3)
output = stage.process(batch)
assert len(set([t._uuid for t in output]))==3    
```

ray-curator/ray_curator/stages/modules/add_id.py
VibhuJawa: Added

ray-curator/ray_curator/stages/modules/add_id.py
VibhuJawa: Added

VibhuJawa:
@ayushdg / @praateekmahajan / @sarahyurick  ready for another look.

ray-curator/tests/stages/io/writer/test_jsonl.py
```diff
@@ -48,10 +48,12 @@ def test_jsonl_writer(
                 assert mock_get_deterministic_hash.call_count == 1
                 # Verify get_deterministic_hash was called with correct arguments
                 mock_get_deterministic_hash.assert_called_once_with(source_files, document_batch.task_id)
-                assert mock_uuid4.call_count == 0
+                # because we call it once for task, and that should be the only one
+                assert mock_uuid4.call_count <= 1
```
praateekmahajan, line 52: super nit: might as well check mock_uuid.calls and make sure atleast one call is for the old existing stuff and one for the Task

sarahyurick (approved):

PR #879, opened 2025-08-04
Add Ray curator dockerfile and enable testing

thomasdhc:
## Usage

```python
```
## Checklist


ray-curator/pyproject.toml
```diff
@@ -65,6 +65,11 @@ video = [
     "opencv-python",
 ]
 
+all = [
+    "ray_curator[text]",
+    "ray_curator[video]",
```
praateekmahajan, line 77: let's also add cuda12x because that's where the real check would be... I'm concerned since we don't have GPU tests we might not be able to test things thoroughly whether this docker image indeed works.. If anyone (@sarahyurick / @ayushdg ) have good ideas what to do in the mean time to ensure this docker image is good wrt rapids imports pls opine

sarahyurick: Yes, agree about the `cuda12x` dependencies. The only text GPU tests we have at the moment for Ray Curator are in https://github.com/NVIDIA-NeMo/Curator/pull/881, which I have run locally but not on GitHub.

We may want to merge `ray-api` into `main` soon, if there is no other work planned on `main` anyway... thoughts?

.github/workflows/cicd-main.yml
```diff
@@ -160,206 +160,208 @@ jobs:
...
-#          azure-subscription-id: ${{ secrets.AZURE_SUBSCRIPTION_ID }}
-#          PAT: ${{ secrets.PAT }}
-#          timeout: 20
+  cicd-gpu-tests:
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - script: L0_Unit_Test_GPU
+            runner: linux-amd64-gpu-rtxa6000-latest-2-nemo
```
sarahyurick: Guessing this should be updated to reflect https://github.com/NVIDIA-NeMo/Curator/pull/884?

ray-curator/Dockerfile
```diff
@@ -0,0 +1,11 @@
...
+# NeMo Curator requires Python 3.12, Ubuntu 22.04/20.04, and CUDA 12 (or above)
+ARG CUDA_VER=12.8.1
+ARG LINUX_VER=ubuntu24.04
+
+FROM nvidia/cuda:${CUDA_VER}-cudnn-devel-ubuntu${LINUX_VER}
+
+COPY . .
+
+RUN cd ray-curator && \
+    pip install -e ".[all]"
```
sarahyurick: If `cuda12x` dependencies are added then this will need `--extra-index-url https://pypi.nvidia.com`.

ayushdg: @suiyoubi can confirm but video may need additional system level deps before those pipeline tests pass.

Dockerfile
```diff
@@ -1,11 +1,44 @@
...
 # NeMo Curator requires Python 3.12, Ubuntu 22.04/20.04, and CUDA 12 (or above)
-ARG CUDA_VER=12.8.1
-ARG LINUX_VER=ubuntu24.04
+ARG CUDA_VER=12.5.1
+ARG LINUX_VER=ubuntu22.04
+ARG PYTHON_VER=3.12
+ARG IMAGE_LABEL
 
-FROM nvidia/cuda:${CUDA_VER}-cudnn-devel-ubuntu${LINUX_VER}
+FROM rapidsai/ci-conda:cuda${CUDA_VER}-${LINUX_VER}-py${PYTHON_VER} AS curator-update
```
praateekmahajan: i know the PR isn't ready for review but just as an FYI, rapidsai/ci-conda is not recommended to be used as the base image for non-ci things

Dockerfile
```diff
@@ -1,11 +1,44 @@
...
+# Install the minimal libcu* libraries needed by NeMo Curator
+RUN conda create -y --name curator -c nvidia/label/cuda-${CUDA_VER} -c conda-forge \
+  python=3.12 \
+  cuda-cudart \
+  libcufft \
+  libcublas \
+  libcurand \
+  libcusparse \
+  libcusolver \
+  cuda-nvvm && \
```
praateekmahajan: i believe in theory we should never need this

thomasdhc:
@praateekmahajan The new dockerfile is under ray-curator/Dockerfile please refresh the PR

ray-curator/Dockerfile
```diff
@@ -0,0 +1,78 @@
...
+    python3 \
+    python3-dev \
+    python3-venv \
+    python-is-python3\
+    curl \
+    ffmpeg \
+    git \
+    libsm6 \
+    libxext6 && \
+    rm -rf /var/lib/apt/lists/*
```
praateekmahajan: WDYT of using pixi? I believe in the near future we might want to install non pip things at system level which pixi might be able to help with that.. 
Similarly for our tutorials, we expect users to do bunch of apt-get install, and I'm hoping in future these tutorials could be its own sub environment
an e.g. here
https://github.com/NVIDIA-NeMo/Curator/blob/10caa668acdebf7f36c4eab0ce371a3178a9a9f2/tutorials/dapt-curation/README.md?plain=1#L50-L61

I understand this might require more thought so wanted to start this discussion

thomasdhc: I'm not too familiar with pixi, but I can look into it.
Are you saying pixi is like a venv on the system level?

Regarding these separate installation tutorials, is the expectation that those will be included in the curator dockerfile?

ray-curator/Dockerfile
```diff
@@ -0,0 +1,78 @@
...
+
+FROM curator_dep AS nemo_curator
+
+# Install NeMo-Curator
+# ARG TARGET_ARCH
+# ARG CURATOR_COMMIT
+# ARG CURATOR_VERSION
+
+# ENV NEMO_FW_CURATOR_COMMIT=$CURATOR_COMMIT
+# ENV NEMO_FW_CURATOR_VERSION=$CURATOR_VERSION
```
sarahyurick: Are these still relevant?

thomasdhc: They might be. On the FW side we have environment variables that indicate with version and commit of the software uses are on. This was a frequent customer request.

I think it'll be good to have on the curator side as well.

ray-curator/Dockerfile
praateekmahajan: Hey sorry i didn't respond to this but I see you're trying pixi which is awesome.

> is the expectation that those will be included in the curator dockerfile 
No, not what I was thinking. but rather make the user install. 

I don't know a lot about pixi, but here is what i have in mind and happy to brainstorm

1. pixi can be used to do pip + system installations (whether from conda or apt-get or only conda?)
   - i wonder for our own package that we publish to pypi how does it handle system installations?
3. pixi uses uv for python
4. i'm wondering if it generates lock files (given it uses uv, maybe it does)
5. if it does have concept of lock files, does it have a concept of environments?
6. if so, for our tutorials, i wonder if we can create environments and share the lock / enviornment.yaml such that a user
   - who had done a `pip / pixi install` of our repo, if they decide to try out a tutorial  using, they can just do another `pip / pixi` install using the `env.yaml` or `.lock` file
   - who were using a docker container, can also just `cd` into the tutorial and do a `pip / pixi` install using the `env.yaml` or `.lock` file

ray-curator/Dockerfile
```diff
@@ -0,0 +1,77 @@
...
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL
+# THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+# FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+# DEALINGS IN THE SOFTWARE.
+
+ARG CUDA_VER=12.8.1
+ARG LINUX_VER=ubuntu24.04
+
+FROM nvidia/cuda:${CUDA_VER}-cudnn-devel-${LINUX_VER} AS cuda
```
sarahyurick, line 24: I am still curious about the best base image to use. Not `devel`?

thomasdhc: We can always change the base image

Which do you think is more appropriate?
https://hub.docker.com/r/nvidia/cuda/tags?name=12.8.1

sarahyurick: Thanks for sharing the link. I was wondering if we can get away with `nvidia/cuda:12.8.1-base-ubuntu22.04`, or is this too minimal for PyTorch?

Maybe `nvidia/cuda:12.8.1-runtime-ubuntu22.04` would be the sweet spot.

.github/actions/test-template/action.yml
```diff
@@ -225,17 +224,17 @@ runs:
...
-    - name: Upload artifacts
-      uses: actions/upload-artifact@v4
-      if: ${{ steps.check.outputs.coverage_report != 'none' }}
-      with:
-        name: ${{ steps.check.outputs.coverage_report }}
-        path: |
-          coverage.xml
-          .coverage
-        include-hidden-files: true
+    # - name: Test coverage
```
sarahyurick, line 227: Are we able to uncomment?

thomasdhc: Let me do it in a follow up, the gpu tests don't have coverage just yet.

I'll have to update some of these commands to run with the pixi environment

sarahyurick: Got it, okay with me.

thomasdhc: Thanks!

ray-curator/Dockerfile
sarahyurick: @thomasdhc could you try with each of these and see if they work?

thomasdhc: There's no way to discern functionality with the different base images at the moment. I think the installation will be successful. Seems like devel has the most amount of features. I think when we have more gpu tests in place we can experiment with the different base containers to find the minimal one.

Let's also do this in a follow-up, don't think it should block this PR.

sarahyurick: @thomasdhc I think we have aligned to use `runtime`.

ray-curator/pyproject.toml
```diff
@@ -69,8 +71,36 @@ video = [
     "opencv-python",
 ]
 
+all = [
+    "ray_curator[cuda12x]",
```
ayushdg: This needs to be updated to `deduplication_cuda12x` with #837.

PR #881, opened 2025-08-05
[Ray] Classifier tests

sarahyurick:
Tests for https://github.com/NVIDIA-NeMo/Curator/pull/753. Note that the model stage requires GPU testing to be enabled.

We should add more generic tests for `TokenizerStage`, `ModelStage`, etc.

### DIFF
--- a/ray-curator/tests/stages/text/classifiers/test_classifiers.py
+++ b/ray-curator/tests/stages/text/classifiers/test_classifiers.py
@@ -1,0 +1,423 @@
+# Copyright (c) 2025, NVIDIA CORPORATION.  All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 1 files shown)


VibhuJawa (approved):
Should we delete the https://github.com/NVIDIA-NeMo/Curator/blob/main/tests/test_classifiers.py as part of this PR? 

I think now is as good a time to start deleting stuff . Other than that this PR looks good to me.

sarahyurick:
Thanks @VibhuJawa ! Sorry for the confusion, I deleted the `test_classifiers.py` file in https://github.com/NVIDIA-NeMo/Curator/pull/753. I probably should have held off and waited to delete it in this PR for better tracking.

VibhuJawa (approved):
LGTM, thanks for adding tests around padding, naming of stages etc too.

sarahyurick:
Some type of Hugging Face connection error is causing the GPU CI to fail: https://github.com/NVIDIA-NeMo/Curator/actions/runs/16995881978/job/48186855579

I added skips for now. Verified that the tests work locally.

PR #890, opened 2025-08-08
Initial Shuffle + LSH PR

ayushdg:
## Usage

```python
```
## Checklist


sarahyurick:
Conflict is because the directory structure is now `stages/text/io/reader`. Sorry for the inconvenience.

ray-curator/ray_curator/backends/experimental/ray_actor_pool/executor.py
```diff
@@ -51,7 +53,7 @@ def execute(self, stages: list["ProcessingStage"], initial_tasks: list[Task] | N
         try:
             # Initialize Ray and register loguru serializer
             register_loguru_serializer()
-            ray.init(ignore_reinit_error=True)
+            ray.init(ignore_reinit_error=True, runtime_env=self.config.get("runtime_env", None))
```
praateekmahajan, line 57: Curious why this?

ayushdg: Makes it easier to set UCX specific config within the executor config which might be necessary debugging/some system setups which have different networking.

The alternative might be  to have the user call `ray.init` with these args

ayushdg:
ci tests here are failing with `No module named 'ray_curator'` which is a bit odd. @thomasdhc any suggestions?

thomasdhc:
@ayushdg I'm taking a look

ray-curator/ray_curator/backends/experimental/ray_actor_pool/executor.py
```diff
@@ -166,6 +171,49 @@ def _create_raft_actor_pool(self, stage: "ProcessingStage", num_actors: int, ses
...
+        for actor_idx in range(num_actors):
+            actor = ShuffleStageAdapter.options(
+                num_cpus=stage.resources.cpus,
+                num_gpus=stage.resources.gpus,
+                name=f"{stage.name}-Worker_{actor_idx}",
+            ).remote(stage=stage, rank=actor_idx, nranks=num_actors, num_input_tasks=num_tasks)
+            actors.append(actor)
+
+        # Setup UCXX communication
+        logger.info("    Setting up UCXX communication...")
```
VibhuJawa, line 190:
```suggestion
        logger.info("Setting up UCXX communication...")
```

ray-curator/ray_curator/backends/experimental/ray_actor_pool/executor.py
```diff
@@ -166,6 +171,49 @@ def _create_raft_actor_pool(self, stage: "ProcessingStage", num_actors: int, ses
...
+
+        # initialize the first actor as the root remotely in the cluster
+        root_address_bytes = ray.get(actors[0].setup_root.remote())
+
+        # setup the workers in the cluster including root
+        ray.get([actor.setup_worker.remote(root_address_bytes) for actor in actors])
+
+        logger.info("    UCXX setup complete")
```
VibhuJawa, line 198:
```suggestion
        logger.info("UCXX setup complete")
```

ray-curator/ray_curator/stages/deduplication/gpu_utils.py
```diff
@@ -0,0 +1,5 @@
+def align_memory(memory_size: int) -> int:
```
VibhuJawa: Will suggest renaming function or adding an arg to the multiple of alignment as we are aligning down to 256 and not the nearest multiple if you may. 

```suggestion
def align_down_to_256(memory_size: int) -> int:
```

ray-curator/ray_curator/stages/deduplication/fuzzy/lsh/lsh.py
```diff
@@ -0,0 +1,295 @@
...
+        Number of minhashes per band.
+    id_column
+        Name of the ID column in input data.
+    minhash_field
+        Name of the minhash field in input data.
+    output_path
+        Path to write output files.
+    rmm_pool_size
+        Size of the RMM memory pool in bytes.
+    spill_device
```
VibhuJawa: Can we create a better name for this to relfect that we need to pass in a limit , also improve docstring to tell what are options here etc

ray-curator/ray_curator/stages/shuffler/rapidsmpf_shuffler.py
```diff
@@ -0,0 +1,247 @@
...
+        Number of ranks in the communication group.
+    total_nparts
+        Total number of output partitions.
+    shuffle_on
+        List of column names to shuffle on.
+    output_path
+        Path to write output files.
+    rmm_pool_size
+        Size of the RMM memory pool in bytes.
+    spill_device
```
VibhuJawa: Add doc string documenting auto behavior

VibhuJawa (commented):
Took an intial look, still understanding the logic behind it, added some comments around documentation that i found confusing

ray-curator/ray_curator/backends/experimental/ray_actor_pool/executor.py
ayushdg: Following the same indentation pattern as the Raft Actors: https://github.com/NVIDIA-NeMo/Curator/blob/f39e4c12d81f44e738e58fcd9f979ee68e88a1e2/ray-curator/ray_curator/backends/experimental/ray_actor_pool/executor.py#L160

@praateekmahajan I presume this is for better visual appearance when reading logs?

ray-curator/ray_curator/backends/experimental/ray_actor_pool/executor.py
```diff
@@ -194,15 +239,98 @@ def _process_stage_with_pool(
...
                 ray.get(actor.teardown.remote())
                 ray.kill(actor)
             except (ray.exceptions.RayActorError, ray.exceptions.RaySystemError) as e:  # noqa: PERF203
                 logger.warning(f"      Warning: Error cleaning up actor {i}: {e}")
+
+    def _cleanup_actor_pool(self, actor_pool: ActorPool) -> None:
+        """Clean up actors in the pool."""
+
+        # Get all actors from the pool
+        all_actors = list(actor_pool._idle_actors) + [actor for actor, _ in actor_pool._future_to_actor.items()]
```
sarahyurick, line 319: I was wondering if we are worried about race conditions here?

ray-curator/ray_curator/backends/experimental/ray_actor_pool/shuffle_adapter.py
```diff
@@ -0,0 +1,123 @@
...
+            self._batch_size = 1
+
+        # Auto set total_nparts if not set
+        stage_class_kwargs = self.stage.actor_kwargs.copy()
+        if stage_class_kwargs.get("total_nparts") is None:
+            if num_input_tasks is None:
+                err_msg = "Shuffle total_nparts could not be set automatically. Please set it manually during stage initialization."
+                raise ValueError(err_msg)
+
+            self.output_nparts = max(1, 2 ** math.floor(math.log2(num_input_tasks)))
```
sarahyurick, line 62: Could you explain more where this calculation is coming from?

ray-curator/ray_curator/backends/experimental/ray_actor_pool/shuffle_adapter.py
```diff
@@ -0,0 +1,123 @@
...
+        _, self.root_address = self.stage._actor_obj.setup_root()
+        return self.root_address
+
+    def setup_worker(self, root_address: bytes) -> None:
+        """Setup UCXX communication."""
+        if self.root_address is None:
+            self.root_address = root_address
+        elif self.root_address != root_address:
+            logger.warning(f"Root address mismatch during worker setup: {self.root_address} != {root_address}")
+        self.stage._actor_obj.setup_worker(root_address)
```
sarahyurick: Is this okay to just be a warning (as opposed to an error)? Also, should a re-assignment between `self.root_address` and `root_address` happen in this case?

ray-curator/ray_curator/backends/experimental/ray_actor_pool/shuffle_adapter.py
```diff
@@ -0,0 +1,123 @@
...
+    def setup_on_node(self) -> None:
+        """
+        Note: This method is not used in the current implementation since we use
+        the Ray Data pattern of calling setup_on_node before actor creation.
+        """
+        super().setup_on_node(self.node_info, self.worker_metadata)
+
+    def setup(self) -> None:
+        """Note: This method is unused"""
+        super().setup(worker_metadata=self.worker_metadata)
```
sarahyurick: Are we able to remove?

ray-curator/ray_curator/stages/deduplication/fuzzy/lsh/lsh.py
```diff
@@ -0,0 +1,299 @@
...
+
+        if band_range[0] < 0 or band_range[1] > self.num_bands or band_range[0] >= band_range[1]:
+            msg = f"Invalid band range: {band_range}, must be in range [0, {self.num_bands})"
+            raise ValueError(msg)
+
+        # Process files in batches
+        minhash_df = self.read_minhash(filepaths)
+        # Skip processing if the batch is empty
+        if minhash_df is None or len(minhash_df) == 0:
+            print("Skipping empty batch")
```
sarahyurick: Use logger?

ray-curator/ray_curator/stages/deduplication/fuzzy/lsh/lsh.py
```diff
@@ -0,0 +1,299 @@
...
+        """
+        # Get consistent column names for all batches
+        column_names = [self.id_column, "_bucket_id"]
+
+        if not filepaths:
+            # Return column names even if no paths were provided
+            return column_names
+
+        if band_range[0] < 0 or band_range[1] > self.num_bands or band_range[0] >= band_range[1]:
+            msg = f"Invalid band range: {band_range}, must be in range [0, {self.num_bands})"
```
sarahyurick:
```suggestion
            msg = f"Invalid band range: {band_range}, must be in range [0, {self.num_bands}]"
```
nit?

ray-curator/ray_curator/stages/shuffler/rapidsmpf_shuffler.py
```diff
@@ -0,0 +1,249 @@
...
+        Setup the UCXX communication and a shuffle operation.
+
+        Parameters
+        ----------
+        root_address_bytes
+            Address of the root worker for UCXX initialization.
+        """
+        super().setup_worker(root_address_bytes)
+
+        # Initialize the RMM memory resource
```
sarahyurick, line 92: Should there be some type of cleanup before initializing?

ray-curator/ray_curator/backends/experimental/ray_actor_pool/executor.py
ayushdg: I've left this logic as is from #893. I'll let @praateekmahajan chime in as well.

In theory it is possible that if idle actors are changing while this line is executed there could be race conditions. In practice it's less likely since all actors should have completed processing by this point since this executor waits for all outputs to materialize before reaching this step.

ray-curator/ray_curator/backends/experimental/ray_actor_pool/shuffle_adapter.py
ayushdg: This has been carried over from the old logic here: https://github.com/NVIDIA-NeMo/Curator/blob/ray-api/nemo_curator/modules/fuzzy_dedup/lsh.py#L182. 

The main motivation for this was to optimize the shuffle perf which is sensitive to num output partitions. The actual math is based more on empirical evidence where we could get away by rounding down to the nearest power of 2 (generally) without running into OOM issues at scale.

ray-curator/ray_curator/backends/experimental/ray_actor_pool/shuffle_adapter.py
ayushdg: Updated to error out if there is a mismatch.

ray-curator/ray_curator/backends/experimental/ray_actor_pool/shuffle_adapter.py
ayushdg: Replaced with a not implemented Error in case it gets accidentally called.

ray-curator/ray_curator/stages/shuffler/rapidsmpf_shuffler.py
ayushdg: This should be a fresh process so no cleanup is needed. Might look into some cleanup if issues come up in the future

sarahyurick (approved):
Lgtm, thank you!

ray-curator/ray_curator/backends/experimental/ray_actor_pool/executor.py
```diff
@@ -182,18 +236,17 @@ def _process_stage_with_pool(
...
                 logger.warning(
                     f"Stage {_stage.name} is a RAFT stage but has a batch size of {stage_batch_size}. Ignoring batch size."
                 )
             num_actors = len(actor_pool._idle_actors)
-            task_batches = [batch.tolist() for batch in np.array_split(tasks, num_actors) if len(batch) > 0]
+            task_batches = self._generate_task_batches(tasks, num_output_tasks=num_actors)
         else:
             # For non-RAFT stages, we batch it based on the stage batch size
-            task_batches = [tasks[i : i + stage_batch_size] for i in range(0, len(tasks), stage_batch_size)]
+            task_batches = self._generate_task_batches(tasks, stage_batch_size)
```
praateekmahajan: nit
```suggestion
            task_batches = self._generate_task_batches(tasks, batch_size=stage_batch_size)
```

ray-curator/ray_curator/backends/experimental/ray_actor_pool/executor.py
```diff
@@ -214,15 +267,98 @@ def _process_stage_with_pool(
...
+            tasks: List of Task objects to process
+            band_range: Band range for LSH shuffle
+        Returns:
+            List of processed Task objects
+        """
 
-        # Get all actors from the pool
-        all_actors = list(actor_pool._idle_actors) + [actor for actor, _ in actor_pool._future_to_actor.items()]
+        actor_pool = ActorPool(actors)
+        stage_batch_size: int = ray.get(actors[0].get_batch_size.remote())
```
praateekmahajan, line 286: Is the concept of batch_size is relevant for LSH actors?

ray-curator/ray_curator/backends/experimental/ray_actor_pool/executor.py
```diff
@@ -214,15 +267,98 @@ def _process_stage_with_pool(
 
         return all_results
 
-    def _cleanup_actor_pool(self, actor_pool: ActorPool) -> None:
-        """Clean up actors in the pool."""
+    def _process_lsh_stage_with_rapidsmpf_actors(
+        self,
+        actors: list[ray.actor.ActorHandle],
+        _stage: "ProcessingStage",
```
praateekmahajan: Unused?

ray-curator/ray_curator/stages/deduplication/fuzzy/lsh/lsh.py
```diff
@@ -0,0 +1,300 @@
...
+    """
+
+    def __init__(  # noqa: PLR0913
+        self,
+        nranks: int,
+        total_nparts: int,
+        num_bands: int,
+        minhashes_per_band: int,
+        id_column: str = CURATOR_DEDUP_ID_STR,
+        minhash_column: str = "_minhash_signature",
```
praateekmahajan: column or field?

ray-curator/ray_curator/stages/deduplication/fuzzy/lsh/lsh.py
```diff
@@ -0,0 +1,300 @@
...
+            List of paths to minhash files.
+        band_range
+            Tuple of (start_band, end_band) to process.
+
+        Returns
+        -------
+            Column names of the output table.
+        """
+        # Get consistent column names for all batches
+        column_names = [self.id_column, "_bucket_id"]
```
praateekmahajan: I'm not sure if I'm following why `column_names` is returned? I see it later gets set to `self.output_columns` but is there a reason that's not set from the beginning to `self.id_field, "_bucket_id"`

ray-curator/ray_curator/stages/deduplication/fuzzy/lsh/lsh.py
```diff
@@ -0,0 +1,300 @@
...
+            when building an LSH index that needs to maintain all documents.
+
+        Returns
+        -------
+            DataFrame with bucket IDs and lists of document IDs.
+        """
+        if not include_singles:
+            # Find bucket_ids that appear more than once (have multiple documents)
+            # Keep only rows with buckets that are duplicated
+            df = df[df["_bucket_id"].duplicated(keep=False)]
```
praateekmahajan, line 240: Do you want to add a todo for helping prioritize which dupe to keep? might be hard given the input columns but I solved it in semantic here https://github.com/NVIDIA-NeMo/Curator/pull/911 using `RankingStratergy`

ray-curator/ray_curator/stages/shuffler/rapidsmpf_shuffler.py
```diff
@@ -0,0 +1,249 @@
...
+
+        if isinstance(spill_memory_limit, int):
+            self.spill_memory_limit = align_down_to_256(spill_memory_limit)
+        elif spill_memory_limit == "auto":
+            self.spill_memory_limit = align_down_to_256(0.8 * self.rmm_pool_size)
+        elif spill_memory_limit is None:
+            self.spill_memory_limit = None
+        else:
+            err_msg = f"Invalid spill_memory_limit: {spill_memory_limit}"
+            raise ValueError(err_msg)
```
praateekmahajan, line 77: I realised I'm not using any of these things in Semantic, maybe I should.. will create a ticket

ray-curator/tests/stages/deduplication/fuzzy/test_lsh_stage.py
```diff
@@ -0,0 +1,414 @@
...
+from ray_curator.stages.deduplication.id_generator import CURATOR_DEDUP_ID_STR
+from ray_curator.tasks import FileGroupTask
+
+
+@pytest.fixture(scope="module")
+def ray_fixture():
+    """Initialize and shutdown Ray for tests."""
+    ray.init(ignore_reinit_error=True)
+    yield
+    ray.shutdown()
```
praateekmahajan: I don't think you need to do this. You can just use the fixture from our conftest

ray-curator/ray_curator/backends/experimental/ray_actor_pool/executor.py
ayushdg: not at the moment but would like to keep this in place and let the ProcessingStage eventually call `process` per task in case some shuffle operations decide to implement a process_batch

ray-curator/ray_curator/stages/deduplication/fuzzy/lsh/lsh.py
ayushdg: Updated to field. will standardize the others modules that were missed in a followup

ray-curator/ray_curator/stages/deduplication/fuzzy/lsh/lsh.py
ayushdg: Yup it's not needed here. updated so that this returns None. 
I think the idea was that insert method could return column names that can be used by the extractor to assign those names back to the pylibcudf frames but we know these column names at the class level already.

ray-curator/ray_curator/stages/shuffler/rapidsmpf_shuffler.py
ayushdg: Spill limit is probably not relevant to RaftActors since they need to persist to memory. Would recommend checking out setting a regular rmm pool and setting cupy to use the same pool to see if has any perf impact.

ray-curator/tests/stages/deduplication/fuzzy/test_lsh_stage.py
ayushdg: updated

ray-curator/ray_curator/stages/deduplication/fuzzy/lsh/lsh.py
ayushdg: This is def a todo but not here rather in the stage after connected components. 
Added todo here to support building an LSH index which allows for easier incremental dedup without shuffling the whole dataset again.

ayushdg:
Merging as the failing test is unrelated to LSH.

cc: @sarahyurick might be flaky failure from model download? https://github.com/NVIDIA-NeMo/Curator/actions/runs/17112349374/job/48536856214?pr=890

PR #892, opened 2025-08-09
Fix exception when blocksize is set

robinhad:
## Description
If blocksize is set instead of files_per_partition, this line raised an exception.

## Usage

```python
```
## Checklist


abhinavg4:
@thomasdhc Looks like the ray-api tag should not be present for this PR?

ayushdg:
> @thomasdhc Looks like the ray-api tag should not be present for this PR?

@abhinavg4 The name is confusing but the tag doesn't denote that this PR is related to the Ray-api. The label sets up the cherry pick workflow to ensure that any new changes to main are also merged into the ray-api branch, ensuring that we don't miss any changes/updates during the migration.

PR #893, opened 2025-08-11
[Ray] Add Ray Actor Pool Exceuctor 

praateekmahajan:
## Description
Adds a `RayActorPool` executor which supports a vanilla adapter and a `raft` adapter. 

The executor in its current state is aimed to work with our dedup pipelines. Currently the way it works is
1. For each stage create a pool based on stage's resources and available resources in the cluster (i.e. use all resources available for the stage)
   - It also takes into account num_input_tasks for the stage, i.e if we have only 2 input tasks but 10 gpus for stage with gpus==1, instead of an actor pool with 10 actors it'll create only 2 actors
3. Submit tasks to the actor pool using map_unordered
   - this means we serialize the output to the driver. This could be **very bad** if your stage output is heavy. 
   - for our dedup pipelines we do FileGroupTask as input and output so this works
   - We use map_unordered so that we are able to handle skew as well. We'll have to do some hacking to leverage something like map_unordered but not return to driver ([internal slack message](https://nvidia.slack.com/archives/C08QJ76JCUV/p1754432244322229))
 3. After execution of each stage it kills all the actors and moves to the next

### Raft / Comms

The exeuctor supports a vanilla adapter (`RayActorPoolStageAdapter`) and a Raft Adapter (called `RayActorPoolRAFTAdapter`)
The Raft Adapter does a lil bit more than our other adapters where it 
1. creates bunch of actors all who know how many total actors exist in the pool i.e. `pool-size` and each actor knowing it's own `index`
1. designates a "root" actor
2. let all the other actor know who the "root" actor is i.e. `broadcast_root_unique_id`
3. Perform setup on all actors
   - This also calls `_setup_nccl` / `_setup_raft`

## Usage

```python
```
## Checklist

### DIFF
--- a/ray-curator/ray_curator/stages/video/clipping/clip_extraction_stages.py
+++ b/ray-curator/ray_curator/stages/video/clipping/clip_extraction_stages.py
@@ -9,7 +9,7 @@ from cosmos_xenna.ray_utils.resources import _get_local_gpu_info, _make_gpu_reso
 from loguru import logger
 
 from ray_curator.backends.base import WorkerMetadata
-from ray_curator.backends.experimental.ray_data.utils import RayStageSpecKeys
+from ray_curator.backends.experimental.utils import RayStageSpecKeys
 from ray_curator.stages.base import ProcessingStage
 from ray_curator.stages.resources import Resources, _get_gpu_memory_gb
 from ray_curator.tasks.video import Clip, Video, VideoTask
@@ -36,6 +36,7 @@ class ClipTranscodingStage(ProcessingStage[VideoTask, VideoTask]):
         verbose: Whether to print verbose logs.
         ffmpeg_verbose: Whether to print FFmpeg verbose logs.
     """
+
     num_cpus_per_worker: float = 6.0
     encoder: str = "libx264"
     encoder_threads: int = 1
@@ -67,7 +68,9 @@ class ClipTranscodingStage(ProcessingStage[VideoTask, VideoTask]):
                 gpu_info = _get_local_gpu_info()[0]
                 nvencs = _make_gpu_resources_from_gpu_name(gpu_info.name).num_nvencs
                 gpu_memory_gb = _get_gpu_memory_gb()
-                self._resources = Resources(nvencs=nvencs // self.nb_streams_per_gpu, gpu_memory_gb=gpu_memory_gb // self.nb_streams_per_gpu)
+                self._resources = Resources(
+                    nvencs=nvencs // self.nb_streams_per_gpu, gpu_memory_gb=gpu_memory_gb // self.nb_streams_per_gpu
+                )
             else:
                 self._resources = Resources(gpus=1)
         else:
@@ -160,17 +163,16 @@ class ClipTranscodingStage(ProcessingStage[VideoTask, VideoTask]):
             output_tasks.append(subtask)
         logger.info(f"Creating {len(clip_chunks)} tasks for downstream from {video.input_video}.")
 
-
         return output_tasks
 
     def _extract_clips(
-            self,
-            working_dir: pathlib.Path,
-            video_filename: str,
-            *,
-            force_pix_fmt: bool,
-            use_bit_rate: str | None,
-            clips: list[Clip],
+        self,
+        working_dir: pathlib.Path,
+        video_filename: str,
+        *,
+        force_pix_fmt: bool,
+        use_bit_rate: str | None,
+        clips: list[Clip],
     ) -> None:
         """Extract clips using ffmpeg."""
         # construct ffmpeg command
@@ -183,11 +185,11 @@ class ClipTranscodingStage(ProcessingStage[VideoTask, VideoTask]):
         self._read_clips_to_memory(working_dir, clips)
 
     def _build_ffmpeg_command(
-            self,
-            video_filename: str,
-            clips: list[Clip],
-            force_pix_fmt: bool,
-            use_bit_rate: str | None,
+        self,
+        video_filename: str,
+        clips: list[Clip],
+        force_pix_fmt: bool,
+        use_bit_rate: str | None,
     ) -> list[str]:
         """Build the ffmpeg command for extracting clips."""
         command = [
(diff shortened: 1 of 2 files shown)


ray-curator/ray_curator/backends/experimental/ray_actor_pool/executor.py
```diff
@@ -0,0 +1,195 @@
...
+
+                if not current_tasks:
+                    msg = f"{stage} - No tasks to process, can't continue"
+                    raise ValueError(msg)  # noqa: TRY301
+
+                # Create actor pool for this stage
+                num_actors = calculate_optimal_actors_for_stage(
+                    stage,
+                    len(current_tasks),
+                    reserved_cpus=self.config.get("reserved_cpus", 0.0),
```
ayushdg, line 77: Orthogonal to this PR but it's hard to know what keys can be set/supported for an executor. I wonder if there's a better way to represent the concrete set of keys. (Maybe a config dataclass).

ray-curator/ray_curator/backends/experimental/ray_actor_pool/executor.py
```diff
@@ -0,0 +1,195 @@
...
+            logger.info("Shutting down Ray to clean up all resources...")
+            ray.shutdown()
+
+    def _create_actor_pool(self, stage: "ProcessingStage", num_actors: int) -> ActorPool:
+        """Create an ActorPool for a specific stage."""
+        actors = []
+        for _ in range(num_actors):
+            actor = RayActorPoolStageAdapter.options(
+                num_cpus=stage.resources.cpus,
+                num_gpus=stage.resources.gpus,
```
ayushdg: prefer adding the name here similar to the other case. The dashboard shows this name more cleanly rather than a generic RayActorPoolStageAdapter.
```suggestion
                num_gpus=stage.resources.gpus,
                name=f"{stage.name}-{actor_idx}",
```

ray-curator/ray_curator/backends/experimental/ray_actor_pool/utils.py
```diff
@@ -0,0 +1,48 @@
...
+    max_actors_resources = min(max_actors_cpu, max_actors_gpu)
+
+    # Ensure we don't create more actors than configured maximum
+    max_actors_resources = min(max_actors_resources, stage.num_workers() or _LARGE_INT)
+
+    # Don't create more actors than tasks
+    optimal_actors = min(num_tasks, max_actors_resources)
+
+    # Ensure at least 1 actor if we have tasks
+    optimal_actors = max(1, optimal_actors) if num_tasks > 0 else 0
```
ayushdg, line 56: If the computation above returns 0 and we have more than 1 task should we wait/raise/warn here or spin up one actor anyway?

ray-curator/ray_curator/backends/internal/raft/ray_comms.py
```diff
@@ -0,0 +1,80 @@
...
+        """
+        Construct a new CommsContext instance
+
+        Parameters
+        ----------
+        verbose : bool
+                  Print verbose logging
+        nccl_root_location : string
+                  Indicates where the NCCL's root node should be located.
+                  ['client', 'worker', 'scheduler' (default), 'ray-actor']
```
ayushdg: Nit: Fix comment

ray-curator/ray_curator/backends/experimental/utils.py
```diff
@@ -1,9 +1,62 @@
...
+        node_id = node["NodeID"]
+        node_info = NodeInfo(node_id=node_id)
+        worker_metadata = WorkerMetadata(worker_id="", allocation=None)
+        logger.info(f"Executing setup on node {node_id} for {len(stages)} stages")
+        for stage in stages:
+            # Create NodeInfo and WorkerMetadata for this node
+
+            ray_tasks.append(
+                _setup_stage_on_node.options(
+                    num_cpus=stage.resources.cpus if stage.resources is not None else 1,
```
ayushdg, line 57: nice!

ayushdg (commented):
Minor comments

ray-curator/ray_curator/backends/experimental/ray_actor_pool/executor.py
praateekmahajan: That's a good point, will create a ticket for future, some data class that makes it easy to understand `map[exeuctor, config keys]`

ray-curator/ray_curator/backends/experimental/ray_actor_pool/executor.py
praateekmahajan: Good call, let me see if that is better or overriding __name__ or __qual_name__ like we do for Xenna / RayData

ray-curator/ray_curator/backends/experimental/ray_actor_pool/utils.py
ayushdg: Micro-optimization: What if we don't sleep on `get_available_resources` but only add a backoff/retry logic that sleeps and retries this calculation if max_actor_resources=0?

ayushdg: Thinking about this a bit more, maybe it does make sense to still sleep in the original get resources method. If stage 1 takes k CPUs. It's likely that without the sleep stage 2 will see only N-k cpus available. Which is less than optimal as well.

ray-curator/ray_curator/backends/experimental/ray_actor_pool/executor.py
praateekmahajan: I ended up doing mix of both
image

ray-curator/ray_curator/backends/experimental/ray_actor_pool/utils.py
```diff
@@ -0,0 +1,92 @@
...
+
+    # Ensure we don't create more actors than configured maximum
+    max_actors_resources = min(max_actors_resources, stage.num_workers() or _LARGE_INT)
+
+    if max_actors_resources == 0:
+        msg = f"No resources available for stage {stage.name}."
+        raise ValueError(msg)
+
+    # Don't create more actors than tasks
+    optimal_actors = min(num_tasks, max_actors_resources)
```
ayushdg: Another edge case. 
Let's say a stage has `batch size=8` & `num_tasks=32`. Assuming infinite resources. We would want optimal_actors here to be `4` rather than `32`. Maybe this becomes `min(num_tasks//stage.batch_size, max_actor_resources)` ?

praateekmahajan: Great catch!

PR #896, opened 2025-08-11
api autodoc pointer changes for ray curator

lbliii:
we need the api pointers to change first to unblock doc builds. let me know if you'd like this done differently


docs/_templates/autodoc2_index.rst
sarahyurick, line 1: I was wondering the best way to view this/if you could provide a screenshot? https://github.com/NVIDIA-NeMo/Curator/blob/d673f8743301a32fe1f21d006f32ad9c3775ceb8/docs/_templates/autodoc2_index.rst looks a bit wonky to me.

lbliii: I'm hoping for per-PR build previews in the future. Until then, I can include a screenshot. You can also preview locally using `make docs-live-ga` (assuming you've already done `make docs-env` first at least once)

image


I believe we can configure even deeper than this for the api if we want to. I'm still learning autodoc haha.

PR #907, opened 2025-08-13
Add IdGenerator to JsonlReader + IdGenerator tests / write_to_disk / from_disk

praateekmahajan:
## Description

This pull request introduces significant improvements to the 
- deduplication ID generator, 
- enhances test reliability via a shared Ray cluster
- adds new features to the JSONL reader stages for ID assignment and generation. 

**Deduplication ID generator improvements:**

* Refactored `IdGenerator` into a base class (`IdGeneratorBase`) and a Ray actor, added disk persistence methods (`to_disk`, `from_disk`), and provided utility functions to create, retrieve, and kill the ID generator actor. This enables robust state management and easier testing. 
* Updated `minhash.py` to use the new `get_id_generator_actor` function for improved actor management and testability. [[1]](diffhunk://#diff-2975d744dc9a860628d2f157d99c56c22f4796d14ce05c449b8fbb6ba8796412L11-R11) [[2]](diffhunk://#diff-2975d744dc9a860628d2f157d99c56c22f4796d14ce05c449b8fbb6ba8796412L257-R256)

**JSONL reader enhancements:**

* Added `_generate_ids` and `_assign_ids` options to `JsonlReader` and `JsonlReaderStage`, with corresponding logic to generate or assign deduplication IDs using the ID generator actor when reading data. These features are currently only supported with the pandas backend. 

**Test infrastructure improvements:**

* Introduced a shared Ray cluster fixture (`shared_ray_cluster`) for all deduplication stage tests, improving performance and resource utilization while ensuring proper isolation.
* Refactored MinHashStage tests to use the new shared Ray client and ID generator management, simplifying setup and teardown while ensuring actor cleanup. 

**CI configuration:**

* Updated the CI workflow to ignore deduplication stage tests in coverage and test runs for CPU jobs, streamlining test execution.
## Usage

```python
```
## Checklist


ray-curator/ray_curator/stages/text/io/reader/jsonl.py
```diff
@@ -126,14 +128,63 @@ def _read_with_pandas(
...
+            return self.generate_ids_func(file_paths, df)
+        if self._assign_ids:
+            return self.assign_ids_func(file_paths, df)
+        return df
+
+    def _assign_ids_func(self, filepath: str | list[str], df: pd.DataFrame) -> pd.DataFrame:
+        import numpy as np
+        import ray
+
+        from ray_curator.stages.deduplication.id_generator import CURATOR_DEDUP_ID_STR, get_id_generator_actor
```
VibhuJawa: Lets use global import. Dont know why our linter does not catch these.

ray-curator/ray_curator/stages/text/io/reader/jsonl.py
```diff
@@ -126,14 +128,63 @@ def _read_with_pandas(
...
+            min_id, max_id = ray.get(id_generator.get_batch_range.remote(filepath, None))
+            df[CURATOR_DEDUP_ID_STR] = np.arange(min_id, max_id + 1)
+
+        return df
+
+    def _generate_ids_func(self, filepath: str | list[str], df: pd.DataFrame) -> pd.DataFrame:
+        import numpy as np
+        import ray
+
+        from ray_curator.stages.deduplication.id_generator import CURATOR_DEDUP_ID_STR, get_id_generator_actor
```
VibhuJawa: Same here

ray-curator/ray_curator/stages/deduplication/id_generator.py
```diff
@@ -37,3 +56,64 @@ def get_batch_range(self, files: str | list[str] | None, key: str | None) -> tup
...
+                name=CURATOR_ID_GENERATOR_ACTOR_NAME, namespace=CURATOR_ID_GENERATOR_ACTOR_NAME, lifetime="detached"
+            ).remote(start_id=data["next_id"], batch_registry=data["batch_registry"])
+    except Exception as e:
+        logger.error(f"Error creating id generator actor: {e}")
+        raise
+
+    finally:
+        # Shutdown Ray to allow future pipelines to call ray.init with their own configuration
+        logger.info("Shutting down Ray for create_id_generator_actor")
+        ray.shutdown()
```
VibhuJawa: Shutting down on the fly seems finicky, will this work if we start a local ray cluster ? Wont it shut down

ray-curator/ray_curator/stages/deduplication/id_generator.py
```diff
@@ -1,18 +1,37 @@
...
-@ray.remote
-class IdGenerator:
-    def __init__(self, start_id: int = 0):
+class IdGeneratorBase:
+    """Base IdGenerator class without Ray decorator for testing and direct use."""
+
+    def __init__(self, start_id: int = 0, batch_registry: dict[str, tuple[int, int]] | None = None):
         self.next_id = start_id
-        self.batch_registry = {}  # {batch_hash: (min_id, max_id)}
+        self.batch_registry = batch_registry or {}  # {batch_hash: (min_id, max_id)}
```
VibhuJawa, line 35: Minor unrelated PR comment is to sort the list of files in `hash_files` to ensure determinism for same list of values and also add tests for it

ray-curator/tests/stages/deduplication/conftest.py
```diff
@@ -0,0 +1,120 @@
...
+        "--dashboard-port",
+        str(dashboard_port),
+        "--ray-client-server-port",
+        str(ray_client_server_port),
+        "--temp-dir",
+        str(temp_dir),
+        "--num-cpus",
+        "11",
+        "--num-gpus",
+        "2",
```
VibhuJawa: Should not hard code here, either  set it to min of current gpu available. We can get number of gpus using something like below. 


```suggestion
        def get_gpu_count():
            pynvml.nvmlInit()
            return pynvml.nvmlDeviceGetCount()
```

ray-curator/ray_curator/stages/text/io/reader/jsonl.py
```diff
@@ -126,14 +128,63 @@ def _read_with_pandas(
             return None
 
         # Concatenate all dataframes
-        return pd.concat(dfs, ignore_index=True)
+        df = pd.concat(dfs, ignore_index=True)
+        if self._generate_ids:
+            return self.generate_ids_func(file_paths, df)
```
VibhuJawa:
```suggestion
            return self._generate_ids_func(file_paths, df)
```

ray-curator/ray_curator/stages/text/io/reader/jsonl.py
```diff
@@ -126,14 +128,63 @@ def _read_with_pandas(
...
+        if self._assign_ids:
+            return self.assign_ids_func(file_paths, df)
```
VibhuJawa:
```suggestion
            return self._assign_ids_func(file_paths, df)
```

ray-curator/ray_curator/stages/deduplication/id_generator.py
```diff
@@ -37,3 +56,64 @@ def get_batch_range(self, files: str | list[str] | None, key: str | None) -> tup
...
+            _ = IdGenerator.options(
+                name=CURATOR_ID_GENERATOR_ACTOR_NAME, namespace=CURATOR_ID_GENERATOR_ACTOR_NAME, lifetime="detached"
+            ).remote(start_id=data["next_id"], batch_registry=data["batch_registry"])
+    except Exception as e:
+        logger.error(f"Error creating id generator actor: {e}")
+        raise
+
+    finally:
+        # Shutdown Ray to allow future pipelines to call ray.init with their own configuration
+        logger.info("Shutting down Ray for create_id_generator_actor")
```
VibhuJawa:
```suggestion
        logger.info(f"Shutting down Ray for {CURATOR_ID_GENERATOR_ACTOR_NAME}")
```

ray-curator/tests/stages/deduplication/conftest.py
praateekmahajan: Will keep it as is for now. Not using dynamic allocation for both cpu and gpu because otherwise on bigger clusters it'll just end up using all the CPUs / GPUs for unittests (like our lab machines which isn't ideal)

ray-curator/ray_curator/stages/deduplication/id_generator.py
praateekmahajan: Not at all minor glad you're thinking of that too.
It's already taken care of in the FilePartitioningStage -> get_all_files_paths_under -> fs.find(...) already sorts and returns the result

ray-curator/ray_curator/stages/deduplication/id_generator.py
praateekmahajan: For `IdGenerator` (and maybe everything else) we'll recommend that you start a ray cluster outside the current subprocess, in which case ray.shutdown behaves like a disconnect rather than shutdown.
This allows us to run multiple pipelines at once within a running cluster and keep resetting envvars.

ray-curator/ray_curator/stages/text/io/reader/jsonl.py
praateekmahajan: The reason I have intentionally kept it like this is because
a) JsonlReader(generate_id) must be used only for dedup
b) dedup imports can be very expensive and might require GPU (i.e. the cuda12x depdencies)
c) Ideally we don't have IdGenerator leaking into JsonlReader, but for SemDedup it has to be (so that embedding generation can be run using Ray Data / Xenna, while remaining stages can be run using RayActorPoolExecutor)

ray-curator/ray_curator/stages/text/io/reader/jsonl.py
```diff
@@ -126,14 +128,63 @@ def _read_with_pandas(
...
+
+        from ray_curator.stages.deduplication.id_generator import CURATOR_DEDUP_ID_STR, get_id_generator_actor
+
+        if CURATOR_DEDUP_ID_STR not in df.columns:
+            # Get the ID generator actor and retrieve the previously registered ID range
+            id_generator = get_id_generator_actor()
+            min_id, max_id = ray.get(id_generator.get_batch_range.remote(filepath, None))
+            df[CURATOR_DEDUP_ID_STR] = np.arange(min_id, max_id + 1)
+
+        return df
```
sarahyurick: I was wondering if there should be an `else` block here to log if the column exists already? Would it be useful to have a warning/error here?

ray-curator/ray_curator/stages/text/io/reader/jsonl.py
```diff
@@ -126,14 +128,63 @@ def _read_with_pandas(
...
 
     def _read_with_pyarrow(
         self, file_paths: list[str], reader_kwargs: dict[str, Any], columns: list[str] | None
     ) -> pa.Table | None:
         """Read JSONL files using PyArrow."""
 
         tables = []
+        if self._generate_ids or self._assign_ids:
+            msg = "Generating or assigning IDs is not supported for PyArrow reader"
+            raise NotImplementedError(msg)
```
sarahyurick, line 206: Any specific reason why not? Is it just because dedupe only uses the Pandas reader?

praateekmahajan: No specific reason except scope-spill.

ray-curator/ray_curator/stages/text/io/reader/jsonl.py
praateekmahajan: Good call, also moved `id_generator = get_id_generator_actor()` to setup

ray-curator/ray_curator/stages/deduplication/id_generator.py
```diff
@@ -37,3 +56,63 @@ def get_batch_range(self, files: str | list[str] | None, key: str | None) -> tup
             key = self.hash_files(files)
 
         return self.batch_registry[key]
+
+    def to_disk(self, filepath: str) -> None:
+        with open(filepath, "w") as f:
```
ayushdg, line 61: Could you open a followup to track remote support here.

ray-curator/ray_curator/stages/deduplication/id_generator.py
```diff
@@ -37,3 +56,63 @@ def get_batch_range(self, files: str | list[str] | None, key: str | None) -> tup
...
+
+def get_id_generator_actor() -> ActorHandle[IdGenerator]:
+    return ray.get_actor(name=CURATOR_ID_GENERATOR_ACTOR_NAME, namespace=CURATOR_ID_GENERATOR_ACTOR_NAME)
+
+
+def kill_id_generator_actor() -> None:
+    ray.kill(get_id_generator_actor())
+
+
+def create_id_generator_actor(filepath: str | None = None) -> None:
```
ayushdg, line 90: Nit: Add docstring here to explain filepath.

ray-curator/ray_curator/stages/deduplication/id_generator.py
```diff
@@ -37,3 +56,63 @@ def get_batch_range(self, files: str | list[str] | None, key: str | None) -> tup
...
+            # First load the data from disk
+            with open(filepath) as f:
+                data = json.load(f)
+            # Create actor with loaded data
+            _ = IdGenerator.options(
+                name=CURATOR_ID_GENERATOR_ACTOR_NAME, namespace=CURATOR_ID_GENERATOR_ACTOR_NAME, lifetime="detached"
+            ).remote(start_id=data["next_id"], batch_registry=data["batch_registry"])
+    except Exception as e:
+        logger.error(f"Error creating id generator actor: {e}")
+        raise
```
ayushdg, line 116: Is there a way of catching the specific exception where an ID generator actor already exists and warning/logging differently.

ray-curator/ray_curator/stages/deduplication/id_generator.py
```diff
@@ -37,3 +56,63 @@ def get_batch_range(self, files: str | list[str] | None, key: str | None) -> tup
...
+
+    finally:
+        # Shutdown Ray to allow future pipelines to call ray.init with their own configuration
+        ray.shutdown()
```
ayushdg, line 120: Unrelated: Just realized that the docs should heavily recommend starting your own ray cluster or using the get_client api rather than just creating pipelines and calling pipeline.run() which will result in shutdown shutting down the cluster and losing the actor.

ray-curator/ray_curator/stages/text/io/reader/jsonl.py
```diff
@@ -40,13 +42,28 @@ class JsonlReaderStage(ProcessingStage[FileGroupTask, DocumentBatch]):
     reader: str = "pandas"  # "pandas" or "pyarrow"
     reader_kwargs: dict[str, Any] = field(default_factory=dict)
     _name: str = "jsonl_reader"
+    _generate_ids: bool = False
+    _assign_ids: bool = False
```
ayushdg, line 58: Docstrings with what these two args do would be helpful.

ray-curator/ray_curator/stages/text/io/reader/jsonl.py
```diff
@@ -40,13 +42,28 @@ class JsonlReaderStage(ProcessingStage[FileGroupTask, DocumentBatch]):
...
 
     def inputs(self) -> tuple[list[str], list[str]]:
         return [], []
 
     def outputs(self) -> tuple[list[str], list[str]]:
         return ["data"], self.columns or []
 
+    def setup(self, _: WorkerMetadata | None = None) -> None:
+        if self._generate_ids or self._assign_ids:
```
ayushdg, line 72: Either in post init or here it's also worth checking if both are set to True since that's not a case we support

ray-curator/ray_curator/stages/text/io/reader/jsonl.py
```diff
@@ -40,13 +42,28 @@ class JsonlReaderStage(ProcessingStage[FileGroupTask, DocumentBatch]):
     reader: str = "pandas"  # "pandas" or "pyarrow"
     reader_kwargs: dict[str, Any] = field(default_factory=dict)
     _name: str = "jsonl_reader"
+    _generate_ids: bool = False
```
ayushdg, line 57: Since this initial version of this is primarily targeting id's for dedup, does it make sense to rename these to `_generate_deduplication_ids` and `_assign_deduplication_ids` to ensure folks don't accidentally use it?

ray-curator/ray_curator/stages/text/io/reader/jsonl.py
```diff
@@ -186,6 +243,8 @@ class JsonlReader(CompositeStage[_EmptyTask, DocumentBatch]):
     reader_kwargs: dict[str, Any] | None = None
     storage_options: dict[str, Any] | None = None
     task_type: Literal["document", "image", "video", "audio"] = "document"
+    _generate_ids: bool = False
+    _assign_ids: bool = False
```
ayushdg, line 264: same as above

ray-curator/tests/conftest.py
```diff
@@ -0,0 +1,227 @@
...
+    with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as s:
+        s.bind(("", 0))
+        return s.getsockname()[1]
+
+
+def gpu_available() -> bool:
+    """Check if GPU is available on the system."""
+    try:
+        # Use full path to nvidia-smi for security
+        result = subprocess.run(  # noqa: S603
```
ayushdg, line 40: Can we add a todo/follow up look into improving this in the future. Possibly with some direct API rather than nvidia-smi checks

ray-curator/tests/conftest.py
```diff
@@ -0,0 +1,227 @@
...
+
+    This checks:
+    1. If GPU tests are explicitly being run (via -m gpu)
+    2. If we're in a GPU test environment (CUDA_VISIBLE_DEVICES set)
+    3. If any collected test has the gpu marker
+    """
+    # Check environment variables that indicate GPU testing
+    if "CUDA_VISIBLE_DEVICES" in os.environ:
+        logger.info("CUDA_VISIBLE_DEVICES detected, enabling GPU cluster")
+        return True
```
ayushdg: This might be true in CI but will this always be true in local?

ray-curator/tests/conftest.py
```diff
@@ -0,0 +1,227 @@
...
+
+    # Check if running with -m gpu marker
+    gpu_marker = config.getoption("-m", default="")
+    if gpu_marker:
+        if "not gpu" in gpu_marker:
+            logger.info("'not gpu' marker detected, disabling GPU cluster")
+            return False
+        elif "gpu" in gpu_marker:
+            logger.info("GPU marker detected in test selection, enabling GPU cluster")
+            return True
```
ayushdg, line 74: I'm not too familiar with contest. But -m gpu, -m "not gpu" already works today. Is this logic needed?

ayushdg (changes requested):

ray-curator/tests/conftest.py
thomasdhc: @ayushdg for pytest functions that need to run on gpu you can add @pytest.mark.gpu
Pytest will pick it up accordingly

ray-curator/ray_curator/stages/deduplication/id_generator.py
praateekmahajan: Maybe I don't follow? We have `write_id_generator_to_disk`, does that work and also a test

ray-curator/ray_curator/stages/deduplication/id_generator.py
praateekmahajan: Probably but didn't look too much into catching the right exception with the hope that it'll be raised and be explicit enough, wdyt?

ray-curator/ray_curator/stages/deduplication/id_generator.py
praateekmahajan: Agreed, @VibhuJawa had similar question. Let me create a dedup ticket with p0
image

ray-curator/ray_curator/stages/text/io/reader/jsonl.py
praateekmahajan: I'd prefer to keep it this way in case we see a use case of non-dedup needing to generate monotonic ids. For now I'm hoping the "_" hides it for most use cases

ray-curator/tests/conftest.py
praateekmahajan: That's a valid question, but the key distinction is in the function name: `session_needs_gpu`. This function determines whether the current test session requires GPU resources for the Ray cluster setup, not just test filtering. While `-m gpu` markers help `pytest` filter which tests to run, we need a separate runtime decision to detect if any collected tests need GPU and accordingly provision either a GPU-enabled or CPU-only Ray cluster before execution begins.

ray-curator/tests/conftest.py
praateekmahajan: Fair point, removed it, and we can just look at the filtered list of tests to see if any of the tests have GPU, because in some sense this was buggy i.e. one could do `CUDA_VISIBLE_DEVICES=0,1 pytest -m "not gpu"` and that would've still started a GPU cluster

ray-curator/tests/conftest.py
thomasdhc: Gotcha, thanks for the clarification. Do you think it is necessary to have the gpu_available check? The way CI is setup tests marked with m gpu will run on the gpu environment. Will ray throw an error if gpu is configured but not provided?

thomasdhc: Thinking if that could help unblock you immediately

praateekmahajan: I'm unblocked now, thanks! Was able to use pynvml to get the count (which is a reminder we need a pynvml / gpustat as an explicit dependency).

ray-curator/tests/conftest.py
praateekmahajan: Made it slightly better using pynvml

ray-curator/ray_curator/stages/deduplication/id_generator.py
ayushdg: Sorry I mean reading/writing from remote storage (fsspec)

PR #912, opened 2025-08-15
[SemDedup] Add KMeans

praateekmahajan:
## TODO: Fill description

For now reviewer can look at `TestKMeansStageIntegration` inside test_kmeans, which creates a pipeline and runs it, and then work backwards.

Major improvements over dask implementation is, assuming enough GPU memory we'll never run into the cudf 2bn row limit. 

## Usage

```python
```
## Checklist


ray-curator/ray_curator/backends/internal/raft/ray_comms.py
```diff
@@ -58,7 +58,7 @@ def __init__(
 
     def __del__(self) -> None:
         if self.nccl_initialized:
-            self.destroy()
+            self.nccl_initialized = False
```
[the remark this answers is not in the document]
praateekmahajan: @ayushdg any thoughts? It seems like self.destroy isn't defined in the og example here either https://github.com/rapidsai/ray-rapids/blob/51a9ffecb53fcdad34fb522e7637d4ce98947584/rapids-experiments/ray_comms.py#L67-L69

ray-curator/ray_curator/stages/deduplication/semantic/kmeans.py
```diff
@@ -0,0 +1,371 @@
...
+        verbose: bool = False,
+        max_iter: int = 300,
+        tol: float = 1e-4,
+        random_state: int = 42,
+        init: Literal["k-means||", "random"] | np.ndarray = "k-means||",
+        n_init: int | Literal["auto"] = 1,
+        oversampling_factor: float = 2.0,
+        max_samples_per_batch: int = 1 << 15,
+        # I/O args
+        read_kwargs: dict[dict] | None = None,
```
VibhuJawa, line 53: Perfect, I also independently ended up aligning on read_kwargs . Not the bext name but at least we are all being consistent.  


https://github.com/NVIDIA-NeMo/Curator/pull/918/files#diff-d037626ba7459ddf25e54ad344bd6bf37de9f5a8c85ff1f530c35c758afa1da1R48

ray-curator/ray_curator/stages/deduplication/semantic/kmeans.py
```diff
@@ -0,0 +1,371 @@
...
+# Column names
+L2_DIST_TO_CENT_COL = "l2_dist_to_cent"
+COSINE_DIST_TO_CENT_COL = "cosine_dist_to_cent"
+
+FILETYPE_TO_EXTENSIONS = {
+    "parquet": [".parquet"],
+    "jsonl": [".jsonl", ".json"],
+}
+# Create a flat extension -> filetype mapping
+EXTENSION_TO_FILETYPE = {ext: filetype for filetype, exts in FILETYPE_TO_EXTENSIONS.items() for ext in exts}
```
VibhuJawa: Create a util for this and put it in file utils please , we will all use from this and ensure consitent matching

ray-curator/ray_curator/stages/deduplication/semantic/utils.py
```diff
@@ -0,0 +1,61 @@
...
+    """Break parquet files into groups to avoid cudf 2bn row limit."""
+    if embedding_dim is None:
+        # Default aggressive assumption of 1024 dimensional embedding
+        embedding_dim = 1024
+
+    cudf_max_num_rows = 2_000_000_000  # cudf only allows 2bn rows
+    cudf_max_num_elements = cudf_max_num_rows / embedding_dim  # cudf considers each element in an array to be a row
+
+    import pyarrow.parquet as pq
+    from fsspec.parquet import open_parquet_file
```
VibhuJawa: Lets import please at top, these are light enough and also move this to file utils. 

Like we are doing this for list column but might want to do it for other columns too etc so better to make it generic

ray-curator/ray_curator/stages/deduplication/semantic/kmeans.py
```diff
@@ -0,0 +1,371 @@
...
+        if not file_extensions:
+            msg = f"Unsupported filetype: {self.input_filetype}"
+            raise ValueError(msg)
+
+        return [
+            FilePartitioningStage(
+                file_paths=self.input_path,
+                file_extensions=file_extensions,
+                files_per_partition=1,  # We set this to one, and then the RaftActor will break it up into smaller groups
+            ),
```
VibhuJawa, line 343: I wonder if we should pass storage options here in FilePartitioningStage. 

I passed it here: 

https://github.com/NVIDIA-NeMo/Curator/pull/918/files#diff-6ec921fa8f277a57d283334f28937023416eb1c0318c72899fc6490f5a3a3989R172

ray-curator/ray_curator/backends/experimental/ray_actor_pool/executor.py
```diff
@@ -180,18 +182,31 @@ def _process_stage_with_pool(
             List of processed Task objects
         """
         stage_batch_size: int = ray.get(actor_pool._idle_actors[0].get_batch_size.remote())
+        if _stage.ray_stage_spec().get(RayStageSpecKeys.IS_RAFT_ACTOR, False):
+            # For a raft stage we want to make sure we have as many tasks as there are actors
+            if stage_batch_size is not None:
+                logger.warning(
+                    f"Stage {_stage.name} is a RAFT stage but has a batch size of {stage_batch_size}. Ignoring batch size."
+                )
+            stage_batch_size = math.ceil(len(tasks) / len(actor_pool._idle_actors))
```
ayushdg: Assuming 11 tasks and 5 actors the math will come out to `stage_batch_size=3` and we'll end up with 4 batches instead of 5 which can be problematic.
Happy to address this here or on my followup in #915 which strictly requires `num_tasks=num_actors` rather than `num_tasks<=num_actors`

ray-curator/ray_curator/stages/deduplication/semantic/kmeans.py
```diff
@@ -0,0 +1,371 @@
...
+
+        if not tasks:
+            return []
+
+        # Collect all files from all tasks
+        all_files = [file for task in tasks for file in task.data]
+
+        # Get filetype from file extension
+        filename = all_files[0]
+        filetype = next((EXTENSION_TO_FILETYPE[ext] for ext in EXTENSION_TO_FILETYPE if filename.endswith(ext)), None)
```
VibhuJawa: Lets do it via a util here

ray-curator/tests/stages/deduplication/semantic/test_kmeans.py
```diff
@@ -0,0 +1,498 @@
...
+
+        # Run single-GPU baseline for this test
+        single_gpu_assignments = run_single_gpu_baseline(self.input_dir, file_format=self.file_format)
+        # Read the multi-gpu output data
+        multi_gpu_assignments = (
+            cudf.read_parquet(self.output_dir).sort_values("id", ignore_index=True)["centroid"].to_numpy()
+        )
+
+        # Compare results with multi-GPU baseline
+        from sklearn.metrics import adjusted_rand_score
```
VibhuJawa: Move it on top

VibhuJawa (changes requested):
Took a initial look, mostly looks good.

ray-curator/ray_curator/backends/internal/raft/ray_comms.py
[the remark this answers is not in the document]
ayushdg: I think it's remnants of the core class here: https://github.com/rapidsai/raft/blob/branch-25.10/python/raft-dask/raft_dask/common/comms.py. 

We can explore readding all the cleanup functions in a followup if needed.

ray-curator/ray_curator/stages/deduplication/semantic/kmeans.py
```diff
@@ -0,0 +1,371 @@
...
+from loguru import logger
+
+# Column names
+L2_DIST_TO_CENT_COL = "l2_dist_to_cent"
+COSINE_DIST_TO_CENT_COL = "cosine_dist_to_cent"
+
+FILETYPE_TO_EXTENSIONS = {
+    "parquet": [".parquet"],
+    "jsonl": [".jsonl", ".json"],
+}
```
sarahyurick: Could this be useful in a shared `utils.py` or `constants.py` file?

ray-curator/ray_curator/stages/deduplication/semantic/kmeans.py
```diff
@@ -0,0 +1,371 @@
...
+            # Create result task for this subgroup
+            results.append(
+                _EmptyTask(
+                    task_id=output_filename,
+                    dataset_name=f"kmeans_group_{i}",
+                    _metadata=None,
+                    _stage_perf=[],
+                    data=None,
+                )
+            )
```
sarahyurick, line 217: What am I missing here? Why not return a `FileGroupTask` instead?

ray-curator/ray_curator/stages/deduplication/semantic/kmeans.py
```diff
@@ -0,0 +1,371 @@
...
+        n_clusters (int): The number of clusters to create.
+        id_field (str): The column name of the id column.
+        embedding_field (str): The column name of the embedding column.
+        input_path (str | list[str]): The path to the input directory.
+        output_path (str): The path to the output directory.
+        metadata_fields (list[str] | None): The columns to keep in the output. These columns can be used later to prioritize deduplication.
+        verbose (bool): Whether to print verbose output.
+        embedding_dim (int | None): The dimension of the embedding. This helps us read data into smaller chunks.
+        input_filetype (Literal["jsonl", "parquet"]): The type of the input file
+        input_file_extensions (list[str] | None): The file extensions of the input files. If not provided, we will infer it from the filetype.
```
sarahyurick:
```suggestion
```
Not a listed parameter.

ray-curator/ray_curator/backends/experimental/ray_actor_pool/executor.py
praateekmahajan: Good catch, I think for the RAFT actor stage we can do 
`task_batches = [batch.tolist() for batch in np.array_split(tasks, num_actors) if len(batch) > 0]`
which will in your e.g. return `3,2,2,2,2` len batches

ray-curator/ray_curator/stages/deduplication/semantic/kmeans.py
praateekmahajan: This is very KMeans specific (and maybe dedup specific) so don't want to move it up, because everywhere else we use FilePartitioningStage which accepts `file_extensions` while KMeans accepts `filetype` (since the I/O is different).

ray-curator/ray_curator/stages/deduplication/semantic/utils.py
praateekmahajan: I'm not sure if non-dedup pipelines will suffer from this since they in theory can just make `files_per_partition` lower. It's only an issue for Dedup Pipelines which have a RAFT actor stage (that's the stage where all tasks are evenly divided into num_actors) and hence some actors might receive bunch of coalesced files where num rows >> 2bn (then also it depends on how the RAFT actor is performing the read). For instance my KMeans actor extracts all files from list[FileGroupTask] and then tries to read.. 

I'd say this already generalizes in the sense that `break_parquet_partition_into_groups(files, ebedding_dim=1)`will work for non-array columns. Happy ro rename embedding_dim to something else but yea, for now I think it is generic.

@ayushdg in theory I imagine CC might run into this issue too, or no?

Moved the import to top

ray-curator/ray_curator/stages/deduplication/semantic/kmeans.py
praateekmahajan: Good catch, added it

ray-curator/ray_curator/stages/deduplication/semantic/kmeans.py
praateekmahajan: @ayushdg i imagine you might need something like this too?

ray-curator/ray_curator/stages/deduplication/semantic/kmeans.py
praateekmahajan: Because we write out using `partition_by` so its hard to truly know which files were written out

ray-curator/ray_curator/stages/deduplication/semantic/kmeans.py
praateekmahajan: Good catch

ray-curator/ray_curator/stages/deduplication/semantic/kmeans.py
praateekmahajan: Related to https://github.com/NVIDIA-NeMo/Curator/pull/912#discussion_r2285837712 will address it together

ray-curator/ray_curator/stages/deduplication/semantic/kmeans.py
praateekmahajan: Related to https://github.com/NVIDIA-NeMo/Curator/pull/912#discussion_r2285837712 will address it together

ayushdg: This mapping might be useful as a general shared constant in utils outside dedup. I suspect different readers can use this mapping to set the write extensions to filter on when calling filePartitioningStage.

ray-curator/ray_curator/stages/deduplication/semantic/kmeans.py
ayushdg: I'm not sure other readers outside dedup need the reverse mapping since it sounds like you're inferring the filetype based on the extension.
This could live in `dedup` utils.

In Minhash I end up accepting a `filetype` arg in the init rather than inferring it based on the files inside that directory. I personally prefer having the users be explicit about the input filetype because folks may come with non-standard extensions and things get difficult to infer very quickly.

ray-curator/ray_curator/stages/deduplication/semantic/utils.py
ayushdg: CC may also run into OOM's unless we read each file one by one. CC doesn't suffer from the cuDF 2bn row limit so the strict size checking isn't needed.

ray-curator/ray_curator/stages/deduplication/semantic/kmeans.py
ayushdg: Somewhat related. I've been bitten by the default extension for `FilePartitioningTask` being `.jsonl`. Would prefer the default behavior there to not filter and each reader (jsonl, parquet etc.) can set whatever extension they want to filter on

ray-curator/ray_curator/backends/internal/raft/ray_comms.py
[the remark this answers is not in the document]
praateekmahajan: For now setting `nccl_initialized=False`... is that okay for now? And follow up could be running `_func_destroy_all` / `_func_destroy_scheduler_session`

It's also interesting that `__del__` is likely never called in the Ray world otherwise this would've thrown an error

ray-curator/ray_curator/stages/deduplication/semantic/kmeans.py
VibhuJawa: Plus one, should be general otuside dedup for readers

praateekmahajan: > prefer the default behavior there to not filter and each reader (jsonl, parquet etc.) can set whatever extension

Agreed in this case I guess in @VibhuJawa 's PR we want to amke sure JsonlReader exposes `file_extensions` attribbutes that is then passed to `FilePartitioningStage`

> This mapping might be useful as a general shared constant in utils outside dedup

I can see if above is what we're doing then any reader can just get the defaults unless specified

ray-curator/ray_curator/stages/deduplication/semantic/kmeans.py
praateekmahajan: I'm happy to match what you have in `Minhash` where its called `read_format`. I'll name it to `filetype` to be consistent with file_extensions in FilePartitioningStage. Lmk if not, but I think both SemDedup and Minhash (along with rest of API) should have similar naming conventions

ray-curator/ray_curator/stages/deduplication/semantic/kmeans.py
praateekmahajan: Made the change

ray-curator/ray_curator/stages/deduplication/semantic/kmeans.py
praateekmahajan: Made the change!

ray-curator/ray_curator/stages/deduplication/semantic/kmeans.py
praateekmahajan: Removed it altogether

ray-curator/ray_curator/stages/deduplication/semantic/kmeans.py
VibhuJawa: Just noting that this will fail today (which is fine, we can merge my PR after this). so we are good.

VibhuJawa (approved):
LGTM , thanks for the work here

PR #915, opened 2025-08-18
Add Connected components

ayushdg:
## Description

This PR adds a connected component stage which leverages RAFT actors for comms.
Key difference vs the previous implementation:
1. Avoiding encoding, joins & edgelist drop duplicates
  a. The current implementation can avoid those joins/merges since we use integer IDs now. Additionally it also pushes duplicate edge removal and renumbering (supported for ints) down to the libcugraph layer.
2. Uses pylibcugraph's MGGraph directly instead of dask cugraph. In a followup (post 25.08 dependencies) all references to cugraph will be replaced with pylibcugraph.

Benchmark results:

Dataset - Edgelist generated from 1.4TB common crawl snapshot going through dedup.
Dask Version - 60s
Current Ray version - 28-32s + result shuffle time (will be implemented in a followup PR)

## Usage

```python
```
## Checklist


ray-curator/ray_curator/stages/deduplication/fuzzy/connected_components.py
```diff
@@ -0,0 +1,209 @@
...
+        dfs = []
+        for input_file in input_files:
+            dfs.append(self.read_parquet(input_file, columns=edgelist_columns, **self.read_kwargs))
+        df = cudf.concat(dfs)
+        # remove duplicate edges
+        df = df.drop_duplicates(subset=edgelist_columns, ignore_index=True)
+        vertices, labels = self.weakly_connected_components(df, self.source_column, self.destination_column)
+        df = cudf.DataFrame()
+        df[CURATOR_DEDUP_ID_STR] = vertices
+        df["_duplicate_group_id"] = labels
```
sarahyurick: Maybe we would want to store column names like this in a `constants.py` file somewhere?

ray-curator/ray_curator/stages/deduplication/fuzzy/connected_components.py
```diff
@@ -0,0 +1,209 @@
...
+        Args:
+            output_dir: The directory to write the resulting connected components to.
+            source_column: The column name containing the document ids of the source of the edge.
+            destination_column: The column name containing the document ids of the destination of the edge.
+            read_kwargs: Keyword arguments to pass for reading the input files.
+            write_kwargs: Keyword arguments to pass for writing the output files.
+        """
+        # Initialize parent classes
+        ProcessingStage.__init__(self)
+        DeduplicationIO.__init__(self, id_generator=None)
```
praateekmahajan: Do we need these? Or should super().__init__ take care? I notice my Kmeans doesn't have either :sus:

ray-curator/ray_curator/stages/deduplication/fuzzy/connected_components.py
```diff
@@ -0,0 +1,209 @@
...
+
+        self.source_column = source_column
+        self.destination_column = destination_column
+        self.read_kwargs = read_kwargs if read_kwargs is not None else {}
+        self.write_kwargs = write_kwargs if write_kwargs is not None else {}
+
+        self._name = self.__class__.__name__
```
praateekmahajan, line 61: Why this vs just "ConnectedComponents"?

ray-curator/ray_curator/stages/deduplication/fuzzy/connected_components.py
```diff
@@ -0,0 +1,209 @@
...
+        self._batch_size = None
+
+        # Handle output directory cleanup logic
+        self.output_fs = get_fs(output_dir, self.write_kwargs.get("storage_options", {}))
+        self.output_dir = self.output_fs.sep.join([output_dir, self.name])
+        if is_not_empty(self.output_dir, self.output_fs):
+            logger.warning(f"Output directory {self.output_dir} is not empty. Deleting it.")
+            delete_dir(self.output_dir, self.output_fs)
+        self.output_fs.mkdirs(self.output_dir, exist_ok=True)
+
```
praateekmahajan: Can you move this to a file_utils? I could also use it in kmeans / pairwise

ray-curator/ray_curator/stages/deduplication/fuzzy/connected_components.py
```diff
@@ -0,0 +1,209 @@
...
+        dg = cugraph.Graph(directed=False)
+
+        src_array = df[src_col]
+        dst_array = df[dst_col]
+
+        rhandle = ResourceHandle(self._raft_handle.getHandle())
+
+        graph_props = GraphProperties(
+            is_multigraph=False,  # dg.properties.multi_edge (what is multi_edge)
```
praateekmahajan: what's the comment?

ray-curator/ray_curator/stages/deduplication/fuzzy/connected_components.py
VibhuJawa: I think i have not seen it at other places too .  :suspect:

ray-curator/ray_curator/stages/deduplication/fuzzy/connected_components.py
```diff
@@ -0,0 +1,209 @@
...
+
+if TYPE_CHECKING:
+    from ray_curator.backends.base import WorkerMetadata
+
+
+class ConnectedComponentsStage(ProcessingStage[FileGroupTask, FileGroupTask], DeduplicationIO):
+    def __init__(
+        self,
+        output_dir: str,
+        source_column: str = f"{CURATOR_DEDUP_ID_STR}_x",
```
praateekmahajan, line 42: column or field?

ray-curator/ray_curator/stages/deduplication/fuzzy/connected_components.py
```diff
@@ -0,0 +1,209 @@
...
+            Columns to read from input file
+        """
+        input_files = []
+        for task in tasks:
+            input_files.extend(task.data)
+        output_file = self.output_fs.sep.join([self.output_dir, f"{tasks[0].task_id}.parquet"])
+        edgelist_columns = [self.source_column, self.destination_column]
+        dfs = []
+        for input_file in input_files:
+            dfs.append(self.read_parquet(input_file, columns=edgelist_columns, **self.read_kwargs))
```
praateekmahajan, line 181: Won't this be much slower than just doing `self.read_parquet(input_files)`? Why are we reading 1 by 1? Some memory thing I assume, if so how much is it, can you share? I might need to be mindful too.
FWIW I'm reading like ~300 files at a time using read_parquet and that's been fine, the biggest bottleneck has been the write_parquet step which seems to have way more of a memory overhead. see https://nvidia.slack.com/archives/CCWT6NJ0K/p1747335283494139

ray-curator/ray_curator/stages/deduplication/fuzzy/connected_components.py
```diff
@@ -0,0 +1,209 @@
...
+        df = cudf.concat(dfs)
+        # remove duplicate edges
+        df = df.drop_duplicates(subset=edgelist_columns, ignore_index=True)
+        vertices, labels = self.weakly_connected_components(df, self.source_column, self.destination_column)
+        df = cudf.DataFrame()
```
praateekmahajan: any opinions on lining and using a new variable?

```python
del df
output_df = cudf.DataFrame({CURATOR_DEDUP_ID_STR : verticies, _duplicate_group_id : labels})
```

ray-curator/ray_curator/stages/deduplication/fuzzy/connected_components.py
praateekmahajan: We had something like this here

https://github.com/ayushdg/ray-curator/blob/92d1280010e9953ef8c602ea9441aaf9fecfddeb/ray_curator/utils/file_utils.py#L63C5-L76

FWIW s3's mkdirs doesn't work IIRC, which is why the code above has `fs.touch(os.path.join(dir_path, ".empty"))`

ray-curator/ray_curator/stages/deduplication/fuzzy/connected_components.py
```diff
@@ -0,0 +1,209 @@
...
+        for task in tasks:
+            input_files.extend(task.data)
+        output_file = self.output_fs.sep.join([self.output_dir, f"{tasks[0].task_id}.parquet"])
+        edgelist_columns = [self.source_column, self.destination_column]
+        dfs = []
+        for input_file in input_files:
+            dfs.append(self.read_parquet(input_file, columns=edgelist_columns, **self.read_kwargs))
+        df = cudf.concat(dfs)
+        # remove duplicate edges
+        df = df.drop_duplicates(subset=edgelist_columns, ignore_index=True)
```
VibhuJawa, line 184: Don't we drop multi edges  here , is it internal to do it twice in cuDF and hand off layer ?

```
        drop_multi_edges=True
```

VibhuJawa (commented):
Mostly looks good, had a question about 2 drop duplicates

ray-curator/ray_curator/stages/deduplication/fuzzy/connected_components.py
VibhuJawa: Best guess: 

https://github.com/rapidsai/cugraph/blob/84db1a5f79bb3355f3425f53e8264ac87e0940d1/cpp/include/cugraph/graph_generators.hpp#L421

ray-curator/ray_curator/stages/deduplication/fuzzy/connected_components.py
ayushdg: I've generally been following the convention of naming this `MinHashStage, LSHStage etc.` instead of just ConnectedComponents or LSH. I haven't looked at the other stage names in too much detail but don't have strong opinions. Happy to have it be ConnectedComponents

ray-curator/ray_curator/stages/deduplication/fuzzy/connected_components.py
ayushdg: Agreed. I'll do this in the followup composite stage for dedup

ray-curator/ray_curator/stages/deduplication/fuzzy/connected_components.py
ayushdg: `super.init()` would fail since any args/kwargs passed would be propagated to both parents. And if the parent is expecting/not expecting a kwarg it receives it will error out. The alternative is to call the super of each parent separately.

In this case it I realize that I can omit since we don't need to call parent's init methods at all.

ray-curator/ray_curator/stages/deduplication/fuzzy/connected_components.py
ayushdg: The drop multi-edge here is doing it at the partition level before handing off to cugraph. 
Cugraph will handle dropping multi edges at the global graph level. Main motivation for adding this call is to reduce the data being passed to cugraph in case there's some global memory overhead from cugraphs drop multi-edge impl.

ray-curator/ray_curator/stages/deduplication/fuzzy/connected_components.py
ayushdg: It is much slower but comes with lower memory overhead. 
Here are results from reading a 8GB dataset split into 128 files.

```python
def read_one_at_a_time(files):
    dfs = []
    for file in files:
        dfs.append(cudf.read_parquet(file))
    df = cudf.concat(dfs, ignore_index=True)
    del(dfs)
    return df

def read_all(files):
    return cudf.read_parquet(files)
```

```
Read All: ~1s, 30GB peak memory usage
Read one at a time: ~9s, 16GB peak memory usage
```

ray-curator/ray_curator/stages/deduplication/fuzzy/connected_components.py
VibhuJawa: Sounds good.

ray-curator/ray_curator/stages/deduplication/fuzzy/connected_components.py
ayushdg: Based on the comments here: https://github.com/fsspec/s3fs/issues/401 mkdirs is a no-op for s3 which is fine since directly writing out to a path without making directories is allowed. I don't think a `touch` is needed in this case.

ayushdg: Moved the boilerplate to a util

ray-curator/ray_curator/stages/deduplication/fuzzy/connected_components.py
ayushdg: Updated. For context most cugraph examples set `is_multigraph=dg.properties.multi_edge` however here we just set it False. Our graph doesn't need more than one edge between the same pair of docs, we want to drop duplicate edges anyway.

ray-curator/ray_curator/stages/deduplication/fuzzy/connected_components.py
ayushdg: Will standardize in a followup since there's some ongoing discussion

ray-curator/ray_curator/utils/file_utils.py
```diff
@@ -61,6 +62,29 @@ def delete_dir(
...
+def create_or_overwrite_dir(
+    path: str, fs: fsspec.AbstractFileSystem | None = None, storage_options: dict[str, str] | None = None
+) -> None:
+    """
+    Creates a directory if it does not exist and overwrites it if it does.
+    Warning: This function will delete all files in the directory if it exists.
+    """
+    if fs is None and storage_options is None:
+        err_msg = "fs or storage_options must be provided"
+        raise ValueError(err_msg)
```
praateekmahajan, line 74: I noticed all our functions have this check, but I guess isn't it fine if both are None? In that case it should just goto fs = get_fs(path, storage_options=None) which should return LocalFileSystem

praateekmahajan (approved):
LGTM, except the minor nit on how we read pq files.

ray-curator/ray_curator/stages/deduplication/fuzzy/connected_components.py
```diff
@@ -0,0 +1,203 @@
...
+            raise ValueError(msg)
+
+        self._setup_post()
+
+    def ray_stage_spec(self) -> dict[str, Any]:
+        return {
+            RayStageSpecKeys.IS_RAFT_ACTOR: True,
+        }
+
+    def __get_2D_div(self, ngpus: int) -> tuple[int, int]:  # noqa: N802
```
praateekmahajan, line 82: why `__`?

ray-curator/ray_curator/stages/deduplication/fuzzy/connected_components.py
```diff
@@ -0,0 +1,203 @@
...
+
+        self.source_column = source_column
+        self.destination_column = destination_column
+        self.read_kwargs = read_kwargs if read_kwargs is not None else {}
+        self.write_kwargs = write_kwargs if write_kwargs is not None else {}
+
+        self._name = self.__class__.__name__
+        self._resources = Resources(cpus=1.0, gpus=1.0)
+        self._batch_size = None
+
```
praateekmahajan, line 64: idk if you need to use `check_disallowed_args` here to enshure read_args / write_args don't have a param that you can't allow

ray-curator/ray_curator/utils/file_utils.py
VibhuJawa: +1, We should probably allow this function to work for local file systems too (and if it does) we have to allow  this to go through.

ray-curator/ray_curator/stages/deduplication/fuzzy/connected_components.py
ayushdg: Keeping the same convention as the source from cugraph where I adapted this from.

PR #917, opened 2025-08-18
[Ray/Examples] Port the TinyStories and PEFT curation tutorials to Ray API

Maghoumi:
## Description

This commit uses the Ray API to curate the TinyStories and the PEFT curation datasets.

## Usage
```python
python ray-curator/ray_curator/examples/tinystories/main.py
python ray-curator/ray_curator/examples/peft-curation/main.py
```
## Checklist


ray-curator/ray_curator/examples/peft-curation/main.py
```diff
@@ -0,0 +1,95 @@
...
+            modifier_fn=[UnicodeReformatter(), UnicodeReformatter(), UnicodeReformatter()],
+            text_field=["subject", "body", "category"],
+        )
+    )
+    pipeline.add_stage(
+        Modify(
+            modifier_fn=AddPeriod(),
+            text_field=["category"],
+        )
+    )
```
ayushdg: Just noting that it's a matter of preference but you could also create a list of stages and then do:
`pipeline = Pipeline(...., stages=...)`

sarahyurick: Small nit: can you make it consistent as to whether the `text_field` or filter/modifier field is listed first?

ray-curator/ray_curator/examples/peft-curation/stage_download.py
```diff
@@ -0,0 +1,157 @@
...
+        }
+
+    def input_columns(self) -> list[str]:
+        return ["id", "raw_content"]
+
+    def output_columns(self) -> list[str]:
+        return ["id", "subject", "body", "category"]
+
+
+class EnronEmailsDownloadExtractStage(DocumentDownloadExtractStage):
```
sarahyurick: Very nice extension of the download and extract modules, thanks!

ray-curator/ray_curator/examples/peft-curation/README.md
```diff
@@ -0,0 +1,13 @@
+# TinyStories
```
sarahyurick: Looks like the wrong file was committed here?

ray-curator/ray_curator/examples/peft-curation/main.py
```diff
@@ -0,0 +1,95 @@
...
+    # TODO: Add token counter and apply_chat_template once Modifier supports that use-case.
+    executor = XennaExecutor()
```
ayushdg: With #922  you don't need to create this anymore. Just `pipeline.run()` should suffice.

ray-curator/ray_curator/examples/peft-curation/main.py
VibhuJawa: With https://github.com/NVIDIA-NeMo/Curator/pull/921 I changed the name to input_filed, you will have to update this PR.

ray-curator/ray_curator/examples/peft-curation/main.py
```diff
@@ -0,0 +1,95 @@
...
+        ScoreFilter(
+            text_field=["subject", "body", "category"],
+            filter_obj=[FilterEmptyEmails(), FilterEmptyEmails(), FilterEmptyEmails()],
+            invert=True,
+        )
+    )
+    pipeline.add_stage(
+        Modify(
+            modifier_fn=[UnicodeReformatter(), UnicodeReformatter(), UnicodeReformatter()],
+            text_field=["subject", "body", "category"],
```
VibhuJawa: With https://github.com/NVIDIA-NeMo/Curator/pull/921 I changed the name to input_filed, you will have to update this PR. 
```suggestion
            text_fields=["subject", "body", "category"],
```

ray-curator/ray_curator/examples/peft-curation/main.py
```diff
@@ -0,0 +1,95 @@
...
+        )
+    )
+    pipeline.add_stage(
+        Modify(
+            modifier_fn=AddPeriod(),
+            text_field=["category"],
```
VibhuJawa:
```suggestion
            input_field=["category"],
```

ray-curator/ray_curator/examples/peft-curation/main.py
```diff
@@ -0,0 +1,95 @@
...
+    # TODO: Add token counter and apply_chat_template once Modifier supports that use-case.
+    executor = XennaExecutor()
+
+    print("Starting the curation pipeline")
+    start_time = time.time()
+    results = pipeline.run(executor)
+    end_time = time.time()
+    execution_time = end_time - start_time
+    # Combine the results and output
+    combined_data = pd.concat([document_batch.data for document_batch in results])
```
VibhuJawa: We should probably also write using a stage (to ensure folks don't do collection) on driver.

ray-curator/ray_curator/examples/peft-curation/main.py
Maghoumi: Sarah, it's done now!

Vibhu, fixed!

ray-curator/ray_curator/examples/peft-curation/stage_download.py
Maghoumi: Cool!

ray-curator/ray_curator/examples/peft-curation/README.md
Maghoumi: Oops, fixed!

ray-curator/ray_curator/examples/peft-curation/main.py
Maghoumi: Good suggestion, done!

ray-curator/ray_curator/examples/peft-curation/main.py
Maghoumi: Good to know, fixed.

ray-curator/ray_curator/examples/peft-curation/main.py
Maghoumi: Should be fixed now.

ray-curator/ray_curator/examples/peft-curation/main.py
Maghoumi: Same as above

ray-curator/ray_curator/examples/peft-curation/main.py
Maghoumi: @VibhuJawa 
Good suggestion. I hadn't seen `JsonlWriter`. Modified both tutorials. Some suggestions:

1. Looks like there is no way to specify a name (or a name stem) for the `JsonlWriter`. Currently it uses a random uuid name. My suggestion is to make it possible for users to specify a name stem `e.g. MyDataset` then have this writer do `MyDataset-0.jsonl, -1.jsonl` and so on. This would be slightly better than UUID strings in the name.

2. Now that you have the multi-col thing merged, for the UnicodeReformatter, is it possible to specify the columns it should operate on? You might notice I have 3 separate instances of UnicodeReformatter, along with 3 separate column names (one for each instance). I wonder if there's a way to do this already but I just don't know it.

Maghoumi:
> General open question: I think these will eventually live in tutorials/, not sure if it would be better to move them there now or as a follow-up (once there is agreement among the team)?

@sarahyurick Sorry just saw this. I initially put everything inside tutorials, but then I thought fundamentally the tutorials are examples, so we might wanna unify?

I don't have a strong preference tbh, whatever makes the most sense.

ray-curator/ray_curator/examples/peft-curation/main.py
VibhuJawa: 2. is it possible to specify the columns it should operate on? 

Yes, this is a good example test:  

https://github.com/NVIDIA-NeMo/Curator/blob/8e6371752f969ff96fc9f720c458636b38123077/ray-curator/tests/stages/text/modules/test_modifiers.py#L563-L576


1. Looks like there is no way to specify a name (or a name stem) for the JsonlWriter. 

Thanks for feedback, Let me add that as part of https://github.com/NVIDIA-NeMo/Curator/pull/918

Maghoumi: Did you mean something like this?

```
Modify(
            modifier_fn=UnicodeReformatter(),
            input_fields=[["subject", "body", "category"]],
            output_fields=[["subject", "body", "category"]],
        ),
```

Asking because this doesn't work.

```
    df[output_field_i] = [inner_modify_fn(**rec) for rec in df[cols].to_dict("records")]
                          ^^^^^^^^^^^^^^^^^^^^^^
TypeError: UnicodeReformatter.modify_document() got an unexpected keyword argument 'subject'
```

Sounds like UnicodeReformatter itself needs to change so if there are multiple columns passed to it, it would reformat every column and return each column.

Maghoumi: I guess this also applies to other `DocumentModifier` implementations as well.

VibhuJawa: You will have to create a new modifier which accepts 3 args,  not a single car if that makes sense. 

Like create a new modifier which will accept multiple inputs. If you want to apply a single modifier to all , then repeating it is fine.  This is for new Modifiers rather than the existing modifiers, if that makes sense.

Maghoumi: Got it. Will keep it as-is then

Maghoumi:
@ayushdg @sarahyurick @VibhuJawa is this PR in good shape to merge?

Regarding Sarah's comment above and my response on folder placement, `tutorials` vs `examples`, I have a different opinion now :smile: I'd saying having all the "how to use" stuff under `examples` makes the most sense. That ways users won't have to jump across different folders to discover things.

This seems to align well with other repositories too. But ultimately up to you.

ayushdg:
> Sorry just saw this. I initially put everything inside tutorials, but then I thought fundamentally the tutorials are examples, so we might wanna unify?

For now I'm fine leaving them inside tutorials. We may move them out to examples or someplace else depending on the final layout we decide based on @lbliii recommendations. 

> I have a different opinion now 😄 I'd saying having all the "how to use" stuff under examples makes the most sense. That ways users won't have to jump across different folders to discover things.

Would love your thoughts when we discuss this but I believe there is some confusion b/w `examples/tutorials/scripts` today so the goal was to have everything live in a `tutorials` folder that had subdirectories containing `how-to-guides, getting-starting-notebooks` etc. The final layout still needs to be finalized.

ray-curator/ray_curator/examples/tinystories/stages.py
```diff
@@ -0,0 +1,207 @@
...
+
+        Args:
+            url: URL to download
+            path: Local path to save file
+
+        Returns:
+            Tuple of (success, error_message). If success is True, error_message should be None.
+            If success is False, error_message should contain the error details.
+        """
+        print(f"Downloading TinyStories dataset from '{url}'...")
```
praateekmahajan: instead of prints lets use loguru.logger and use `logger.debug`

VibhuJawa (approved):
Looks good to merge to me.

ray-curator/ray_curator/examples/tinystories/stages.py
Maghoumi: Replaced all print statements with logger.info and logger.debug.

Maghoumi:
> > Sorry just saw this. I initially put everything inside tutorials, but then I thought fundamentally the tutorials are examples, so we might wanna unify?
> 
> For now I'm fine leaving them inside tutorials. We may move them out to examples or someplace else depending on the final layout we decide based on @lbliii recommendations.
> 
> > I have a different opinion now 😄 I'd saying having all the "how to use" stuff under examples makes the most sense. That ways users won't have to jump across different folders to discover things.
> 
> Would love your thoughts when we discuss this but I believe there is some confusion b/w `examples/tutorials/scripts` today so the goal was to have everything live in a `tutorials` folder that had subdirectories containing `how-to-guides, getting-starting-notebooks` etc. The final layout still needs to be finalized.

@ayushdg Moved the tutorials to the `tutorials` folder, please take another look!

sarahyurick:
LGTM, thanks! The main thing I wanted to verify:

- TinyStories uses PII redaction and exact duplicates, which are not ready for tutorials yet
- PEFT uses PII redaction

We are aligned to merge without them for now?

Maghoumi:
> LGTM, thanks! The main thing I wanted to verify:
> 
> * TinyStories uses PII redaction and exact duplicates, which are not ready for tutorials yet
> * PEFT uses PII redaction
> 
> We are aligned to merge without them for now?

Yes, spoke with Ayush previously about this, and he said it's fine to merge without PII for now.

sarahyurick:

Maghoumi:
> Hi, please target `main` instead of `ray-api` now, thanks!

Rebased onto main, and revised the PR. Let me know if this is good to merge now, thanks!

Maghoumi:
@sarahyurick Any other concerns, or can this be merged now?

sarahyurick:
Hi @Maghoumi yes, all file imports should be `nemo_curator` now and not `ray_curator`. Thanks!

Issue #706, opened 2025-05-12
Add features from Curriculum Dataprep tutorial to NeMo Curator modules


sarahyurick:
While some of the classes in https://github.com/NVIDIA/NeMo-Curator/pull/695 are very specific to the dataset being curated, others could be useful for a variety of datasets and should be added as NeMo Curator modules.

In particular, I think it would be nice to add these as modules: `NonEnglishFilter` (could be generalized to any language), `TokenCountFilter`, and `CompletionTokenCountFilter`.

sarahyurick:
TODO:

- [x] Update [this comment](https://github.com/NVIDIA/NeMo-Curator/blob/main/tutorials/llama-nemotron-data-curation/main.py#L427)
- [x] Add description to `tutorials/README.md`

sarahyurick:
TODO items fixed by: https://github.com/NVIDIA-NeMo/Curator/pull/771.

sarahyurick:
See https://github.com/NVIDIA-NeMo/Curator/pull/921#discussion_r2289024810.

PR #921, opened 2025-08-20
[REVIEW] Add modifiers to enable multiple columns

VibhuJawa:
This pull request refactors and generalizes the document modification logic to support both single-input and multi-input fields, making the `Modify` processing stage more flexible and robust. 

**Core API and Logic Refactoring**

* The `DocumentModifier` abstract base class and its `modify_document` method now support both single and multiple input values, allowing for more flexible document transformations. [[1]](diffhunk://#diff-b0ca5adfb02d9ee5a0d2975cb9c4b2f5e09f482841be0290e41d9b16e34de8f1L22-R27) [[2]](diffhunk://#diff-b0ca5adfb02d9ee5a0d2975cb9c4b2f5e09f482841be0290e41d9b16e34de8f1L35-R39)
* The `Modify` stage now uses `input_fields` and `output_fields` instead of `text_field`, supporting single fields, lists of fields, and per-modifier multi-input fields. Output behavior is clarified: in-place modification is only allowed when each modifier has a single input field; otherwise, explicit output fields are required.

**Validation and Normalization Helpers**

* New helper functions `_normalize_input_fields` and `_normalize_output_fields` handle the expansion and validation of input and output field specifications, ensuring correct mapping between modifiers and their corresponding fields.
* The `_validate_and_normalize_modifiers` function now validates against the new `input_fields` argument and enforces stricter checks for field-modifier correspondence.

### DIFF
--- a/ray-curator/ray_curator/backends/experimental/ray_actor_pool/executor.py
+++ b/ray-curator/ray_curator/backends/experimental/ray_actor_pool/executor.py
@@ -1,6 +1,7 @@
 import uuid
 from typing import TYPE_CHECKING
 
+import numpy as np
 import ray
 from loguru import logger
 from ray.util.actor_pool import ActorPool
@@ -88,6 +89,7 @@ class RayActorPoolExecutor(BaseExecutor):
                 else:
                     actor_pool = self._create_actor_pool(stage, num_actors)
 
+                logger.info(f"Created actor pool for {stage.name} with {num_actors} actors")
                 # Process tasks through this stage using ActorPool
                 current_tasks = self._process_stage_with_pool(actor_pool, stage, current_tasks)
 
@@ -139,7 +141,7 @@ class RayActorPoolExecutor(BaseExecutor):
                 .options(
                     num_cpus=stage.resources.cpus,
                     num_gpus=stage.resources.gpus,
-                    name=f"{stage.name}-{actor_idx}",
+                    name=f"{stage.name}Actor-{actor_idx}",
                 )
                 .remote(
                     stage=stage,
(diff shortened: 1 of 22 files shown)


ray-curator/tests/stages/text/modules/test_modifiers.py
```diff
@@ -490,34 +490,118 @@ def test_docmodifier_single_normalization_and_name(self) -> None:
...
+        m.setup()
+        out = m.process(doc_batch)
+
+        assert out is not None
+        assert out.data.columns.tolist() == ["text", "stripped", "shouted"]
+        assert out.data["text"].tolist() == ["  hello  ", " world"]
+        assert out.data["stripped"].tolist() == ["hello", "world"]
+        assert out.data["shouted"].tolist() == ["  HELLO  ", " WORLD"]
+
+    def test_multi_input_callable_success(self) -> None:
```
sarahyurick, line 563: Very nice!

VibhuJawa: @sarahyurick , @Maghoumi  asked if he can create modifiers that work on multiple columns. 

The main use-case for this is for implementing a bunch of custom ops. Like a modifier that can add a token counter to the dataset. Or a modifier that applies a chat template to a text based on multiple fields.

Maghoumi: Yeah, also it was a feature I requested in our Live VR a while back, and it's great to see it implemented natively now :)

sarahyurick: Got it, thanks! Sounds like we can try to use it for https://github.com/NVIDIA-NeMo/Curator/tree/main/tutorials/llama-nemotron-data-curation eventually. I can tag it along with the follow-ups from here: https://github.com/NVIDIA-NeMo/Curator/issues/706.

PR #924, opened 2025-08-21
[WIP] Text Removal Stage

praateekmahajan:
## Usage

```python
```
## Checklist


ray-curator/ray_curator/stages/deduplication/removal.py
```diff
@@ -0,0 +1,102 @@
...
+        Our deduplicator should've written out a parquet file with the IDs to remove.
+        We read that file, filter the input dataframe to only include the IDs to remove,
+        and return the filtered dataframe.
+        We optimize by not loading the whole ids to remove into memory, but only loading the ids that are in the range of the input dataframe.
+        """
+        df = task.to_pandas()
+        input_df_t0 = time.perf_counter()
+        min_id = df[self.id_field].min()
+        max_id = df[self.id_field].max()
+        input_df_time = time.perf_counter() - input_df_t0
```
VibhuJawa: Better name: 

```suggestion
        input_df_min_max_time = time.perf_counter() - input_df_t0
```

ray-curator/ray_curator/stages/deduplication/removal.py
```diff
@@ -0,0 +1,102 @@
...
+        # Filter the parquet files for IDs to remove within this range
+        read_dupes_t0 = time.perf_counter()
+        removal_df = pd.read_parquet(
+            self.ids_to_remove_path,
+            filters=[("id", ">=", min_id), ("id", "<=", max_id)],
```
sarahyurick: Maybe "id" should not be hardcoded here?

ray-curator/ray_curator/stages/deduplication/removal.py
```diff
@@ -0,0 +1,102 @@
...
+            columns=["id"],
+            **self.read_kwargs,
+            storage_options=self.read_kwargs.get("storage_options"),
```
VibhuJawa: We should probably default to arrow read and engine="pyarrow"  and "dtype_backend"="pyarrow"  here unless overwritten by user, helped provide speedup earlier

ray-curator/ray_curator/stages/deduplication/removal.py
```diff
@@ -0,0 +1,102 @@
...
+        )
+        read_dupes_time = time.perf_counter() - read_dupes_t0
+        removal_ids = set(removal_df["id"].tolist())
+
+        # Filter out documents with IDs in the removal set using pandas
+        time_to_remove_t0 = time.perf_counter()
+        df = df[~df[self.id_field].isin(removal_ids)]
```
VibhuJawa: Assuming we have benchmarked and this is faster than joins ?

VibhuJawa (dismissed):
Mostly looks good, have feedback regarding reading files, and set operation

ray-curator/ray_curator/stages/deduplication/removal.py
```diff
@@ -0,0 +1,102 @@
...
+        input_df_time = time.perf_counter() - input_df_t0
+        # Filter the parquet files for IDs to remove within this range
+        read_dupes_t0 = time.perf_counter()
+        removal_df = pd.read_parquet(
+            self.ids_to_remove_path,
+            filters=[("id", ">=", min_id), ("id", "<=", max_id)],
+            columns=["id"],
+            **self.read_kwargs,
+            storage_options=self.read_kwargs.get("storage_options"),
+        )
```
sarahyurick: I was wondering if this could be generalized to fall back to reading the full `self.ids_to_remove_path` dataset, if this fails for some reason?

Also, maybe it would be good to check the ID column type is an int?

ray-curator/ray_curator/stages/deduplication/removal.py
praateekmahajan: Let me check that, good idea

ray-curator/ray_curator/stages/deduplication/removal.py
praateekmahajan: `id` is the field is written out by the `identify_duplicates.py` i..e is hardcoded irrespective of `id_field`

Let's see what @ayushdg  has in fuzzy and we can change it then

ray-curator/ray_curator/stages/deduplication/removal.py
praateekmahajan: I'd rather fail loudly if this read fails for some reason just because this only partially loads the dataset rather than all of it

ray-curator/ray_curator/stages/deduplication/removal.py
praateekmahajan: The cost of this is negligible. The whole stage is faster than our JsonlRead and ParquetWrite stage, and within the stage the cost of read_dupes >>>> input_df_time ~ time_to_remove_time

ray-curator/ray_curator/stages/deduplication/removal.py
sarahyurick: All the more reason, it could be a configurable parameter and default to "id" or something else as needed.

ray-curator/ray_curator/stages/deduplication/removal.py
sarahyurick: I meant, will we always expect the ID field to behave the same here, even for other modalities that will use dedupe? It could be more generic if the filter is optional somehow, unless we always expect it to follow this convention and we always expect the column to be a number?

ray-curator/ray_curator/stages/deduplication/removal.py
praateekmahajan: I think it makes sense, I'll add id field now

ray-curator/ray_curator/stages/deduplication/removal.py
praateekmahajan: even if id column is not an int this will still work. And we do expect ids to remove to be parquet so metadata filtering will always work

praateekmahajan: Also added test to check that tests for str field

ray-curator/ray_curator/stages/deduplication/removal.py
praateekmahajan: Added along with test,  thanks for the callout

ray-curator/ray_curator/stages/deduplication/removal.py
sarahyurick: Good with me, thanks!

sarahyurick (approved):

ray-curator/ray_curator/stages/deduplication/removal.py
VibhuJawa: Sounds good.

ray-curator/ray_curator/stages/deduplication/removal.py
```diff
@@ -0,0 +1,104 @@
...
+        max_id = df[self.id_field].max()
+        input_df_min_max_time = time.perf_counter() - input_df_t0
+        # Filter the parquet files for IDs to remove within this range
+        read_dupes_t0 = time.perf_counter()
+        removal_df = pd.read_parquet(
+            self.ids_to_remove_path,
+            filters=[(self.duplicate_id_field, ">=", min_id), (self.duplicate_id_field, "<=", max_id)],
+            columns=[self.duplicate_id_field],
+            **self.read_kwargs,
+            storage_options=self.read_kwargs.get("storage_options"),
```
ayushdg: The kwarg for `storage_options` will get repeated twice since we're also passing `**self.read_kwargs`?

sarahyurick:

ray-curator/ray_curator/stages/deduplication/removal.py
```diff
@@ -0,0 +1,104 @@
...
+
+import pandas as pd
+
+from ray_curator.stages.base import ProcessingStage
+from ray_curator.stages.deduplication.id_generator import CURATOR_DEDUP_ID_STR
+from ray_curator.tasks import DocumentBatch
+
+
+@dataclass
+class DuplicatesRemovalStage(ProcessingStage[DocumentBatch, DocumentBatch]):
```
praateekmahajan: @sarahyurick / @huvunvidia i'll move this to `stages/text/deduplication/removal.py` and call it TextDuplicatesRemovalStage, and in Image you can follow suite.

ray-curator/ray_curator/stages/deduplication/removal.py
praateekmahajan: Good catch; removed it

ray-curator/ray_curator/stages/text/deduplication/removal.py
```diff
@@ -0,0 +1,103 @@
...
+            filters=[(self.duplicate_id_field, ">=", min_id), (self.duplicate_id_field, "<=", max_id)],
+            columns=[self.duplicate_id_field],
+            **self.read_kwargs,
+        )
+        read_dupes_time = time.perf_counter() - read_dupes_t0
+
+        # Filter out documents with IDs in the removal set using pandas
+        time_to_remove_t0 = time.perf_counter()
+        removal_ids = set(removal_df[self.duplicate_id_field].tolist())
+        df = df[~df[self.id_field].isin(removal_ids)]
```
ayushdg, line 86: Pandas allows for the removal_ids arg to be a df/series here. Idk if that matters given perf but might be able to avoid going to a list.

PR #928, opened 2025-08-21
ci: Migrate ray-curator docker to uv

thomasdhc:
## Usage

```python
```
## Checklist


ayushdg:
@thomasdhc Would it be possible to also include conda in the Dockerfile? That way the tutorials can leverage conda (mambaforge/miniforge) without needing to set conda up.

ray-curator/docker/Dockerfile
```diff
@@ -0,0 +1,121 @@
...
+
+# Install uv
+ENV UV_VERSION="0.7.2"
+RUN curl -LsSf https://astral.sh/uv/${UV_VERSION}/install.sh | sh
+ENV PATH="/root/.local/bin:$PATH"
+ENV UV_PROJECT_ENVIRONMENT=/opt/venv
+ENV UV_CACHE_DIR=/opt/uv_cache
+ENV PATH="$UV_PROJECT_ENVIRONMENT/bin:$PATH"
+ENV UV_LINK_MODE=copy
+RUN uv venv ${UV_PROJECT_ENVIRONMENT} --system-site-packages
```
abhinavg4, line 55: Just wondering, can't we do pip install `nemo-curator` here, which installs uv automatically. Else just `pip install uv`.

Thinking from a user's perspective, it would be nice to have this file just be a
```
pip install nemo-curator
bash ffmpeg.sh
bash ....
```

ray-curator/docker/Dockerfile
```diff
@@ -0,0 +1,121 @@
...
+
+FROM build as nemo_curator
+
+# install video dependency
+RUN apt-get update && apt-get install -y \
+    libcrypt-dev \
```
abhinavg4: I'm a bit worried about each of these. 

1. We need OSRB for each and ensure they have GPL.
2. Can we put them inside something like `install_video.sh`. Basically I just want us to provide a simple interface to the end user

ray-curator/docker/Dockerfile
```diff
@@ -0,0 +1,121 @@
...
+    texinfo \
+    vainfo \
+    wget \
+    yasm \
+    zlib1g-dev \
+    && rm -rf /var/lib/apt/lists/*
+
+# Install ffmpeg
+COPY ray-curator/docker/common/install_ffmpeg.sh .
+RUN bash install_ffmpeg.sh --FFMPEG_VERSION=7.0.1 --NVCODEC_VERSION=12.1.14.0 && \
```
abhinavg4: Let's put these versions inside `install_video.sh`?

ray-curator/docker/Dockerfile
```diff
@@ -0,0 +1,121 @@
...
+    rm install_ffmpeg.sh
+
+# Copy ray-curator
+COPY ray-curator /opt/ray-curator
+
+# Install dependencies
+RUN cd ray-curator && \
```
abhinavg4: Saem comment as above. Can we do something like 

`pip install nemo-curator[cuda]` or `pip install nemo-curator[cpu]` ?

abhinavg4 (changes requested):
My main comment is to look into making the docker file as simple as possible. 

If we want someone to use curator outside the container, we aim at providing just 2 commands:
`pip install nemo-curator` and `bash setup.sh`

Why can't we do the same in the dockerfile ?

ray-curator/docker/Dockerfile
```diff
@@ -0,0 +1,121 @@
...
+
+# Install base dependency
+ENV PIP_BREAK_SYSTEM_PACKAGES=1
+ENV DEBIAN_FRONTEND=noninteractive
+RUN apt-get update && apt-get install -y --no-install-recommends \
+    python3.12-dev \
+    python3 \
+    python3-dev \
+    python3-venv \
+    python-is-python3\
```
sarahyurick, line 40: Not sure if this matters, but should all of these specify Python 3.12 instead of just Python 3?

ray-curator/docker/Dockerfile
sarahyurick: Agree, any logic that is strictly related to video can be moved to its own file.

praateekmahajan: ➕

ray-curator/docker/Dockerfile
```diff
@@ -0,0 +1,121 @@
...
+
+FROM cuda AS build
+
+ARG CURATOR_ENV=ci
+ENV CURATOR_ENVIRONMENT=${CURATOR_ENV}
+
+ENV NVIDIA_PRODUCT_NAME="NeMo Curator"
+
+# Install base dependency
+ENV PIP_BREAK_SYSTEM_PACKAGES=1
```
praateekmahajan, line 33: What's this for?

ray-curator/docker/Dockerfile
praateekmahajan, line 55: Can we add a comment why we do this, IIUC its to install the dependencies at that path above.. if so should we call it something else? e.g. base?

praateekmahajan: And if the env variable `UV_PROJECT_ENVIRONMENT` is set do we need to pass it each time we do `uv venv`?

ray-curator/docker/Dockerfile
```diff
@@ -0,0 +1,121 @@
...
+    zlib1g-dev \
+    && rm -rf /var/lib/apt/lists/*
+
+# Install ffmpeg
+COPY ray-curator/docker/common/install_ffmpeg.sh .
+RUN bash install_ffmpeg.sh --FFMPEG_VERSION=7.0.1 --NVCODEC_VERSION=12.1.14.0 && \
+    rm install_ffmpeg.sh
+
+# Copy ray-curator
+COPY ray-curator /opt/ray-curator
```
praateekmahajan, line 65: Idk enough about docker layer caching but  I read here that its recommended to copy over pyproject.toml and lock file first then run `uv sync --locked` before copying over the remaining code? Is that relevant?

https://devblogs.microsoft.com/ise/dockerizing-uv/

ray-curator/docker/Dockerfile
```diff
@@ -0,0 +1,121 @@
+# Copyright (c) 2025, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
praateekmahajan: Can you also update README / Contributing.md to show how to locally develop moving forward? 
I'd believe it'll be like install uv in root (conda or something) and then do `uv pip isntall -e .`?

ray-curator/docker/Dockerfile
praateekmahajan: Discussed offline, in the container we want to install from the lock file to guarantee dependencies that we've developed and tested with.

ray-curator/docker/Dockerfile
praateekmahajan: Discussed offline that the user use case is still that, while for container and local dev we should be using `uv sync --lock` ` uv add` etc

ray-curator/docker/Dockerfile
thomasdhc: Yeah, let me update the docs

ray-curator/docker/Dockerfile
thomasdhc: It allows pip install to override packages already installed on system

ray-curator/docker/Dockerfile
thomasdhc: some of the packages can't specify more specific version ie `python-is-python3`, install python 3.12 links python3 to use 3.12

ray-curator/docker/Dockerfile
thomasdhc: No uv venv is setup here once and added to the system path. It is defined here to further run configuration `uv venv ${UV_PROJECT_ENVIRONMENT} --system-site-packages` within the container itself users don't need to run anything specific to venv

thomasdhc: This is used for container setup and configuration, not instructions users will follow to install.

ray-curator/docker/Dockerfile
thomasdhc: 1. Yes, if we want video curator will have to do so.
2. Ok let me move the installation

ray-curator/docker/Dockerfile
thomasdhc: If you look at install_ffmpeg.sh the version are already defined within the file, arguments will let us override in case we want to test different versions. I can remove it if you don't want these to be surfaced.

ray-curator/docker/Dockerfile
thomasdhc: Yeah, we can further optimize

ray-curator/docker/Dockerfile
thomasdhc: If we don't install python3, commands such as python3 will break, users will have to run python3.12

praateekmahajan (approved):
LGTM! Please wait for approval from others who haven't reviewed yet.

How is the lock file generated? Which all groups does it have?
We should have a fast follow of having some consistency for groups that we have since we'll also need to add `cosmos-xenna[gpu]` so should it be `_cuda` or `_cuda12x` or `gpu` (and now we have `nvenc` )

I can't imagine a user will have it easy to understand which one they need if they wish to work on a submodule especially video cc @suiyoubi

ray-curator/CONTRIBUTING.md
```diff
@@ -10,27 +10,44 @@
...
+# Sync all (includes dali, deduplication_cuda12x, text, video, video_cuda)
+uv sync --extra all
+```
 
 ### Dev Pattern
 
 - All upstream work/changes with the new API and ray backend should target the `NeMo-Curator/ray-api` branch.
 - When re-using code already in `NeMo-Curator/nemo_curator`, use `git mv` to move those source files into the `ray-curator/ray_curator` namespace.
 - Sign and signoff commits with `git commit -sS`. (May be relaxed in the future)
+- If project dependencies are updated a new uv lock file needs to be generated. Run `uv lock` and add the changes of the new uv.lock file.
```
thomasdhc, line 50: @praateekmahajan the lock file is generated via uv lock command

thomasdhc: The lock file is generated for all dependencies

ray-curator/docker/Dockerfile
```diff
@@ -0,0 +1,87 @@
...
```
sarahyurick: Nit: this looks like a different copyright from the rest of our files?

abhinavg4 (approved):
Looks good. Thanks

PR #937, opened 2025-08-22
Add FuzzyDeduplicationPipeline

ayushdg:
## Description

Depends on changes from #945.
Adds an end to end workflow to run fuzzy duplicate identification.

The workflow has the following expectations:

1. No ID generator actor already exists. This workflow will create one specifically for fuzzy dedup and use that for deduplication.
2. `input_path` points to the input dataset to deduplicate
3. `cache_path` points to the path where fuzzy deduplication intermediates such has minhashes, buckets etc are written.
4. `output_path` points to the path where the Ids to remove and the Idgenerator mapping (for future removal) will be written.

## Usage

```python
```
## Checklist

### DIFF
--- a/ray-curator/ray_curator/stages/video/clipping/transnetv2_extraction.py
+++ b/ray-curator/ray_curator/stages/video/clipping/transnetv2_extraction.py
@@ -89,7 +89,7 @@ class TransNetV2ClipExtractionStage(ProcessingStage[VideoTask, VideoTask]):
     def __post_init__(self) -> None:
         self._resources = Resources(gpu_memory_gb=self.gpu_memory_gb)
 
-    def process(self, task: VideoTask) -> VideoTask:
+    def process(self, task: VideoTask) -> VideoTask:  # noqa: C901
         video: Video = task.data
         video_name = video.input_video
         if not video.has_metadata():
@@ -104,6 +104,8 @@ class TransNetV2ClipExtractionStage(ProcessingStage[VideoTask, VideoTask]):
         if frames is None:
             error_msg = "Run `FrameExtractionStage` stage prior to `TransNetV2ClipExtractionStage`!"
             raise ValueError(error_msg)
+        if tuple(frames.shape[1:4]) == (48, 27, 3):
+            frames = frames.transpose(0, 2, 1, 3)  # transpose weight and height
         if tuple(frames.shape[1:4]) != (27, 48, 3):
             error_msg = f"Expected frames of shape 27x48x3, got {frames.shape[1:4]}."
             raise ValueError(error_msg)
--- a/ray-curator/ray_curator/stages/video/clipping/video_frame_extraction.py
+++ b/ray-curator/ray_curator/stages/video/clipping/video_frame_extraction.py
@@ -29,7 +29,7 @@ try:
     from ray_curator.utils.nvcodec_utils import PyNvcFrameExtractor
 
     _PYNVC_AVAILABLE = True
-except ImportError:
+except (ImportError, RuntimeError):
     logger.warning("PyNvcFrameExtractor not available, PyNvCodec mode will fall back to FFmpeg")
     PyNvcFrameExtractor = None
     _PYNVC_AVAILABLE = False
--- a/ray-curator/ray_curator/utils/nvcodec_utils.py
+++ b/ray-curator/ray_curator/utils/nvcodec_utils.py
@@ -35,7 +35,7 @@ try:
         Nvc.Pixel_Format.YUV444: cvcuda.ColorConversion.YUV2RGB,  # type: ignore[import-untyped]
         Nvc.Pixel_Format.NV12: cvcuda.ColorConversion.YUV2RGB_NV12,  # type: ignore[import-untyped]
     }
-except ImportError:
+except (ImportError, RuntimeError):
     logger.warning("PyNvVideoCodec is not installed, some features will be disabled.")
     Nvc = None
     cvcuda = None


ray-curator/ray_curator/stages/deduplication/fuzzy/connected_components.py
```diff
@@ -39,22 +40,22 @@ class ConnectedComponentsStage(ProcessingStage[FileGroupTask, FileGroupTask], De
     def __init__(
         self,
         output_dir: str,
```
VibhuJawa: Lets use path here ?

ray-curator/ray_curator/stages/deduplication/fuzzy/fuzzy_deduplication.py
```diff
@@ -0,0 +1,235 @@
...
+from ray_curator.utils.file_utils import get_fs
+
+
+@dataclass
+class FuzzyDeduplicationConfig:
+    """
+    Configuration for MinHash based fuzzy duplicates detection.
+    Parameters
+    ----------
+    input_dir: str | list[str]
```
VibhuJawa:
```suggestion
    input_paths: str | list[str]
```

ray-curator/ray_curator/stages/deduplication/fuzzy/fuzzy_deduplication.py
```diff
@@ -0,0 +1,235 @@
...
+        Directory or list of files containing the input dataset.
+    cache_dir: str
+        Directory to store deduplication intermediates such as minhashes/buckets etc.
+    output_dir: str
```
VibhuJawa:
```suggestion
    output_path: str
```

ray-curator/ray_curator/stages/deduplication/fuzzy/fuzzy_deduplication.py
```diff
@@ -0,0 +1,235 @@
...
+    use_64_bit_hash: bool
+        Whether to use a 32bit or 64bit hash function for minhashing.
+    bands_per_iteration: int
+        Number of bands/buckets to shuffle concurrently.
+        Larger values process larger batches by processing multiple bands
+        but might lead to memory pressures and related errors.
+    """
+
+    # I/O config
+    input_dir: str | list[str]
```
VibhuJawa:
```suggestion
    input_path: str | list[str]
```

ray-curator/ray_curator/stages/deduplication/fuzzy/fuzzy_deduplication.py
```diff
@@ -0,0 +1,235 @@
...
+    cache_dir: str
+    output_dir: str | None = None
```
VibhuJawa:
```suggestion
    output_path: str | None = None
```

ray-curator/ray_curator/stages/deduplication/fuzzy/fuzzy_deduplication.py
```diff
@@ -0,0 +1,235 @@
...
+    text_field: str = "text"
+    perform_removal: bool = False
+
+    # Minhash + LSH Config
+    seed: int = 42
+    char_ngrams: int = 24
+    num_bands: int = 20
+    minhashes_per_band: int = 13
+    use_64_bit_hash: bool = False
+    bands_per_iteration: int = 5
```
VibhuJawa: Set these via a constant please.

ray-curator/ray_curator/stages/deduplication/fuzzy/fuzzy_deduplication.py
```diff
@@ -0,0 +1,235 @@
...
+                    file_extensions=self.fuzzy_deduplication_config.input_file_extensions,
+                    files_per_partition=30,  # TODO: Replace with blocksize
+                    storage_options=self.fuzzy_deduplication_config.read_kwargs.get("storage_options", {})
+                    if self.fuzzy_deduplication_config.read_kwargs is not None
+                    else None,
+                ),
+            )
+        stages.append(
+            MinHashStage(
+                output_dir=self.fuzzy_deduplication_config.cache_dir,
```
VibhuJawa:
```suggestion
               output_path=self.fuzzy_deduplication_config.cache_dir,
```

ray-curator/ray_curator/stages/deduplication/fuzzy/minhash.py
```diff
@@ -216,8 +214,8 @@ class MinHashStage(ProcessingStage[FileGroupTask, FileGroupTask], DeduplicationI
     def __init__(  # noqa: PLR0913
         self,
         output_dir: str,
```
VibhuJawa:
```suggestion
        output_path: str,
```

ray-curator/ray_curator/tasks/file_group.py
```diff
@@ -28,4 +28,7 @@ def validate(self) -> bool:
         if len(self.data) == 0:
             logger.warning(f"No files to process in task {self.task_id}")
             return False
+        if not isinstance(self.data, list) or not all(isinstance(item, str) for item in self.data):
+            err = f"Invalid data type in task {self.task_id}"
+            raise ValueError(err)
```
ayushdg: Found a bug where we could pass in a string here and accessing elements like `task.data[0]` for fsspec checks etc would return the first char rather then the first file.

Should this be a reassignment instead of an error here?

ray-curator/ray_curator/stages/deduplication/fuzzy/fuzzy_deduplication.py
```diff
@@ -0,0 +1,259 @@
...
+        input_path: str | list[str],
+        cache_path: str,
+        output_path: str | None = None,
+        input_filetype: Literal["jsonl", "parquet"] = "parquet",
+        input_file_extensions: list[str] | None = None,
+        read_kwargs: dict[str, Any] | None = None,
+        cache_kwargs: dict[str, Any] | None = None,
+        write_kwargs: dict[str, Any] | None = None,
+        text_field: str = "text",
+        perform_removal: bool = False,
```
ayushdg: Will remove this if @praateekmahajan the remove duplicates pipeline for text works here as well.

ray-curator/ray_curator/stages/deduplication/fuzzy/generate_duplicate_ids.py
```diff
@@ -0,0 +1,136 @@
...
+            spill_memory_limit=spill_memory_limit,
+            enable_statistics=enable_statistics,
+        )
+
+    def _get_removal_ids(self, df: "cudf.DataFrame") -> "cudf.DataFrame":
+        """
+        Get the removal ids for the given dataframe.
+        """
+        removal_ids = df[df[self.duplicate_group_field].duplicated(keep="first")][self.document_id_field]
+        removal_ids = removal_ids.sort_values(ignore_index=True)
```
ayushdg: @praateekmahajan PTAL if we need to add anything more here for efficient removal

ray-curator/ray_curator/stages/deduplication/fuzzy/generate_duplicate_ids.py
```diff
@@ -0,0 +1,136 @@
...
+        write_kwargs["index"] = write_kwargs.get("index", False)
+
+        result_tasks = []
+        for partition_id, partition in self._actor_obj.extract():
+            shuffled_partition_df = pylibcudf_to_cudf_dataframe(partition, column_names=self.output_columns)
+            num_groups = shuffled_partition_df[self.duplicate_group_field].nunique()
+            removal_ids = self._get_removal_ids(shuffled_partition_df)
+
+            output_file = self.output_fs.sep.join([self.output_path, f"part.{partition_id}.parquet"])
+            removal_ids.to_parquet(output_file, **write_kwargs)
```
ayushdg: Do you also recommend setting row group sizes here?

ray-curator/ray_curator/stages/deduplication/fuzzy/fuzzy_deduplication.py
```diff
@@ -118,43 +31,149 @@ class FuzzyDeduplicationPipeline:
...
     def _generate_minhash_pipeline(self, generate_input_filegroups: bool) -> Pipeline:
         stages = []
         if generate_input_filegroups:
             stages.append(
                 FilePartitioningStage(
-                    file_paths=self.fuzzy_deduplication_config.input_dir,
-                    file_extensions=self.fuzzy_deduplication_config.input_file_extensions,
+                    file_paths=self.input_path,
+                    file_extensions=self.input_file_extensions,
                     files_per_partition=30,  # TODO: Replace with blocksize
```
VibhuJawa: So blocksize works now , if you wnat to replace

ray-curator/ray_curator/stages/deduplication/fuzzy/fuzzy_deduplication.py
```diff
@@ -163,36 +182,41 @@ def _generate_minhash_pipeline(self, generate_input_filegroups: bool) -> Pipelin
...
-        cache_dir_fs = get_fs(self.fuzzy_deduplication_config.cache_dir, self.fuzzy_deduplication_config.cache_kwargs)
+        cache_dir_fs = get_fs(self.cache_path, self.cache_kwargs)
         return Pipeline(
             name="lsh_duplicate_identification_pipeline",
             stages=[
                 FilePartitioningStage(
-                    file_paths=cache_dir_fs.sep.join([self.fuzzy_deduplication_config.cache_dir, "MinHashStage"]),
+                    file_paths=cache_dir_fs.sep.join([self.cache_path, "MinHashStage"]),
                     file_extensions=[".parquet"],
                     files_per_partition=8,
```
VibhuJawa: Unsure if this logic requires fpp or blocksize 
```suggestion
                    blocksize="128MB",
```

ray-curator/ray_curator/stages/deduplication/fuzzy/generate_duplicate_ids.py
praateekmahajan: Yup we should always sort + write with row_groups..
in pandas there is `row_group_size` so i just do max(1, len(removal_ids) // 10) s.t. we have 10 row groups

sarahyurick:

ray-curator/ray_curator/backends/experimental/ray_actor_pool/shuffle_adapter.py
```diff
@@ -16,7 +16,8 @@
     from ray_curator.stages.shuffler.stage import ShuffleStage
 
 
-@ray.remote
+# TODO: Remove once UCX memory usage with GPU staging buffers is fixed.
```
praateekmahajan, line 19: Do we have a ticket in the projcet for this?

ray-curator/ray_curator/stages/deduplication/fuzzy/buckets_to_edges.py
```diff
@@ -46,7 +46,7 @@ class BucketsToEdgesStage(ProcessingStage[FileGroupTask, FileGroupTask]):
 
     def __init__(
         self,
-        output_dir: str,
+        output_path: str,
         doc_id_field: str = CURATOR_DEDUP_ID_STR,
```
praateekmahajan, line 50: sort of nit now, but id_field or doc_id_field

ray-curator/ray_curator/stages/deduplication/fuzzy/fuzzy_deduplication.py
```diff
@@ -0,0 +1,288 @@
...
+            logger.warning("output_path will be unused as perform_removal is False")
+
+    def _generate_minhash_pipeline(self, generate_input_filegroups: bool) -> Pipeline:
+        stages = []
+        if generate_input_filegroups:
+            stages.append(
+                FilePartitioningStage(
+                    file_paths=self.input_path,
+                    file_extensions=self.input_file_extensions,
+                    blocksize="1GiB",  # compressed size of 1GiB
```
sarahyurick: Should this be configurable with default "1GiB"?

ray-curator/ray_curator/stages/deduplication/fuzzy/fuzzy_deduplication.py
```diff
@@ -0,0 +1,288 @@
...
+                read_kwargs=self.read_kwargs,
+                write_kwargs=self.cache_kwargs,
+            ),
+        )
+        return Pipeline(
+            name="minhash_pipeline",
+            stages=stages,
+        )
+
+    def _generate_lsh_pipeline(self) -> Pipeline:
```
praateekmahajan: nit : _generate_pipeline or _create_pipeline

ray-curator/ray_curator/stages/deduplication/fuzzy/fuzzy_deduplication.py
```diff
@@ -0,0 +1,288 @@
...
+        cache_dir_fs = get_fs(self.cache_path, self.cache_kwargs)
+        return Pipeline(
+            name="lsh_duplicate_identification_pipeline",
+            stages=[
+                FilePartitioningStage(
+                    file_paths=cache_dir_fs.sep.join([self.cache_path, "MinHashStage"]),
+                    file_extensions=[".parquet"],
+                    blocksize="2GiB",
```
sarahyurick: Similar comment here, and I was wondering why they are different?

ray-curator/ray_curator/stages/deduplication/gpu_utils.py
```diff
@@ -1,5 +1,25 @@
...
+
+
 def align_down_to_256(memory_size: int) -> int:
     """
     Aligns a memory size down to the nearest multiple of 256.
     """
     return (memory_size // 256) * 256
+
+
+def get_device_free_memory() -> int | None:
```
sarahyurick, line 12: I thought we had this function floating around somewhere else in the repo already?

ray-curator/ray_curator/backends/experimental/ray_actor_pool/shuffle_adapter.py
ayushdg: Added here: https://github.com/orgs/NVIDIA-NeMo/projects/15/views/13?sliceBy%5Bvalue%5D=DeDupe&filterQuery=&pane=issue&itemId=126257829 

The original issue is in an nvbug

ray-curator/ray_curator/stages/deduplication/fuzzy/buckets_to_edges.py
ayushdg: The reason this is `doc_id_field` is because the input df also contains `_bucket_id` referring to the bucket id.

ray-curator/ray_curator/stages/deduplication/fuzzy/fuzzy_deduplication.py
```diff
@@ -0,0 +1,288 @@
+import time
```
praateekmahajan: should we call it fuzzy_dup.py or base.py.. i'll make the same change (not that it matters too much since user will do from ray_curator.stages.deduplication import FuzzyDuplicationWorkflow)

ray-curator/ray_curator/stages/deduplication/fuzzy/__init__.py
```diff
@@ -0,0 +1,17 @@
+from ray_curator.stages.deduplication.fuzzy.buckets_to_edges import BucketsToEdgesStage
+from ray_curator.stages.deduplication.fuzzy.connected_components import ConnectedComponentsStage
+from ray_curator.stages.deduplication.fuzzy.fuzzy_deduplication import (
+    FuzzyDeduplicationWorkflow,
+)
+from ray_curator.stages.deduplication.fuzzy.generate_duplicate_ids import GenerateRemovalIDs
+from ray_curator.stages.deduplication.fuzzy.lsh.stage import LSHStage
+from ray_curator.stages.deduplication.fuzzy.minhash import MinHashStage
+
+__all__ = [
```
praateekmahajan: Do you foresee user having to use any / all of the classes except `FuzzyDeduplicationWorkflow` if not I'd say let's just expose through init the ones we want user to interface with while advanced users can fetch from absolute path

ray-curator/ray_curator/stages/deduplication/gpu_utils.py
ayushdg: Found a variant of it here: https://github.com/NVIDIA-NeMo/Curator/blob/main/ray-curator/ray_curator/stages/resources.py that does it on the client instead of workers and returns  total memory.

Opened an item to track unifying: https://github.com/orgs/NVIDIA-NeMo/projects/15/views/13?sliceBy%5Bvalue%5D=Core%2FBase&filterQuery=&pane=issue&itemId=126258989

ray-curator/ray_curator/stages/deduplication/fuzzy/fuzzy_deduplication.py
```diff
@@ -0,0 +1,288 @@
...
+                    output_path=self.cache_path,
+                    read_kwargs=self.read_kwargs,
+                    write_kwargs=self.cache_kwargs,
+                ),
+                ConnectedComponentsStage(
+                    output_path=self.cache_path,
+                    read_kwargs=self.read_kwargs,
+                    write_kwargs=self.cache_kwargs,
+                ),
+                GenerateRemovalIDs(
```
praateekmahajan: let's come up with a common name for this stage? mine is IdentifyDuplicatesStage?

ray-curator/ray_curator/stages/deduplication/fuzzy/fuzzy_deduplication.py
ayushdg: Personal preference: I prefer `fuzzy_deduplication/fuzzyduplicates.py` since I associate `base` with a base class/parent class type implementation, but as you mentioned it shouldn't matter too much.

ray-curator/ray_curator/stages/deduplication/fuzzy/__init__.py
ayushdg: I would anticipate most users working with the Workflow directly and only advanced/slurm/multi-node users working with the individual stages. Happy to make that change

ray-curator/ray_curator/stages/deduplication/fuzzy/fuzzy_deduplication.py
```diff
@@ -0,0 +1,288 @@
...
+
+        Args:
+            initial_tasks:
+            Set of FileGroupTasks generated by a previous stage pointing to the dataset to be deduplicated.
+            If not provided, the pipeline will generate the input tasks based on the input_dir and input_file_extensions.
+        """
+        self._validate_initial_tasks(initial_tasks)
+        executor = RayActorPoolExecutor(config=self.executor_config)
+        try:
+            get_id_generator_actor()
```
praateekmahajan: what's the use case to get and not create always? I worry get might allow bad case where user might've had one lingering from previous run or something

ray-curator/ray_curator/stages/deduplication/fuzzy/generate_duplicate_ids.py
```diff
@@ -0,0 +1,142 @@
...
+from ray_curator.stages.shuffler.rapidsmpf_shuffler import pylibcudf_to_cudf_dataframe
+from ray_curator.stages.shuffler.stage import ShuffleStage
+from ray_curator.tasks import FileGroupTask
+from ray_curator.utils.file_utils import get_fs
+
+if TYPE_CHECKING:
+    import cudf
+
+
+class GenerateRemovalIDs(ShuffleStage):
```
praateekmahajan: nit : IDs or Ids

ray-curator/ray_curator/stages/deduplication/fuzzy/fuzzy_deduplication.py
ayushdg: The prev fuzzy dedup impl would read the minhashes with a blocksize of `2GiB` before shuffling. This matches the same pattern. 
Side note: RapidsMPFShuffle is sensitive to the number of partitions so a larger blcocksize is preferred here

ray-curator/ray_curator/stages/deduplication/fuzzy/fuzzy_deduplication.py
ayushdg: That works for me. I'll update to `IdentifyDuplicatesStage`

ray-curator/ray_curator/stages/deduplication/fuzzy/fuzzy_deduplication.py
ayushdg: Ah, I figured we could re-use from old runs rather than have the user restart a cluster (or delete the actor) b/w exact & fuzzy & semdedup. But maybe that's not preferred?

ray-curator/ray_curator/stages/deduplication/fuzzy/generate_duplicate_ids.py
```diff
@@ -0,0 +1,142 @@
...
+        write_kwargs = self.write_kwargs.copy()
+        write_kwargs["index"] = write_kwargs.get("index", False)
+
+        result_tasks = []
+        for partition_id, partition in self._actor_obj.extract():
+            shuffled_partition_df = pylibcudf_to_cudf_dataframe(partition, column_names=self.output_columns)
+            num_groups = shuffled_partition_df[self.duplicate_group_field].nunique()
+            removal_ids = self._get_removal_ids(shuffled_partition_df)
+
+            output_file = self.output_fs.sep.join([self.output_path, f"part.{partition_id}.parquet"])
```
praateekmahajan: I'm unclear on the the details of a ShuffleStage, but is partition_id guaranteed to be unique across actors; i.e. is there a case where we have different processes overwriting the same files

ayushdg: Yes `partition_id` is guaranteed to be unique among actors. If a shuffle has `N` output partitions the shuffler assigns `~N/K` partition_id's to each actor/process. That actors receives the specific partition post shuffle.

ray-curator/ray_curator/stages/deduplication/fuzzy/generate_duplicate_ids.py
```diff
@@ -0,0 +1,142 @@
...
+            # If user has not specified row_group_size_rows, set it to the lower of 10% of the number of removal ids or 1M (default) or a minimum of 100k (for small datasets)
+            write_kwargs["row_group_size_rows"] = write_kwargs.get(
```
praateekmahajan: If there are 99k rows, IIUC we'll end up with 1 row group. Is there a reason to not split into 10 still?

Essentially asking if we should lower the threshold much more?

ray-curator/ray_curator/stages/deduplication/gpu_utils.py
praateekmahajan, line 12: Want to add a test for this?

ray-curator/ray_curator/stages/deduplication/fuzzy/generate_duplicate_ids.py
ayushdg: Happy to lower it to 1000 or some other lower value. This number was arbitrary. Just want to avoid cases (like CI) or small datasets where we're writing out 10 row groups for a 100 row or 1000 row dataset.

Does 1k sounds resonable?

ray-curator/ray_curator/stages/deduplication/gpu_utils.py
ayushdg: open to ideas here: Since CI/dev gpu memory can differ (esp free memory), I'm not sure how to test.

ray-curator/ray_curator/stages/deduplication/fuzzy/fuzzy_deduplication.py
ayushdg: updated

ray-curator/ray_curator/stages/deduplication/fuzzy/__init__.py
ayushdg: updated

ray-curator/ray_curator/stages/deduplication/fuzzy/fuzzy_deduplication.py
praateekmahajan: What about `workflow.py`

ray-curator/ray_curator/stages/deduplication/fuzzy/fuzzy_deduplication.py
praateekmahajan: Ah basically, I think we might be at risk where the IdGenerator keys from each other get overwritten / keys get too big. So I'd imagine we do something like this

```python

create_id_gen()
run_fuzzy()
write_id_gen(f"{dataset_name}_fpp-{fpp}_fuzzy.json")


create_id_gen()
run_exact()
write_id_gen(f"{dataset_name}_fpp-{fpp}_exact.json")

```
And  ideally

```python

with IdGenerator.context(f"{dataset_name}_fpp-{fpp}_fuzzy.json"):
    run_fuzzy()

with IdGenerator.context(f"{dataset_name}_fpp-{fpp}_fuzzy.json"):
    run_exact()

```

ray-curator/ray_curator/stages/deduplication/fuzzy/fuzzy_deduplication.py
```diff
@@ -0,0 +1,295 @@
...
+                f"Connected components pipeline completed in {connected_components_end_time - connected_components_start_time} seconds"
+            )
+            num_removed_documents = sum(
+                task._metadata.get("num_removal_ids", 0) for task in connected_components_tasks
+            )
+            logger.info(f"Number of documents removed: {num_removed_documents}")
+
+        end_time = time.time()
+
+        logger.info(f"Fuzzy deduplication pipeline completed in {end_time - start_time} seconds")
```
praateekmahajan: I notice you haven't written out the IdGenerator path too, which we should, right? Or should the user do that outside?
I can see bothways, but maybe for now let's be prescriptive, and based on the decision we make we can decide on the above thread whether it should be get or create... i.e. if you decide to go ahead with the fact that id gen stuff happens in example.py
```
# example.py
create_id_gen(...)
run_fuzzy_workflow() # this code
write_id_gen(...)
```

or 
```
```
# example.py
run_fuzzy_workflow() # this code does id gen creation + write
```

File: ray-curator/ray_curator/stages/deduplication/fuzzy/buckets_to_edges.py [hunk 29]
```diff
@@ -57,14 +57,11 @@ def __init__(
         self.read_storage_options = read_kwargs.get("storage_options") if read_kwargs is not None else None
```
praateekmahajan, line 57: i haven't looked at all the stages but as an fyi we have [`check_disallowed_kwargs`
](https://github.com/NVIDIA-NeMo/Curator/blob/efe76b9d181cb7d855d09bd789d9a7abb7377200/ray-curator/ray_curator/utils/file_utils.py#L370-L393) here, which you might want to use.

ray-curator/ray_curator/stages/deduplication/fuzzy/fuzzy_deduplication.py
```diff
@@ -0,0 +1,295 @@
...
+        # I/O config
+        input_path: str | list[str],
+        cache_path: str,
+        output_path: str | None = None,
+        input_filetype: Literal["jsonl", "parquet"] = "parquet",
+        input_blocksize: str | int = "1GiB",
+        input_file_extensions: list[str] | None = None,
+        read_kwargs: dict[str, Any] | None = None,
+        cache_kwargs: dict[str, Any] | None = None,
+        write_kwargs: dict[str, Any] | None = None,
```
praateekmahajan: Seems related to this question I had
https://nvidia.slack.com/archives/C098W91J3ND/p1755909117350379

Is there only one cache_kwargs? Do you prefer cache_kwargs or just read/write_kwargs. To mimic, I think we should also have the same for SemDedup

ray-curator/ray_curator/stages/deduplication/fuzzy/fuzzy_deduplication.py
ayushdg: Good point. I'll also write out the IDGenerator state to the fuzzy dedup cache directory.

ray-curator/ray_curator/stages/deduplication/fuzzy/fuzzy_deduplication.py
ayushdg: I was viewing the pipeline to be as follows:
1. `read_kwargs` kwargs for the input
2. `cache_kwargs` kwargs for all the intermediates (minhashes etc.) written out by dedup. This will suffer from the same issue described on the thread where some stages might want different kwargs than others.
3. `write_kwargs` - This is currently unused because removal is a separate part of the workflow but the goal was to have `output_path` and `write_kwargs` be used for the removal and final write of the resulting dataset.

ray-curator/ray_curator/stages/deduplication/fuzzy/fuzzy_deduplication.py
ayushdg: I see. Until the context manager is implemented. Should Fuzzy worfklow etc also be responsible for deleting/cleaning up that actor after writing out the json file?

ray-curator/ray_curator/stages/deduplication/fuzzy/workflow.py
```diff
@@ -0,0 +1,320 @@
...
+        executor = RayActorPoolExecutor(config=self.executor_config)
+        try:
+            create_id_generator_actor()
+        except ValueError:
+            err_msg = """
+            An existing id generator actor was found. Please remove or save the existing id generator with
+            `ray_curator.stages.deduplication.id_generator.write_id_generator_to_disk` (if needed) and remove the actor with
+            `ray_curator.stages.deduplication.id_generator.kill_id_generator_actor` before running the fuzzy deduplication pipeline.
+            """
+            raise RuntimeError(err_msg) from None
```
ayushdg, line 272: @praateekmahajan as discussed this workflow only creates a new actor and raises if one exists.

ray-curator/ray_curator/stages/deduplication/fuzzy/workflow.py
```diff
@@ -0,0 +1,320 @@
...
+                    read_kwargs=self.read_kwargs,
+                    write_kwargs=self.cache_kwargs,
+                ),
+                ConnectedComponentsStage(
+                    output_path=self.cache_path,
+                    read_kwargs=self.read_kwargs,
+                    write_kwargs=self.cache_kwargs,
+                ),
+                IdentifyDuplicatesStage(
+                    output_path=self.output_path,
```
ayushdg, line 234: Writes removal ids to output_path now

ray-curator/ray_curator/stages/deduplication/fuzzy/workflow.py
```diff
@@ -0,0 +1,320 @@
...
+                    self.output_path,
+                    self.write_kwargs.get("storage_options") if self.write_kwargs is not None else None,
+                )
+                id_generator_path = output_fs.sep.join([self.output_path, ID_GENERATOR_OUTPUT_FILENAME])
+                write_id_generator_to_disk(
+                    id_generator_path,
+                    storage_options=self.write_kwargs.get("storage_options")
+                    if self.write_kwargs is not None
+                    else None,
+                )
```
ayushdg, line 315: writes id_gen to output_path.

ray-curator/ray_curator/stages/deduplication/fuzzy/workflow.py
```diff
@@ -0,0 +1,320 @@
...
+                logger.info(f"Id generator written to {id_generator_path}")
+            end_time = time.time()
+            logger.info(f"Fuzzy deduplication pipeline completed in {end_time - start_time} seconds")
+        finally:
+            kill_id_generator_actor()
```
ayushdg, line 320: kill id gen after the job is complete

ray-curator/ray_curator/stages/deduplication/fuzzy/workflow.py
```diff
@@ -0,0 +1,320 @@
...
+                num_removed_documents = sum(
+                    task._metadata.get("num_removal_ids", 0) for task in connected_components_tasks
+                )
+                logger.info(f"Number of documents removed: {num_removed_documents}")
+                output_fs = get_fs(
+                    self.output_path,
+                    self.write_kwargs.get("storage_options") if self.write_kwargs is not None else None,
+                )
+                id_generator_path = output_fs.sep.join([self.output_path, ID_GENERATOR_OUTPUT_FILENAME])
+                write_id_generator_to_disk(
```
ayushdg, line 310: TODO: Might consider moving this write after minhash if minhashes succeeds. That way users can re-use this separately if something fails and they decide to run the other stages manually outside the workflow

praateekmahajan: that's a good idea actually.. moving it just after minhash, so that they can resume from LSH

PR #943, opened 2025-08-23
Translate text documents using a multi-step LLM

TsukiSama9292:
### Added an LLM-based multi-step text translation feature.

- TranslationDataGenerator Workflows:
  - Create class instance (variable or YAML): Initialize OpenAIClient (for initial translation, reflection, and refinement of translation) & Tokenizer (used to calculate token count for long texts)...
  - Perform translation:
    - Long text segmentation: Softly segment text based on content, Tokenizer, and configuration parameters (this method first splits by sentence, then applies a soft token length limit).
    - Translate each chunk one by one, referencing the implementation from [Github: andrewyng/translation-agent](https://github.com/andrewyng/translation-agent):
      - Initial Translation Agent: Uses LLM to perform an initial translation based on source language, source text, and target language.
      - Reflection Agent: Uses LLM with the same inputs, but also includes the initial translation and target country, to reflect on and evaluate the initial translation.
      - Improvement Agent: Uses all of the above to improve the translation, enhancing overall quality.

## Usage

```python
from nemo_curator.synthetic.translate import TranslationDataGenerator

# Example usage for both function and YAML config
if __name__ == "__main__":
    # Function-based usage
    text = "Once upon a time, there were three little pig brothers..."
    generator = TranslationDataGenerator(
        base_url="http://localhost:11434/v1",                   # (Change this) Base URL for local API (P.S: Ollama supports the OpenAI API format.)
        api_key="",                                             # API key (empty if not required)
        init_translate_model="gpt-oss:latest",                  # Initial translation model
        reflection_model="gpt-oss:latest",                      # Reflection model for improvement
        improvement_model="gpt-oss:latest",                     # Model for translation improvement
        hf_tokenizer="openai/gpt-oss-20b" ,                     # (Change this) HuggingFace model for tokenization
        hf_token=None,                                          # (Change this) HuggingFace authentication token
        temperature=1.0,                                        # Sampling temperature for generation
        top_p=1.0,                                              # Nucleus sampling parameter
        max_tokens=24576,                                       # Maximum tokens for input
        stop=["<|return|>","<|endoftext|>", "<|call|>"],        # (Change this) Stop TOKEN sequences
        max_token_per_chunk=5000,                               # Max tokens per chunk for translation
        source_lang="English",                                  # Source language
        target_lang="Traditional Chinese",                      # Target language
        country="Taiwan",                                       # (Optional) Country context for translation
    )
    translations = generator.generate(text)
    print(generator.parse_response(translations))

    # YAML-based usage (parameters only, text provided in code)
    generator_yaml = TranslationDataGenerator.from_yaml("config/translation_config.yaml")
    print(generator_yaml.generate_from_yaml("config/translation_config.yaml", text))

    # Pipeline DataFrame usage
    import pandas as pd
    df = pd.DataFrame(
        {
            "text": [
                "Once upon a time, there were three little pig brothers...",
                "The quick brown fox jumps over the lazy dog.",
            ]
        }
    )
    df_translated = generator_yaml.generate_from_dataframe(df, text_column="text", batch_size=16)
    print(df_translated.head())

    # Async test for async_generate_from_dataframe
    import asyncio
    async def test_async_generate_from_dataframe():
        df_translated = await generator_yaml.async_generate_from_dataframe(df, text_column="text", batch_size=16)
        print("[Async]\n", df_translated.head())

    asyncio.run(test_async_generate_from_dataframe())
```

Requesting review from NLP maintainers:
@ericharper @ekmb @yzhang123 @VahidooX @vladgets @okuchaiev

## Checklist

### DIFF
--- a/nemo_curator/synthetic/translate.py
+++ b/nemo_curator/synthetic/translate.py
@@ -41,7 +41,7 @@ class TextSplitter:
     """
     Utility class for splitting text into chunks based on token count.
     """
-    def __init__(self, model_name: str = "", hf_token: str = "", max_token_per_chunk: int = 4096, tokenizer: AutoTokenizer = None, spacy_model: str = "en_core_web_lg"):
+    def __init__(self, model_name: str = "", hf_token: str = "", max_token_per_chunk: int = 4096, tokenizer: AutoTokenizer = None, spacy_model: str = "xx_sent_ud_sm"):
         self.model_name = model_name
         self.hf_token = hf_token
         self.max_token_per_chunk = max_token_per_chunk
@@ -106,7 +106,7 @@ class TranslationDataGenerator(SyntheticDataGenerator):
         source_lang: str = "English",
         target_lang: str = "Traditional Chinese",
         country: str = "Taiwan",
-        spacy_model: str = "en_core_web_lg"
+        spacy_model: str = "xx_sent_ud_sm"
     ):
         super().__init__()
         # Initialize parameters for translation pipeline
@@ -326,11 +326,11 @@ if __name__ == "__main__":
     # Function-based usage
     text = "Once upon a time, there were three little pig brothers..."
     generator = TranslationDataGenerator(
-        base_url="http://localhost:11434/v1",                   # (Change this) Base URL for local API (P.S: Ollama supports the OpenAI API format.)
+        base_url="http://localhost:8080/v1",                   # (Change this) Base URL for local API (P.S: Ollama supports the OpenAI API format.)
         api_key="",                                             # API key (empty if not required)
-        init_translate_model="gpt-oss:latest",                  # Initial translation model
-        reflection_model="gpt-oss:latest",                      # Reflection model for improvement
-        improvement_model="gpt-oss:latest",                     # Model for translation improvement
+        init_translate_model="/home/user/.cache/llama.cpp/unsloth_gpt-oss-20b-GGUF_gpt-oss-20b-F16.gguf",                  # Initial translation model
+        reflection_model="/home/user/.cache/llama.cpp/unsloth_gpt-oss-20b-GGUF_gpt-oss-20b-F16.gguft",                      # Reflection model for improvement
+        improvement_model="/home/user/.cache/llama.cpp/unsloth_gpt-oss-20b-GGUF_gpt-oss-20b-F16.gguf",                     # Model for translation improvement
         hf_tokenizer="openai/gpt-oss-20b" ,                     # (Change this) HuggingFace model for tokenization
         hf_token=None,                                            # (Change this) HuggingFace authentication token
         temperature=1.0,                                        # Sampling temperature for generation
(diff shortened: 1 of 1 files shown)


shuoyangd:
Hi @TsukiSama9292 ! Thanks for the great contribution. We have a few requests before we can merge this one but me, @ayushdg and @abhinavg4 will support you along the way.

The main request we have as of now is that we are migrating away from Dask and switching to Ray as the new backend, so it would be nice if you can refactor your current workflow into something compatible with the rest of the repo. Good starting points would be the [API design doc](https://github.com/NVIDIA-NeMo/Curator/blob/ray-api/ray-curator/api-design.md) and our [quick start guide](https://github.com/NVIDIA-NeMo/Curator/blob/ray-api/ray-curator/ray_curator/examples/quickstart.py). Please take a look and let us know if you have questions!

After finishing the refactoring, let's do some benchmarks to ensure the default recipe generates good results.

In addition, I have two high-level questions upon glancing over your changes:

* Could you please switching to some existing sentence segmenter rather than [inventing your own](https://github.com/NVIDIA-NeMo/Curator/blob/a7bf3f1f7cc80656ed76e42197b665871909fc2f/nemo_curator/synthetic/translate.py#L55)? Those regex-based solutions tend to break when you scale to more languages. I'm fine with anything like [spacy](https://github.com/explosion/spaCy) or [ersatz](https://github.com/rewicks/ersatz) as long as it has good multilingual support.
* Correct me if I'm wrong, but currently your pipeline seems to assume that an LLM server has already been started (I only see client queries with OpenAI client but nothing that starts an LLM server, e.g. vllm or trtllm)? Did you give any thoughts about self-hosting cases?

TsukiSama9292:
> Hi @TsukiSama9292 ! Thanks for the great contribution. We have a few requests before we can merge this one but me, @ayushdg and @abhinavg4 will support you along the way.
> 
> The main request we have as of now is that we are migrating away from Dask and switching to Ray as the new backend, so it would be nice if you can refactor your current workflow into something compatible with the rest of the repo. Good starting points would be the [API design doc](https://github.com/NVIDIA-NeMo/Curator/blob/ray-api/ray-curator/api-design.md) and our [quick start guide](https://github.com/NVIDIA-NeMo/Curator/blob/ray-api/ray-curator/ray_curator/examples/quickstart.py). Please take a look and let us know if you have questions!
> 
> After finishing the refactoring, let's do some benchmarks to ensure the default recipe generates good results.
> 
> In addition, I have two high-level questions upon glancing over your changes:
> 
> * Could you please switching to some existing sentence segmenter rather than [inventing your own](https://github.com/NVIDIA-NeMo/Curator/blob/a7bf3f1f7cc80656ed76e42197b665871909fc2f/nemo_curator/synthetic/translate.py#L55)? Those regex-based solutions tend to break when you scale to more languages. I'm fine with anything like [spacy](https://github.com/explosion/spaCy) or [ersatz](https://github.com/rewicks/ersatz) as long as it has good multilingual support.
> * Correct me if I'm wrong, but currently your pipeline seems to assume that an LLM server has already been started (I only see client queries with OpenAI client but nothing that starts an LLM server, e.g. vllm or trtllm)? Did you give any thoughts about self-hosting cases?


I’m planning to migrate llm-translate from Dask to ray-api. I also updated `split_sentences` to use spaCy for sentence splitting, allowing users to choose different spaCy models.

nemo_curator/synthetic/translate.py
```diff
@@ -107,6 +106,7 @@ def __init__(
         source_lang: str = "English",
         target_lang: str = "Traditional Chinese",
         country: str = "Taiwan",
+        spacy_model: str = "en_core_web_lg"
```
shuoyangd: Mmm, this won't work for multilingual cases. Looks like there's a multilingual model for SpaCy, but it doesn't have the most permissive license.

Do you think you can switch models based on the input language? Alternatively, you could switch to ersatz -- their default multilingual model is pretty decent.

shuoyangd: Multilingual model for SpaCy: https://spacy.io/models/xx

TsukiSama9292: The default model has already been switched to the multilingual one.  
For better sentence segmentation in real use cases, it’s generally recommended to use a source-language-specific model instead.  
In most cases, `xx_sent_ud_sm` provides a good balance between performance and speed.

TsukiSama9292: Regarding the license, I briefly looked at [CC BY-SA 3.0](https://creativecommons.org/licenses/by-sa/3.0/) and [CC BY-SA 4.0](https://creativecommons.org/licenses/by-sa/4.0/). The license seems to apply to the model itself. In usage scenarios like sentence segmentation, since no additional content is generated, there likely isn’t a need to attribute the tool or be concerned about non-commercial restrictions.

PR #945, opened 2025-08-25
Generic shuffle stage

ayushdg:
## Description

Adds a generic shuffle stage that can be utilized by other shuffle steps. Follows a similar pattern to LSH but uses the BulkShufflingActor directly.

## Usage

```python
```
## Checklist


ray-curator/ray_curator/backends/experimental/ray_actor_pool/executor.py
```diff
@@ -86,17 +85,25 @@ def execute(self, stages: list["ProcessingStage"], initial_tasks: list[Task] | N
                     logger.info(
                         f" {stage} - Creating {num_actors} actors (CPUs: {stage.resources.cpus}, GPUs: {stage.resources.gpus})"
                     )
-
+                    # TODO: Clean up branching logic and handling here
```
ayushdg, line 88: This section is getting a bit messy. Will revisit it post release.

ray-curator/ray_curator/stages/shuffler/stage.py
```diff
@@ -0,0 +1,133 @@
...
+
+from ray_curator.backends.experimental.utils import RayStageSpecKeys
+from ray_curator.stages.base import ProcessingStage
+from ray_curator.stages.resources import Resources
+from ray_curator.stages.shuffler.rapidsmpf_shuffler import BulkRapidsMPFShuffler
+from ray_curator.tasks import FileGroupTask
+from ray_curator.utils.file_utils import create_or_overwrite_dir
+
+
+class ShuffleStage(ProcessingStage[FileGroupTask, FileGroupTask]):
```
ayushdg, line 25: Will update LSH to inherit from this in the future.

ray-curator/ray_curator/backends/experimental/ray_actor_pool/shuffle_adapter.py
```diff
@@ -58,8 +59,10 @@ def __init__(
             if num_input_tasks is None:
                 err_msg = "Shuffle total_nparts could not be set automatically. Please set it manually during stage initialization."
                 raise ValueError(err_msg)
-
-            self.output_nparts = max(1, 2 ** math.floor(math.log2(num_input_tasks)))
+            if self.stage.ray_stage_spec().get(RayStageSpecKeys.IS_LSH_STAGE, False):
+                self.output_nparts = max(1, 2 ** math.floor(math.log2(num_input_tasks)))
```
praateekmahajan, line 65: I remember there was some function that used to calculate this? Maybe call that? If no function exists can you document what it does iiuc it rounds up a number to closest power of 2 so 127 becomes 128, 129 becomes 256

ray-curator/ray_curator/stages/shuffler/stage.py
praateekmahajan: Once we make that change LSH stage should become much simpler right? What's the LOE to do that?

praateekmahajan: It'll also be nice if we can move some LSH logic outside of executor if we're able to generalize some generic concept of a shuffler

ray-curator/tests/stages/shuffler/test_shuffle_stage.py
```diff
@@ -0,0 +1,182 @@
...
+    def test_data_df(self, test_data: list[FileGroupTask]) -> cudf.DataFrame:
+        """Create a dataframe from the test data."""
+        return cudf.concat([cudf.read_parquet(task.data) for task in test_data], ignore_index=True)
+
+    @pytest.mark.parametrize(
+        ("shuffle_on", "total_nparts"),
+        [
+            (["category"], None),
+            (["category"], 1),
+            (["category", "value"], 3),
```
praateekmahajan, line 93: What if `total_nparts >> len(input_tasks)`

ray-curator/ray_curator/stages/shuffler/stage.py
```diff
@@ -0,0 +1,133 @@
+# Copyright (c) 2025, NVIDIA CORPORATION.
```
sarahyurick: Nit about adding the copyright header everywhere.

ray-curator/ray_curator/stages/shuffler/stage.py
ayushdg: Yeah I plan to address both in a followup. The LOE for LSHStage to inherit from this stage isn't too high but it won't reduce the lines of code for LSH by much. I'll open a followup task to see if the executor logic can be cleaned up a bit.

ray-curator/ray_curator/backends/experimental/ray_actor_pool/shuffle_adapter.py
ayushdg: No function it was from the old code here: https://github.com/NVIDIA-NeMo/Curator/blob/ray-api/nemo_curator/modules/fuzzy_dedup/lsh.py#L182.

I can add a comment for what it does.

ray-curator/tests/stages/shuffler/test_shuffle_stage.py
ayushdg: Added a test case for that case as well.

Issue #762, opened 2025-07-01
[Ray] Enable YAML filtering pipelines


sarahyurick:
As a follow-up to https://github.com/NVIDIA-NeMo/Curator/pull/746, we need to re-enable YAML reading with `build_filter_pipeline`... I think this should be pretty straightforward.

This will require enabling general YAML logic in Ray as well.

sarahyurick:
Also, in this PR we should delete:

- https://github.com/NVIDIA-NeMo/Curator/blob/ray-api/examples/identify_languages.py
- https://github.com/NVIDIA-NeMo/Curator/blob/ray-api/nemo_curator/scripts/filter_documents.py

And any other relevant files that use filtering modules.

sarahyurick:
Closing in favor of https://github.com/NVIDIA-NeMo/Curator/issues/1144, since its description is more accurate.

PR #957, opened 2025-08-26
[Ray] YAML pipelines

sarahyurick:
Closes https://github.com/NVIDIA-NeMo/Curator/issues/762.


config/run.py
```diff
@@ -0,0 +1,70 @@
+# Copyright (c) 2025, NVIDIA CORPORATION.  All rights reserved.
```
abhinavg4, line 1: Maybe move this to nemo_curator/core or nemo_curator/scripts?

ray-curator/ray_curator/config.py
```diff
@@ -1,3 +1,17 @@
+# Copyright (c) 2025, NVIDIA CORPORATION.  All rights reserved.
```
abhinavg4, line 1: Wait should this file exist?

abhinavg4 (changes requested):
Left a couple of small comments

ray-curator/ray_curator/config.py
sarahyurick: Not sure if it's being used anywhere tbh...

config/run.py
sarahyurick: Agree it shouldn't be under `config`, but what about under `tutorials`? It is a nice simple example that should be able to serve as a guide for users to expand themselves.

(Plus, it won't be strictly required by code freeze.)

sarahyurick:
Closing and tracking as: https://github.com/NVIDIA-NeMo/Curator/issues/1144.

PR #958, opened 2025-08-26
Refactor Curator to use Ray backend

sarahyurick:


.github/workflows/cicd-main.yml
```diff
@@ -17,7 +17,7 @@ on:
     - cron: 0 0 * * *
   push:
     branches:
-      - main
+      - ray-api
```
ayushdg: Should this be main now

sarahyurick: Hmm let me check.

sarahyurick: Updating here: https://github.com/NVIDIA-NeMo/Curator/pull/962.

PR #1000, opened 2025-08-29
Add TextSemanticDeduplicationWorkflow

praateekmahajan:
## Description
One monolithy stage for text that
1. **Embedding Generation** - Reads data, generates embeddings, writes to parquet (xenna or ray data with id generator)
2. **KMeans** Reads, KMeans, Writes (ray actor pool)
3. **Pairwise** Reads Pairwise, IdentifyDuplicates, Write (xenna or ray data)
4. **Removal** Reads, Removes, Writes  (xenna or ray data; uses id generator; ray data preferred)

## Usage

```python
from nemo_curator.stages.text.deduplication.semantic import TextSemanticDeduplicationWorkflow
from nemo_curator.backends.xenna import XennaExecutor

# Run semantic deduplication workflow
workflow = TextSemanticDeduplicationWorkflow(
    input_path="./data/documents",
    output_path="./output/",
    cache_path="./cache/",
    perform_removal=True,
    use_id_generator=True,
    eps=0.1,
    n_clusters=100,

)

results = workflow.run(XennaExecutor())
```

## Other fixes
1. Added `clear_output` in SemanticDeduplicationWorkflow (so that we don't always clear output directories)
7. Don't write `_curator_dedup_id` in pairwise column since that's anyway written as `id`
## Followups
1. Xenna <> Dedup actor can't be run one after the other (unclear what the order is when it breaks)

## Checklist


nemo_curator/stages/text/deduplication/semantic.py
```diff
@@ -0,0 +1,524 @@
...
+    output_filetype: Literal["jsonl", "parquet"] = "parquet"
+    output_file_extension: str | None = None
+    output_fields: list[str] | None = None
+    read_kwargs: dict[str, Any] = field(default_factory=dict)
+    cache_kwargs: dict[str, Any] = field(default_factory=dict)
+    write_kwargs: dict[str, Any] = field(default_factory=dict)
+    # Execution parameters
+    verbose: bool = True
+    clear_output: bool = True
+    """
```
sarahyurick, line 115: I noticed that these are parameters that are not in the docstring:
```
id_generator_state_file: str | None = None
output_file_extension: str | None = None
output_fields: list[str] | None = None
cache_kwargs: dict[str, Any] = field(default_factory=dict)
```
I noticed that this is in the docstring but not parameters:
```
executor: Single executor for all stages or tuple of (embedding, pairwise, removal) executors
```

praateekmahajan: Thanks for catching that!

sarahyurick (approved):

PR #1004, opened 2025-09-02
ci: Re-organize curator dep

thomasdhc:
## Usage

```python
```
## Checklist


pyproject.toml
```diff
@@ -75,8 +78,13 @@ text = [
...
 ]
 
+text = [
+    "nemo_curator[gpu]",
+    "nemo_curator[text_cpu]",
+]
+
 # Installs CPU + GPU text curation modules
-deduplication_cuda12x = [
+deduplication = [
```
ayushdg: I personally still prefer a `cuda12/cu12` type suffix. If we need to support cuda 13 going forward I find transitioning from `cu12->cu12,cu13` variants to be easier.

pyproject.toml
```diff
@@ -137,19 +128,44 @@ internvideo2 = [
...
+video_cpu = [
+    "nemo_curator[internvideo2]",
+    "av==13.1.0",
+    "opencv-python",
+    "torchvision",
+    "einops",
+    "easydict",
+]
+
+video_cuda = [
```
ayushdg: Should standardize and probably use `cuda` everywhere instead of `gpu` or vice versa

thomasdhc: @ayushdg gotcha let me start on thread on this. There's some other points I want to highlight

pyproject.toml
```diff
@@ -137,19 +128,44 @@ internvideo2 = [
...
+    "nemo_curator[gpu]",
+    "cvcuda_cu12",
+    "pycuda",
+    "vllm==0.9.2; (platform_machine == 'x86_64' and platform_system != 'Darwin')",
+    "flash-attn<=2.8.3; (platform_machine == 'x86_64' and platform_system != 'Darwin')",
+    "torch<=2.8.0",
+    "torchaudio",
+]
+
+video_nvenc = ["PyNvVideoCodec==2.0.2; (platform_machine == 'x86_64' and platform_system != 'Darwin')"]
```
suiyoubi: Should we have this included in the `video` as well or you are intentionally make this optional ?

pyproject.toml
sarahyurick: I think calling it `text_cuda12x` instead of `deduplication` could be cleaner? If we are trying to keep it generic enough to be included in multiple dependencies, then maybe `_deduplication_cuda12x` followed by:
```
text_cuda12x = [nemo_curator[_deduplication_cuda12x], ...]
image_cuda12x = [nemo_curator[_deduplication_cuda12x], ...]
```

pyproject.toml
thomasdhc: There's an outstanding nvenc import for this right? We'll have to validate, and then we can add back to video

pyproject.toml
```diff
@@ -90,25 +73,42 @@ audio = [
     "nemo_toolkit[asr]",
```
sarahyurick, line 73: Nit: I am not sure if audio dependencies should follow a similar `audio_cpu`/`audio_cuda12` convention?

praateekmahajan: +1.
What happens if someone wants to run Audio on GPU? Do we just need two entires audio_cpu / audio_gpu both pointing to nemo_toolkit?

pyproject.toml
```diff
@@ -137,19 +137,34 @@ internvideo2 = [
     "wandb>=0.18.3",
 ]
 
-# NVIDIA DALI (simplified; update the package to match your CUDA version if needed)
-dali = [
-    "nvidia-dali-cuda120",
+video_nvenc = ["PyNvVideoCodec==2.0.2; (platform_machine == 'x86_64' and platform_system != 'Darwin')"]
```
praateekmahajan: When is this installed? Is there a reason this isn't installed with `video_cuda12`?
cc @suiyoubi

pyproject.toml
thomasdhc: nemo_toolkit[asr] was the only thing defined for audio. Did we want cpu and cuda to point to the same dependencies?

pyproject.toml
ayushdg: https://github.com/NVIDIA-NeMo/Curator/pull/1004#discussion_r2316376680

thomasdhc: PyNvVideoCodec is not supported on H100, I think there's an import guard so we may be able to add it back in

pyproject.toml
```diff
@@ -137,19 +137,34 @@ internvideo2 = [
...
+
+video_cpu = [
+    "nemo_curator[internvideo2]",
```
ayushdg, line 144: Can be handled in a followup but should internvideo2 live inside the gpu dependencies? 
Is there any video pipeline that can be run CPU only?

thomasdhc: @suiyoubi ?

pyproject.toml
praateekmahajan: I'd say so, assuming nemo_toolkit[asr] is smart enough that when cpu only it just installs cpu packages and for gpu it installs gpu packages.
cc @huvunvidia

huvunvidia: @karpnv 
Can you have a look? Do we need separate package for nemo_toolkit[asr] for CPU and GPU, or it should figure out itself the right package to install?

pyproject.toml
```diff
@@ -75,40 +106,13 @@ text = [
...
-    "flash-attn<=2.8.3; (platform_machine == 'x86_64' and platform_system != 'Darwin')",
-    "torch<=2.8.0",
-    "torchaudio",
+text_cuda12 = [
+    "nemo_curator[cuda12]",
+    "nemo_curator[deduplication_cuda12]",
+    "nemo_curator[text_cpu]",
 ]
 
-video_nvenc = ["PyNvVideoCodec==2.0.2; (platform_machine == 'x86_64' and platform_system != 'Darwin')"]
```
sarahyurick, line 110: It looks like this is still being referenced in the Dockerfile.

thomasdhc: @sarahyurick Addressed! If you could approve when you get a chance

PR #1010, opened 2025-09-02
Merge `examples/` directory into `tutorials/`

sarahyurick:
Follow-up work: fix links in https://huggingface.co/collections/nvidia/nemo-curator-classifier-models-66b25154213dafdcb8bde900.


praateekmahajan:
As part of this can you suffix the tutorials / examples similarly, and that'll help set precedent for future too

E.g.
`audio/fleurs-tutorials` vs `image/image-curation` vs `text/tinystories` vs `video/video-examples`

Either all end with examples|tutorials or none do, is my thought. I'd vote for none since the distinction between a tutorial and an example isn't clear yet.

sarahyurick:
> As part of this can you suffix the tutorials / examples similarly, and that'll help set precedent for future too
> 
> E.g. `audio/fleurs-tutorials` vs `image/image-curation` vs `text/tinystories` vs `video/video-examples`
> 
> Either all end with examples|tutorials or none do, is my thought. I'd vote for none since the distinction between a tutorial and an example isn't clear yet.

Yes, I can rename for better consistency! I think this PR will not be perfect regardless, since most modalities only have a single "tutorial" for now. I am expecting the directory structure to become more refined/iterated upon once we have a better range of tutorials.

praateekmahajan:
Yup agreed re perfection, I think just lets start somewhere and then adapt (and improve if needed) subsequently

PR #1011, opened 2025-09-02
Add benchmarking abilitiy

praateekmahajan:
## Description
Nightly benchmarking runs a configurable matrix of benchmark scripts with per-entry Ray cluster isolation, standardized result outputs, environment capture, and optional logging to MLflow/W&B. It’s designed to provide reproducible, comparable performance measurements across datasets and executors, with clean per-run artifacts and robust failure isolation.

## Usage
```python
python -m nightly_benchmarking.run \
  --matrix nightly_benchmarking/matrix.yaml \
  --datasets nightly_benchmarking/dataset_paths.json
```
## Future TODOs (not ranked by priority)
1. Upload results / artifacts to Google Drive
2. Post to Slack ability
3. Be able to create charts from logged results over time
1. See if we can run each of the entry in the matrix on a separate cluster (ie. increase parallelism)
1. Figure out how to run in CI? If running in CI, then how to get environment variables


praateekmahajan:
@rlratzel finished the effort in #1197

PR #1021, opened 2025-09-03
Small updates for PEFT and Tinystories tutorials

sarahyurick:


tutorials/text/peft-curation/main.py
```diff
@@ -27,13 +27,17 @@
...
 
+from nemo_curator.core.client import RayClient
 from nemo_curator.pipeline import Pipeline
 from nemo_curator.stages.text.io.writer import JsonlWriter
 from nemo_curator.stages.text.modifiers import UnicodeReformatter
 from nemo_curator.stages.text.modules import Modify, ScoreFilter
 
 
 def main(args: argparse.Namespace) -> None:
+    ray_client = RayClient()
```
Maghoumi, line 38: Out of curiosity, what happens if this isn't done?

tutorials/text/peft-curation/main.py
```diff
@@ -27,13 +27,17 @@
...
+    ray_client.start()
```
Maghoumi, line 39: Sounds like an excellent opportunity for wrapping into a context manager.

tutorials/text/peft-curation/main.py
sarahyurick: It is similar to how we had to start a Dask cluster in older versions. I think there may be additional logic already baked into Curator/Cosmos Xenna to help with this too, but being more explicit about it here helps avoid flakiness.

ayushdg (approved):

PR #1059, opened 2025-09-09
Allow RayActorPool (RAFT) to run after Xenna

praateekmahajan:
## Description
Xenna implicitly sets `os.environ["RAY_EXPERIMENTAL_NOSET_CUDA_VISIBLE_DEVICES"]` on client. This means RayActorPool when starting after Xenna won't be able to allocate `CUDA_VISIBLE_DEVICE` on its own. This could result in RAFT Actor Pool failing 

## Root cause

```python
client.start() # noset hasn't been defined
run_xenna() # noset HAS been defined on client but not on workers
run_kmeans()  # same as above, but since its not on workers this works fine
client.stop() # all workers die; noset is still present on client

client.start() # noset is now propogated to workers too
run_kmeans() # fails
```
## Checklist

### DIFF
--- a/tests/backends/experimental/ray_actor_pool/executor.py
+++ b/tests/backends/experimental/ray_actor_pool/executor.py
@@ -1,0 +1,19 @@
+from unittest import mock
+
+from nemo_curator.backends.experimental.ray_actor_pool.executor import _parse_runtime_env
+
+
+class TestRayActorPoolExecutor:
+    def test_parse_runtime_env(self):
+        # With noset defined we should override it to be empty
+        with_noset_defined = {"env_vars": {"RAY_EXPERIMENTAL_NOSET_CUDA_VISIBLE_DEVICES": mock.ANY}}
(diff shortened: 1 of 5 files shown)


tests/stages/deduplication/semantic/test_kmeans.py
```diff
@@ -107,46 +107,6 @@ def create_clustered_dataset(  # noqa: PLR0913
     return input_dir, y_true
 
 
-def create_kmeans_pipeline(
```
praateekmahajan, line 110: unused method, deleted it

tests/backends/experimental/ray_actor_pool/executor.py
```diff
@@ -0,0 +1,19 @@
+from unittest import mock
```
sarahyurick: Copyright header?

praateekmahajan: also realized filename should be `test_executor.py`

PR #1063, opened 2025-09-10
Llama Nemotron Data Curation tutorial

sarahyurick:
Re-adds https://github.com/NVIDIA-NeMo/Curator/tree/dask/tutorials/llama-nemotron-data-curation with Ray backend instead of Dask.

Related issue: https://github.com/NVIDIA-NeMo/Curator/issues/706.

TODO:

- [x] Verify reasoning on versus reasoning off composition of the input data
```
syurick@ipp1-3302:~$ jq -r '.reasoning' /raid/syurick/Llama-Nemotron-Post-Training-Dataset/SFT/chat/chat.jsonl | awk '{counts[$0]++} END {for (k in counts) print counts[k], k}' | sort -nr
31218 off
8574 on
syurick@ipp1-3302:~$ jq -r '.reasoning' /raid/syurick/Llama-Nemotron-Post-Training-Dataset/SFT/math/math_v1.1.jsonl | awk '{counts[$0]++} END {for (k in counts) print counts[k], k}' | sort -nr
2225427 on
```
- [x] Verify correctness for heuristic filters
- [x] Verify correctness for model filters
- [x] Verify end-to-end run
- [ ] If time allows: generate filtering statistics

### DIFF
--- a/docs/_extensions/rich_metadata/__init__.py
+++ b/docs/_extensions/rich_metadata/__init__.py
@@ -1,0 +1,477 @@
+"""
+Rich Metadata Extension for Sphinx
+
+Injects SEO-optimized metadata into HTML <head> based on frontmatter.
+Supports Open Graph, Twitter Cards, JSON-LD structured data, and standard meta tags.
+
+Frontmatter fields supported:
+- description: Page description for meta tags
+- tags: Keywords for SEO
+- personas: Target audience information
+- difficulty: Content difficulty level
+- content_type: Type of content (tutorial, concept, reference, etc.)
+- modality: Content modality (text-only, image-only, video-only, multimodal, universal)
+- cascade.product.name: Product name
+- cascade.product.version: Product version
+"""
+
+import json
+import os
+from typing import Any
+
+from docutils import nodes
+from sphinx.application import Sphinx
+from sphinx.config import Config
+from sphinx.util import logging
+
+# Import YAML for frontmatter parsing
+try:
+    import yaml
+    YAML_AVAILABLE = True
+except ImportError:
+    YAML_AVAILABLE = False
+    yaml = None
+
+logger = logging.getLogger(__name__)
+
+
+def extract_frontmatter(env, docname: str) -> dict[str, Any]:  # noqa: ANN001
+    """
+    Extract frontmatter from markdown source file.
+
+    Uses the same approach as the json_output extension (proven to work).
+
+    Args:
+        env: Sphinx build environment
+        docname: Document name
+
+    Returns:
+        Dictionary of frontmatter fields
+    """
+    metadata = {}
+
+    if not YAML_AVAILABLE:
+        logger.warning("PyYAML not available, frontmatter cannot be parsed")
+        return metadata
+
+    try:
+        # Get source path - this returns the full absolute path
+        source_path = env.doc2path(docname)
+
+        # Only process markdown files
+        if source_path and str(source_path).endswith(".md"):
+            with open(source_path, encoding="utf-8") as f:
+                content = f.read()
+
+            # Check for valid frontmatter format (YAML between --- markers)
+            if content.startswith("---"):
+                end_marker = content.find("\n---\n", 3)
+                if end_marker != -1:
+                    frontmatter_text = content[3:end_marker]
+                    metadata = yaml.safe_load(frontmatter_text) or {}
+                    logger.debug(f"rich_metadata: Extracted {len(metadata)} fields from {docname}")
+
+    except yaml.YAMLError as e:
+        logger.warning(f"rich_metadata: YAML parsing error in {docname}: {e}")
+    except OSError as e:
+        logger.warning(f"rich_metadata: File error for {docname}: {e}")
+
+    return metadata
+
+
+def build_page_title(_metadata: dict[str, Any], context: dict[str, Any], _pagename: str) -> str | None:
+    """
+    Build enhanced page title in format: Page Title: Section - Site | NVIDIA
+
+    Args:
+        metadata: Page frontmatter metadata
+        context: Sphinx page context
+        pagename: Document name
+
+    Returns:
+        Enhanced title string or None to keep default
+    """
+    # Get base title
+    page_title = context.get("title", "")
+    if not page_title:
+        return None
+
+    # Build title components
+    components = [page_title]
+
+    # Add parent section if available (from breadcrumbs or parents)
+    if context.get("parents"):
+        # Get the immediate parent (last item in parents list)
+        parent = context["parents"][-1]
+        if "title" in parent and parent["title"] != page_title:
+            components.append(parent["title"])
+
+    # Add site name (from Sphinx's project config)
+    site_name = context.get("docstitle", "")
+    if site_name and site_name != page_title:
+        components.append(site_name)
+
+    # Build final title with format depending on available components
+    if len(components) == 1:
+        # Just page: "Page Title | NVIDIA"
+        return f"{components[0]} | NVIDIA"
+    elif len(components) == 2:  # noqa: PLR2004
+        # Page and site: "Page Title - Site | NVIDIA"
+        return f"{components[0]} - {components[1]} | NVIDIA"
+    else:
+        # Page, section, and site: "Page Title: Section - Site | NVIDIA"
+        return f"{components[0]}: {components[1]} - {components[2]} | NVIDIA"
+
+
+def _add_basic_fields(metadata: dict[str, Any]) -> list[str]:
+    """Add basic SEO meta tags."""
+    basic_tags = []
+    if "description" in metadata:
+        description = metadata["description"]
+        basic_tags.append(f'<meta name="description" content="{description}">')
+
+    if "tags" in metadata:
+        keywords = metadata["tags"]
+        if isinstance(keywords, list):
+            keywords_str = ", ".join(keywords)
+            basic_tags.append(f'<meta name="keywords" content="{keywords_str}">')
+    return basic_tags
+
+
+def _add_opengraph_fields(metadata: dict[str, Any], context: dict[str, Any]) -> list[str]:
+    """Add Open Graph meta tags."""
+    og_tags = []
+    if "description" in metadata:
+        og_tags.append(f'<meta property="og:description" content="{metadata["description"]}">')
+
+    og_tags.append('<meta property="og:type" content="article">')
+
+    enhanced_title = context.get("pagetitle") or context.get("title", "")
+    if enhanced_title:
+        og_tags.append(f'<meta property="og:title" content="{enhanced_title}">')
+
+    if "pageurl" in context:
+        url = context["pageurl"]
+        og_tags.append(f'<meta property="og:url" content="{url}">')
+
+    return og_tags
+
+
+def _add_twitter_fields(metadata: dict[str, Any], context: dict[str, Any]) -> list[str]:
+    """Add Twitter Card meta tags."""
+    twitter_tags = []
+    if "description" in metadata:
+        twitter_tags.append(f'<meta name="twitter:description" content="{metadata["description"]}">')
+
+    enhanced_title = context.get("pagetitle") or context.get("title", "")
+    if enhanced_title:
+        twitter_tags.append(f'<meta name="twitter:title" content="{enhanced_title}">')
+
+    twitter_tags.append('<meta name="twitter:card" content="summary">')
+    return twitter_tags
+
+
+def _add_personas_tag(metadata: dict[str, Any]) -> list[str]:
+    """Add personas/audience metadata tag."""
+    if "personas" not in metadata:
+        return []
+
+    personas = metadata["personas"]
+    if not isinstance(personas, list):
+        return []
+
+    audience_map = {
+        "data-scientist-focused": "Data Scientists",
+        "mle-focused": "Machine Learning Engineers",
+        "admin-focused": "Cluster Administrators",
+        "devops-focused": "DevOps Professionals",
+    }
+    audiences = [audience_map.get(p, p) for p in personas]
+    audience_str = ", ".join(audiences)
+    return [f'<meta name="audience" content="{audience_str}">']
+
+
+def _add_product_tags(metadata: dict[str, Any]) -> list[str]:
+    """Add product metadata tags from cascade."""
+    if "cascade" not in metadata:
+        return []
+
+    cascade = metadata["cascade"]
+    if not isinstance(cascade, dict) or "product" not in cascade:
+        return []
+
+    product = cascade["product"]
+    if not isinstance(product, dict):
+        return []
+
+    tags = []
+    if product.get("name"):
+        tags.append(f'<meta name="product-name" content="{product["name"]}">')
+    if product.get("version"):
+        tags.append(f'<meta name="product-version" content="{product["version"]}">')
+    return tags
+
+
+def _add_custom_fields(metadata: dict[str, Any]) -> list[str]:
+    """Add custom NVIDIA/content metadata tags."""
+    custom_tags = []
+
+    # Add personas/audience tags
+    custom_tags.extend(_add_personas_tag(metadata))
+
+    # Add simple content metadata fields
+    if "content_type" in metadata:
+        custom_tags.append(f'<meta name="content-type-category" content="{metadata["content_type"]}">')
+
+    if "difficulty" in metadata:
+        custom_tags.append(f'<meta name="difficulty" content="{metadata["difficulty"]}">')
+
+    if "modality" in metadata:
+        custom_tags.append(f'<meta name="modality" content="{metadata["modality"]}">')
+
+    # Add product information from cascade
+    custom_tags.extend(_add_product_tags(metadata))
+
+    return custom_tags
+
+
+def build_meta_tags(metadata: dict[str, Any], context: dict[str, Any]) -> dict[str, list[str]]:
+    """
+    Build HTML meta tags from frontmatter metadata, organized by category.
+
+    Args:
+        metadata: Frontmatter metadata dictionary
+        context: Sphinx HTML context
+
+    Returns:
+        Dictionary with categorized meta tag lists
+    """
+    return {
+        "basic": _add_basic_fields(metadata),
+        "opengraph": _add_opengraph_fields(metadata, context),
+        "twitter": _add_twitter_fields(metadata, context),
+        "custom": _add_custom_fields(metadata)
+    }
+
+
+def build_json_ld(metadata: dict[str, Any], context: dict[str, Any]) -> str | None:  # noqa: C901, PLR0912
+    """
+    Build JSON-LD structured data for SEO.
+
+    Args:
+        metadata: Frontmatter metadata dictionary
+        context: Sphinx HTML context
+
+    Returns:
+        JSON-LD script tag string or None
+    """
+    # Base structure
+    structured_data = {
+        "@context": "https://schema.org",
+        "@type": "TechArticle",
+    }
+
+    # Add title
+    if "title" in context:
+        structured_data["headline"] = context["title"]
+        structured_data["name"] = context["title"]
+
+    # Add description
+    if "description" in metadata:
+        structured_data["description"] = metadata["description"]
+
+    # Add keywords
+    if "tags" in metadata and isinstance(metadata["tags"], list):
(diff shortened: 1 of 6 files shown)


tutorials/text/llama-nemotron-data-curation/README.md
```diff
@@ -0,0 +1,124 @@
...
+
+The above script applies basic filtering to the input dataset:
+
+- Only take samples used for Nemotron Nano training
+- Remove empty and malformed samples
+- Remove non-English samples
+- Remove samples with total length (system prompt, input, and output responses) longer than 16k tokens (with chat template applied via the tokenizer)
+- Remove samples with output responses longer than 8k tokens (with chat template applied via the tokenizer)
+- Remove any columns specified by the `--remove-columns` parameter. We recommend removing the columns specified above (so that the remaining columns are the "input", "output", and "completion_token_count" columns)
```
sarahyurick: Is this reasonable? Does `training.jsonl` need to keep any additional columns?

tutorials/text/llama-nemotron-data-curation/filters/model_filters.py
```diff
@@ -0,0 +1,371 @@
...
+        outpt_copy = outpt.copy()
+
+        templates_list = outpt_copy.apply(
+            lambda text: self.tokenizer.apply_chat_template(
+                [{"role": "assistant", "content": text}],
+                # For maximum accuracy it should be tokenize=True here,
+                # but for speedups we just use the length of the base string instead of tokenizing it
+                # (we consider this to be acceptable since 1 token is approx 4 characters for English
+                # and we are only using the CompletionTokenCountFilter as a proxy for text complexity)
+                tokenize=False,
```
sarahyurick, line 247: IIRC we aligned on this when implementing the Dask tutorial. Just wanted to double check that this is still okay.

tutorials/text/llama-nemotron-data-curation/filters/model_filters.py
```diff
@@ -0,0 +1,371 @@
...
+            self._setup(local_files_only=False)
+        except Exception as e:
+            msg = f"Failed to download {self.tokenizer_identifier}"
+            raise RuntimeError(msg) from e
+
+    # We use the _setup function to ensure that everything needed for the tokenizer is downloaded and loaded properly
+    def _setup(self, local_files_only: bool = True) -> None:
+        self.tokenizer = AutoTokenizer.from_pretrained(
+            self.tokenizer_identifier,
+            use_fast=True,  # TODO: Test without this
```
sarahyurick: This is from the Dask implementation. I want to verify that it still helps.

tutorials/text/llama-nemotron-data-curation/utils/jsonl_utils.py
```diff
@@ -0,0 +1,71 @@
...
+
+import itertools
+import os
+from collections.abc import Generator
+
+import ray
+
+
+@ray.remote
+def split_jsonl_by_size(input_file: str, target_size_mb: int, output_dir: str, output_prefix: str) -> None:
```
sarahyurick, line 20: We need this function to essentially enable partitions of single JSONL files. It is basically mimicking `reshard_jsonl` which was deprecated in the Dask to Ray migration: https://github.com/NVIDIA-NeMo/Curator/blob/dask/nemo_curator/scripts/make_data_shards.py

I think it is okay for just the tutorial for now, but we should add it back to the codebase eventually.

tutorials/text/llama-nemotron-data-curation/utils/jsonl_utils.py
```diff
@@ -0,0 +1,71 @@
...
+    gen2 = stream_jsonl_files(dir2)
+
+    with open(out_path, "w") as out:
+        for line1, line2 in itertools.zip_longest(gen1, gen2):
+            if line1 is not None:
+                out.write(line1 + "\n")
+            if line2 is not None:
+                out.write(line2 + "\n")
+
+    print(f"Interleaved datasets from directories {dir1} and {dir2} into file {out_path}")
```
sarahyurick: This function interleaves 2 datasets, row by row. In the Dask version, we had the option to interleave per row or per partition of data (with interleaving per partition of data as the default). Should we maintain that logic in the Ray version too?

sarahyurick (commented):
@Maghoumi left some open comments, but nothing major IMO. Please let me know what you think.

tutorials/text/llama-nemotron-data-curation/filters/model_filters.py
sarahyurick: No time difference here, so I will remove `use_fast=True`.

tutorials/text/llama-nemotron-data-curation/main.py
```diff
@@ -0,0 +1,288 @@
...
+    parser.add_argument(
+        "--tokenizer",
+        type=str,
+        default="meta-llama/Llama-3.1-8B-Instruct",
+        help="Hugging Face tokenizer",
+    )
+    parser.add_argument(
+        "--hf-token",
+        type=str,
+        help="Hugging Face token",
```
sarahyurick:
```suggestion
        help="Hugging Face token (if needed)",
```

tutorials/text/llama-nemotron-data-curation/README.md
```diff
@@ -0,0 +1,124 @@
...
+### Download input dataset
+
+The input dataset can be downloaded from Hugging Face here: https://huggingface.co/datasets/nvidia/Llama-Nemotron-Post-Training-Dataset
+
+The following commands can be used to download the dataset:
+
+```bash
+git lfs install
+git clone https://huggingface.co/datasets/nvidia/Llama-Nemotron-Post-Training-Dataset
+```
```
sarahyurick, line 59: This may need `apt-get update && apt-get install -y git-lfs` too.

tutorials/text/llama-nemotron-data-curation/main.py
```diff
@@ -0,0 +1,288 @@
...
+        input_files = [filename for filename in input_files if any(s in filename for s in args.filename_filter)]
+
+    # Grab integer from blocksize string, e.g., "100mb" -> 100
+    json_blocksize = int("".join(c for c in args.json_blocksize if c.isdigit()))
+    input_dir = os.path.join(args.input_dir, "split")
+    try:
+        os.makedirs(input_dir, exist_ok=False)
+    except FileExistsError as e:
+        msg = f"Split directory already exists: {input_dir}. Please delete it and try again."
+        raise FileExistsError(msg) from e
```
sarahyurick:
```suggestion
    input_dir = os.path.join(args.output_dir, "input_data_shards")
    os.makedirs(input_dir, exist_ok=False)
```

tutorials/text/llama-nemotron-data-curation/filters/heuristic_filters.py
```diff
@@ -0,0 +1,103 @@
...
+# Filter by thinking ON or OFF
+class ThinkingOnFilter(DocumentFilter):
+    def __init__(self):
+        super().__init__()
+
+    def score_document(self, text: str) -> bool:
+        return "on" in text.lower()
+
+    def keep_document(self, score: bool) -> bool:
+        return score
```
praateekmahajan, line 30: unrelated to the PR, but could DocumentFilter's base method of keep-document / score_document always just return score?

tutorials/text/llama-nemotron-data-curation/filters/model_filters.py
```diff
@@ -0,0 +1,370 @@
...
+        self.system_prompt_field = system_prompt_field
+
+    def inputs(self) -> tuple[list[str], list[str]]:
+        return ["data"], [self.input_field, self.output_field, self.system_prompt_field]
+
+    def outputs(self) -> tuple[list[str], list[str]]:
+        return ["data"], []
+
+    def ray_stage_spec(self) -> dict[str, Any]:
+        return {"is_actor_stage": True}
```
praateekmahajan: not needed but okay to have it explicitly.. if you see in ray_data/executors.py the way it infers if something is actor stage is by seeing if setup_on_node / setup is overridden which I think it'll be?

tutorials/text/llama-nemotron-data-curation/filters/heuristic_filters.py
sarahyurick: Yes I agree it may make sense to add logic to always return `score` if it is a boolean. But if score is an integer, etc. then they are not the same, e.g., https://github.com/NVIDIA-NeMo/Curator/blob/main/nemo_curator/stages/text/filters/heuristic_filter.py#L47

It just so happens for this tutorial that most of the filters use a boolean scoring metric.

tutorials/text/llama-nemotron-data-curation/filters/model_filters.py
```diff
@@ -0,0 +1,370 @@
...
+                # and we are only using the CompletionTokenCountFilter as a proxy for text complexity)
+                tokenize=False,
+                add_generation_prompt=False,
+                truncation=False,
+            )
+        ).tolist()
+        tokenized = self.tokenizer(templates_list)
+        scores = pd.Series([len(tokens) for tokens in tokenized["input_ids"]], index=outpt_copy.index)
+        df["completion_token_count"] = scores
+        mask = (scores > 0) & (scores <= self.max_completion_token_count)
```
praateekmahajan, line 255: Shouldn't this be max_completion_token_count / 4 then based on description above?

tutorials/text/llama-nemotron-data-curation/utils/jsonl_utils.py
praateekmahajan: WDYT of moving this atleast to where the other MR is being merged for `split_parquet_by_size`

tutorials/text/llama-nemotron-data-curation/utils/jsonl_utils.py
```diff
@@ -0,0 +1,71 @@
...
+                bytes_written = 0
+            out.write(line)
+            bytes_written += len(line.encode("utf-8"))
+
+    if out:
+        out.close()
+
+
+def stream_jsonl_files(jsonl_dir: str) -> Generator[str, None, None]:
+    # Confusingly, Ray's write_json function names the files with the .json extension,
```
praateekmahajan, line 48: Where is ray the writer here s.t. extensions are .json?

tutorials/text/llama-nemotron-data-curation/utils/jsonl_utils.py
```diff
@@ -0,0 +1,71 @@
...
+        with open(os.path.join(jsonl_dir, fname)) as f:
+            for line in f:
+                yield line.rstrip("\n")
+
+
+def interleave_datasets(dir1: str, dir2: str, out_path: str) -> None:
+    gen1 = stream_jsonl_files(dir1)
+    gen2 = stream_jsonl_files(dir2)
+
+    with open(out_path, "w") as out:
```
praateekmahajan, line 69: So we create one large file from all the files in two separate directories?

tutorials/text/llama-nemotron-data-curation/main.py
```diff
@@ -0,0 +1,284 @@
...
+            output_field="output",
+            system_prompt_field="system_prompt",
+        ),
+    ]
+    for tokenizer_step in tokenizer_steps:
+        pipeline_thinking_on.add_stage(tokenizer_step)
+        pipeline_thinking_off.add_stage(tokenizer_step)
+
+    # No specific columns are accessed after this point, so we can drop any that the user specifies
+    if args.remove_columns:
```
praateekmahajan: I believe the writers also have an attribute called field to ensure we only write selective columns, so maybe we won't need this

tutorials/text/llama-nemotron-data-curation/filters/model_filters.py
sarahyurick: Yeah that is based on internal conversations about where tokenization should and should not happen, etc. The value used is correct, whereas "token" is used too loosely in the name `max_completion_token_count`. Can try to remove the discrepancy from the comment.

tutorials/text/llama-nemotron-data-curation/utils/jsonl_utils.py
sarahyurick:
```
# Sort datasets
thinking_on_ds = ray.data.read_json(thinking_on_unsorted_path, lines=True)
thinking_on_ds = thinking_on_ds.sort("completion_token_count")
thinking_on_sorted_path = os.path.join(args.output_dir, "thinking_on_sorted")
thinking_on_ds.write_json(thinking_on_sorted_path, orient="records", lines=True)

thinking_off_ds = ray.data.read_json(thinking_off_unsorted_path, lines=True)
thinking_off_ds = thinking_off_ds.sort("completion_token_count")
thinking_off_sorted_path = os.path.join(args.output_dir, "thinking_off_sorted")
thinking_off_ds.write_json(thinking_off_sorted_path, orient="records", lines=True)
```
^ this writes JSONL files as intended, but if you look at the output directories they are all called `*.json`.

tutorials/text/llama-nemotron-data-curation/utils/jsonl_utils.py
sarahyurick: Yes, in order to be used by NeMo it needs to be a single JSONL file.

tutorials/text/llama-nemotron-data-curation/main.py
```diff
@@ -0,0 +1,284 @@
...
+    thinking_off_unsorted_path = os.path.join(args.output_dir, "thinking_off_unsorted")
+    pipeline_thinking_on.add_stage(JsonlWriter(thinking_on_unsorted_path))
+    pipeline_thinking_off.add_stage(JsonlWriter(thinking_off_unsorted_path))
+
+    # Run pipelines
+    _thinking_on_output = pipeline_thinking_on.run()
+    _thinking_off_output = pipeline_thinking_off.run()
+
+    # Sort datasets
+    thinking_on_ds = ray.data.read_json(thinking_on_unsorted_path, lines=True)
```
praateekmahajan, line 175: What's the use here? Do we need a guarantee that the output files are produced in sorted order? 
Because a global sort + write in a distributed setting, might not guarantee that. Each file would be sorted but what but if it outputs N files, they must be read in a certain way to ensure read is also sorted?
For instance spark creates files with UUID, therefore when you read the N files back there isn't a guarantee that the whole dataset in memory is sorted. That's only possible if you enforce a `coalesce` (`repartition(1)`)

tutorials/text/llama-nemotron-data-curation/main.py
```diff
@@ -0,0 +1,284 @@
...
+    thinking_off_sorted_path = os.path.join(args.output_dir, "thinking_off_sorted")
+    thinking_off_ds.write_json(thinking_off_sorted_path, orient="records", lines=True)
+
+    # Interleave datasets and combine into a single output file
+    interleave_datasets(
+        thinking_on_sorted_path,
+        thinking_off_sorted_path,
+        os.path.join(args.output_dir, "training.jsonl"),
+    )
+
```
praateekmahajan, line 192: IIUC above, would you something like this work? Obviously you're t the risk of getting everything to the one worker node due to coalesce (but due to the global sort comment above, that  might be needed) 

```python
df1 = ( 
    read_json(...).coalesce()
    .assign(priority=1)
    .sort(by=[token_count])
    .assign(id=np.arange(len(df1)
)

df2 = ( 
    read_json(...).coalesce()
    .assign(priority=2)
    .sort(by=[token_count])
    .assign(id=np.arange(len(df2)
)

df = df1.union(df2).sort(by=[id, priority])

df.to_json(...)
```

tutorials/text/llama-nemotron-data-curation/main.py
sarahyurick: When I examined the output files, they are sorted alphabetically by name, also corresponding to the global sort operation that happened. So it worked as intended.

Toy example (not how the file names or dataset would look):
```
abc.json
   completion_token_count: 3
   completion_token_count: 7
   completion_token_count: 8

def.json
   completion_token_count: 11
   completion_token_count: 20
   completion_token_count: 90

ghi.json
   completion_token_count: 91
   completion_token_count: 92
   completion_token_count: 98
```

This is why later when I interleave the 2 sorted datasets, I use `files = sorted(f for f in os.listdir(jsonl_dir) if f.endswith(".json"))` to ensure that they are read in the correct (sorted) order.

tutorials/text/llama-nemotron-data-curation/main.py
sarahyurick: The global sort method works correctly and it works fast, so I am not worried.

tutorials/text/llama-nemotron-data-curation/utils/jsonl_utils.py
sarahyurick: I would rather not, this is not a great function by any means (takes a few minutes to run on 40 GB). Once we have JSONL sharding support, I am sure it will be better and should be dropped into this tutorial.

tutorials/text/llama-nemotron-data-curation/main.py
praateekmahajan: Hmm I see how ray data does it.

1. When ray data reads a dataset it splits into blocks (similar to dask's partition). 
2. Then it globally sorts (with some shuffle) maintaining each block has sorted data. 
3. The write then writes based on `{uuid}_{block_idx}.json`.  Code here https://github.com/ray-project/ray/blob/b5a6ad9300ce5f3da9fab9d5f875babc2cc7be2e/python/ray/data/dataset.py#L3722-L3725

To be guaranteed that ray data writes files in a way that when we read those files lexicographically sorted, Ray Data must in step 2 above respect that block `i` contains only values less than or equal to block `i+1`.

And looking at the test I think it does
https://github.com/ray-project/ray/blob/b5a6ad9300ce5f3da9fab9d5f875babc2cc7be2e/python/ray/data/tests/test_sort.py#L35-L51


Having said all this (a fun dive) I think maybe your interleaving code can leverage that by doing a union and sort in ray data (so that its parallel) and then you can create one large file using some bash magic (`cat *.json > new_file.json`)

sarahyurick: Nice, thanks for sharing.

IIRC out of the full tutorial (~60 minutes total), sorting takes about 1-2 minutes and interleaving takes less than a minute, so I am not worried about speeding it up. I would rather keep the current code as a simpler solution and easier for folks to follow (maybe just my preference).

In the Dask version, we did originally have some CLI command for merging into the single file (the flow was sort 2 datasets -> interleave them -> write in many partitions -> use CLI to combine all partitions into a single file), but the goal was to eventually include a Python solution. The current Python function allows us to interleave correctly and write to a single file all in one go.

tutorials/text/llama-nemotron-data-curation/main.py
sarahyurick: Accidentally resolved this, yes I can add this.

tutorials/text/llama-nemotron-data-curation/filters/model_filters.py
sarahyurick: Accidentally resolved this, yes makes sense, I will double check then remove.

tutorials/text/llama-nemotron-data-curation/main.py
praateekmahajan: I'm okay if you're okay to merge as you know more about what the tutorial does and the scale it runs. 
If we expect the tutorial to run at scale or in a distributed world, I'm sure you can see why the interleaving might be slow given its a single process workflow. But apart from that no hesitation for me

tutorials/text/llama-nemotron-data-curation/utils/jsonl_utils.py
praateekmahajan: If there isn't a ticket can you create one?

praateekmahajan (approved):

tutorials/text/llama-nemotron-data-curation/utils/jsonl_utils.py
sarahyurick: Yes, I think we can use https://github.com/NVIDIA-NeMo/Curator/issues/1156 to track this. I'll link this comment in that issue too.

PR #1064, opened 2025-09-10
Improve text embedding generation

praateekmahajan:
## Description
~3x throughput on model forward pass
Main changes 
1. autocast
2. not using cupy / cudf for embedding concatenation (that was code from dask world)
3. changing default batch size to 1024 for embedding generation (~10% improvement on an a100). we can go higher but we don't see more benefits, plus this allows this to work on GPU skus with lower memory footprint too. users can always customize this regardless.

image

image

**Other changes**
1. make autocast a parameter in the base class ModelStage.
2. this means all other domain classifiers / embedders don't need to have `with torch.autocast` as base class will do it @sarahyurick might need your help to see i caught all cases

## Usage

```python
```
## Checklist

### DIFF
--- a/tests/stages/text/embedders/test_base.py
+++ b/tests/stages/text/embedders/test_base.py
@@ -292,15 +292,17 @@ class TestEmbeddingCreatorStage:
         assert embedding_stage.pooling == stage.embedding_pooling
 
     @pytest.mark.parametrize("pooling_strategy", ["mean_pooling", "last_token"])
+    @pytest.mark.parametrize("autocast", [True, False])
     @pytest.mark.gpu
     def test_embedding_creator_stage_with_reference_embeddings(
-        self, pooling_strategy: str, sample_data: DocumentBatch
+        self, pooling_strategy: str, sample_data: DocumentBatch, autocast: bool
     ) -> None:
         """Test embeddings match reference implementation (requires GPU and model download)."""
         stage = EmbeddingCreatorStage(
             model_identifier="sentence-transformers/all-MiniLM-L6-v2",
             embedding_pooling=pooling_strategy,
             model_inference_batch_size=32,
+            autocast=autocast,
         )
 
         # Decompose and setup stages
@@ -362,7 +364,7 @@ class TestEmbeddingCreatorStage:
             )
             inputs = {k: v.to("cuda") for k, v in inputs.items()}
 
-            with torch.no_grad(), torch.autocast(device_type="cuda"):
+            with torch.no_grad():
                 outputs = model(**inputs)
 
             if pooling_strategy == "last_token":


nemo_curator/stages/text/models/model.py
```diff
@@ -154,7 +158,13 @@ def process(self, batch: DocumentBatch) -> DocumentBatch:
...
+                if self.autocast:
+                    with torch.autocast(device_type="cuda"):
+                        if self.unpack_inference_batch:
+                            outputs = self.model(**model_input_batch)
+                        else:
+                            outputs = self.model(model_input_batch)
+                elif self.unpack_inference_batch:
                     outputs = self.model(**model_input_batch)
                 else:
                     outputs = self.model(model_input_batch)
```
sarahyurick: Let's do something like
```
def _call_model(self, model_input_batch):
    if self.unpack_inference_batch:
        return self.model(**model_input_batch)
    return self.model(model_input_batch)

if self.autocast:
    with torch.autocast(device_type="cuda"):
        outputs = self._call_model(model_input_batch)
else:
    outputs = self._call_model(model_input_batch)
```
instead.

nemo_curator/stages/text/embedders/base.py
```diff
@@ -62,33 +60,23 @@ def outputs(self) -> tuple[list[str], list[str]]:
...
-            return self._get_last_token(outputs, model_input_batch[ATTENTION_MASK_COLUMN])
+            return self._get_last_token(outputs, model_input_batch[ATTENTION_MASK_COLUMN]).cpu()
 
-    def collect_outputs(self, processed_outputs: list[torch.Tensor]) -> cp.ndarray:
-        """Collect embeddings into a cupy array."""
-        # TODO : benchmarking this and maybe stay in cpu land
-        cupy_array_embeddings = [cp.asarray(emb_chunk) for emb_chunk in processed_outputs]
-        return cp.concatenate(cupy_array_embeddings, axis=0)
+    def collect_outputs(self, processed_outputs: list[torch.Tensor]) -> list[list[float]]:
+        return torch.cat(processed_outputs, dim=0).numpy().tolist()
```
VibhuJawa, line 75: Is there a requirement here of this being float32/float16 for the downstream models.  With autocast this might become `float16/bfloat16`  etc and unsure if that is a problem for the other downstream models ? 


```suggestion
        return torch.cat(processed_outputs, dim=0).cpu().to(torch.float32).numpy().tolist()
```

nemo_curator/stages/text/embedders/base.py
```diff
@@ -62,33 +60,23 @@ def outputs(self) -> tuple[list[str], list[str]]:
...
-        self.model.to("cuda")
+        self.model.eval().to("cuda")
 
     def process_model_output(
         self, outputs: torch.Tensor, model_input_batch: dict[str, torch.Tensor] | None = None
     ) -> torch.Tensor:
         """Process model outputs to create embeddings."""
         if self.pooling == "mean_pooling":
-            return self._mean_pooling(outputs, model_input_batch[ATTENTION_MASK_COLUMN])
+            return self._mean_pooling(outputs, model_input_batch[ATTENTION_MASK_COLUMN]).cpu()
```
VibhuJawa, line 70: Do we need to upcast/force float32 here too ?

VibhuJawa (changes requested):
Mostly looks good, can we add a test for any of the models to verify that with and without autocast the results are similar ?

nemo_curator/stages/text/embedders/base.py
praateekmahajan: We don't have to cast it to torch.float32 as it still maintains that. I guess that makes sense since autocast only does certain ops in float16 and not all.

```python
# when using autocast=True
type(o)=<class 'torch.Tensor'> o.dtype=torch.float32 o.shape=torch.Size([28783, 384])
type(o_np)=<class 'numpy.ndarray'> o_np.dtype=dtype('float32') o_np.shape=(28783, 384)
Model fwd pass took 9.46 seconds i.e. 3041.84 r/s.

# when using autocast=False
type(o)=<class 'torch.Tensor'> o.dtype=torch.float32 o.shape=torch.Size([28783, 384])
type(o_np)=<class 'numpy.ndarray'> o_np.dtype=dtype('float32') o_np.shape=(28783, 384)
	Model fwd pass took 25.18 seconds i.e. 1142.98 r/s.
```

nemo_curator/stages/text/embedders/base.py
praateekmahajan: Same as above https://github.com/NVIDIA-NeMo/Curator/pull/1064#discussion_r2349553904

nemo_curator/stages/text/embedders/base.py
VibhuJawa: Sounds good. I just was not sure how it behaved . 

P.S: Which GPU was this on ?

VibhuJawa (approved):
LGTM, thanks for the work on consolidating this

PR #1066, opened 2025-09-16
ci: Address cve and update packages

thomasdhc:
## Usage

```python
```
## Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -52,7 +52,7 @@ dependencies = [
     "pyarrow",
     "ray[default,data]>=2.49",
     "torch",
-    "transformers==4.53.1",
+    "transformers>=4.55.0",
 ]
 
 [project.optional-dependencies]
@@ -136,7 +136,7 @@ internvideo2 = [
     "torchaudio>=2.4.1",
     "torchvision>=0.19.1",
     "tqdm>=4.66.5",
-    "transformers==4.53.1",
+    "transformers>=4.55.0",
     "wandb>=0.18.3",
 ]
 
@@ -158,7 +158,14 @@ video_cuda12 = [
     "PyNvVideoCodec==2.0.2; (platform_machine == 'x86_64' and platform_system != 'Darwin')",
     "torch<=2.8.0",
     "torchaudio",
-    "vllm==0.9.2; (platform_machine == 'x86_64' and platform_system != 'Darwin')",
+    "vllm==0.10.1.1; (platform_machine == 'x86_64' and platform_system != 'Darwin')",
+]
+
+
+# CVE fixes
+security = [
+    "protobuf>=4.25.8",
+    "xgrammar>=0.1.21",
 ]
 
 # All dependencies
@@ -167,6 +174,7 @@ all = [
     "nemo_curator[image_cuda12]",
     "nemo_curator[text_cuda12]",
     "nemo_curator[video_cuda12]",
+    "nemo_curator[security]",
 ]
 
 [dependency-groups]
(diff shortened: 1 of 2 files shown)


sarahyurick (approved):
LGTM since classifier tests seem to pass.

pyproject.toml
```diff
@@ -158,7 +158,14 @@ video_cuda12 = [
...
     "torchaudio",
-    "vllm==0.9.2; (platform_machine == 'x86_64' and platform_system != 'Darwin')",
+    "vllm==0.10.1.1; (platform_machine == 'x86_64' and platform_system != 'Darwin')",
+]
+
+
+# CVE fixes
+security = [
+    "protobuf>=4.25.8",
+    "xgrammar>=0.1.21",
```
praateekmahajan: wdyt of adding them to `dependencies` with a comment stating they're for CVE rather than a new group.. i'm ok with this too but thought i'd ask

thomasdhc: yeah we can do that. The dependencies could come from specific group though, ie video, audio. So for some modules we'll be installing additional dependencies where they are not needed. Let me actually see if I can trace the dependencies to their specific module and move the requirements under those modules.

thomasdhc: @praateekmahajan I've moved the dependencies to appropriate groups. Have a look when you get a chance.

pyproject.toml
```diff
@@ -167,6 +169,7 @@ all = [
     "nemo_curator[image_cuda12]",
     "nemo_curator[text_cuda12]",
     "nemo_curator[video_cuda12]",
+    "nemo_curator[security]",
```
praateekmahajan: we don't need this anymore right?

pyproject.toml
```diff
@@ -158,7 +159,8 @@ video_cuda12 = [
     "PyNvVideoCodec==2.0.2; (platform_machine == 'x86_64' and platform_system != 'Darwin')",
     "torch<=2.8.0",
     "torchaudio",
-    "vllm==0.9.2; (platform_machine == 'x86_64' and platform_system != 'Darwin')",
+    "vllm==0.10.1.1; (platform_machine == 'x86_64' and platform_system != 'Darwin')",
+    "xgrammar>=0.1.21",
```
praateekmahajan: can you add a comment here why the pin, e.g. `# due to CVE-2025-123456`

pyproject.toml
```diff
@@ -49,10 +49,11 @@ dependencies = [
     "loguru",
     "mecab-python3",
     "pandas>=2.1.0",
+    "protobuf>=4.25.8",
```
praateekmahajan: can you add a comment here why the pin, e.g. # due to CVE-2025-123456

pyproject.toml
praateekmahajan: lgtm! as long as a non-video doesn't need `xgrammar` i think we should be fine but i imagine there is a risk that if tomorrow audio adds a transitive dependency to `xgrammar` then we'll be exposed.. 
however since we build the container with `--all-extras` then i don't think this is an issue

praateekmahajan: Did a quick google search seems like instead of adding it as an explicit requirement either of the following could work.. I think `constraints` is what we need but could be wrong
```
[tool.uv]
overrides = [
    "x>=Y" 
]
```
```
[tool.uv]
constraints = [
    "x>=Y" 
]
```

instead of an explicit dependency

thomasdhc: Addressed

pyproject.toml
thomasdhc: Addressed

pyproject.toml
thomasdhc: Addressed

pyproject.toml
thomasdhc: Addressed

PR #1068, opened 2025-09-16
Allow jsonl/parquet to rewrite the same file

praateekmahajan:
## Description
The reason we can always rewrite because

1. In grand scheme of things file writing is potentially much cheaper than rest of pipeline
2. We added this to mimic checkpoint behavior however I believe that should be solved differently
    - possibly a stage after FilePartitioningStage that removes tasks if the output of them is found
4. This is already bug prone for instance if a user reruns a pipeline but adds a new stage to it.. rn the writer won't rewrite the files i.e. the effect of new stage wouldn't have taken place

## Usage

```python
```
## Checklist


tests/stages/text/io/writer/test_jsonl.py
```diff
@@ -222,3 +223,45 @@ def test_jsonl_writer_overwrite_mode(self, pandas_document_batch: DocumentBatch,
...
         pd.testing.assert_frame_equal(df, pandas_document_batch.to_pandas())
+
+    @pytest.mark.parametrize("consistent_filename", [True, False])
+    def test_jsonl_writer_overwrites_existing_file(
+        self,
+        pandas_document_batch: DocumentBatch,
+        consistent_filename: bool,
+        tmpdir: str,
+    ):
+        """Test JsonlWriter with different data types."""
```
VibhuJawa:
```suggestion
        """Test that JsonlWriter overwrites existing files when writing to the same path."""

```

nemo_curator/stages/text/io/writer/base.py
```diff
@@ -95,12 +95,11 @@ def process(self, task: DocumentBatch) -> FileGroupTask:
         file_extension = self.get_file_extension()
         file_path = self.fs.sep.join([self._fs_path, f"{filename}.{file_extension}"])
 
-        # Skip if file already exists (idempotent writes)
         if self.fs.exists(file_path):
-            logger.debug(f"File {file_path} already exists, skipping")
-        else:
-            self.write_data(task, file_path)
-            logger.debug(f"Written {task.num_items} records to {file_path}")
+            logger.debug(f"File {file_path} already exists, overwriting it")
```
sarahyurick, line 99: What is the use of `check_output_mode` if we are always overwriting anyway?

VibhuJawa: `check_output_mode` works on a directory level, https://github.com/NVIDIA-NeMo/Curator/blob/8993985374297abeafb822045c6b977495de5f75/nemo_curator/utils/file_utils.py#L302-L342
and we should clean up directories under overwrite mode. 

Above livne works at a file level

sarahyurick: I see, thanks. Although I am still curious, from the PR description a primary reason for this change is that "This is already bug prone for instance if a user reruns a pipeline but adds a new stage to it.. rn the writer won't rewrite the files i.e. the effect of new stage wouldn't have taken place" - does using `mode="overwrite"` not already solve this?

VibhuJawa: Yeah,  mode="overwrite" does solve it but its not default.  The `ignore` stage is probably use-less now,  my understanding is it was tripping up because xenna was pre-empting while writing than anything else.

praateekmahajan: `mode=overwrite` (similarly `error` and `ignore`) are overloaded here. That is for directory level stuff since it cleans the directory (or not). I am not sure `append` should exist, as its non trivial to do file-level appends in a distributed way. Read [SPARK-18199](https://issues.apache.org/jira/browse/SPARK-18199?focusedCommentId=16069304&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-16069304)

This PR aims to solve at a file level given a directory (has been recreated or not) should we overwrite an existing file that exists with the same name. We might have a file with the same name because since the pipeline started maybe another stage wrote a file with that name too (or xenna got pre-empted). And the answer to that question is yes, we should always do that.

PR #1101, opened 2025-09-24
Enhance RayClient shutdown process by resetting RAY_ADDRESS

abhinavg4:
Adding RAY_ADDRESS env reset.
## Usage

```python
```
## Checklist


nemo_curator/core/client.py
```diff
@@ -127,6 +127,8 @@ def stop(self) -> None:
         if self.ray_process:
             self.ray_process.kill()
             self.ray_process.wait()
+            # reset the ray address
+            os.environ.pop("RAY_ADDRESS", None)
```
praateekmahajan, line 137: if we always pop, then would this scenario fail

```python
ray start --port 1234
export RAY_ADDRESS=1234

python run.py # this relies on RAY_ADDRESS therefore doesn't start a new ray start because of line 96/97

python run.py # since RAY_ADDRESS is popped it'll start a new cluster
```

ayushdg: Yup. Mentioned in my issue but also adding the text here:

>  the stop method should also track if the env was set before calling cluster.start and if so it should not unset it.

abhinavg4: Wait, I think if the ray_process is not set, why will this env var get popped ? For your case prateek, the ray_process is always none right?

praateekmahajan: Aah yeah sorry missed the conditional..

praateekmahajan: Can you actually move the environ variable setting from init_cluster to this file and set it here after we call init_cluster, therefore both the setting and unsetting is happening here

nemo_curator/core/client.py
```diff
@@ -120,13 +121,18 @@ def start(self) -> None:
                 self.num_cpus,
                 self.enable_object_spilling,
             )
+            # Set environment variable for RAY_ADDRESS
+            ip_address = socket.gethostbyname(socket.gethostname())
```
ayushdg: nit: rather than also do this inside `init_cluster` what if we pass the ip_address to the `init_cluster` method that it can re-use

sarahyurick (approved):

nemo_curator/core/client.py
```diff
@@ -119,14 +120,21 @@ def start(self) -> None:
                 self.num_gpus,
                 self.num_cpus,
                 self.enable_object_spilling,
+                block=True,
```
ayushdg, line 123: Any reason you added the `block` here?
In theory it shouldn't matter but I don't know enough about this to be sure if it's not going to be an issue. Since we've already tested a fair bit without it what do you think about removing it?

ayushdg: nvm looks like `block`=True was the default you just explicitly pass it here.

ayushdg (approved):

Issue #1123, opened 2025-09-26
Allow `RayClient` to be used as a context manager


jrbourbeau:
`RayClient` usage today has explicit setup (`.start()`) and teardown (`.stop()`) methods.

```python
client = RayClient()
client.start()

pipeline = ...

client.stop()
```

In these cases, I typically see a context manager used (or at least it's an option).

```python
with RayClient():
    pipeline = ...
```

which would ensure teardown always happens (though I see there's an `atexit` handler in this case) and is a little less typing : )

PR #1155, opened 2025-10-01
Context manager support for ``RayClient``

jrbourbeau:
This PRs makes it so users can use ``RayClient`` as a context manager instead of manually calling `.start()` / `.stop()` methods. 

Closes https://github.com/NVIDIA-NeMo/Curator/issues/1123

### DIFF
--- a/docs/project.json
+++ b/docs/project.json
@@ -1,1 +1,1 @@
-{"name": "nemo-curator", "version": "0.25.7"}
+{"name": "nemo-curator", "version": "25.09"}
--- a/docs/versions1.json
+++ b/docs/versions1.json
@@ -1,7 +1,12 @@
 [
     {
         "preferred": true,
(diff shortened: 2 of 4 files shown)


tests/core/test_get_ray_client.py
```diff
@@ -79,3 +79,30 @@ def test_get_ray_client_multiple_start():
             os.environ["RAY_ADDRESS"] = initial_address
         else:
             os.environ.pop("RAY_ADDRESS", None)
+
+
+def wait_for_ray_cluster_start(client: RayClient, timeout: int = 10):
```
jrbourbeau: The other tests in this file as marked as allowed to fail due to ray cluster startup time variability. Happy to update those tests to use this little utility too

sarahyurick (approved):
Thanks!

PR #1160, opened 2025-10-02
Add Nemotron VLM support in video captioning

ronjer30:
## Description
Adds support for the Nemotron Nano 12B V2 VLM

## Usage
1) Download model checkpoints
2) Run
```bash
python tutorials/video/getting-started/video_split_clip_example.py \
  --video-dir </path/to/videos_directory/> \
  --output-clip-path ./outputs \
  --generate-captions \
  --captioning-algorithm nemotron \
  --captioning-model-path </path/to/checkpoints/> \
  --captioning-batch-size 8 \
  --no-generate-embeddings 
```
## Checklist

### DIFF
--- a/nemo_curator/models/nemotron_h_vl.py
+++ b/nemo_curator/models/nemotron_h_vl.py
@@ -1,0 +1,225 @@
+# Copyright (c) 2025, NVIDIA CORPORATION.  All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import json
+import os
+from pathlib import Path
+from typing import Any
+
+from loguru import logger
+
+from nemo_curator.models.base import ModelInterface
+from nemo_curator.utils import grouping
+from nemo_curator.utils.hf_download_utils import download_model_from_hf
+
+# Constants for prompt processing
+VIDEO_TAG_SPLIT_MAX = 1
+EXPECTED_VIDEO_TAG_PARTS = 2
+
+try:
+    from vllm import LLM, SamplingParams
+
+    VLLM_AVAILABLE = True
+except ImportError:
+    VLLM_AVAILABLE = False
+
+    class LLM:
+        pass
+
+    class SamplingParams:
+        pass
+
+
+# HuggingFace model IDs (will be updated when models are published)
+_NEMOTRON_H_NANO_MODEL_ID = None
+_NEMOTRON_H_NANO_MODEL_REVISION = None
+
+
+class NemotronHVL(ModelInterface):
+    """NemotronH hybrid Mamba-Attention VLM for video captioning."""
+
+    def __init__(  # noqa: PLR0913
+        self,
+        model_dir: str,
+        model_variant: str = "nemotron",
+        caption_batch_size: int = 8,
+        max_output_tokens: int = 512,
+        stage2_prompt_text: str | None = None,
+        verbose: bool = False,
+    ):
+        self.model_dir = model_dir
+        self.model_variant = model_variant
+        self.caption_batch_size = caption_batch_size
+        self.max_output_tokens = max_output_tokens
+        self.stage2_prompt = stage2_prompt_text if stage2_prompt_text else "Please refine this caption: "
+        self.verbose = verbose
+
+        if _NEMOTRON_H_NANO_MODEL_ID is not None:
+            self.weight_file = str(Path(model_dir) / _NEMOTRON_H_NANO_MODEL_ID)
+        else:
+            # Local checkpoint: model_dir is the checkpoint path itself
+            self.weight_file = str(Path(model_dir))
+
+    @property
+    def model_id_names(self) -> list[str]:
+        """Return model ID from config.json or HuggingFace ID."""
+        if _NEMOTRON_H_NANO_MODEL_ID is not None:
+            return [_NEMOTRON_H_NANO_MODEL_ID]
+
+        # Read from config.json if available
+        try:
+            config_path = Path(self.weight_file) / "config.json"
+            with open(config_path) as f:
+                config = json.load(f)
+            return [config.get("_name_or_path", Path(self.weight_file).name)]
+        except (FileNotFoundError, json.JSONDecodeError, KeyError):
+            return [Path(self.weight_file).name]
+
+    def setup(self) -> None:
+        if not VLLM_AVAILABLE:
+            msg = "vllm is required for NemotronHVL but is not installed. Please install vllm: pip install vllm"
+            raise ImportError(msg)
+
+        # Use V0 engine to avoid flashinfer issues for now
+        os.environ["VLLM_USE_V1"] = "0"
+        logger.info("Using vLLM V0 engine.")
+
+        os.environ["PYTORCH_CUDA_ALLOC_CONF"] = "expandable_segments:True"
+
+        self.model = LLM(
+            model=self.weight_file,
+            trust_remote_code=True,
+            tensor_parallel_size=1,
+            gpu_memory_utilization=0.9,
+            max_model_len=32768,
+            limit_mm_per_prompt={"video": 1},
+        )
+
+        self.sampling_params = SamplingParams(
+            temperature=0.6,
+            max_tokens=self.max_output_tokens,
+            top_p=0.95,
+            stop=["</s>", "<|endoftext|>", "<SPECIAL_12>", "</think>"],
+        )
+
+        logger.info(
+            f"NemotronHVL initialized: variant={self.model_variant}, "
+            f"TP=1, GPU_util=0.9, max_len=32768"
+        )
+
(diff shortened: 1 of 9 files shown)


nemo_curator/stages/video/caption/caption_preparation.py
```diff
@@ -133,13 +140,16 @@ def process(self, task: VideoTask) -> VideoTask:
...
-                        qwen_llm_input=llm_input,
-                    ),
+                window = _Window(
+                    window_frame_info.start,
+                    window_frame_info.end,
+                    mp4_bytes=window_bytes,
                 )
+                window.llm_inputs[self.model_variant] = llm_input
+
+                if self.model_variant == "qwen":
```
suiyoubi: could we have this unified such that we can both qwen and nemotronH stores the input in `window.llm_inputs` ?

nemo_curator/tasks/video.py
```diff
@@ -41,8 +41,9 @@ class _Window:
     end_frame: int
     # MP4 bytes for this window
     mp4_bytes: bytes | None = None
+    llm_inputs: dict[str, dict[str, Any]] = field(default_factory=dict)
     # Qwen LLM input for this window
-    qwen_llm_input: dict[str, Any] | None = None
+    qwen_llm_input: dict[str, Any] | None = None  # Backward compatibility
```
suiyoubi: Same above, this should be removed

pyproject.toml
```diff
@@ -121,13 +121,14 @@ video_cpu = [
...
     "easydict",
+    "timm"
 ]
 
 video_cuda12 = [
     "nemo_curator[video_cpu]",
     "nemo_curator[cuda12]",
     "cvcuda_cu12",
-    "flash-attn<=2.8.3; (platform_machine == 'x86_64' and platform_system != 'Darwin')",
+    "flash-attn>=2.7.1,<=2.8.2; (platform_machine == 'x86_64' and platform_system != 'Darwin')",
```
suiyoubi: Any reason for this ? Does the nemotronH not worked with 2.8.3 ?

tutorials/video/getting-started/video_split_clip_example.py
```diff
@@ -215,9 +219,16 @@ def create_video_splitting_pipeline(args: argparse.Namespace) -> Pipeline:  # no
                 )
             )
 
+        # Determine model directory
+        model_dir = (
+            args.captioning_model_path
```
suiyoubi: I would suggest name this as `nemotronh_vl_model_path`. Once we have the nemotronh uploaded to HF, we can simplify the logic here

nemo_curator/stages/video/caption/caption_preparation.py
ronjer30: Sure, I've updated the captioning code to consistently use `window.llm_inputs`

pyproject.toml
ronjer30: That is correct, using 2.8.3 results in missing symbol errors at runtime

ronjer30:
Thanks @suiyoubi, I've addressed the issues, added tests and updated existing ones as well.

nemo_curator/models/nemotron_h_vl.py
```diff
@@ -0,0 +1,277 @@
...
+        if self._normalized_variant not in _NEMOTRON_VARIANTS_INFO:
+            valid_variants = ", ".join(_NEMOTRON_VARIANTS_INFO.keys())
+            msg = f"Invalid model_variant: {model_variant}. Valid options are: {valid_variants}"
+            raise ValueError(msg)
+
+        self.model_dir = model_dir
+        self.model_variant = model_variant
+        self.caption_batch_size = caption_batch_size
+        self.max_output_tokens = max_output_tokens
+        self.stage2_prompt = stage2_prompt_text if stage2_prompt_text else "Please refine this caption: "
```
weijiac0619, line 110: could you add the same fallback for QWEN?

tutorials/video/getting-started/video_split_clip_example.py
```diff
@@ -648,25 +656,28 @@ def create_video_splitting_argparser() -> argparse.ArgumentParser:  # noqa: PLR0
         dest="captioning_stage2_caption",
         action="store_true",
         default=False,
-        help="If set, generated captions are used as input prompts again into QwenVL to refine them",
+        help="If set, generated captions are refined with a second model pass (works for both Qwen and Nemotron)",
```
weijiac0619, line 628: I got TypeError: unsupported operand type(s) for +: 'NoneType' and 'str' when running qwen + captioning-stage2-caption 

CUDA_VISIBLE_DEVICES=1 RAY_ACCEL_ENV_VAR_OVERRIDE_ON_ZERO=0 TMPDIR=/mnt/data/weijiac/tmp/pipeline_tmp     .venv/bin/python tutorials/video/getting-started/video_split_clip_example.py     --video-dir /mnt/data/weijiac/tmp/nemo_video_test/input     --model-dir /mnt/data/weijiac/tmp/nemo_video_test/models_fixed     --output-path /mnt/data/weijiac/tmp/nemo_video_test/out_qwen_s1     --splitting-algorithm fixed_stride     --no-generate-embeddings     --generate-captions     --captioning-algorithm qwen     --captioning-stage2-caption  --verbose

weijiac0619:
When running with --captioning-algorithm nemotron-bf16, vLLM's nano_nemotron_vl.py accesses config.dtype directly during model initialization:
`  self.vision_model = self.get_vit_model_from_radio_config(config).to(config.dtype)`
However, the pinned model revision (5d250e2e) of NVIDIA-Nemotron-Nano-12B-v2-VL-BF16 never sets self.dtype in NemotronHConfig.__init__, causing:

 `  AttributeError: 'NemotronHConfig' object has no attribute 'dtype'`

weijiac0619:
1. we may want to force  --captioning-model-does-preprocess to be true when using nemotron  model.
2. When using nemotron, the video frames are first resized to dimensions divisible by 28(from Qwen2.5-VL's patch granularity), then vLLM resizes them again to 512×512. we may want to avoid the intermediate resizing.

nemo_curator/models/nemotron_h_vl.py
```diff
@@ -0,0 +1,277 @@
...
+    def model_id_names(self) -> list[str]:
+        """Return HuggingFace model ID for the selected variant."""
+        return [self._hf_model_id]
+
+    def setup(self) -> None:
+        if not VLLM_AVAILABLE:
+            msg = "vllm is required for NemotronHVL but is not installed. Please install vllm: pip install vllm"
+            raise ImportError(msg)
+
+        os.environ["PYTORCH_CUDA_ALLOC_CONF"] = "expandable_segments:True"
```
suiyoubi: could you please explain why we need this ?

nemo_curator/models/cosmos_embed1.py
```diff
@@ -38,6 +41,191 @@
 }
 
 
+_HEAD_MASK_VECTOR_DIM: Final[int] = 1
```
suiyoubi: do we need all the changes to the cosmos_embed1.py files ?

weijiac0619: yes, they're all needed for` trust_remote_code=True` to work with newer transformers versions. If the file bulk is a concern, they can be moved to a separate _cosmos_embed1_compat.py.

nemo_curator/models/nemotron_h_vl.py
weijiac0619: The env var is not needed on my 48GB GPU. i believe it's a safeguard for smaller GPUs where the 12B BF16 model can trigger OOM. we can move this to docs rather than hardcoding it here.

suiyoubi: Isee. Thanks for explaining. I would prefer to have the min requirement in the docs rather than the safeguard here

nemo_curator/models/cosmos_embed1.py
weijiac0619: discussed with @suiyoubi offline. the patches are not needed with latest cosmos model. will clean this up

suiyoubi:
/ok to test [d8ea9ee](https://github.com/NVIDIA-NeMo/Curator/pull/1160/commits/d8ea9ee453d1623023db46a376dbf1d5d970637b)

pyproject.toml
```diff
@@ -167,6 +167,9 @@ video_cuda12 = [
     "PyNvVideoCodec==2.0.2; (platform_machine == 'x86_64' and platform_system != 'Darwin')",
     "torch<=2.9.1",
     "torchaudio",
+    # Nemotron VL support
+    "transformers>=4.56.0",
+    "vllm>=0.13; (platform_machine == 'x86_64' and platform_system != 'Darwin')",
```
thomasdhc: Looks like vllm already has its own dependency group, and video_cuda12 already inherits it via nemo_curator[vllm]. Can the requirement here be consolidated?

pyproject.toml
```diff
@@ -167,6 +167,9 @@ video_cuda12 = [
     "PyNvVideoCodec==2.0.2; (platform_machine == 'x86_64' and platform_system != 'Darwin')",
     "torch<=2.9.1",
     "torchaudio",
+    # Nemotron VL support
+    "transformers>=4.56.0",
```
thomasdhc: The current transformers version is being overwritten via override-dependency: `"transformers<=4.55.2", # Else Cosmos Embed imports fail`. Is this ok?

pyproject.toml
weijiac0619: nice catch, removed

pyproject.toml
weijiac0619: good catch. removed that override-dependency since the cosmos patch was gone.

thomasdhc (approved):

Issue #1156, opened 2025-10-01
Add utility to shard large datasets into smaller files.


ayushdg:
**Is your feature request related to a problem? Please describe.**
Currently Curator readers support combining multiple small files into a larger batch with `blocksize` or `files_per_partition` but do not support splitting/sharding a larger file into smaller batches during the read.

**Describe the solution you'd like**
Longer term exploration for supporting splitting larger files into smaller batches needs to be explored, but meanwhile a utility that can shard large files in to smaller ones and write them to disk might be helpful for use cases where the original dataset contains large files.

**Describe alternatives you've considered**

**Additional context**
Found while exploring #1152

sarahyurick:
Not completed, we do not have JSONL splitting yet.

sarahyurick:
For JSONL sharding: https://github.com/NVIDIA-NeMo/Curator/pull/1063/files#r2361325477

Once we have support for it I would also like to drop the function I used in https://github.com/NVIDIA-NeMo/Curator/pull/1063.

sarahyurick:
Closing in favor of https://github.com/NVIDIA-NeMo/Curator/issues/1355.

PR #1161, opened 2025-10-02
[WIP] Add large file splitting script

jrbourbeau:
This PR adds a little utility script to split large files (bad for processing) into multiple files with a more reasonable size. Marking as WIP and pushing up early for feedback -- cc @ayushdg @sarahyurick 

For testing locally, here's a little snippet to make a large-ish Parquet file (~7 GB in memory):

```python
import numpy as np
import pandas as pd

num_rows = 200_000_000
data = {
    "id": np.arange(num_rows),
    "value1": np.random.rand(num_rows),
    "value2": np.random.randint(0, 1000, num_rows),
    "category": np.random.choice(["A", "B", "C", "D"], num_rows),
    "timestamp": pd.to_datetime("2023-01-01") + pd.to_timedelta(np.arange(num_rows), unit="s"),
}
df = pd.DataFrame(data)
df.to_parquet("example.parquet")
```

And then you can run the script here like this:

```
python nemo_curator/utils/split_large_file.py --infile example.parquet --outdir foo  --verbose --target-size-mb 128
```

which has output like this

```
Splitting file into smaller ones...

Input file: example.parquet (~7152.56 MB)
Output directory: foo
Target size: 128 MB

Splitting into 64 output files
Saved foo/example_0.parquet with approximate size 111.76 MB
Saved foo/example_1.parquet with approximate size 111.76 MB
Saved foo/example_2.parquet with approximate size 111.76 MB
Saved foo/example_3.parquet with approximate size 111.76 MB
...
...
```

Some room for improvement:

- [ ] Add JSON file support
- [ ] Run in parallel?
- [ ] Support for non-local paths (e.g. using `fsspec` or [`UPath`](https://github.com/fsspec/universal_pathlib))
- [ ] Add support for specifying target size as a string (e.g. "150 MiB"). Definitely not needed, just a small ergonomic thing. 

Closes https://github.com/NVIDIA-NeMo/Curator/issues/1156

### DIFF
--- a/nemo_curator/utils/split_large_file.py
+++ b/nemo_curator/utils/split_large_file.py
@@ -1,0 +1,79 @@
+# Copyright (c) 2025, NVIDIA CORPORATION.  All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import argparse
+import os
+
+import pyarrow as pa
+import pyarrow.parquet as pq
+
+
+def _split_pyarrow_table(table: pa.Table, target_size_mb: int) -> list[pa.Table]:
+    target_size_bytes = target_size_mb * 1024 * 1024
+    # Slice full table into two chunks
+    tables = [table.slice(0, table.num_rows // 2), table.slice(table.num_rows // 2, table.num_rows)]
+    results = []
+    for t in tables:
+        if t.nbytes > target_size_bytes:
+            # If above the target size, continue chunking until chunks
+            # are below the target size
+            results.extend(_split_pyarrow_table(t, target_size_mb=target_size_mb))
+        else:
+            results.append(t)
+    return results
+
+
+def _split_parquet_file_by_size(infile: str, outdir: str, target_size_mb: int, verbose: bool) -> None:
+    root, ext = os.path.splitext(infile)
+    if not ext:
+        ext = ".parquet"
+    outfile_prefix = os.path.basename(root)
+    table = pq.read_table(infile)
+    if verbose:
+        print(f"""Splitting file into smaller ones...
+
+Input file: {infile} (~{table.nbytes / (1024 * 1024):.2f} MB)
+Output directory: {outdir}
+Target size: {target_size_mb} MB
+""")
+    results = _split_pyarrow_table(table, target_size_mb=target_size_mb)
+    if verbose:
+        print(f"Splitting into {len(results)} output files")
+    for idx, t in enumerate(results):
+        # Write chunk to a new Parquet file
+        output_file = os.path.join(outdir, f"{outfile_prefix}_{idx}{ext}")
+        pq.write_table(t, output_file)
+        if verbose:
+            print(f"Saved {output_file} with approximate size {t.nbytes / (1024 * 1024):.2f} MB")
+
+
+def main(args: argparse.Namespace) -> None:
+    os.makedirs(args.outdir, exist_ok=True)
+    _split_parquet_file_by_size(
+        infile=args.infile, outdir=args.outdir, target_size_mb=args.target_size_mb, verbose=args.verbose
+    )
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--infile", type=str, required=True, help="Path to input file to split")
+    parser.add_argument("--outdir", type=str, required=True, help="Output directory to store split files")
+    parser.add_argument("--target-size-mb", type=int, default=128, help="Target size (in MB) of split output files")
+    parser.add_argument(
+        "--verbose",
+        action="store_true",
+        help="Enable verbose output",
+    )
+    args = parser.parse_args()
+    main(args)


nemo_curator/utils/split_large_file.py
```diff
@@ -0,0 +1,79 @@
...
+            results.append(t)
+    return results
+
+
+def _split_parquet_file_by_size(infile: str, outdir: str, target_size_mb: int, verbose: bool) -> None:
+    root, ext = os.path.splitext(infile)
+    if not ext:
+        ext = ".parquet"
+    outfile_prefix = os.path.basename(root)
+    table = pq.read_table(infile)
```
praateekmahajan: nit, i think we can be slightly better about this by reading row group wise so that we are not constrained by memory, and then split at row group or combine row groups (or parts of it)

praateekmahajan: very bad pseudocode that looks like this
```
current_chunk = None
current_chunk_size_remaining = target_size 
row_groups = pq.read_metadata(file).num_row_groups()
for row_group_content in read_row_group(file, row_group):
    if size(row_group_content) > current_chunk_size_remaining:
         desired_row_group, remaining_row_group = split_row_group(current_chunk_size_remaining)
         current_chunk.append(desired_row_group)

```

nemo_curator/utils/split_large_file.py
```diff
@@ -0,0 +1,79 @@
...
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--infile", type=str, required=True, help="Path to input file to split")
+    parser.add_argument("--outdir", type=str, required=True, help="Output directory to store split files")
+    parser.add_argument("--target-size-mb", type=int, default=128, help="Target size (in MB) of split output files")
+    parser.add_argument(
+        "--verbose",
+        action="store_true",
+        help="Enable verbose output",
+    )
+    args = parser.parse_args()
```
praateekmahajan: What if when given a directory we should start a ray cluster and then do `[ray.remote(func(file)) for file in get_all_file_paths_under(director)]``

jrbourbeau: Sure, that sounds good. Any reason to not always go the ray route? Asking because it might make the implementation a little more straightforward

praateekmahajan: Sorry if it wasn't clear, I was saying let's always go the ray route. I saw in your main you only had `-infile` which I assumed only does it for each file. We should probably provide users the ability to run this on a directory which will need parallelization and hence ray

nemo_curator/utils/split_large_file.py
jrbourbeau: Ah yeah. thanks for catching, I forgot `read_table` loads everything in memory and doesn't just extract metadata. You're right, iterating over row groups and splitting / merging accordingly would be nice to minimize memory constraints. I'll push something up for that

nemo_curator/utils/split_large_files.py
```diff
@@ -0,0 +1,126 @@
+# Copyright (c) 2025, NVIDIA CORPORATION.  All rights reserved.
```
praateekmahajan, line 1: Thanks for adding this, can we add some quick tests? 
If we you want to test out ray functionality and need a ray cluster you can use the fixture `shared_ray_client` (defined in conftest.py)

If I were you, I'd just test the python functions `(_split_table, _write_table_to_file  and split_parquet_file_by_size)` and assume thing work on ray..

jrbourbeau: Just added some test that check file splitting is reasonable for parquet files with various row group sizes (both splitting row group and merging row group cases)

jrbourbeau: Hmmm tests seem to not be running for some reason...

jrbourbeau:
When someone gets a chance, could I get a restart for this failing build https://github.com/NVIDIA-NeMo/Curator/actions/runs/18230133314/job/52090467756?pr=1161 Looks like a sporadic failure unrelated to the changed in this PR

ayushdg:
I've triggered a rerun. @thomasdhc any ideas on the sporadic docker errors we see from runs?

nemo_curator/utils/split_large_files.py
```diff
@@ -0,0 +1,130 @@
...
+        ext = ".parquet"
+    outfile_prefix = os.path.basename(root)
+
+    print(f"""Splitting parquet file...
+
+Input file: {input_file}
+Output directory: {outdir}
+Target size: {target_size_mb} MB
+""")
+
```
praateekmahajan: This is okay for now but maybe we can remove the print or move to loguru? 
As an FYI we mostly we use loguru (that comes with some overhead of registering loguru's serializer after ray client has started). 

See https://github.com/NVIDIA-NeMo/Curator/blob/b50d916312aed4ea4b38055effd74a85bffd85e0/nemo_curator/core/utils.py#L86-L90

Since you use RayClient below you get the loguru handled for you so i'd just say use loguru.debug here so that users can disable logs too

nemo_curator/utils/split_large_files.py
```diff
@@ -0,0 +1,130 @@
...
+            if row_group.nbytes > target_size_bytes:
+                # Large row group case. Split into smaller chunks to get below target size.
+                chunks = _split_table(row_group, target_size=target_size_bytes)
+                for chunk in chunks:
+                    file_idx = _write_table_to_file(
+                        chunk, outdir=outdir, output_prefix=outfile_prefix, ext=ext, file_idx=file_idx
+                    )
+                row_group_idx += 1
+            elif row_group.nbytes + current_size > target_size_bytes and current_size > 0:
```
praateekmahajan: Why do we need the second `current_size > 0:`?

tests/utils/test_split_large_files.py
```diff
@@ -0,0 +1,80 @@
...
+        rng = np.random.default_rng(seed=2)
+        df = pd.DataFrame(
+            {
+                "id": np.arange(num_rows),
+                "value1": rng.random(num_rows),
+                "value2": rng.integers(0, 1000, num_rows),
+                "category": rng.choice(["A", "B", "C", "D"], num_rows),
+                "timestamp": pd.to_datetime("2023-01-01") + pd.to_timedelta(np.arange(num_rows), unit="s"),
+            }
+        )
```
praateekmahajan: Can you add a comment about how much this data is in memory? And to confirm we split by uncompressed file size or file size? I believe the former

tests/utils/test_split_large_files.py
```diff
@@ -0,0 +1,80 @@
...
+
+    # Ensure the original and split data is the same
+    pd.testing.assert_frame_equal(expected, result)
+
+    # Check that split data files have expected sizes
+    sizes_mb = [pq.read_table(f).nbytes / (1024 * 1024) for f in outdir.rglob("*")]
+    # Below the target size
+    assert all(s_mb < target_size_mb for s_mb in sizes_mb)
+    # More than half the target (ignoring the last file, which can sometimes be small)
+    assert all(s_mb > target_size_mb / 2 for s_mb in sizes_mb[:-1])
```
praateekmahajan, line 83: Awesome

nemo_curator/utils/split_large_files.py
jrbourbeau: Sure, happy to switch over to loguru

nemo_curator/utils/split_large_files.py
jrbourbeau: Ah, good catch. This was left over from a previous iteration when I had this check at the top of this `if/elif/else` block, but it isn't needed anymore. Removed it in the latest commit.

tests/utils/test_split_large_files.py
jrbourbeau: Just added a comment with the in memory size

> we split by uncompressed file size or file size? I believe the former

Yeah, we use `pyarrow.Table(...).nbytes` to determine splitting / merging logic (so the in memory number of bytes in the table)

jrbourbeau:
Thanks for the feedback @praateekmahajan. I think I've addressed all your comments, but let me know if not

jrbourbeau:
@sarahyurick @praateekmahajan it looks like CI never ran here, so the auto-merge never happened. Maybe it needs a comment like this https://github.com/NVIDIA-NeMo/Curator/pull/1161#issuecomment-3366718585 ?

PR #1162, opened 2025-10-02
Increase timeout for CPU tests

sarahyurick:
CPU tests for all stages currently takes ~19 minutes.

### DIFF
--- a/tutorials/audio/fleurs/pipeline.yaml
+++ b/tutorials/audio/fleurs/pipeline.yaml
@@ -1,3 +1,13 @@
+defaults:
+  - _self_
+  - override hydra/job_logging: none
+  - override hydra/hydra_logging: none
+
+hydra:
+  run:
+    dir: .
+  output_subdir: null
+
 documentation: |
   FLEURS
   ######
--- a/tutorials/audio/fleurs/run.py
+++ b/tutorials/audio/fleurs/run.py
@@ -12,8 +12,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import sys
-
 import hydra
 from loguru import logger
 from omegaconf import DictConfig, OmegaConf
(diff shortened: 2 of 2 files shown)


ayushdg (approved):

.github/workflows/cicd-main.yml
```diff
@@ -156,12 +156,14 @@ jobs:
...
-        timeout-minutes: 20
+        timeout-minutes: 30
         run: |
           cd InternVideo && patch -p1 < ../external/intern_video2_multimodal.patch && cd .. && sed -i "/InternVideo/d" .gitignore
           uv sync --link-mode copy --locked --extra audio_cpu --extra text_cpu --extra video_cpu --group test
           uv add InternVideo/InternVideo2/multi_modality
-          uv run coverage run --branch --source=nemo_curator -m pytest -v tests/${{ matrix.folder }} -m "not gpu"
+          FOLDER="${{ matrix.folder }}"
+          FOLDER="${FOLDER/stages-/stages/}"
+          uv run coverage run --branch --source=nemo_curator -m pytest -v "tests/$FOLDER" -m "not gpu"
```
sarahyurick, line 166: "Upload artifacts" step complains if there is a slash in the folder name and experiencing weirdness when I tried to add a `replace` function. Editing it this way seems to work fine though.

.github/workflows/cicd-main.yml
```diff
@@ -156,12 +156,14 @@ jobs:
         with:
           python-version: ${{ matrix.python-version }}
       - name: Run tests ${{ matrix.folder }} (CPU)
-        timeout-minutes: 20
+        timeout-minutes: 30
```
thomasdhc, line 159: Do we still need to increase the timeout?

.github/workflows/cicd-main.yml
thomasdhc: Looks good!

.github/workflows/cicd-main.yml
sarahyurick: I would like to keep it conservative if that's okay. Text stages can still take ~16 minutes, and tests are still actively being added.

thomasdhc: Gotcha

thomasdhc (approved):

PR #1165, opened 2025-10-03
citation file

lbliii:


CITATION.cff
```diff
@@ -1,25 +1,63 @@
...
 authors:
   - family-names: Jennings
     given-names: Joseph
+    affiliation: "NVIDIA"
   - family-names: Patwary
     given-names: Mostofa
+    affiliation: "NVIDIA"
   - family-names: Subramanian
     given-names: Sandeep
+    affiliation: "NVIDIA"
```
ayushdg: They've left Nvidia now so I'm not sure what affiliation to give. They contributed while they were here.

CITATION.cff
```diff
@@ -1,25 +1,63 @@
...
   - family-names: Jawa
     given-names: Vibhu
+    affiliation: "NVIDIA"
   - family-names: Liu
     given-names: Jiwei
+    affiliation: "NVIDIA"
   - family-names: Wolf
     given-names: Ryan
+    affiliation: "NVIDIA"
```
ayushdg: Same for Ryan as well

lbliii: They were NVIDIA while contributing (signed off commits using nvidia email) so I think it's acceptable. I'm not sure what else we could put but open to ideas. Maybe general contributor? OSS?

ayushdg: No strong opinions. Not sure what the general convention is either. Maybe `NVIDIA (at the time of contribution)`?

Open to suggestions

PR #1173, opened 2025-10-07
Fix bug for running filters on empty batches

sarahyurick:
Depending on what operation(s) a `DocumentFilter` is computing, our current logic for handling empty data batches can fail with an error that looks like:
```
ValueError: Task DocumentBatch(task_id='file_group_3_processed_ThinkingOnFilter_NanoFilter', dataset_name='split', data=Empty DataFrame
```
This PR resolves that error.

Discovered and debugged while I was working on https://github.com/NVIDIA-NeMo/Curator/pull/1063, which has multiple aggressive filtering stages. This fix allows the entire pipeline to run as intended.

### DIFF
--- a/nemo_curator/stages/text/modules/score_filter.py
+++ b/nemo_curator/stages/text/modules/score_filter.py
@@ -64,8 +64,8 @@ class Score(ProcessingStage[DocumentBatch, DocumentBatch]):
 
     def ray_stage_spec(self) -> dict[str, Any]:
         requires_setup = any(
-            hasattr(score_fn, "load_model") or hasattr(score_fn, "load_tokenizer") 
-            for score_fn in self.score_fn 
+            hasattr(score_fn, "load_model") or hasattr(score_fn, "load_tokenizer")
+            for score_fn in self.score_fn
             if isinstance(score_fn, DocumentFilter)
         )
         return {"is_actor_stage": requires_setup}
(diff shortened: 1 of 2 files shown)


nemo_curator/stages/text/modules/score_filter.py
```diff
@@ -62,6 +62,15 @@ def inputs(self) -> tuple[list[str], list[str]]:
...
         return ["data"], self.text_field + self.score_field
 
+    def ray_stage_spec(self) -> dict[str, Any]:
+        for score_fn in self.score_fn:
+            if isinstance(score_fn, DocumentFilter):
+                if hasattr(score_fn, "load_model"):
+                    return {"is_actor_stage": True}
+                if hasattr(score_fn, "load_tokenizer"):
+                    return {"is_actor_stage": True}
+        return {}
```
praateekmahajan: I think when its not an actor_stage we need to be explicit that it's not otherwise it'll assume it is

https://github.com/NVIDIA-NeMo/Curator/blob/c70b31a0f42b59c48fae677c8b5c932d0c726af7/nemo_curator/backends/experimental/ray_data/adapter.py#L87C59-L87C90


```python
requires_setup = any(
    hasattr(score_fn, "load_model") or hasattr(score_fn, "load_tokenizer") 
    for score_fn in self.score_fn 
    if isinstance(score_fn, DocumentFilter)
)
return {"is_actor_stage" : requires_setup}
```

Also maybe add a simple test that just creates two scorefilters one where is_actor_stage is True and one where its false?

sarahyurick: Makes sense, updated!

PR #1181, opened 2025-10-15
Introduce ray status verification to get client

abhinavg4:
- Added functionality to wait for ray status to become active before moving forward

## Usage

```python
```
## Checklist

### DIFF
--- a/benchmarking/runner/ray_cluster.py
+++ b/benchmarking/runner/ray_cluster.py
@@ -22,6 +22,7 @@ from typing import Any
 
 import ray
 from loguru import logger
+from runner.utils import run_shm_size_check
 
 from nemo_curator.core.client import RayClient
 from nemo_curator.core.utils import check_ray_responsive
@@ -76,6 +77,7 @@ def setup_ray_cluster_and_env(  # noqa: PLR0913
 
         _ensure_ray_client_process_started(client, ray_client_start_timeout_s, ray_client_start_poll_interval_s)
         responsive = check_ray_responsive()
+        run_shm_size_check(human_readable=True)
         if not responsive:
             logger.info("Ray cluster did not become responsive in time, stopping client and retrying...")
             client.stop()


nemo_curator/core/utils.py
```diff
@@ -68,6 +71,47 @@ def _logger_custom_deserializer(
     return logger
 
 
+@tenacity.retry(
+    wait=tenacity.wait_fixed(1),
+    stop=tenacity.stop_after_delay(RAY_CLUSTER_START_VERIFICATION_TIMEOUT),
+    retry=tenacity.retry_if_result(lambda x: x is False),
+    reraise=True,
+)
```
sarahyurick: This is essentially operating like a `while True` and sleep right? Could we use it in this test https://github.com/NVIDIA-NeMo/Curator/blob/main/tests/core/test_get_ray_client.py#L84 ?

pyproject.toml
```diff
@@ -61,6 +61,7 @@ dependencies = [
     "pandas>=2.1.0",
     "pyarrow",
     "ray[default,data]>=2.49",
+    "tenacity",
```
sarahyurick: uv.lock needs updated?

thomasdhc (approved):
Dependency change LGTM

nemo_curator/core/utils.py
```diff
@@ -18,13 +18,16 @@
 from typing import TYPE_CHECKING
 
 import ray
+import tenacity
 from loguru import logger
+from ray._private.services import canonicalize_bootstrap_address, find_gcs_addresses
```
ayushdg: Nit: I'm slightly hesitant to use private APIs given that they might change. Is there a public API alternative we could use here?

If not, we can go with this for now, and possibly ask ray folks if there's a public equivalent that can be exposed.

abhinavg4: None that i know of. I agree that probably using the private API is not stable. We can always copy the code. Maybe we can ask the ray folks and depending upon their reply we can move the code or use public APIs. Resolving it for now.

nemo_curator/core/utils.py
abhinavg4: Probably yes. This is not a blocker, though, right? I can address in my followup PR to this. The multinode integration, since that will touch the tests a lot.

sarahyurick:
Hmm not sure why the tests would be failing, maybe a rerun will help?

pyproject.toml
abhinavg4: Done

abhinavg4:
And the test failing seems pretty irrelevant to this PR?

abhinavg4:
/ok to test [c4da331](https://github.com/NVIDIA-NeMo/Curator/pull/1181/commits/c4da331942e8e502277b1cbdd9616f83d440049d)

nemo_curator/core/utils.py
ayushdg: This is a valid concern. I'm curious if we can replace this by instead polling ray health check or ray status until we have a good result

nemo_curator/core/utils.py
```diff
@@ -25,12 +26,61 @@
...
+            else:
+                logger.debug("Ray cluster IS responsive")
+                responsive = True
+
+        except subprocess.CalledProcessError:
+            logger.debug("Ray cluster is not responsive ('ray status' command failed)")
+
+        except subprocess.TimeoutExpired:
+            logger.debug("Ray cluster is not responsive ('ray status' command timed out)")
+
```
rlratzel, line 74: a very common problem preventing the Ray client session from starting is that the object store size is unattainable, which is based on the configuration chosen and the amount of shared memory available on the system.  The original code included a call to `run_shm_size_check()` which printed the state of the shared memory on the system as the Ray cluster is starting, which helped tremendously in debugging.
However, the output of `run_shm_size_check` was a bit clunky, so I can understand why we wouldn't want it showing from the product.  
Can we somehow keep the information from `run_shm_size_check` either in this function as a debug-level log output, or by adding an explicit call to `run_shm_size_check` in benchmarking/runner/ray_cluster.py::setup_ray_cluster_and_env right after the call to `check_ray_responsize` so we don't lose that info for benchmarks?

nemo_curator/core/utils.py
```diff
@@ -25,12 +26,61 @@
...
+                stderr=subprocess.STDOUT,
+                text=True,
+                timeout=10,
+            )
+
+            # Clean stdout to remove any new lines and carriage returns
+            result.stdout = result.stdout.replace("\n", "").replace("\r", "")
+            if "No cluster status" in result.stdout or "Error" in result.stdout:
+                logger.debug("Ray cluster is not responsive ('No cluster status' returned or Error in output)")
+            elif "Found multiple active Ray instances" in result.stdout:
```
rlratzel, line 59: Nice!

abhinavg4: Thanks

rlratzel (approved):

nemo_curator/core/client.py
```diff
@@ -149,10 +154,11 @@ def start(self) -> None:
...
             )
             # Set environment variable for RAY_ADDRESS
-
             os.environ["RAY_ADDRESS"] = f"{ip_address}:{self.ray_port}"
-            # Register atexit handler only when we have a ray process
-            atexit.register(self.stop)
+            # Verify that Ray cluster actually started successfully
+            if not check_ray_responsive():
+                msg = "Ray cluster did not become responsive in time. Please check the logs for more information."
+                raise RuntimeError(msg)
```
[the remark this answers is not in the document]
abhinavg4: atexit will stop it

nemo_curator/core/client.py
```diff
@@ -88,6 +89,10 @@ def __post_init__(self) -> None:
 
     def start(self) -> None:
         """Start the Ray cluster if not already started, optionally capturing stdout/stderr to a file."""
+
+        # register atexit handler to stop the Ray cluster when the program exits
+        atexit.register(self.stop)
```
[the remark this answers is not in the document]
abhinavg4: It is an overhead but that's ok since we wanna make sure the ray cluister is stopped at any cost

PR #1197, opened 2025-10-23
Adds initial benchmarking framework

rlratzel:
Initial version of Curator benchmarking framework, based off of [PR1011](https://github.com/NVIDIA-NeMo/Curator/pull/1011).

### DIFF
--- a/benchmarking/tools/gen_runscript_vars.py
+++ b/benchmarking/tools/gen_runscript_vars.py
@@ -41,7 +41,7 @@ VOLUME_MOUNTS = []
 
 
 def print_help(script_name: str) -> None:
-    """Print usage and help message for the run script (not this script)to stderr."""
+    """Print usage and help message for the run script (not this script) to stderr."""
     sys.stderr.write(f"""
   Usage: {script_name} [OPTIONS] [ARGS ...]
 


docker/common/install_ffmpeg.sh
```diff
@@ -94,4 +94,4 @@ ldconfig
 cd /
 rm -rf /tmp/ffmpeg*
 rm -rf /tmp/nv-codec-headers*
-rm -rf /var/lib/apt/lists/
```
praateekmahajan, line 97: What's the reason for this?
Just because we're touching the common dockerfile, I'd want to be more cautious and let @thomasdhc also take a look

docker/Dockerfile
```diff
@@ -40,8 +40,7 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
     python-is-python3\
```
praateekmahajan, line 40: @thomasdhc can you TAL., the changes here seem inconsequential

benchmarking/run.py
```diff
@@ -0,0 +1,263 @@
...
+def ensure_dir(dir_path: Path) -> None:
+    """Ensure dir_path and parents exists, creating them if necessary."""
+    dir_path.mkdir(parents=True, exist_ok=True)
+
+
+def create_or_overwrite_dir(dir_path: Path) -> None:
+    """Create directory, removing it if it exists."""
+    if dir_path.exists():
+        shutil.rmtree(dir_path, ignore_errors=True)
+    dir_path.mkdir(parents=True, exist_ok=True)
```
praateekmahajan: you can change this PR or add a todo, but this can just be `create_or_overwrite_dir` from `nemo_curator.utils.file_utils`

benchmarking/run.py
```diff
@@ -0,0 +1,263 @@
...
+    # For each of the metric compute mean/std/sum and flatten the dict
+    for stage_name, stage_data in tasks_metrics.items():
+        for metric_name, values in stage_data.items():
+            for agg_name, agg_func in [("sum", sum), ("mean", mean), ("std", stdev)]:
+                stage_key = stage_name if prefix is None else f"{prefix}_{stage_name}"
+                if len(values) > 0:
+                    metrics[f"{stage_key}_{metric_name}_{agg_name}"] = float(agg_func(values))
+                else:
+                    metrics[f"{stage_key}_{metric_name}_{agg_name}"] = 0.0
+    return metrics
```
praateekmahajan: Similar future ticket, we can just move this to `nemo_curator.tasks.utils`

benchmarking/Dockerfile
```diff
@@ -0,0 +1,143 @@
...
+    rm install_ffmpeg.sh
+
+ARG INTERN_VIDEO_COMMIT=09d872e5093296c6f36b8b3a91fc511b76433bf7
+COPY external/intern_video2_multimodal.patch .
+# Clone InternVideo (Video curation dependency)
+RUN git clone https://github.com/OpenGVLab/InternVideo.git && \
+    cd InternVideo && \
+    git checkout ${INTERN_VIDEO_COMMIT} && \
+    patch -p1 < /opt/intern_video2_multimodal.patch && \
+    rm /opt/intern_video2_multimodal.patch
```
praateekmahajan: nit, but is there a way to use this stage or the base from the common Dockerfile.. just to make sure our dockerfile does atleast what the common one does and more.. 
primarily because, i worry things will get changed there and we won't be in sync

benchmarking/Dockerfile
```diff
@@ -0,0 +1,143 @@
...
+COPY --from=curator_setup_deps /opt /opt
+
+# Install Curator, which includes benchmarking tools
+COPY . /opt/Curator
+RUN cd /opt/Curator \
+ && uv sync --link-mode copy --locked --extra all --all-groups \
+ && uv add /opt/InternVideo/InternVideo2/multi_modality \
+ && uv add rich \
+ && uv cache prune \
+ && uv cache clean distance
```
praateekmahajan: Similarly here

benchmarking/Dockerfile
```diff
@@ -0,0 +1,143 @@
...
+    python3-venv \
+    python3.12-dev \
+    vim
+WORKDIR /opt
+
+# Install micromamba
+ENV MAMBA_ROOT_PREFIX=/opt/micromamba
+ENV PATH=$MAMBA_ROOT_PREFIX/bin:$PATH
+RUN curl -Ls https://micromamba.snakepit.net/api/micromamba/linux-64/latest | tar -xvj -C /usr/local/bin --strip-components=1 bin/micromamba && \
+    micromamba shell init -s bash -r $MAMBA_ROOT_PREFIX
```
praateekmahajan: do we need mamba?

benchmarking/Dockerfile
```diff
@@ -0,0 +1,143 @@
...
+    curl \
+    git \
+    less \
+    openssh-client \
+    python-is-python3 \
+    python3 \
+    python3-dev \
+    python3-pip \
+    python3-venv \
+    python3.12-dev \
```
praateekmahajan: I could be wrong but I think if you have `uv` you don't need python as base..


we can do
```python
uv venv --python 3.12 --seed
source .venv/bin/activate
```
which will get the python + pip + setuptools and activate that environment

benchmarking/tools/build_docker.sh
```diff
@@ -0,0 +1,32 @@
...
+CURATOR_DIR="$(cd ${THIS_SCRIPT_DIR}/../.. && pwd)"
+
+# Build the base image
+#docker build -f ${CURATOR_DIR}/benchmarking/Dockerfile --target curator_system_base --tag=curator_system_base ${CURATOR_DIR}
+
+# Build the deps image
+#docker build -f ${CURATOR_DIR}/benchmarking/Dockerfile --target curator_setup_deps --tag=curator_setup_deps ${CURATOR_DIR}
+
+# Build the benchmarking image
+docker build -f ${CURATOR_DIR}/benchmarking/Dockerfile --target curator_benchmarking --tag=curator_benchmarking ${CURATOR_DIR}
```
praateekmahajan: maybe we can add a date as a tag s.t. if a machine has many its easy to resolve

benchmarking/runner/env_capture.py
```diff
@@ -0,0 +1,71 @@
...
+        env_data["pip_freeze_txt"] = freeze_txt_path
+    except Exception as e:  # noqa: BLE001
+        logger.warning(f"Failed to capture pip freeze: {e}")
+    try:
+        # Try micromamba first, then conda as fallback
+        cmd = None
+        if shutil.which("micromamba"):
+            cmd = ["micromamba", "list", "--explicit"]
+        elif shutil.which("conda"):
+            cmd = ["conda", "list", "--explicit"]
```
praateekmahajan: if we have uv i think we won't need this..

i believe pip freeze might be failing because uv is installed without `--seed`

benchmarking/runner/env_capture.py
```diff
@@ -0,0 +1,71 @@
...
+    (output_path / "env.json").write_text(json.dumps(get_obj_for_json(env_data)))
+    return env_data
+
+
+def get_env() -> dict[str, Any]:
+    return {
+        "hostname": platform.node(),
+        "platform": platform.platform(),
+        "ray_version": os.getenv("RAY_VERSION", "unknown"),
+        "git_commit": os.getenv("GIT_COMMIT", "unknown"),
```
praateekmahajan: would these by env variables or some python / git command?

benchmarking/runner/env_capture.py
```diff
@@ -0,0 +1,71 @@
...
+        "image_digest": os.getenv("IMAGE_DIGEST", "unknown"),
```
praateekmahajan: what about this?

benchmarking/runner/env_capture.py
```diff
@@ -0,0 +1,71 @@
...
+        "python_version": platform.python_version(),
+        "executable": os.getenv("_"),
+        "cuda_visible_devices": os.getenv("CUDA_VISIBLE_DEVICES", ""),
```
praateekmahajan: i think we can replace this with `4 x A100s` or 8 x H100s`

docker/common/install_ffmpeg.sh
rlratzel: It's a minor optimization where I keep the `apt` cache intact until the very end so we only have to call `apt update` once upfront. It's minor though and we can continue to remove the cache if it makes it easier to be assured we're not including the cache in our image.

benchmarking/run.py
rlratzel: Oh thanks, I didn't know about that. Yes I'd rather reuse utils from curator if they're available.

benchmarking/Dockerfile
rlratzel: That's a great point. Yeah we should try to consolidate more so we're benchmarking as much of production expectations as possible. 
Summary from offline conversation: we discussed doing a `FROM` of the full production Dockerfile and then updating as needed for benchmarking using one or more `RUN`s. Or, look into adding a reasonable common stage to the production Dockerfile and `FROM` that, then layer the benchmarking additions on top of that. They both have pros and cons, but I'll work on one of those.

benchmarking/Dockerfile
rlratzel: Oh, possibly not. That was from a prior version but it doesn't look like we need it.  I'll remove it and test.

benchmarking/runner/env_capture.py
rlratzel: Okay thanks, I'll take a look.

benchmarking/runner/env_capture.py
rlratzel: I agree that these would be better provided by a git command or similar.  The env vars were leftover from a prior version so I'll update this.

benchmarking/runner/env_capture.py
rlratzel: Ah yes, same issue as above. I wonder if the previous version set this env var as part of the Docker build? (if so, I didn't notice that)  I'll look into how to get this since it seems useful.

benchmarking/README.md
```diff
@@ -0,0 +1,1011 @@
...
+
+### Example 1: Simple Benchmark
+
+**config.yaml:**
+
+```yaml
+results_dir: "./results"
+entries:
+  - name: quick_test
+    script: test_benchmark.py
```
jnke2016: I think there is a mismatch between the between the argument passed to `test_benchmark.py` in this README.md and the arguments required in the script. 
```
parser.add_argument("--input-path", required=True, help="Path to input data")
parser.add_argument("--output-path", required=True, help="Path to output data")
```

In fact, `test_benchmark.py` requires the above arguments. `--benchmark-results-path` is required in the script however, the framework automatically appends it in matrix.py (L44:L47) and this is why you have a FIXME there I believe.

benchmarking/run.py
```diff
@@ -0,0 +1,245 @@
...
+        "--session-name", default=None, help="Optional human-readable session name (default nightly-run-<timestamp>)"
+    )
+    args = parser.parse_args()
+
+    # Consolidate the configuration from all YAML files into a single dict,
+    # and use by passing individual components the keys they need
+    config_dict = {}
+    for yml_file in args.config:
+        with open(yml_file) as f:
+            config_dicts = yaml.full_load_all(f)
```
jnke2016, line 250:
```
config_dicts = list(yaml.full_load_all(f))  # Force generator to read NOW
```
I believe you need to consume the generator inside the block. Otherwise it may not have read the file yet.

benchmarking/README.md
```diff
@@ -0,0 +1,1011 @@
...
+---
+
+## Examples
+
+### Example 1: Simple Benchmark
+
+**config.yaml:**
+
+```yaml
+results_dir: "./results"
```
jnke2016:
```
datasets: [] # Datasets (required even if empty)
```
Otherwise you will get a KeyError `resolver = DatasetResolver.create_from_dicts(config_dict["datasets"])` in `run.py`

benchmarking/README.md
```diff
@@ -0,0 +1,1011 @@
...
+datasets:
+  - name: sample_data
+    formats:
+      - type: parquet
+        path: /data/sample.parquet
+
+entries:
+  - name: benchmark_v1
+    script: my_benchmark.py
```
jnke2016: `my_benchmark.py` is not included in this repo so it might be odd to have `test_bechmark.py` in Example 1 and a different benchmark script which doesn't exist in Example 2. Maybe you meant `test_benchmark.py` ?

benchmarking/README.md
```diff
@@ -0,0 +1,1011 @@
...
+    args: --input {dataset:sample_data,parquet} --algorithm v1
+
+  - name: benchmark_v2
+    script: my_benchmark.py
+    args: --input {dataset:sample_data,parquet} --algorithm v2
+```
```
jnke2016:
```bash
python benchmarking/run.py --config config.yaml
```

You forgot the above to be consistent with Example 1

benchmarking/run.py
```diff
@@ -0,0 +1,245 @@
...
+        for d in config_dicts:
+            config_dict.update(d)
```
jnke2016: `if d:  # Skip None documents`

Maybe also add a preventive check ensuring that the next line doesn't get executed if the YAML documents are empty before updating the dictionary

benchmarking/README.md
```diff
@@ -0,0 +1,1011 @@
...
+    enabled: true
+  - name: slack
+    webhook_url: ${SLACK_WEBHOOK_URL}
+    enabled: true
+
+datasets:
+  - name: cc_large
+    formats:
+      - type: parquet
+        path: /data/common_crawl_large.parquet
```
jnke2016: I assume the user is expected to provide a path to a `.parquet` file ?

benchmarking/README.md
```diff
@@ -0,0 +1,1011 @@
...
+
+entries:
+  - name: cc_extraction
+    script: common_crawl_benchmark.py
```
jnke2016: Is the user expected to create its own `common_crawl_benchmark.py` or should it be `test_benchmark.py` ?

benchmarking/runner/env_capture.py
rlratzel: Removed, thanks.

benchmarking/Dockerfile
rlratzel: Removed, thanks

benchmarking/run.py
rlratzel: Yes, thanks, should be fixed in the latest commit.

benchmarking/README.md
rlratzel: Thanks, I fixed it so it can tolerate a missing `datasets` block in the config instead.

docker/common/install_ffmpeg.sh
rlratzel: I reverted this as part of the update to use `docker/Dockerfile` as-is then add the benchmark dependencies/entrypoint in a separate Dockerfile for benchmarking.

docker/Dockerfile
rlratzel: I reverted this as part of the update to use `docker/Dockerfile` as-is then add the benchmark dependencies/entrypoint in a separate Dockerfile for benchmarking.

benchmarking/Dockerfile
rlratzel: I ended up reverting `docker/Dockerfile` completely and then updating `benchmarking/Dockerfile` to `FROM` `docker/Dockerfile` and add only the dependencies needed for benchmarking to ensure updates the team makes to `docker/Dockerfile` are used in benchmarks.

If we realize we also need to remove things from the benchmarking container that were added to `docker/Dockerfile`, then we can look into additional `RUN` statements in `benchmarking/Dockerfile` to remove them and/or adding new shared stages to `docker/Dockerfile` to keep things separate.

benchmarking/Dockerfile
rlratzel: The latest version only adds benchmarking dependencies so all changes to the `uv` install steps from `docker/Dockerfile` will be used in the benchmarking image as-is.

benchmarking/Dockerfile
rlratzel: These lines are gone in the latest `benchmarking/Dockerfile`

benchmarking/tools/build_docker.sh
rlratzel: Thanks, good idea. Done.

benchmarking/runner/env_capture.py
rlratzel: I ended up setting this env var via the `tools/run.sh` script since I didn't know of a better way to find it interactively from within the container, or even at docker build time.

benchmarking/README.md
```diff
@@ -0,0 +1,993 @@
...
+  - name: test_benchmark
+    script: test_benchmark.py
+    args: --iterations 100
+    timeout_s: 300
+```
+
+**3. Run the benchmark:**
+
+```bash
+./benchmarking/tools/run.sh
```
praateekmahajan: do we need an arg for providing `my_benchmark.yaml`? or does it just find all yaml files and load them? what if we have many yaml files in our directory, one for adhoc testing and one for nightly?

benchmarking/README.md
```diff
@@ -0,0 +1,993 @@
...
+
+**Setup:**
+
+```bash
+# Install NeMo Curator with all dependencies
+cd <curator_repo_root>
+uv sync --extra all --all-groups
+
+# Install additional benchmarking dependencies
+uv add rich loguru
```
praateekmahajan: loguru is already a dependency in curator so we might get that for free

benchmarking/README.md
```diff
@@ -0,0 +1,993 @@
...
+```
+
+Files are merged in order. Later files override earlier ones for conflicting keys.
+
+**Session naming:**
+
+```bash
+python benchmarking/run.py \
+  --config config.yaml \
+  --session-name my-experiment-v2
```
praateekmahajan, line 218: where does session-name get used? is it that when not provided then the default is `nightly-run__2025-01-23__14-30-00`?
wdyt should each session name still be suffixed by timestamp?

i.e. in this case `my-experiment-v2__2025-01-23__14-30-00` to avoid unintentional overwriting?

benchmarking/README.md
```diff
@@ -0,0 +1,993 @@
...
+from nemo_curator.tasks.utils import TaskPerfUtils
+
+
+def run_benchmark(args):
+    """Main benchmark logic."""
+    start_time = time.time()
+
+    # Your benchmark code here
+    with Task("my_operation", TaskPerfUtils()):
+        result = perform_operation(args.input)
```
praateekmahajan: i think its incorrect syntax here.. 
```suggestion
        output_tasks = run_pipeline(executor)
```

benchmarking/README.md
```diff
@@ -0,0 +1,993 @@
...
+    metrics = {
+        "execution_time_s": execution_time,
+        "items_processed": len(result),
+    }
+    with open(args.benchmark_results_path / "metrics.json", "w") as f:
+        json.dump(metrics, f, indent=2)
+
+    tasks = Task.get_all_tasks()
+    with open(args.benchmark_results_path / "tasks.pkl", "wb") as f:
+        pickle.dump(tasks, f)
```
praateekmahajan: same here
```suggestion
    with open(args.benchmark_results_path / "tasks.pkl", "wb") as f:
        pickle.dump(output_tasks, f)
```

benchmarking/README.md
rlratzel: I just updated the help text to describe that `--config` is required since running without that option is not supported (the README was out-of-date).

benchmarking/README.md
rlratzel: I removed that from the latest docs.

benchmarking/README.md
rlratzel: This should be fixed in the latest docs.

benchmarking/README.md
rlratzel: Many of those were just example names and not actual scripts in the repo, but I've updated the docs to not include those examples anymore.

benchmarking/README.md
rlratzel: Removed in the latest docs.

benchmarking/README.md
rlratzel: The latest docs should describe this better.

benchmarking/README.md
rlratzel: Removed in latest.

benchmarking/README.md
rlratzel: Yeah that was a mistake and that example has been removed in the latest docs.

benchmarking/README.md
rlratzel: The latest docs removes this example as well.

benchmarking/README.md
rlratzel: It gets used in `run.py` [here](https://github.com/NVIDIA-NeMo/Curator/blob/d26b5ca9637eed4782d6768d09baefeae08e1e42/benchmarking/run.py#L261). If it's not provided, the default will look something like this: `benchmark-run__2025-11-06__02-07-46`, so fairly generic and unique by default.

I was assuming the user wanted/needed a specific name if they passed it in, but maybe we should unconditionally append a timestamp? Although I was thinking there could be a case where a very specific name is needed (eg. additional automation, like maybe part of a CI runner, is running things and needs to find the results afterwards?) so I was afraid of making that difficult and assumed they would ensure a unique name if needed.

We could add another arg for the case of needing a meaningful name but still wanting it unique by default called `--session-name-prefix`. Do you think that would be useful, or should we just always append a timestamp?

benchmarking/README.md
```diff
@@ -0,0 +1,568 @@
...
+
+Built-in sinks include:
+- **Slack**: Post results to Slack channels
+- **Google Drive**: Upload results to cloud storage (extensible)
+- **MLflow**: Track experiments and metrics
+
+See [Sinks: Custom Reporting & Actions](#sinks-custom-reporting--actions) for details.
+
+### Entry
+
```
praateekmahajan: nit : move this above sinks? as most users won't play around with sinks but will play with entries

benchmarking/README.md
```diff
@@ -0,0 +1,568 @@
...
+Results are written to the `results_path` specified in your configuration, organized by session timestamp.
+
+---
+
+## Concepts
+
+### Session
+
+A **session** represents a single invocation of the benchmarking framework. Each session:
+- Has a unique name with timestamp (e.g., `nightly-run__2025-01-23__14-30-00`)
```
praateekmahajan: i think default is `benchmark-run__2025-01-23__14-30-00`

benchmarking/README.md
```diff
@@ -0,0 +1,568 @@
...
+- Can reference datasets using template syntax
+- Can pass additional data to sinks to provide for customized operations unique to the entry. For example, the `slack_sink` can accept additional metrics to report for an entry that other entries may not have.
+- Can specify specific requirements that must be met in order to return a passing status. For example, an entry can require that a specific throughput metric meet or exceed a minimum value.
+
+## Configuration
+
+### YAML Configuration Files
+
+The framework uses one or more YAML files to configure benchmark sessions. Multiple configuration files are merged, allowing separation of concerns (e.g., machine-specific paths vs. benchmark definitions).
+
```
praateekmahajan, line 115: IMO we should add a line and also show in the example that we use two different yamls, one for "things that change" and one for "things that are static"

things that change = machine specific = results / artifacts / datasets path + slack env variable / sinks

things that are static = entries.yaml

benchmarking/README.md
```diff
@@ -0,0 +1,568 @@
...
+    enabled: false
+    drive_folder_id: ${GDRIVE_FOLDER_ID}
+    service_account_file: ${GDRIVE_SERVICE_ACCOUNT_FILE}
+
+# Optional: Define datasets for template substitution
+datasets:
+  - name: common_crawl
+    formats:
+      - type: json
+        path: "{datasets_path}/cc_sample.jsonl"  # Can reference base paths
```
praateekmahajan:
```suggestion
        path: "{datasets_path}/cc_jsonl/"  # Can reference base paths
```
(since usually these are directories)

benchmarking/README.md
```diff
@@ -0,0 +1,568 @@
...
+      - type: parquet
+        path: "{datasets_path}/cc_sample.parquet"
```
praateekmahajan:
```suggestion
        path: "{datasets_path}/cc_parquet"
```

benchmarking/README.md
praateekmahajan: discussed offline, can add in followup

benchmarking/README.md
```diff
@@ -0,0 +1,568 @@
...
+args: --input {dataset:common_crawl,parquet}
+```
+
+Resolves to the path defined in the `datasets` section for that dataset and format.
+
+**Session entry directory** - Reference the entry's runtime directory:
+
+```yaml
+args: --output {session_entry_dir}/results
+```
```
praateekmahajan, line 265: after reading this i'm slightly confused with where `artifacts_path` is used? is the base of`session_entry_dir`, `artifacts_path`?
what does `delete_scratch : True` do? can we elaborate on those?

benchmarking/run.py
```diff
@@ -0,0 +1,315 @@
...
+    for metric_name, requirement_dict in requirements.items():
+        reason_not_met = None
+        actual_value = find_result(result_data, metric_name)
+        if actual_value is None:
+            reason_not_met = f"{metric_name} not found in metrics"
+        elif "min_value" in requirement_dict:
+            min_value = requirement_dict["min_value"]
+            if actual_value < min_value:
+                reason_not_met = f"{metric_name} < {min_value}"
+        elif "max_value" in requirement_dict:
```
praateekmahajan: i think if both `min_value` and `max_value` are provided then the code might not work

benchmarking/run.py
```diff
@@ -0,0 +1,315 @@
...
+    entry: MatrixEntry,
+    path_resolver: PathResolver,
+    dataset_resolver: DatasetResolver,
+    session_path: Path,
+    result_data: dict[str, Any],
+) -> bool:
+    session_entry_path = session_path / entry.name
+
+    # scratch_path : This is the directory user can use to store scratch data; it'll be cleaned up after the entry is done
+    # ray_cluster_path : This is the directory where Ray cluster is started; it'll be cleaned up after the entry is done
```
praateekmahajan: this isn't true, since at `stop_ray_node` we copy over the data to this location

benchmarking/README.md
praateekmahajan: now that i think of it, maybe we can have `session_entry_scratch_dir` as a variable too so that users don't have to suffix with `/scratch` each time

benchmarking/run.py
```diff
@@ -0,0 +1,315 @@
...
+        # Execute command with timeout
+        logger.info(f"\t\tRunning command {' '.join(cmd) if isinstance(cmd, list) else cmd}")
+        started_exec = time.time()
+        run_data = run_command_with_timeout(
+            command=cmd,
+            timeout=entry.timeout_s,
+            stdouterr_path=logs_path / "stdouterr.log",
+            env=ray_env,
+            run_id=run_id,
+            fancy=os.environ.get("CURATOR_BENCHMARKING_DEBUG", "0") == "0",
```
praateekmahajan, line 173: i like fancy!!

benchmarking/README.md
rlratzel: good point, done.

benchmarking/README.md
rlratzel: thanks, updated.

benchmarking/README.md
rlratzel: That's a great idea.  I made an attempt at describing this better in the README but I may have been too verbose. I'll try improving it for clarity in a followup PR.

benchmarking/run.py
rlratzel: Thanks, I think I fixed it in the latest commit.

benchmarking/run.py
rlratzel: Thanks. Fixed in latest commit.

benchmarking/README.md
rlratzel: I believe `artifacts_path` isn't used at the moment, I think when I copied over some code from PR1011, I moved the files that were in artifacts (packages.txt files) to the results dir instead.  I can change that back in the followup PR if that's okay.
There's some very brief text about `delete_scratch` in the README [here](https://github.com/NVIDIA-NeMo/Curator/blob/9d7d34ef020ab3c6122d276e1dd5b3770ea33267/benchmarking/README.md?plain=1#L130) but I can elaborate more in the followup PR too.

Also, I can certainly add support for a `session_entry_scratch_dir` template var too.  I haven't tried this yet, but since it's YAML, _I think_ users should also be able to use standard YAML anchors and references to create their own convenience vars if that's preferred.

benchmarking/runner/sinks/__init__.py
```diff
@@ -0,0 +1 @@
+
```
sarahyurick:
```suggestion
```

benchmarking/scripts/common_crawl_benchmark.py
```diff
@@ -0,0 +1,205 @@
...
+        executor = RayDataExecutor()
+    elif args.executor == "ray_actors":
+        from nemo_curator.backends.experimental.ray_actor_pool.executor import RayActorPoolExecutor
+
+        executor = RayActorPoolExecutor()
+    else:
+        msg = f"Invalid executor type: {args.executor}"
+        raise ValueError(msg)
+
+    initial_task = _EmptyTask(task_id="common_crawl_task", dataset_name="common_crawl", data=None)
```
sarahyurick: I was wondering if there is any reason to add this? The default `initial_task` is already an empty task.

benchmarking/scripts/domain_classification_benchmark.py
```diff
@@ -0,0 +1,210 @@
+#!/usr/bin/env python3
```
sarahyurick:
```suggestion
```

benchmarking/scripts/embedding_generation_benchmark.py
```diff
@@ -0,0 +1,228 @@
+#!/usr/bin/env python3
```
sarahyurick:
```suggestion
```

benchmarking/scripts/removal_benchmark.py
```diff
@@ -0,0 +1,246 @@
+#!/usr/bin/env python3
```
sarahyurick:
```suggestion
```

benchmarking/tools/gen_runscript_vars.py
```diff
@@ -0,0 +1,183 @@
+#!/bin/env python
```
sarahyurick:
```suggestion
```

benchmarking/tools/gen_runscript_vars.py
```diff
@@ -0,0 +1,183 @@
...
+HOST_CURATOR_DIR = os.environ.get("HOST_CURATOR_DIR", str(this_script_path.parent.parent.absolute()))
+CURATOR_BENCHMARKING_DEBUG = os.environ.get("CURATOR_BENCHMARKING_DEBUG", "0")
+
+BASH_ENTRYPOINT_OVERRIDE = ""
+ENTRYPOINT_ARGS = []
+VOLUME_MOUNTS = []
+
+
+def print_help(script_name: str) -> None:
+    """Print usage and help message for the run script (not this script)to stderr."""
```
sarahyurick:
```suggestion
    """Print usage and help message for the run script (not this script) to stderr."""
```

PR #1209, opened 2025-10-29
Add ability to ignore_head_node for RayDataExecutor and RayActorPoolExecutor

praateekmahajan:
### Add ignore_head_node to RayDataExecutor and RayActorPoolExecutor

This is done to ensure s.t. on environments like Anyscale where head node can not be scheduled on, we explicitly ignore it. Behind the scenes it does two things
1. Ensure `setup_on_node` is not executed on `head_node`
2. Remove the resources of head node from `get_available_cpu_gpu_resources`
   - For RayActorPoolExecutor : For each stage, this influences how many Actors we should start.
   - For RayDataExecutor : For each stage, this influences the concurrency i.e. `map_batches(concurrency=(min, max))`

#### Con
This _might_ work but one concern I have is
- If we have a 3 node machine, each node having 4 CPUs, so 12 CPUs in total where 4 are from head node 
- We don't want to schedule stages on head node, so `setup_on_node` correctly ignores that node
- But I think the execution of individual stages might not ignore that node (and we might run into runtime error since setup_on_node isn't called)
    - For RayActorPool, if we have a stage which needs one CPU, and we start 8 actors, there isn't a gurantee that it won't schedule on head node (i.e. our calculation just says start 8 actors, not where to start the actors). Does anyscale ensure that headnode is truly not executable?
    - Similarly for RayData, when we set `concurrency=(1, 8)` how does it ensure that it doesn't chose the head node for scheduling.

In simpler words, because for these two executors we don't decide "where" scheduling happens, just the "how many". Does Anyscale restrict any scheduling?
Maybe the way to "solve" this is that users using Curator on Anyscale, must set num_cpus on head_node to 0. 

If we need to solve the "where" then we'll need a bigger refactor in both our executors to possibly pass some NodeAffinity to ensure head_node is not there.

### Adapt dedup workflows to accept executors
All of exact / fuzzy / semantic dedup where `.run(...)` accepts executor config.
For SemanticDedup specifically if a customer wants to run then they can do

```python
# SemanticDedup
workflow = TextSemanticDeduplicationWorkflow(...)

workflow.run(
    streaming_executor=RayDataExecutor(ignore_head_node=True), 
    batch_executor=RayActorPoolExecutor(ignore_head_node=True)
)

# Exact
workflow = ExactDeduplicationWorkflow(...)
workflow.run(executor=RayActorPoolExecutor(ignore_head_node=True))

# Fuzzy
workflow = FuzzyDeduplicationWorkflow()
workflow.run(executor=RayActorPoolExecutor(ignore_head_node=True))

```

### Future Work
1. Make changes in Xenna to allow RayData / RayActorPool
2. Adapt those changes in Curator (keeping the API consistent)
3. Document how to run on Anyscale
4. Have some QA to ensure Anyscale works for us consistently
5. Follow up on RayData performance on Curator

## Usage

```python
```
## Checklist

### DIFF
--- a/benchmarking/runner/ray_cluster.py
+++ b/benchmarking/runner/ray_cluster.py
@@ -90,35 +90,8 @@ def start_ray_head(
         enable_object_spilling=enable_object_spilling,
         ray_dashboard_host="0.0.0.0",  # noqa: S104
     )
-    # Redirect Ray startup output to log file if provided, otherwise suppress it
-    import sys
-
-    if ray_log_path:
-        with open(ray_log_path, "w") as f:
-            # Save original stdout/stderr
-            original_stdout = sys.stdout
-            original_stderr = sys.stderr
-            try:
-                # Redirect to log file
-                sys.stdout = f
-                sys.stderr = f
-                client.start()
-            finally:
-                # Restore original stdout/stderr
-                sys.stdout = original_stdout
-                sys.stderr = original_stderr
-    else:
-        # Suppress Ray startup output by redirecting to devnull
-        with open(os.devnull, "w") as devnull:
-            original_stdout = sys.stdout
-            original_stderr = sys.stderr
-            try:
-                sys.stdout = devnull
-                sys.stderr = devnull
-                client.start()
-            finally:
-                sys.stdout = original_stdout
-                sys.stderr = original_stderr
+    ray_stdouterr_capture_file = str(ray_log_path) if ray_log_path else os.devnull
+    client.start(stdouterr_capture_file=ray_stdouterr_capture_file)
     # Wait for Ray client to start, no longer than timeout
     wait_for_ray_client_start(client, ray_client_start_timeout_s, ray_client_start_poll_interval_s)
     logger.debug(f"RayClient started successfully: pid={client.ray_process.pid}, port={client.ray_port}")
--- a/nemo_curator/core/client.py
+++ b/nemo_curator/core/client.py
@@ -16,7 +16,7 @@ import atexit
 import os
 import socket
 import subprocess
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 
 from loguru import logger
 
@@ -73,10 +73,17 @@ class RayClient:
     num_gpus: int | None = None
     num_cpus: int | None = None
     enable_object_spilling: bool = False
+    ray_stdouterr_capture_file: str | None = None
 
-    ray_process: subprocess.Popen | None = None
+    ray_process: subprocess.Popen | None = field(init=False)
+
+    def __post_init__(self) -> None:
+        if self.ray_stdouterr_capture_file and os.path.exists(self.ray_stdouterr_capture_file):
+            msg = f"Capture file {self.ray_stdouterr_capture_file} already exists."
+            raise FileExistsError(msg)
 
     def start(self) -> None:
+        """Start the Ray cluster, optionally capturing stdout/stderr to a file."""
         if self.include_dashboard:
             # Add Ray metrics service discovery to existing Prometheus configuration
             if is_prometheus_running() and is_grafana_running():
@@ -110,6 +117,7 @@ class RayClient:
                 get_next_free_port=(self.ray_client_server_port == DEFAULT_RAY_CLIENT_SERVER_PORT),
             )
             ip_address = socket.gethostbyname(socket.gethostname())
+
             self.ray_process = init_cluster(
                 self.ray_port,
                 self.ray_temp_dir,
@@ -122,6 +130,7 @@ class RayClient:
                 self.enable_object_spilling,
                 block=True,
                 ip_address=ip_address,
+                stdouterr_capture_file=self.ray_stdouterr_capture_file,
             )
             # Set environment variable for RAY_ADDRESS
 
--- a/nemo_curator/core/utils.py
+++ b/nemo_curator/core/utils.py
@@ -80,6 +80,7 @@ def init_cluster(  # noqa: PLR0913
     enable_object_spilling: bool = False,
     block: bool = True,
     ip_address: str | None = None,
+    stdouterr_capture_file: str | None = None,
 ) -> subprocess.Popen:
     """Initialize a new local Ray cluster or connects to an existing one."""
     # Turn off serization for loguru. This is needed as loguru is not serializable in general.
@@ -121,6 +122,10 @@ def init_cluster(  # noqa: PLR0913
     os.environ["XENNA_RAY_METRICS_PORT"] = str(ray_metrics_port)
     os.environ["XENNA_RESPECT_CUDA_VISIBLE_DEVICES"] = "1"
 
-    proc = subprocess.Popen(ray_command, shell=False)  # noqa: S603
+    if stdouterr_capture_file:
+        with open(stdouterr_capture_file, "w") as f:
+            proc = subprocess.Popen(ray_command, shell=False, stdout=f, stderr=subprocess.STDOUT)  # noqa: S603
+    else:
+        proc = subprocess.Popen(ray_command, shell=False)  # noqa: S603
     logger.info(f"Ray start command: {' '.join(ray_command)}")
     return proc
--- a/tests/core/test_get_ray_client.py
+++ b/tests/core/test_get_ray_client.py
@@ -12,6 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# ruff: noqa: ARG001
+
 import os
 import tempfile
 import time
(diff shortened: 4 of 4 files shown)


nemo_curator/backends/experimental/utils.py
```diff
@@ -38,15 +62,30 @@ def get_worker_metadata_and_node_id() -> tuple[NodeInfo, WorkerMetadata]:
...
+        head_node_id = get_head_node_id()
+        if head_node_id is not None:
+            total_resources = ray.state.total_resources_per_node().get(head_node_id, {})
+            head_node_cpus = total_resources.get("CPU", 0)
+            head_node_gpus = total_resources.get("GPU", 0)
+            logger.info(
+                f"Ignoring head node {head_node_id} with {head_node_cpus} CPUs and {head_node_gpus} GPUs for resource calculation"
+            )
+            available_cpus = max(0, available_cpus - head_node_cpus)
+            available_gpus = max(0, available_gpus - head_node_gpus)
```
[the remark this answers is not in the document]
praateekmahajan: Since we assume Ray Cluster is only being used for a single Curator job, available resources should match total resources. I'll leave a comment for the same

iamjustinhsu:
> In simpler words, because for these two executors we don't decide "where" scheduling happens, just the "how many". Does Anyscale restrict any scheduling?

Ok, so in my research, anyscale workspaces start the head-node with 0 cpus, ie, `ray start --head --num-cpus=0`. This implies any task with `ray.remote(num_cpus=N)`, where `N > 0` will fail because of resource requirements. Simiarly, because of `@ray.remote` defaults documented [here](https://docs.ray.io/en/latest/ray-core/scheduling/resources.html#specifying-task-or-actor-resource-requirements), `_setup_stage_on_node` requires 1 CPU for scheduling, so trying to scheduling that on the head node will also fail. Consider the following script
```python
import ray
from ray.util.scheduling_strategies import NodeAffinitySchedulingStrategy

@ray.remote(num_cpus=0) # This will work
# @ray.remote(num_cpus=1) # < -- This does not work, head node has 0 cpus
# @ray.remote # < -- Similarly, this does not work, default requires 1 cpu.
class _Actor:
    def __init__(self):

        node_id = ray.get_runtime_context().get_node_id()
        print(f"node_id is {node_id}")

ray.init()

nodes = ray.nodes()
head_node = None
for node in nodes:
    if "node:__internal_head__" in node["Resources"]:
        head_node = node["NodeID"]
        break

assert head_node is not None

print(f"head_node is {head_node}")
schedule_on_head = NodeAffinitySchedulingStrategy(node_id=head_node, soft=False)
handle = _Actor.options(scheduling_strategy=schedule_on_head).remote()
ray.get(handle.__ray_ready__.remote())
```
So the TL;DR is that if you _manually_ set `num_cpus=0` for a task, it can run on the head node. With that said, if Nemo can specify `num_cpus>0` for tasks, you don't need to worry about the where

iamjustinhsu:
@praateekmahajan Heads up, I think I ran into a small usability issue while testing your PR. This example https://docs.nvidia.com/nemo/curator/25.09/curate-text/process-data/deduplication/semdedup.html#quick-start doesn't quite work out-of-box because `ignore_head_node=True` isn't being propagated throughout the workflow. The way you can solve this is to specify another executor `batch_executor=RayActorPoolExecutor(ignore_head_node=True)`, but might be annoying. I would expect `streaming_executor=RayDataExecutor(ignore_head_node=True)` and `batch_executor=None` to propagate the `ignore_head_node=True` to other executors.

But then again, if this isn't long-term, I think it's Ok. A solution to this would be to create a global context / env var, but ofc you definitely know what's best here, so I'll defer to you.

praateekmahajan:
@iamjustinhsu thanks for the feedback, I agree an envvar will be much neater so added that [here](https://github.com/NVIDIA-NeMo/Curator/pull/1209/commits/0e199014ed5c2c4c8472addcba271d1a83644883) using `CURATOR_IGNORE_RAY_HEAD_NODE`

nemo_curator/backends/xenna/executor.py
```diff
@@ -43,8 +43,14 @@ def __init__(self, config: dict[str, Any] | None = None):
...
                 - cpu_allocation_percentage: CPU allocation ratio (default: 0.95)
                 - autoscale_interval_s: Auto-scaling interval (default: 180)
+            ignore_head_node (bool, optional): Whether to ignore the head node (default: False)
+                Not supported by XennaExecutor. Ignored if set to True.
         """
-        super().__init__(config)
+        super().__init__(config, ignore_head_node)
+        if self.ignore_head_node:
+            msg = "ignore_head_node is not supported by XennaExecutor. Use RayDataExecutor or RayActorPoolExecutor instead."
+            raise ValueError(msg)
```
sarahyurick, line 52: From the docstring, it seems like we should log a warning instead of raise an error?

nemo_curator/stages/deduplication/exact/workflow.py
```diff
@@ -173,7 +176,12 @@ def run(self, initial_tasks: list[FileGroupTask] | None = None) -> None:
             If not provided, the pipeline will generate the input tasks based on the input_dir and input_file_extensions.
         """
         self._validate_initial_tasks(initial_tasks)
-        executor = RayActorPoolExecutor(config=self.executor_config)
+        if executor is None:
+            executor = RayActorPoolExecutor(config=self.executor_config)
+        else:
+            previous_config = executor.config
+            executor.config = merge_executor_configs(executor.config, self.executor_config)
+            warn_on_env_var_override(previous_config, executor.config)
```
sarahyurick, line 187: I was wondering if you can help provide context for this change?

tests/backends/experimental/ray_data/test_utils.py
```diff
@@ -24,6 +24,7 @@
 
 
 class TestGetAvailableCpuGpuResources:
+    # TODO: Move this to tests/backends/experimental/test_utils.py
```
sarahyurick, line 27: TODO for this PR?

sarahyurick (commented):
I don't have much context on this issue, but happy to help unblock. I left some comments for now.

nemo_curator/backends/experimental/ray_data/adapter.py
```diff
@@ -70,7 +70,7 @@ def _process_batch_internal(self, batch: dict[str, Any]) -> dict[str, Any]:
         # For Task objects, we return them in the 'item' column
         return {"item": results}
 
-    def process_dataset(self, dataset: Dataset) -> Dataset:
+    def process_dataset(self, dataset: Dataset, ignore_head_node: bool = False) -> Dataset:
```
ayushdg, line 73: Could we use the executor config the instantiate this parameter as an attribute? Any preference on that vs having it be configured in `process_dataset`?

nemo_curator/backends/experimental/utils.py
```diff
@@ -17,10 +17,34 @@
...
+        The head node ID if a head node exists, otherwise None.
+    """
+    global _HEAD_NODE_ID_CACHE  # noqa: PLW0603
+
+    if _HEAD_NODE_ID_CACHE is not None:
+        return _HEAD_NODE_ID_CACHE
+
+    # Compute head node ID
+    for node in ray.nodes():
+        if "node:__internal_head__" in node.get("Resources", {}):
```
ayushdg: I know it's a one liner but I would recommend splitting this out into its own method. maybe something like `is_head_node`. That way we have a single source of truth that can be re-used in places like `setup_on_node`

nemo_curator/backends/experimental/utils.py
```diff
@@ -55,19 +99,22 @@ def _setup_stage_on_node(stage: ProcessingStage, node_info: NodeInfo, worker_met
...
-def execute_setup_on_node(stages: list[ProcessingStage]) -> None:
+def execute_setup_on_node(stages: list[ProcessingStage], ignore_head_node: bool = False) -> None:
     """Execute setup on node for a stage."""
-    from ray.util.scheduling_strategies import NodeAffinitySchedulingStrategy
-
+    head_node_id = get_head_node_id()
     ray_tasks = []
     for node in ray.nodes():
         node_id = node["NodeID"]
+        is_head_node = node_id == head_node_id or "node:__internal_head__" in node.get("Resources", {})
```
ayushdg: Do we need the second half of the conditional? If so let's use it through a method call rather than checking this attribute again.

nemo_curator/backends/xenna/executor.py
```diff
@@ -43,8 +43,14 @@ def __init__(self, config: dict[str, Any] | None = None):
                 - execution_mode: 'streaming' or 'batch' (default: 'streaming')
                 - cpu_allocation_percentage: CPU allocation ratio (default: 0.95)
                 - autoscale_interval_s: Auto-scaling interval (default: 180)
+            ignore_head_node (bool, optional): Whether to ignore the head node (default: False)
+                Not supported by XennaExecutor. Ignored if set to True.
         """
-        super().__init__(config)
+        super().__init__(config, ignore_head_node)
+        if self.ignore_head_node:
+            msg = "ignore_head_node is not supported by XennaExecutor. Use RayDataExecutor or RayActorPoolExecutor instead."
```
ayushdg, line 51: Discussion for a followup: Is there a way to check if we're in environments like any scale where head node scheduling is not allowed and we can preemptively raise rather than expect the user to set `ignore_head_node`

ayushdg (changes requested):
Left minor changes/comments. 

Is it possible to add documentation someplace where we call this out?

nemo_curator/stages/deduplication/exact/workflow.py
praateekmahajan: So for both Fuzzy / Exact workflows, we had hardcoded that we'll create a `RayActorPoolExecutor(self.executor_config)` where `self.executor_config` is created using env_vars from the init of the Workflow.

Since we want to support `ignore_head_node` we want to be able to let users provide their own executor
```python
# Exact
workflow = ExactDeduplicationWorkflow(...)
workflow.run(executor=RayActorPoolExecutor(ignore_head_node=True))

# Fuzzy
workflow = FuzzyDeduplicationWorkflow()
workflow.run(executor=RayActorPoolExecutor(ignore_head_node=True))
```

However when they do so, we want to also be able to use the `self.executor_config` that is an attribute of the object, hence we want to override the config of the user provided executor. It'll be unrealistic to expect that the user will know how to set `env_vars` etc just to solve the Anyscale issue.


I'm not super keen on this design, so I'm open to changing it as you suggest.

praateekmahajan: cc @ayushdg since you have more context on `self.executor_config`

tests/backends/experimental/ray_data/test_utils.py
praateekmahajan: Not doing in this PR as the diff will become unreadable and it'll show up like a new file

nemo_curator/backends/experimental/ray_data/adapter.py
praateekmahajan: Good question, thinking out loud, I think an `Adapter` is a "transformation" to a stage for a specific `Executor`. It contains information about how to run the stage. While the executor is the one which says how to run stuff on a cluster. In other words I think ignoring head node is something to do with cluster and less to do with how to transform a stage.
For instance, once we have the change in Xenna too, I'd imagine it'll be an executor config and nothing within the Adapter. I might be wrong tho. WDYT?

nemo_curator/backends/experimental/utils.py
praateekmahajan: We don't, removed it

nemo_curator/stages/deduplication/exact/workflow.py
ayushdg: The main reason I added `env_vars` as an argument is so that users could (at runtime) set ucx type env vars  if needed without having to do this at the cluster level.

The question I have here is why do we want to allow users to bring their own executor? So that they can explicitly say `ignore_head_node=True` for the executor which cannot be passed through easily? 
If so we should also raise if the user passes in anything other than an instance of RayActorPoolExecutor

nemo_curator/backends/experimental/ray_data/adapter.py
ayushdg: In either case we do have some information exchange/bleeding where the adapter needs to know what cluster resources to use for calculating concurrency which needs to be provided by the executor either while creating the adapter object or while calling it's method. I'm okay keeping it this way, but we expand adapters to do more work or need to know this information more broadly than `process_data` it might make sense to have it as an attribute?

nemo_curator/stages/deduplication/exact/workflow.py
praateekmahajan: > So that they can explicitly say ignore_head_node=True for the executor which cannot be passed through easily?

Yup, basically. Will add the error to be raised

nemo_curator/backends/experimental/ray_data/adapter.py
praateekmahajan: I see wym that `calculate_concurrency_for_actors_for_stage` is being called by the adapter so there is some leakage already. 

Let's keep like this for now and we can revisit when we see more leakage

nemo_curator/stages/deduplication/exact/workflow.py
ayushdg: I'm fine with this for the time being.

Issue #1144, opened 2025-09-30
Re-add YAML files


sarahyurick:
YAML support should be implicit in Curator, as demonstrated with our audio tutorial: https://github.com/NVIDIA-NeMo/Curator/tree/main/tutorials/audio/fleurs

I had opened https://github.com/NVIDIA-NeMo/Curator/pull/957 to add scripts for YAML pipelines, but the PR is old with conflicts. We should create a fresh PR with comprehensive examples for filters, etc.

PR #1212, opened 2025-10-30
Add YAML support for text pipelines

sarahyurick:
Closes https://github.com/NVIDIA-NeMo/Curator/issues/1144.

### DIFF
--- a/docs/admin/installation.md
+++ b/docs/admin/installation.md
@@ -66,7 +66,7 @@ Install NeMo Curator from the Python Package Index using `uv` for proper depende
    ```bash
    uv pip install torch wheel_stub psutil setuptools setuptools_scm
    echo "transformers==4.55.2" > override.txt
-   uv pip install  https://pypi.nvidia.com --no-build-isolation "nemo-curator[all]" --override override.txt
+   uv pip install --no-build-isolation "nemo-curator[all]" --override override.txt
    ```
 
 :::
@@ -98,11 +98,17 @@ uv add InternVideo/InternVideo2/multi_modality
 
 :::{tab-item} Container Installation
 
-NeMo Curator is available as a standalone container:
+NeMo Curator is available as a standalone container on NGC: https://catalog.ngc.nvidia.com/orgs/nvidia/containers/nemo-curator. The container includes NeMo Curator with all dependencies pre-installed. You can run it with:
 
-```{note}
-**Container Build**: You can build the NeMo Curator container locally using the provided Dockerfile. A pre-built container will be available on NGC in the future.
+```bash
+# Pull the container from NGC
+docker pull nvcr.io/nvidia/nemo-curator:{{ container_version }}
+
+# Run the container with GPU support
+docker run --gpus all -it --rm nvcr.io/nvidia/nemo-curator:{{ container_version }}
 ```
+
+Alternatively, you can build the NeMo Curator container locally using the provided Dockerfile:
 
 ```bash
 # Build the container locally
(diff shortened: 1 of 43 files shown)


nemo_curator/config/run.py
```diff
@@ -0,0 +1,103 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from typing import Any
+
+import hydra
```
ayushdg, line 17: Currently Hydra is a transitive dependency coming from asr. We should make this a base dependency if all modalities need to use this.

nemo_curator/config/text/fasttext_filter_pipeline.yaml
```diff
@@ -0,0 +1,73 @@
...
+input_path: ???
+output_path: ???
+text_field: text
+# Path to the FastText language id model (e.g., lid.176.bin)
+fasttext_langid_model_path: ???
+# Path to the FastText quality model file
+fasttext_quality_model_path: ???
+
+stages:
+  - input_file_type: jsonl  # This can be jsonl or parquet
```
ayushdg: Curious about this approach vs users explicitly mentioning/using the class directly (JsonlReader) etc. and having access to all the args there.

ayushdg (commented):
Initial comments while I take a look at this in more detail.

nemo_curator/config/text/fasttext_filter_pipeline.yaml
sarahyurick: Yeah I was going back and forth here. The hope was that the user should not have to update the YAML file at all, and can pass any relevant parameters in the CLI command. I don't think passing something like `nemo_curator.stages.text.io.reader.jsonl.JsonlReader` would work, and some type of if-else logic in the YAML file doesn't work either.

I was thinking that since the JSONL and Parquet readers and writers share the same parameters, we can use `input_file_type` and `output_file_type` and just handle it in `run.py`. Although I am happy to just choose one and use the class directly if you prefer. WDYT?

praateekmahajan: My vote would be to use the class, as then it doesn't force us to ensure that the arguments are same or similar. In the dask world we used to have a common code path for json / parquet reader which got too convoluted eventually with multiple if / else.

I like that in a yaml we're essentially defining a pipeline just how we would define in code, and even in code that does mean the user will have to swap out the classes.

nemo_curator/config/run.py
```diff
@@ -0,0 +1,103 @@
...
+    else:
+        msg = "Invalid YAML configuration. Please define stages to add to a pipeline or a workflow to execute."
+        raise RuntimeError(msg)
+
+
+@hydra.main(version_base=None)
+def main(cfg: DictConfig) -> None:
+    ray_client = create_ray_client_from_yaml(cfg)
+    ray_client.start()
+
```
praateekmahajan, line 68: for future, we probably might need a call to `create_id_generator` given some user config

nemo_curator/config/text/fasttext_filter_pipeline.yaml
sarahyurick: Thanks! Updating now.

ayushdg (approved):
Changs lgtm! 

Few improvements/changes that might help these:

1. Support for other executors
2. more comprehensive pytest coverage across all text pytests & testing the deduplication yaml workflows as well.
3. Examples of overriding resources/resource usage by a stage

sarahyurick:
Thanks @ayushdg ! I like all of those suggestions. I created an issue to track follow-up work: https://github.com/NVIDIA-NeMo/Curator/issues/1359

PR #1223, opened 2025-11-07
Update user-facing variable names and add override checks for `name`, `resources`, and `batch_size`

sarahyurick:


docs/about/concepts/video/abstractions.md
```diff
@@ -58,12 +58,6 @@ Composite stages provide a user-facing convenience API and decompose into one or
 
 ```python
 class MyStage(ProcessingStage[X, Y]):
-    @property
-    def name(self) -> str: ...
-
-    @property
-    def resources(self) -> Resources: ...
-
     def inputs(self) -> tuple[list[str], list[str]]: ...
```
abhinavg4, line 64: Should we still tell them how to use the name and resources? Like as a variable instead of func?

sarahyurick: Thanks! Updated.

Yeah I tried to be as comprehensive as possible with the docs, although I could have missed things. I am hoping the docs reviews by the engineers can catch any lingering mistakes.

Assuming our pytests are comprehensive, then all changes to the codebase should be correct. I updated the tutorials as well.

PR #1230, opened 2025-11-10
Bug fix in Text Semantic Dedup Worklfow  to allow output_filetype to be jsonl

praateekmahajan:
## Description
We were incorrectly checking `output_filetype` inside embedding generation while `output_filetype` is a concept only relevant if you have `perform_removal=True`

## Usage

```python
```
## Checklist


sarahyurick (commented):
Should it emit a warning/error if the user specifies `output_filetype="jsonl"`?

Edit: Ah I see, it looks like your edit only changes an intermediate output stage.

nemo_curator/stages/text/deduplication/semantic.py
ayushdg: For these do you want to update the comment to allude to the fact that you're referring to the deduplicated output?

PR #1233, opened 2025-11-17
Adds FuzzyDedup identification and Removal benchmarks

rlratzel:
* Adds initial duplicate and removal benchmarks.
* Sets the docker max system memory to be host memory but no more than 1TB, sets SHM size and object store to be 50% of docker max system memory.
* Adds arg to functions that start ray to allow object store size to be specified.
* Updates Slack sink to add indentation, format time values.
* Installs all optional deps to Docker benchmarking image.
* Refactors benchmark runner classes for clarity: renamed classes from Matrix* to Session and Entry since those seem to be the words devs use to describe them.
* Adds updated cluster polling function to check that the Ray cluster is indeed running and retries to start it if not.

### DIFF
--- a/benchmarking/nightly-benchmark.yaml
+++ b/benchmarking/nightly-benchmark.yaml
@@ -11,14 +11,25 @@ artifacts_path: /path/where/artifacts/are/stored
 datasets_path: /path/to/datasets
 
 datasets:
-  - name: "tinystories_train"
+  - name: "tinystories"
     formats:
     - type: "parquet"
-      path: "{datasets_path}/clean/tinystories_train_parquet"
-  - name: "cleaned_exact_dedup_all_cc"
+      path: "{datasets_path}/tinystories/parquet_data"
+    - type: "jsonl"
+      path: "{datasets_path}/tinystories/jsonl_data"
+  - name: "commoncrawl"
     formats:
     - type: "jsonl"
-      path: "{datasets_path}/cleaned_exact_dedup_all_cc"
+      path: "{datasets_path}/commoncrawl/jsonl_data"
+  - name: "commoncrawl_id_map"
+    formats:
+    - type: "json"
+      path: "{datasets_path}/commoncrawl/id_generator.json"
+  - name: "commoncrawl_ids"
+    formats:
+    - type: "parquet"
+      path: "{datasets_path}/commoncrawl/IDs/parquet_data"
+
 
 default_timeout_s: 7200
 
@@ -46,7 +57,7 @@ entries:
     script: domain_classification_benchmark.py
     args: >-
       --executor=ray_data
-      --input-path={dataset:tinystories_train,parquet}
+      --input-path={dataset:tinystories,parquet}
       --dataset-size-gb=10
       --model-inference-batch-size=1024
     timeout_s: 20000
@@ -68,7 +79,7 @@ entries:
     script: domain_classification_benchmark.py
     args: >-
       --executor=xenna
-      --input-path={dataset:tinystories_train,parquet}
+      --input-path={dataset:tinystories,parquet}
       --dataset-size-gb=10
       --model-inference-batch-size=1024
     timeout_s: 20000
@@ -78,7 +89,7 @@ entries:
     script: embedding_generation_benchmark.py
     args: >-
       --executor=ray_data
-      --input-path={dataset:tinystories_train,parquet}
+      --input-path={dataset:tinystories,parquet}
       --dataset-size-gb=10
       --model-identifier=sentence-transformers/all-MiniLM-L6-v2
       --model-inference-batch-size=1024
@@ -96,7 +107,7 @@ entries:
     script: embedding_generation_benchmark.py
     args: >-
       --executor=xenna
-      --input-path={dataset:tinystories_train,parquet}
+      --input-path={dataset:tinystories,parquet}
       --dataset-size-gb=10
       --model-identifier=sentence-transformers/all-MiniLM-L6-v2
       --model-inference-batch-size=1024
@@ -106,7 +117,7 @@ entries:
     enabled: true
     script: fuzzy_dedup_identification_benchmark.py
     args: >-
-      --input-path={dataset:cleaned_exact_dedup_all_cc,jsonl}
+      --input-path={dataset:commoncrawl,jsonl}
       --cache-path={session_entry_dir}/scratch/cache
       --output-path={session_entry_dir}/output
       --input-filetype=jsonl
@@ -123,9 +134,9 @@ entries:
     enabled: true
     script: dedup_removal_benchmark.py
     args: >-
-      --input-path={dataset:cleaned_exact_dedup_all_cc,jsonl}
-      --id-generator-path={datasets_path}/cleaned_exact_dedup_all_cc_fuzzy_output_nightly_container_paths/fuzzy_id_generator.json
-      --ids-to-remove-path={datasets_path}/cleaned_exact_dedup_all_cc_fuzzy_output_nightly_container_paths/FuzzyDuplicateIds/
+      --input-path={dataset:commoncrawl,jsonl}
+      --id-generator-path={dataset:commoncrawl_id_map,json}
+      --ids-to-remove-path={dataset:commoncrawl_ids,parquet}
       --output-path={session_entry_dir}/scratch/output
       --executor=ray_data
       --input-filetype=jsonl
@@ -143,9 +154,9 @@ entries:
     enabled: true
     script: dedup_removal_benchmark.py
     args: >-
-      --input-path={dataset:cleaned_exact_dedup_all_cc,jsonl}
-      --id-generator-path={datasets_path}/cleaned_exact_dedup_all_cc_fuzzy_output_nightly_container_paths/fuzzy_id_generator.json
-      --ids-to-remove-path={datasets_path}/cleaned_exact_dedup_all_cc_fuzzy_output_nightly_container_paths/FuzzyDuplicateIds/
+      --input-path={dataset:commoncrawl,jsonl}
+      --id-generator-path={dataset:commoncrawl_id_map,json}
+      --ids-to-remove-path={dataset:commoncrawl_ids,parquet}
       --output-path={session_entry_dir}/scratch/output
       --executor=xenna
       --input-filetype=jsonl


praateekmahajan:
I know it's in draft, but as a heads up, for text we have three duplicate identification algorithm supported fuzzy, exact and semantic
Here are some thoughts

1. I believe right now your script runs Fuzzy Dedup, so we would want to have a dedicated file named that "fuzzy_dedup_identification.py"  (or similar)
2. Both Fuzzy and Exact only support RayActorPool as an executor. For future reference semantic runs on `RayActorPool` for part of it and `RayData|Xenna` for the other part of it. (So for fuzzy / exact we just can start with one entry)
3. Removal can be benchmarked just once, and not for all different types of identification.. 
4. Since each deduplication type (Fuzzy/Exact/Semantic) runs as a workflow that can contain multiple pipelines, we need to think carefully about how we handle `tasks.pkl`. Right now, each pipeline produces a task_output that gets dumped to `tasks.pkl`. But with workflows, we'll have multiple task_output objects - one from each pipeline in the workflow. So if a workflow runs pipelines A, B, and C, we need to merge `task_output_a`, `task_output_b`, and `task_output_c` into a single `tasks.pkl` file. The tricky part is that task_output is keyed by stage names. If pipeline A and pipeline B both have a stage called S, we can't just merge them naively or we'll overwrite the metrics from stage S in pipeline A with the metrics from stage S in pipeline B. I think we should prefix or suffix the stage names with their pipeline identifier to avoid collisions. This logic should probably live in `script.py`. It would be helpful to create a utility function like `combine_task_metrics(pipeline_name_to_task_output: list[tuple[str, list[Task]]])` that handles this merging intelligently.

benchmarking/runner/ray_cluster.py
```diff
@@ -27,123 +29,152 @@
...
+        # Create and start the Ray client
+        client = RayClient(
+            ray_temp_dir=str(short_temp_path),
+            include_dashboard=include_dashboard,
+            num_gpus=num_gpus,
+            num_cpus=num_cpus,
+            enable_object_spilling=enable_object_spilling,
+            ray_dashboard_host="0.0.0.0",  # noqa: S104
+            ray_stdouterr_capture_file=ray_stdouterr_capture_file,
+            object_store_memory=object_store_memory,
```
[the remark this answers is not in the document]
rlratzel: Greptile noticed that this PR currently requires changes from https://github.com/NVIDIA-NeMo/Curator/pull/1274

benchmarking/runner/sinks/slack_sink.py
```diff
@@ -250,6 +253,20 @@ def _two_column_row_bold(left_text: str, right_text: str) -> list[dict[str, Any]
...
+    def _get_formatted_metric_value_tuple(metric: str, result: Any) -> tuple[str, str]:  # noqa: ANN401
+        if metric == "exec_time_s":
+            try:
+                hours = int(result // 3600)
+                minutes = int((result % 3600) // 60)
+                seconds = result % 60
+            except (ValueError, TypeError):
+                return (metric, str(result))
+            else:
+                return (metric, f"{result:.3f}s  ({hours:02}h : {minutes:02}m : {seconds:06.3f}s)")
```
praateekmahajan: can we do something else s.t.  if hours or minutes is 0 then we don't print it

```python
formatted_str = "{result:.2f}s}"
if hours > 0 or minutes > 0: formatted_str += " ("
    if hours > 0: formatted_str += "{hours:.02}h :"
    if minutes > 0: formatted_str += "{minutes:.02}m :"
    formatted_str += "{seconds:06.2f}s )"
```

benchmarking/scripts/fuzzy_dedup_identification_benchmark.py
```diff
@@ -0,0 +1,184 @@
...
+            DataContext.get_current().target_max_block_size = 1
+
+    elif executor_name == "xenna":
+        # TODO: Uncomment this when the pipeline is fixed
+        # from nemo_curator.backends.xenna import XennaExecutor
+        # executor = XennaExecutor()
+        pass
+    else:
+        msg = f"Executor {executor_name} not supported"
+        raise ValueError(msg)
```
praateekmahajan: all of this can be removed

benchmarking/scripts/fuzzy_dedup_identification_benchmark.py
```diff
@@ -0,0 +1,184 @@
...
+    parser.add_argument("--executor", default="xenna", choices=["xenna", "ray_data"], help="Executor to use")
+    parser.add_argument("--input-filetype", default="jsonl", choices=["jsonl", "parquet"], help="Input filetype")
+    parser.add_argument(
+        "--bands-per-iteration", type=int, default=20, help="Bands per iteration (for LSH deduplication)"
+    )
+    parser.add_argument("--text-field", default="text", help="Text field to use for duplicate identification")
+    parser.add_argument(
+        "--input-blocksize", type=str, default="1.5GiB", help="Target partition size for input data (e.g. '512MB')"
+    )
+    parser.add_argument("--use-ray-data-settings", action="store_true", help="If set, use ray data settings")
```
praateekmahajan: can be removed

benchmarking/scripts/fuzzy_dedup_identification_benchmark.py
```diff
@@ -0,0 +1,184 @@
...
+        pickle.dump(results["tasks"], f)
+
+
+def main() -> int:
+    parser = argparse.ArgumentParser(description="Duplicate identification benchmark for nightly benchmarking")
+    parser.add_argument("--benchmark-results-path", required=True, help="Path to benchmark results")
+    parser.add_argument("--input-path", required=True, help="Path to input data")
+    parser.add_argument("--cache-path", required=True, help="Path to cache directory")
+    parser.add_argument("--output-path", required=True, help="Output directory for results")
+    parser.add_argument("--executor", default="xenna", choices=["xenna", "ray_data"], help="Executor to use")
```
praateekmahajan: can be removed

benchmarking/tools/gen_runscript_vars.py
```diff
@@ -22,13 +22,19 @@
...
+from runner.path_resolver import (
     CONTAINER_ARTIFACTS_DIR_ROOT,
     CONTAINER_CONFIG_DIR_ROOT,
     CONTAINER_CURATOR_DIR,
     CONTAINER_DATASETS_DIR_ROOT,
     CONTAINER_RESULTS_DIR_ROOT,
 )
+from runner.utils import get_total_memory_bytes
+
+_max_container_memory_bytes = 1099511627776  # 1TB
```
praateekmahajan:
```suggestion
ONE_MB = 1024 * 1024 * 1024
_max_container_memory_bytes = 1024 * 1024 * ONE_MB # 1TB
```

benchmarking/nightly-benchmark.yaml
```diff
@@ -99,40 +102,56 @@ entries:
...
-    enabled: false
-    script: removal_benchmark.py
+  - name: fuzzy_dedup_identification_raydata
+    enabled: true
+    script: fuzzy_dedup_identification_benchmark.py
     args: >-
+      --input-path={dataset:cleaned_exact_dedup_all_cc,jsonl}
+      --cache-path={session_entry_dir}/scratch/cache
+      --output-path={session_entry_dir}/output
       --executor=ray_data
```
praateekmahajan: can remove this

benchmarking/runner/sinks/slack_sink.py
```diff
@@ -138,7 +140,7 @@ def _post(self) -> None:  # noqa: C901
         rows.append(self._two_column_row_bold("OVERALL STATUS", overall_status))
```
praateekmahajan, line 140: can we also add a row for shm / object store / cpu memory etc.. one is plenty

benchmarking/runner/sinks/slack_sink.py
rlratzel: If hours >0 but minutes == 0, seems like we should still print minutes?
eg.: `(01h : 00m : 34.23s)` vs. `(01h : 34.23s)`  the latter seems too easy to read as 1 minute 34 seconds.
And, should we also drop the `h`, `m`, `s` in that format too?
`(01h : 00m : 34.23s)` vs. `(01:00:34.23)`

benchmarking/nightly-benchmark.yaml
```diff
@@ -15,6 +15,10 @@ datasets:
     formats:
     - type: "parquet"
       path: "{datasets_path}/prospector-lm/clean/tinystories_train_parquet"
+  - name: "cleaned_exact_dedup_all_cc"
+    formats:
+    - type: "jsonl"
+      path: "{datasets_path}/prospector-lm/cleaned_exact_dedup_all_cc"
```
sarahyurick: Can it be `{datasets_path}/cleaned_exact_dedup_all_cc` here, and remove `prospector-lm` everywhere? I would like to keep the paths as generic as possible.

Same for:
```
results_path: /raid/curator-team/nightly/results
artifacts_path: /raid/curator-team/nightly/artifacts
datasets_path: /raid
```
Can they not be hardcoded as absolute paths?

benchmarking/runner/sinks/slack_sink.py
rlratzel: I added lines to the "ENVIRONMENT" section for total system memory, shm, and default object store size.

benchmarking/nightly-benchmark.yaml
rlratzel: Chatted offline.  The latest commit replaces those user/machine-specific paths with placeholders, making the file more of a template for users/devs to modify before using.  Another option is to provide a separate .yaml file for the paths that overrides the placeholder values (this is what the nightly run will do), like this:
```bash
./benchmarking/tools/run.sh \
    --config=./benchmarking/nightly-benchmark.yaml \
    --config=/config/for/nightly/run/paths.yaml
```
where `paths.yaml` is just:
```yaml
results_path: /raid/curator-team/nightly/results
artifacts_path: /raid/curator-team/nightly/artifacts
datasets_path: /raid/prospector-lm
```

benchmarking/nightly-benchmark.yaml
```diff
@@ -6,15 +6,19 @@
...
 datasets:
   - name: "tinystories_train"
     formats:
     - type: "parquet"
-      path: "{datasets_path}/prospector-lm/clean/tinystories_train_parquet"
+      path: "{datasets_path}/clean/tinystories_train_parquet"
+  - name: "cleaned_exact_dedup_all_cc"
+    formats:
+    - type: "jsonl"
+      path: "{datasets_path}/cleaned_exact_dedup_all_cc"
```
sarahyurick: Nit but this section could be more generic here as well, and overridden by any additional config file. I was thinking something like:
```
# Placeholder
datasets:
  - name: "parquet_data"
    formats:
    - type: "parquet"
      path: "{datasets_path}/parquet_data"
  - name: "jsonl_data"
    formats:
    - type: "jsonl"
      path: "{datasets_path}/jsonl_data"
```
Plus any additional updates needed to make
```
--input-path={dataset:tinystories_train,parquet}
--input-path={dataset:cleaned_exact_dedup_all_cc,jsonl}
--id-generator-path={datasets_path}/cleaned_exact_dedup_all_cc_fuzzy_output_nightly_container_paths/fuzzy_id_generator.json
--ids-to-remove-path={datasets_path}/cleaned_exact_dedup_all_cc_fuzzy_output_nightly_container_paths/FuzzyDuplicateIds/
```
more generic too?

WDYT?

rlratzel: sounds good, I'll work on that...

PR #1239, opened 2025-11-17
Adds optional filename arg to capture ray stdout/stderr

rlratzel:
Adds optional filename arg to RayClient.start() method to capture ray stdout/stderr. This prevents stray output from the 'ray start' process from appearing in applications running Curator workflows, and allows those applications to capture that output to a separate file for later analysis.

### DIFF
--- a/nemo_curator/core/client.py
+++ b/nemo_curator/core/client.py
@@ -16,8 +16,7 @@ import atexit
 import os
 import socket
 import subprocess
-from dataclasses import dataclass
-from typing import TextIO
+from dataclasses import dataclass, field
 
 from loguru import logger
 
@@ -74,11 +73,16 @@ class RayClient:
     num_gpus: int | None = None
     num_cpus: int | None = None
     enable_object_spilling: bool = False
+    ray_stdouterr_capture_file: str | None = None
 
-    ray_process: subprocess.Popen | None = None
-    ray_stdouterr_capture_file_obj: TextIO | None = None
+    ray_process: subprocess.Popen | None = field(init=False)
 
-    def start(self, stdouterr_capture_file: str | None = None) -> None:
+    def __post_init__(self) -> None:
+        if self.ray_stdouterr_capture_file and os.path.exists(self.ray_stdouterr_capture_file):
+            msg = f"Capture file {self.ray_stdouterr_capture_file} already exists."
+            raise FileExistsError(msg)
+
+    def start(self) -> None:
         """Start the Ray cluster, optionally capturing stdout/stderr to a file."""
         if self.include_dashboard:
             # Add Ray metrics service discovery to existing Prometheus configuration
@@ -113,12 +117,6 @@ class RayClient:
                 get_next_free_port=(self.ray_client_server_port == DEFAULT_RAY_CLIENT_SERVER_PORT),
             )
             ip_address = socket.gethostbyname(socket.gethostname())
-            # Open the stdouterr capture file if provided.
-            # This will be closed when the Ray client is stopped.
-            if stdouterr_capture_file:
-                self.ray_stdouterr_capture_file_obj = open(stdouterr_capture_file, "w")  # noqa: SIM115
-            else:
-                self.ray_stdouterr_capture_file_obj = None
 
             self.ray_process = init_cluster(
                 self.ray_port,
@@ -132,7 +130,7 @@ class RayClient:
                 self.enable_object_spilling,
                 block=True,
                 ip_address=ip_address,
-                stdouterr_capture_file_obj=self.ray_stdouterr_capture_file_obj,
+                stdouterr_capture_file=self.ray_stdouterr_capture_file,
             )
             # Set environment variable for RAY_ADDRESS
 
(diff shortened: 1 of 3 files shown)


nemo_curator/core/client.py
```diff
@@ -75,8 +76,10 @@ class RayClient:
     enable_object_spilling: bool = False
 
     ray_process: subprocess.Popen | None = None
+    ray_stdouterr_capture_file_obj: TextIO | None = None
```
praateekmahajan: I think these both can be
```suggestion
    ray_process: subprocess.Popen | None = field(init=False)
    ray_stdouterr_capture_file_obj: TextIO | None = field(init=False)
```
where field is from `from dataclasses import field`

nemo_curator/core/utils.py
```diff
@@ -121,6 +122,9 @@ def init_cluster(  # noqa: PLR0913
     os.environ["XENNA_RAY_METRICS_PORT"] = str(ray_metrics_port)
     os.environ["XENNA_RESPECT_CUDA_VISIBLE_DEVICES"] = "1"
 
-    proc = subprocess.Popen(ray_command, shell=False)  # noqa: S603
+    if stdouterr_capture_file_obj:
```
praateekmahajan: What if we do the `with` here? something like this, therefore we won't have to worry about passing around an open file object and also ensuring we clean it up

```python
def init_cluster(...., stdout_filepath):

  if stdout_filepath:
     with open(stdout_filepath, "w") as stdout_fileobj:
        proc = subprocess.Popen(ray_command, shell=False, stdout=stdout_fileobj, stderr=subprocess.STDOUT)
    else:
        proc = subprocess.Popen(ray_command, shell=False, stdout=DEFAULT, stderr=subprocess.STDOUT)
```

nemo_curator/core/client.py
```diff
@@ -75,8 +76,10 @@ class RayClient:
...
 
-    def start(self) -> None:
+    def start(self, stdouterr_capture_file: str | None = None) -> None:
```
praateekmahajan: Should this be in `start` or should this be in `dataclass` therefore it also works with the context enter / exit.

tests/core/test_get_ray_client.py
```diff
@@ -81,28 +93,44 @@ def test_get_ray_client_multiple_start():
...
-
-            with open(os.path.join(ray_tmp, "ray_current_cluster")) as f:
-                content = f.read()
-                assert content.split(":")[1].strip() == str(client.ray_port)
+            _assert_ray_cluster_started(client)
 
         assert client.ray_process is None
+
+
+@pytest.mark.xfail(strict=False, reason="Non-deterministic test due to Ray startup timing")
```
praateekmahajan: Let's see if we can make this deterministic? Also TAL at [caplog](https://docs.pytest.org/en/stable/how-to/logging.html#caplog-fixture) which might help us capture the stdout output when the filepath isn't present

nemo_curator/core/client.py
sarahyurick: +1 this is a very good callout.

nemo_curator/core/client.py
```diff
@@ -110,6 +113,13 @@ def start(self) -> None:
                 get_next_free_port=(self.ray_client_server_port == DEFAULT_RAY_CLIENT_SERVER_PORT),
             )
             ip_address = socket.gethostbyname(socket.gethostname())
+            # Open the stdouterr capture file if provided.
+            # This will be closed when the Ray client is stopped.
+            if stdouterr_capture_file:
```
sarahyurick: Do we care if the file already exists?

nemo_curator/core/client.py
rlratzel: Thanks, I was debating that myself too. In the end I chose in `start()` to allow for unique capture files per start call, like this:
```
client = RayClient(ray_temp_dir="/tmp/ray")
client.start(stdouterr_capture_file="/tmp/ray/run1.out")
...
client.stop()
...
client.start(stdouterr_capture_file="/tmp/ray/run2.out")
...
client.stop()
```
but maybe that pattern isn't used and we'd prefer the capture file to just be part of the dataclass?

I think the use case for the CM does indeed justify making it just be specified via the dataclass though, so I'll make that change unless we'd also like to specify it via start (IOW we could do both but I'll just support dataclass for simplicity).

nemo_curator/core/client.py
rlratzel: Thanks, I can check for that. Do we want to error if the file already exists? I can see a case for both error and overwrite being useful.

nemo_curator/core/utils.py
rlratzel: Thanks, that's nicer. I thought it would be a problem when the file obj is closed as the CM exits, but apparently Popen will make a copy of the file descriptor and close it itself, so I'll do this.

rlratzel: lol, looks like greptile was thinking the same thing I was [here](https://github.com/NVIDIA-NeMo/Curator/pull/1239#discussion_r2539762370).  I looked into this (as well as tested) and your suggestion to use the CM is safe: 
```
* POSIX fork() semantics - child inherits copies of parent's file descriptors
* Python's subprocess implementation - which uses fork() + exec() on Unix
* File descriptor independence - closing in parent doesn't close in child
So while the Python docs could be more explicit about this, the behavior is fundamentally guaranteed by the Unix process model that Python's subprocess module is built upon.
```

PR #1249, opened 2025-11-19
Fix ParquetReader and *Writer for Cloud I/O

praateekmahajan:
## Description
1. Parquet Reader shouldn't work on list[files] as per https://github.com/pandas-dev/pandas/issues/62922#issuecomment-3470027819 so unfortunately we have to do a concat
2. Our writer currently strips out the protocol for the path and passes the path w/o protocol to write_data

## Usage

```python
```
## Checklist


nemo_curator/stages/text/io/writer/base.py
```diff
@@ -40,25 +40,12 @@ class BaseWriter(ProcessingStage[DocumentBatch, FileGroupTask], ABC):
     write_kwargs: dict[str, Any] = field(default_factory=dict)
     fields: list[str] | None = None
     mode: Literal["ignore", "overwrite", "append", "error"] = "ignore"
-    name: str = "BaseWriter"
```
sarahyurick, line 43: We could still keep the name if we want.

nemo_curator/stages/text/io/reader/parquet.py
```diff
@@ -59,7 +59,10 @@ def read_data(
         if "dtype_backend" not in read_kwargs:
             update_kwargs["dtype_backend"] = "pyarrow"
         read_kwargs.update(update_kwargs)
-        return pd.read_parquet(paths, **read_kwargs)
+        return pd.concat(
+            (pd.read_parquet(path, **read_kwargs) for path in paths),
```
ayushdg, line 63: Can we open a todo to investigate perf implications of this? I know it's unavoidable for now but this might make us explore other read alternatives?

PR #1289, opened 2025-12-03
 feat: CI perf improvement (manage bottlenecks) #2

pablo-garay:
CPU test fails does not cancel other CPU tests and GPU test is canceled.
When a CPU test fails, the GPU has an ongoing monitoring background job which cancels the GPU test as soon as detected.
Tested.
I ran different test case scenarios & it worked
Example 1: gets canceled: [Revert to pre-merge state (f65996ee) · NVIDIA-NeMo/Curator@d09581a](https://github.com/NVIDIA-NeMo/Curator/actions/runs/19919817121/job/57106357773?pr=1289)
Example 2: [update2 · NVIDIA-NeMo/Curator@65cc7a4](https://github.com/NVIDIA-NeMo/Curator/actions/runs/19919992828/job/57106616906?pr=1289)

### DIFF
--- a/.github/actions/test-template/action.yml
+++ b/.github/actions/test-template/action.yml
@@ -144,8 +144,8 @@ runs:
         start_cpu_monitor() {  # Usage: start_cpu_monitor "kill_cmd" "name"
           local pf="/tmp/cpu_monitor_${RUN_ID}_${2:-default}.pid"
           # Clean up ALL monitors from this run (handles cross-step and retry cleanup)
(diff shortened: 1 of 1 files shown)


.github/workflows/cicd-main.yml
```diff
@@ -203,12 +208,6 @@ jobs:
             ./.coverage
           include-hidden-files: true
 
-      - name: Cancel workflow on failure
```
pablo-garay: this cancels everything; hence we dont do this anymore. Only cancel GPU job if a CPU test fails

PR #1301, opened 2025-12-13
[benchmarking] Updates to support multiple nightly runs, code cleanup

rlratzel:
This PR contains changes to support a nightly run that consists of multiple benchmark builds/runs.
An example would be building the `main` branch and running the regular nightly suite, then building a separate image using newer dependency versions and running a specific subset of benchmarks.
This PR also includes several updates to clean up the code: var name updates, removal of files no longer used, etc.

### Updates
* Adds the benchmark session name to slack report
* Updates env var names to match other top-level scripts for consistency
* No longer tags images with `:latest` by default to prevent multiple users/runs from overwriting images when default names are used
* Removes unused benchmark "dummy" files

### DIFF
--- a/benchmarking/runner/sinks/slack_sink.py
+++ b/benchmarking/runner/sinks/slack_sink.py
@@ -33,7 +33,7 @@ _post_template = """
       "type": "header",
       "text": {
         "type": "plain_text",
-        "text": "Curator Benchmark Summary $SESSION_NAME"
+        "text": "Curator Benchmark Summary$SESSION_NAME"
       }
     },
     {
@@ -123,7 +123,7 @@ class SlackSink(Sink):
             "REPORT_JSON_TEXT": "REPORT_JSON_TEXT",
             "GOOGLE_DRIVE_LINK": "https://google.com",
             "EXECUTIVE_SUMMARY": " ",
-            "SESSION_NAME": f"- {self.session_name}" if self.session_name else "",
+            "SESSION_NAME": f" - {self.session_name}" if self.session_name else "",
         }
         indent = "-    "  # start with a dash since leading whitespace is stripped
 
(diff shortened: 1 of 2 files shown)


benchmarking/runner/sinks/slack_sink.py
```diff
@@ -33,7 +33,7 @@
       "type": "header",
       "text": {
         "type": "plain_text",
-        "text": "Curator Benchmark Summary"
+        "text": "Curator Benchmark Summary $SESSION_NAME"
```
sarahyurick:
```suggestion
        "text": "Curator Benchmark Summary$SESSION_NAME"
```
Nit: avoid extra whitespace when session name is none.

benchmarking/runner/sinks/slack_sink.py
```diff
@@ -123,6 +123,7 @@ def _post(self) -> None:  # noqa: C901
             "REPORT_JSON_TEXT": "REPORT_JSON_TEXT",
             "GOOGLE_DRIVE_LINK": "https://google.com",
             "EXECUTIVE_SUMMARY": " ",
+            "SESSION_NAME": f"- {self.session_name}" if self.session_name else "",
```
sarahyurick:
```suggestion
            "SESSION_NAME": f" - {self.session_name}" if self.session_name else "",
```
See comment above.

benchmarking/runner/utils.py
```diff
@@ -139,7 +139,10 @@ def run_shm_size_check(human_readable: bool = False) -> tuple[int | None, str |
         output = result.stdout
         line = output.strip().split("\n")[-1]
         try:
-            size = int(line.split()[1])  # Size is the second column
+            size = line.split()[1]  # Size is the second column
+            # Convert to a real number if not meant for simply reading by humans
+            if human_readable is False:
```
sarahyurick:
```suggestion
            if not human_readable:
```

sarahyurick (commented):
Left minor comments, thanks.

PR #1303, opened 2025-12-13
feat: add install-test

pablo-garay:
### Summary
New CI workflow that validates NeMo Curator installation for all optional dependencies (CPU and GPU extras) with both pip and uv installers.

### 🏗️ Architecture
Individual extra testing: Each optional extra (cuda12, audio_cpu, video_cuda12, etc.) is tested in isolation
Single base image + isolated containers: Build ONE Docker base image, then run each test in an ephemeral container
Dual installer validation: Tests both pip and uv (lockfile) installations
Clear required vs optional jobs: Only CPU extras (uv) and GPU extras (uv) are required to pass

### 📋 Job Structure
Required (must pass):├── CPU Extras Individual (pip + uv) - Py3.12└── GPU Extras Individual (uv) - Py3.12Optional (can fail, won't block merge):├── OPTIONAL: GPU Extras Individual (pip) - Py3.12  ← pip conflicts expected├── ALL Extras Combined (pip + uv) - Py3.10/3.12└── OPTIONAL: ALL Extras + Import (uv) - Py3.12

### ✨ Features
Error visibility: Actual error messages from stacktrace shown in logs on failure
Disk-efficient: All test artifacts in container overlay, auto-cleaned on exit
Fast GPU tests: ~5-10 min (single base image + docker run per extra)
Proper error detection: Uses set -o pipefail for accurate exit codes
Large download support: UV_HTTP_TIMEOUT=300 for multi-GB CUDA packages

### Triggers
Push to main or pull-request/* branches
Daily at midnight (cron)
Manual dispatch

Expected Results
- CPU extras (pip + uv): all pass
- GPU extras (uv): all pass
- GPU extras (pip): expected failures due to cuda-toolkit vs torch conflicts
- ALL extras: potential expected failures due to transformers version conflicts (nemo_toolkit[asr] vs vllm)

## TESTED: 
Sample Test Runs:
- Last one (updated): https://github.com/NVIDIA-NeMo/Curator/actions/runs/20348903348?pr=1303
- https://github.com/NVIDIA-NeMo/Curator/actions/runs/20344962779
- https://github.com/NVIDIA-NeMo/Curator/actions/runs/20343819227/job/58450444223?pr=1303
- https://github.com/NVIDIA-NeMo/Curator/actions/runs/20329773814/job/58402169989?pr=1303
- https://github.com/NVIDIA-NeMo/Curator/actions/runs/20329782244
- https://github.com/NVIDIA-NeMo/Curator/actions/runs/20325311751/job/58389351222?pr=1303
- https://github.com/NVIDIA-NeMo/Curator/actions/runs/20328600201/job/58398777928?pr=1303
- https://github.com/NVIDIA-NeMo/Curator/actions/runs/20328994495
- https://github.com/NVIDIA-NeMo/Curator/actions/runs/20329168602

### DIFF
--- a/.github/workflows/install-test.yml
+++ b/.github/workflows/install-test.yml
@@ -33,7 +33,7 @@ concurrency:
   cancel-in-progress: true
 
 env:
-  UV_HTTP_TIMEOUT: 300
+  UV_HTTP_TIMEOUT: 300 # 300s timeout needed for large GPU packages (cudf, vllm, flash-attn, torch)
 
 jobs:
   pre-flight:


.github/workflows/install-test.yml
```diff
@@ -0,0 +1,537 @@
...
+              apt-get update && apt-get install -y --no-install-recommends \
+              python${{ matrix.python-version }} python${{ matrix.python-version }}-venv python${{ matrix.python-version }}-dev git curl && \
+              update-alternatives --install /usr/bin/python3 python3 /usr/bin/python${{ matrix.python-version }} 1 && \
+              update-alternatives --install /usr/bin/python python /usr/bin/python${{ matrix.python-version }} 1 && \
+              rm -rf /var/lib/apt/lists/*
+          RUN curl -LsSf https://astral.sh/uv/install.sh | sh
+          ENV PATH="/root/.local/bin:\$PATH"
+          WORKDIR /workspace
+          COPY . .
+          RUN uv sync --locked --link-mode copy --extra all
```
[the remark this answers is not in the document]
pablo-garay: Good catch from Greptile, but actually this isn't needed because we use uv run python (line 563), not bare python:
docker run --rm --gpus all "$CONTAINER" uv run python -c "
uv run automatically activates the venv and runs the command within it, so the PATH isn't needed.
If we were using bare python, then yes, we'd need:
ENV PATH="/workspace/.venv/bin:$PATH"
But since we're using uv run python, it handles venv activation automatically. The latest run shows the import verification passed ✅, confirming this works.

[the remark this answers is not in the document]
pablo-garay: updated

.github/workflows/install-test.yml
```diff
@@ -0,0 +1,654 @@
...
+        with:
+          submodules: recursive
+
+      - name: Free disk space
+        run: |
+          sudo rm -rf /usr/share/dotnet /opt/ghc /usr/local/share/boost
+          sudo rm -rf /usr/local/lib/android
+          df -h
+
+      - name: Build base image
```
thomasdhc, line 216: Why was it decided to build the base image instead of using the github workflow `container: with:`: ie: https://github.com/NVIDIA-NeMo/Automodel/blob/main/.github/workflows/install-test.yml#L67-L68

.github/workflows/install-test.yml
```diff
@@ -0,0 +1,654 @@
...
+          
+          echo ""
+          echo "## CPU Installation Results (pip)"
+          echo "| Extra | Install | Import |"
+          echo "|-------|---------|--------|"
+          echo -e "$RESULTS"
+          
+          [ $FAILED -gt 0 ] && exit 1 || exit 0
+
+      - name: Test each CPU extra (uv)
```
thomasdhc: Logic for uv and pip seem mostly the same, where certain commands just need to be replaced by uv or pip. Is there a way we can condense these blocks of code?

thomasdhc:
@pablo-garay It seems like a lot of the logic is duplicated throughout the stages. Can you review the duplicate parts of the code, check first if there are native github workflow ways to simplify the overall stages, if not check if there are ways to simplify by reusing the code. Please estimate the amount of work that would take and make a suggestion on what you will do. Let's discuss it after that point so there's no confusion.

thomasdhc:
@pablo-garay Please also review the current failures. We've discuss prior that pip install `all` option should be failing. But in the current pipeline, those tests are passing and the individual gpu extras install is failing.

pablo-garay:
> @pablo-garay Please also review the current failures. We've discuss prior that pip install `all` option should be failing. But in the current pipeline, those tests are passing and the individual gpu extras install is failing.

Apologies about that - I had left a bit of previous behavior (remnant) of continue-on-error: true which displays them as green even though they fail with stacktrace. This was prior to adopting convention of showing them as red and having OPTIONAL as prefix in their name. So in summary, there was overlap between the two styles; now Updated to the convention described :)

.github/workflows/install-test.yml
```diff
@@ -0,0 +1,673 @@
...
+
+      - name: Cleanup
+        if: always()
+        run: docker rmi "test-uv-all-py3.12-${{ github.run_id }}" 2>/dev/null || true
+
+  # ============================================================================
+  # Summary
+  # ============================================================================
+  install-test-summary:
+    needs: [pre-flight, cpu-install-test, gpu-install-test-pip, gpu-install-test-uv, all-install-only-test, all-install-import-test]
```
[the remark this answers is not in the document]
pablo-garay: addresed

.github/workflows/install-test.yml
pablo-garay: We keep docker build/run for GPU extras tests because:
- They **need isolation per extra (each of 6 extras runs in a fresh container)** - to be able to install each in isolation in same CI job
- The `container:`  option:  runs ONE container for the whole job - can't isolate between loop iterations
- Heavy packages like cudf, flash-attn benefit from full Docker isolation


For this very specific/particular case only, we need features we cannot use with `container:`
With `container:` :
-no Docker daemon inside container
The job runs inside the container, so:
-No Docker socket available
-Can't run docker build, docker run, docker pull . docker run allows us to isolate, `docker run` creates a completely isolated environment
-Can't create/modify other containers
-"trapped" inside the pre-defined image

pablo-garay: Fortunately : ~30s build time so it'd be negligible

pablo-garay: skip the docker build and use the CUDA base image directly?

we could:
`- run: |    for extra in cuda12 audio_cuda12 ...; do      docker run --rm nvidia/cuda:12.8.1-cudnn-devel-ubuntu24.04 bash -c "        apt-get update && apt-get install python3.12 python3-pip ...  # Install Python each time        pip install '.[$extra]'      "    done`

But it's slower:

Build base image first | Once (~30s) | Just pip install (~2-5min)
Use base image directly | 6 times (~30s × 6 = 3min) | apt-get + pip install

Current approach is faster because:
Python/pip/uv installed once in base image
Each docker run starts with Python ready
Only runs the install command, not system setup
Direct base image would re-install Python for every extra (6 times), adding ~3 minutes of redundant work.
So the docker build step is an optimization - it front-loads the common setup so each extra test is faster.

.github/workflows/install-test.yml
```diff
@@ -0,0 +1,672 @@
...
```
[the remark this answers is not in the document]
pablo-garay: addressed

.github/workflows/install-test.yml
thomasdhc: - Isolation. We are using virtual environments so additional isolation is not needed
- We need to install python in both cases. If we use `container: with:` we can further utilize the github workflow way of installing python: https://github.com/NVIDIA-NeMo/Automodel/blob/main/.github/workflows/install-test.yml#L78. This also allows us to create a matrix of python versions.
- You're running the install test in a for loop so I don't see why python would need to be setup again each time.

thomasdhc: I don't see the advantages of manually setting up the docker + running docker run, when there are existing github workflow actions that allow us to do this. The manual setup decreases readability significantly.

.github/workflows/install-test.yml
```diff
@@ -0,0 +1,614 @@
...
+          echo "| Extra | Install | Import |"
+          echo "|-------|---------|--------|"
+          echo -e "$RESULTS"
+          
+          [ $FAILED -gt 0 ] && exit 1 || exit 0
+
+  # ============================================================================
+  # GPU Extras - Install Only (pip)
+  # ============================================================================
+  gpu-install-test-pip:
```
thomasdhc, line 200: Is it possible to do `matrix: installer: [pip, uv]` for gpu install as well?

pablo-garay: I had been precisely working on that; was wishing you would get to see after i made the change. WIP 
It's on the v2 version : https://github.com/NVIDIA-NeMo/Curator/pull/1329

.github/workflows/install-test.yml
pablo-garay: Ah good point / good suggestions ! I see 

I implemented/moved to : https://github.com/NVIDIA-NeMo/Curator/pull/1335

PR #1344, opened 2025-12-27
Fix ID generator to be non-blocking

### LINKED ISSUE #1343: ID Generator creation is non blocking and can cause failuire for large files
**Describe the bug**

When an IdGenerator is [created](https://github.com/NVIDIA-NeMo/Curator/blob/aaf633038c71de15f6658cd1f9160c14ceab256c/nemo_curator/stages/deduplication/id_generator.py#L116) from a JSON big enough, the IdGenerator actor creation may take a long time, as the constructor parameters are slow to serialize/deserialize. So long that the actor is not constructed and registered when used in the rest of the pipeline, such as FuzzyDeduplicationWorkflow.

Loading a ~3M json serialization file reproduces the issue 90% of the time on my AnyScale/GCS setup. Error happens next time the IdGenerator is used, with the error "Failed to look up actor with name 'curator_deduplication_id_generator'".

**Expected behavior**

As the actor was created before, that should not happen.

**Additional context**

The error is the following: actor creation on Ray is not synchronous. When IdGenerator.remote(....) is called, the actor will be registered in Ray only when the constructor is done. If one search for the actor too soon, the call could fail. Usual fix is to add a function def ready() -> bool: return True in the actor, and calls that before returning the function. Will send a fix soon.

Chyroprase:
Fixes #1343

  ## Summary
  - Brief description of your changes

  ## Test plan
  - [x] Ran unit tests
  - [x] Verified changes work as expected

### DIFF
--- a/nemo_curator/stages/text/io/writer/__init__.py
+++ b/nemo_curator/stages/text/io/writer/__init__.py
@@ -13,9 +13,11 @@
 # limitations under the License.
 
 from nemo_curator.stages.text.io.writer.jsonl import JsonlWriter
+from nemo_curator.stages.text.io.writer.megatron_tokenizer import MegatronTokenizerWriter
 from nemo_curator.stages.text.io.writer.parquet import ParquetWriter
 
 __all__ = [
     "JsonlWriter",
+    "MegatronTokenizerWriter",
     "ParquetWriter",
 ]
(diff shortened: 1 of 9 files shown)


ayushdg:
/ok to test [9eb82b8](https://github.com/NVIDIA-NeMo/Curator/pull/1344/commits/9eb82b8a5ce53ae49d4ad18f5fda6ed61533e9cb)

PR #1327, opened 2025-12-18
Add vllm embedding support

VibhuJawa:
This pull request adds support for benchmarking and using VLLM as an embedding backend in the NeMo Curator pipeline, alongside the existing HuggingFace transformers backend. It introduces a new benchmarking script for comparing embedding backends, implements a VLLM embedding stage, and updates the embedding pipeline to allow backend selection and configuration.

**VLLM Embedding Backend Support:**

* Added a new `VLLMEmbeddingModelStage` in `nemo_curator/stages/text/embedders/vllm.py` that uses VLLM to generate embeddings, submitting all texts at once for optimal throughput. The stage manages model loading, resource allocation, and batch processing.
* Updated `EmbeddingCreatorStage` in `nemo_curator/stages/text/embedders/base.py` to support a `backend` argument (`"transformers"` or `"vllm"`), and to initialize the appropriate embedding stage with backend-specific options. [[1]](diffhunk://#diff-3ca91b0db00eb504da92a051679c039dc22edc1fd8664954e74dc0b7bc618217R103-R130) [[2]](diffhunk://#diff-3ca91b0db00eb504da92a051679c039dc22edc1fd8664954e74dc0b7bc618217R139-R147) [[3]](diffhunk://#diff-3ca91b0db00eb504da92a051679c039dc22edc1fd8664954e74dc0b7bc618217R169-R185)

**Benchmarking and Pipeline Enhancements:**

* Added a new benchmarking script `benchmarking/scripts/embedding_backend_benchmark.py` to compare the performance of transformers and VLLM embedding backends, with configurable model, backend, and execution options.

**Benchmarks**

VLLM on A100:
```
2025-12-18 08:17:42,180 - INFO - Finished | elapsed=326.68s | backend=vllm | model=google/embeddinggemma-300m | executor=XennaExecutor | files=8 | fpp=1 | docs=80000 | throughput=244.88 docs/s | out=embedding_output.parquet
2025-12-18 08:17:42,180 - INFO - ============================================================
2025-12-18 08:17:42,180 - INFO - BENCHMARK RESULTS
2025-12-18 08:17:42,180 - INFO - ============================================================
2025-12-18 08:17:42,180 - INFO - Backend: vllm
2025-12-18 08:17:42,180 - INFO - Documents processed: 80000
2025-12-18 08:17:42,181 - INFO - Total time: 326.68 seconds
2025-12-18 08:17:42,181 - INFO - Throughput: 244.88 docs/second
[...]

PR #1346, opened 2025-12-31
Add vLLM and Sentence Transformers support for embedding generation

praateekmahajan:
## Description
Tried out few models 

1. `sentence_transformer` : Current implementation of TokenizerStage \+ ModelStage  
2. `vllm_text` : vLLM with text as input \- Single stage of vLLM only  
   1. `vllm_text_4cpus` : Same as above, except we dedicate 4 CPUs to the stage  
3. `vllm_tokens` : vLLM with tokens as input, tokenization done by TokenizerStage  
4. `vllm_text_pretokenized` :  vLLM with text as input, tokenization done within the stage  
   

The experiment ran on 5gb of common crawl data and the findings were

1. Dependent on model size  
   1. Larger model (embedding gemma): vLLM wins  
   2. Smaller model (miniLM) : Sentence Transformer Current implementation wins  
2. When measuring average time per task / partition (combining tokenization and modeling) the fastest, irrespective of model size is `vllm_text_pretokenized`. This suggests given a large number of tasks and amortized startup cost, we should see `vllm_text_pretokenized` come out ahead.   
3. For large model  
   1. Among vLLM variations, total time taken is similar, but that includes startup time, so we should measure average time spent per task.  
4. For small model  
   1. Among vLLM variations, `vllm_text` is slowest, likely due to tokenization happening sentence by sentence leading to higher GPU idle time, and not justifying the small model runtime on GPU. Increasing cpu allocation for the stage doesn’t improve runtimes.  
5. vllm\_tokens is also fast for both small / large models, but it increases the following cost for a negligible overhead compared to `vllm_text_pretokenized`;   
   1. maintenance overhead of maintaining a separate tokenization stage   
   2. serialization overhead of transferring data from one actor to another

image
image

## Usage

```python
```
## Checklist


nemo_curator/stages/text/embedders/vllm.py
```diff
@@ -0,0 +1,138 @@
...
+from nemo_curator.stages.base import ProcessingStage
+from nemo_curator.stages.resources import Resources
+from nemo_curator.stages.text.models.utils import format_name_with_suffix
+from nemo_curator.tasks import DocumentBatch
+
+if TYPE_CHECKING:
+    from transformers import AutoTokenizer
+
+
+class VLLMEmbeddingModelStage(ProcessingStage[DocumentBatch, DocumentBatch]):
```
VibhuJawa, line 31: I dont understand why we have a separate VLLMEmbeddingModelStage . 

From a user workflow perspective, should we not always use the same interface and let user choose the backend ?

VibhuJawa: Like, when the eventual switch happens for VLLM as default,  this will make users/tutorials go and change their workflow.

VibhuJawa, line 31: Also, should we not import this logic for ModelStage here ???

nemo_curator/stages/text/embedders/vllm.py
```diff
@@ -0,0 +1,138 @@
...
+
+            self.tokenizer = AutoTokenizer.from_pretrained(self.model_identifier)
+
+    def process(self, batch: DocumentBatch) -> DocumentBatch:
+        df = batch.to_pandas()
+        input_data = df[self.text_field].tolist()
+        metrics = {}
+
+        if self.pretokenize:
+            from vllm.inputs import TokensPrompt
```
sarahyurick, line 121: Move to top-level imports?

nemo_curator/stages/text/embedders/vllm.py
```diff
@@ -0,0 +1,138 @@
...
+            vllm_init_kwargs["runner"] = "pooling"
+        if "model_impl" not in vllm_init_kwargs:
+            # TODO: Once transformers is bumpted to 5.0 then we should also support transformers
+            vllm_init_kwargs["model_impl"] = "vllm"
+
+        # Reduce verbosity when not in verbose mode
+        if not self.verbose and "disable_log_stats" not in vllm_init_kwargs:
+            vllm_init_kwargs["disable_log_stats"] = True
+
+        self.model = LLM(model=self.model_identifier, **vllm_init_kwargs)
```
sarahyurick, line 89: These are the parameters I was experimenting with:
```
self.model = vllm.LLM(
    model=self.model_identifier,
    max_num_seqs=self.model_inference_batch_size,
    max_num_batched_tokens=self.max_num_batched_tokens,
    gpu_memory_utilization=self.gpu_memory_utilization,
    # task="embed",
    tensor_parallel_size=1,
    enforce_eager=False,
    disable_log_stats=True,
    # runner="pooling",
)
```

I am wondering if we should include this logic too:
```
if "max_num_seqs" not in vllm_init_kwargs:
    vllm_init_kwargs["max_num_seqs"] = self.model_inference_batch_size
```

nemo_curator/stages/text/embedders/vllm.py
```diff
@@ -0,0 +1,138 @@
...
+        if self.model is None:
+            self._initialize_vllm()
+        if self.pretokenize:
+            from transformers import AutoTokenizer
+
+            self.tokenizer = AutoTokenizer.from_pretrained(self.model_identifier)
+
+    def process(self, batch: DocumentBatch) -> DocumentBatch:
+        df = batch.to_pandas()
+        input_data = df[self.text_field].tolist()
```
sarahyurick, line 117:
```suggestion
        input_data = df[self.text_field].tolist()

        if self.max_chars is not None:
            input_data = [input_data[:self.max_chars] for text in input_data]
```
?

nemo_curator/stages/text/embedders/vllm.py
```diff
@@ -0,0 +1,138 @@
...
+                raise ValueError(msg)
+
+            t0 = time.perf_counter()
+            max_model_len = self.model.model_config.max_model_len
+            tokenized_data = self.tokenizer.batch_encode_plus(input_data, truncation=True, max_length=max_model_len)
+            input_data = [TokensPrompt(prompt_token_ids=ids) for ids in tokenized_data.input_ids]
+            metrics["tokenization_time"] = time.perf_counter() - t0
+
+        t0 = time.perf_counter()
+        vllm_output = self.model.embed(input_data, truncate_prompt_tokens=-1, use_tqdm=self.verbose)
```
sarahyurick, line 136: Is there an advantage to doing tokenization then the model forward pass within the same stage? I was thinking that we should either break tokenization into a separate CPU-only stage or always let `self.model.embed(...)` handle it like:

```suggestion
        vllm_output = self.model.embed(input_data, truncate_prompt_tokens=self.max_seq_length, use_tqdm=self.verbose)
```

nemo_curator/stages/text/embedders/vllm.py
sarahyurick, line 1: I think the file shouldn't be called `vllm.py`, to avoid possible import weirdness with the `vllm` library.

tests/stages/text/embedders/test_vllm.py
```diff
@@ -0,0 +1,120 @@
...
+
+pytest.importorskip("vllm", reason="VLLMEmbeddingModelStage tests require vllm")
+
+import numpy as np
+import pandas as pd
+import torch
+from sentence_transformers import SentenceTransformer
+
+from nemo_curator.stages.text.embedders.vllm import VLLMEmbeddingModelStage
+from nemo_curator.tasks import DocumentBatch
```
sarahyurick: Instead of `pytest.importorskip`, let's use a `with suppress(ImportError):` block for any GPU-dependent imports.

sarahyurick (commented):
Added some comments I had from https://github.com/NVIDIA-NeMo/Curator/pull/1327. Really excited about this work!

nemo_curator/stages/text/embedders/vllm.py
praateekmahajan: I haven't benchmarked the perf for these kwargs so I feel reluctant to add them, I'd prefer if users pass it through model_kwargs

nemo_curator/stages/text/embedders/vllm.py
praateekmahajan: I can add `max_chars` here

nemo_curator/stages/text/embedders/vllm.py
praateekmahajan: > Is there an advantage to doing tokenization then the model forward pass within the same stage

Yup, if you see the description, the `vllm_tokens`, `vllm_text` and `vllm_pretokenize`.. `vllm_pretokenize` comes in most ahead, followed by `vllm_text`. `vllm_tokens` is also close but we have no reason to have an extra stage which incurs serialization overhead and increase dev mantainenance .

nemo_curator/stages/text/embedders/vllm.py
praateekmahajan: Do you have thoughts on what it should be? Given the import needs to be `from x.y.z.vllm import vLLMEmbeddingStage` I don't see it to be an issue

nemo_curator/stages/text/embedders/vllm.py
praateekmahajan: I'm not a fan of top level imports especially in-case of imports that don't need to get resolved. For instance in case of `pretokenize=False` there is no need to for user to resolve `from transformers import AutoTokenizer` which will have its own plethora of imports.

nemo_curator/stages/text/embedders/vllm.py
praateekmahajan: Our `ModelStage` is currently more like `PreTokenizedSortedTextModelStage` as it has a lot of helpers from there. We don't really reuse many functions from there, so there isn't a need to extend that.

However if the intent is we should unify model interfaces of our repos.. 100%, we're not doing great on that and we should, however that warrants a bigger overhaul something that also considers image / audio / video modalities

praateekmahajan: Good question I am **very** reulctant to having a monolithy stage where everything is seemingly controlled by `"backend=vllm/sentence_transformers/nn_module"` because in reality it doesn't end up like that.Here is my reasoning

- The backends have materially different surfaces (pooling options for `nn_module`, vLLM-specific kwargs like `pretokenize` or `vllm_init_kwargs`, sentence-transformers lacking pooling and requiring `model_inference_batch_size` etc.). A single “backend” flag forces one stage to carry many backend-only parameters that are dead or invalid in other paths, making validation, docs, UX and developer experience worse.
- We’ve seen this pattern go wrong (e.g., in our Dask’s code base we had one  `read(..., format=...)`): the code became convoluted, and users still had to learn format-specific caveats. Changing a default later still required users to adjust arguments because of backend-only options.
- Separate stages keep each interface tight, validated, and testable; errors stay specific to the backend. It also keeps tutorials and maintenance simpler—no giant monolith of conditional logic.

This bloat however will show up in `TextSemanticWorkflow` which does many things including emebedding generation. If we want to support both vllm and sentence transformers, we'll have the same argument bloat in our workflow construction (and there is already so much bloat because it also does kmeans / pairwise / removal so all those arguments), in my opinion, we should support one backend by default; users who need another backend can compose/override the embedding stage explicitly.

VibhuJawa: I understand the point that vLLM / sentence-transformers / nn_module have materially different parameter surfaces (pooling semantics, vLLM init kwargs, pretokenization options, etc.). Where I disagree is the conclusion that this forces _**multiple public stages**_ or forces users to _**manually re-compose workflows.**_

I’d prefer we optimize for user effort by keeping a single default “Embedding” stage as the public entry point, but avoid the “giant backend-specific kwargs soup” by structuring it as a facade + typed backend configs:

- EmbeddingStage(model_identifier=..., backend="vllm", backend_config=VllmConfig(...))
- EmbeddingStage(..., backend="sentence_transformers", backend_config=SentenceTransformersConfig(...))

This gives us:

- a stable, easy tutorial/workflow story (“use EmbeddingStage”) with reasonable defaults,
- backend-specific validation and docs (each config is tight and testable),
- and avoids a single constructor accumulating dead/invalid params across backends.

It also reduces migration burden in workflows: new backends become new config types rather than new stage classes + new tutorial branches. Users with advanced needs can still instantiate the backend-specific stage/config directly, but the default path stays unified.

nemo_curator/stages/text/embedders/vllm.py
sarahyurick: `vllm_model.py` or `vllm_embedder.py` could work.

nemo_curator/stages/text/embedders/vllm.py
sarahyurick: For most kwargs I agree, yes. I guess it depends on how closely we want to stay aligned with `ModelStage`/`EmbeddingModelStage` with init parameters, like `model_inference_batch_size`.

nemo_curator/stages/text/embedders/vllm.py
sarahyurick: I don't really see a strong reason for allowing multiple options unless we would expect there to be advantages when switching models (i.e., model A performs better with `pretokenize=True` but model B performs better with `pretokenize=False`).

The precedence from before was to delegate tokenization to a CPU-only stage, with serialization overhead not being a huge issue (due to `max_chars` and `max_seq_len`). In the current implementation here, tokenization is done during a GPU stage no matter what.

nemo_curator/stages/text/embedders/vllm.py
VibhuJawa: We synced offline, the goal of making users use the best/fastest framework for 90% of the cases will be done by doing a fast follow to this PR where the text embedding/semdedup workflows will move on to VLLM backend.

VibhuJawa: We discussed this, the goal indeed is to enforece unification/same signature and some functional contracts b/w all the backend stages.  

We are aligned on that as long as we achieve that. My opinion is to have it in this PR for the text stages, other stages can do a fast follow/unification

PR #1360, opened 2026-01-10
[benchmarking] Adds audio curation benchmark to nightly

rlratzel:
Adds an audio benchmark to the nightly benchmark suite.

This benchmark is based on the [current audio example](https://github.com/NVIDIA-NeMo/Curator/blob/main/tutorials/audio/fleurs/pipeline.py) and adds code to save metadata and results used by the benchmarking framework.

_Note to reviewers: this PR depends on features in https://github.com/NVIDIA-NeMo/Curator/pull/1341, so it has been merged with this branch.  When https://github.com/NVIDIA-NeMo/Curator/pull/1341 is merged, the diff should only include changes needed for adding the audio benchmark._

### DIFF
--- a/benchmarking/runner/utils.py
+++ b/benchmarking/runner/utils.py
@@ -77,9 +77,6 @@ def _replace_env_var(match: re.Match[str]) -> str:
         raise ValueError(msg)
 
 
-_env_var_pattern = re.compile(r"\$\{([^}]+)\}")  # Pattern to match ${VAR_NAME}
-
-
 def resolve_env_vars(data: dict | list | str | object) -> dict | list | str | object:
     """Recursively resolve environment variables in strings in/from various objects.
 


benchmarking/scripts/audio_fleurs_benchmark.py
```diff
@@ -0,0 +1,194 @@
...
+def read_jsonl(file_paths: Path | list[Path], executor: XennaExecutor) -> Pipeline:
+    """Read jsonl files from one or more directories."""
+    p = Pipeline(name="read", description="Read jsonl")
+    # Build a list of directory path strings from Path object(s), as required by JsonlReader
+    file_path_strs = [str(d) for d in (file_paths if isinstance(file_paths, list) else [file_paths])]
+    p.add_stage(JsonlReader(file_paths=file_path_strs))
+    return p.run(executor)
+
+
+def assert_valid_pipeline(pipeline: Pipeline, expected_values: argparse.Namespace) -> None:
```
praateekmahajan: can we remove these? I'm not sure if we need these

benchmarking/scripts/audio_fleurs_benchmark.py
```diff
@@ -0,0 +1,194 @@
...
+            )
+        )
+
+        assert_valid_pipeline(pipeline, args)
+
+        results = pipeline.run(executor)
+        assert len(results) == _expected_num_results
+        predict = read_jsonl(results_dir, executor)
+        assert len(predict) == _expected_num_results
+        success = True
```
praateekmahajan: we don't need the asserts.. we can just have them as metrics and we can verify in our nightly the results indeed match what we expect.. can we also then get rid of read_jsonl function? thanks

benchmarking/scripts/audio_fleurs_benchmark.py
```diff
@@ -0,0 +1,194 @@
...
+
+    except Exception as e:  # noqa: BLE001
+        logger.error(f"Error running audio fleurs benchmark: {e}\n{traceback.format_exc()}")
```
praateekmahajan: well all our benchmarks should follow the same try except.. and i don't think we have it, so can we remove this here too

praateekmahajan (approved):
LGTM!

PR #1372, opened 2026-01-14
Fix bug in Gliner tutorial

sarahyurick:


tutorials/text/gliner-pii-redaction/gliner_pii_redactor.py
```diff
@@ -164,7 +164,7 @@ def process(self, batch: DocumentBatch) -> DocumentBatch | None:
 
         # Run model inference via the GLiNER library
         # This returns a list of dictionaries per document, each containing the entities and their confidence scores
-        entities = self.model.run(
+        entities = self.model.inference(
```
ayushdg, line 167: Is this because of some new release in the gliner library? If so, can we also update the readme slightly to include `install gliner>=version_with_this` or maybe the current version.

sarahyurick: Yes. I can add a comment about the Gliner version.

PR #1373, opened 2026-01-14
Add benchmarking for `ScoreFilter`

sarahyurick:


benchmarking/nightly-benchmark.yaml
```diff
@@ -179,6 +179,34 @@ entries:
...
     timeout_s: 20000
 
+  - name: score_filter_raydata
+    enabled: true
+    script: score_filter_benchmark.py
+    args: >-
+      --benchmark-results-path={session_entry_dir}
+      --executor=ray_data
+      --input-path={dataset:tinystories,parquet}
+      --yaml-config=/opt/Curator/nemo_curator/config/text/heuristic_filter_english_pipeline.yaml
```
praateekmahajan: @rlratzel how can we point this to curator path is `/opt/Curator` the right way to do it? I think `curator_repo_dir` can work too right?

sarahyurick: Yes, I would appreciate advice here. I tried `--yaml-config={curator_repo_dir}/nemo_curator/config/text/heuristic_filter_english_pipeline.yaml` but got an error.

rlratzel: yes, `{curator_repo_dir}` should work (as seen [here](https://github.com/NVIDIA-NeMo/Curator/blob/a086dd436ae830dc82eb07bfdf00777c7801a5f2/benchmarking/nightly-benchmark.yaml#L184)). Support for this was recently added so make sure you have the latest. I have not tried using that placeholder in script args though like you are so there could be a bug.

benchmarking/scripts/score_filter_benchmark.py
```diff
@@ -0,0 +1,196 @@
+# Copyright (c) 2025, NVIDIA CORPORATION.  All rights reserved.
```
rlratzel: I'm assuming this should be 2026

benchmarking/nightly-benchmark.yaml
sarahyurick: Pulling main worked for me, thanks!

Issue #1381, opened 2026-01-15
Track number of filtered/kept documents per `DocumentFilter`


sarahyurick:
It would be great to track the number of filtered and/or kept documents per `Filter`/`ScoreFilter` stage.

Once we support this, we can use it in benchmarks (https://github.com/NVIDIA-NeMo/Curator/pull/1373) and tutorials.

PR #1385, opened 2026-01-16
Add metrics for `ScoreFilter` benchmarks

sarahyurick:
Closes https://github.com/NVIDIA-NeMo/Curator/issues/1381.

### DIFF
--- a/benchmarking/nightly-benchmark.yaml
+++ b/benchmarking/nightly-benchmark.yaml
@@ -248,6 +248,11 @@ entries:
       --yaml-config={curator_repo_dir}/nemo_curator/config/text/heuristic_filter_english_pipeline.yaml
       --overrides="stages.0._target_=nemo_curator.stages.text.io.reader.ParquetReader"
     timeout_s: 20000
+    sink_data:
+      - name: slack
+        additional_metrics:
+          - num_kept_documents
+          - throughput_docs_per_sec
     ray:
       num_cpus: 64
       num_gpus: 0
@@ -260,6 +265,8 @@ entries:
       - metric: num_kept_documents
         min_value: 2090470
         max_value: 2090490
+      - metric: throughput_docs_per_sec
+        min_value: 19000
 
   - name: score_filter_xenna
     enabled: true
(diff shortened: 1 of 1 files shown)


benchmarking/nightly-benchmark.yaml
```diff
@@ -252,6 +252,15 @@ entries:
...
       enable_object_spilling: false
+    requirements:
+      # ensure the total number of documents processed is correct
+      - metric: num_documents_processed
+        min_value: 2119489
+        max_value: 2119489
+    # account for stochastic filters
+      - metric: num_kept_documents
+        min_value: 2090470
+        max_value: 2090490
```
sarahyurick, line 267: This corresponds with the dataset at `/raid/prospector-lm/clean/tinystories_train_parquet`.

benchmarking/nightly-benchmark.yaml
```diff
@@ -252,6 +252,15 @@ entries:
       num_cpus: 64
       num_gpus: 0
       enable_object_spilling: false
+    requirements:
+      # ensure the total number of documents processed is correct
+      - metric: num_documents_processed
+        min_value: 2119489
```
praateekmahajan: I think we also have `exact_value` that @rlratzel added recently

sarahyurick: Thanks! Updated.

benchmarking/nightly-benchmark.yaml
```diff
@@ -252,6 +252,15 @@ entries:
...
+        max_value: 2119489
+    # account for stochastic filters
+      - metric: num_kept_documents
+        min_value: 2090470
+        max_value: 2090490
```
praateekmahajan, line 270: Another requirement to add here would be to catch perf regressions with ray data / xenna would be `throughput_docs_per_sec` which for ray data is `19,808` and xenna `8,716`.. you can multiply that by 0.95 to get a 5% buffer.. so I'd say
```suggestion
    requirements:
      # Observed throughput of 19,800 docs/sec so we allow a 5% buffer to account for variability
      - metric: throughput_docs_per_sec
        min_value: 18810

```

benchmarking/nightly-benchmark.yaml
```diff
@@ -252,6 +252,14 @@ entries:
       num_cpus: 64
       num_gpus: 0
       enable_object_spilling: false
```
praateekmahajan, line 259: Just realised `score_filter` doesn't have

```suggestion
    sink_data:
      - name: slack
        additional_metrics:
          - num_kept_documents
          - throughput_docs_per_sec
```

benchmarking/nightly-benchmark.yaml
sarahyurick: Sure. Going to push slightly different numbers than what you listed. The numbers I was actually seeing were ~20k and ~9k. LMK what you think.

PR #1389, opened 2026-01-16
[REVIEW] Benchmarking Script for E2E Pipeline

VibhuJawa:
This pull request adds a new configuration file for running local end-to-end benchmarking of the Common Crawl pipeline. The configuration includes local paths, global settings, and a single enabled entry for executing the pipeline with specific arguments and resource allocations.

Configuration for local benchmarking:

* Added `benchmarking/local-cc-e2e.yaml` to define local paths for results, datasets, and models, as well as global settings like default timeout and scratch deletion.
* Configured a single pipeline entry (`cc_e2e_pipeline_local`) with script arguments, resource limits (CPUs, GPUs, object store size), and Ray-specific settings for local execution.

### DIFF
--- a/.github/workflows/config/.secrets.baseline
+++ b/.github/workflows/config/.secrets.baseline
@@ -1,0 +1,213 @@
+{
+  "version": "1.5.0",
+  "plugins_used": [
+    {
+      "name": "ArtifactoryDetector"
+    },
+    {
+      "name": "AWSKeyDetector"
+    },
+    {
+      "name": "AzureStorageKeyDetector"
+    },
+    {
+      "name": "Base64HighEntropyString",
+      "limit": 4.5
+    },
+    {
+      "name": "BasicAuthDetector"
+    },
+    {
+      "name": "CloudantDetector"
+    },
+    {
+      "name": "DiscordBotTokenDetector"
+    },
+    {
+      "name": "GitHubTokenDetector"
+    },
+    {
+      "name": "GitLabTokenDetector"
+    },
+    {
+      "name": "HexHighEntropyString",
+      "limit": 3.0
+    },
+    {
+      "name": "IbmCloudIamDetector"
+    },
+    {
+      "name": "IbmCosHmacDetector"
+    },
+    {
+      "name": "IPPublicDetector"
+    },
+    {
+      "name": "JwtTokenDetector"
+    },
+    {
+      "name": "KeywordDetector",
+      "keyword_exclude": ""
+    },
+    {
+      "name": "MailchimpDetector"
+    },
+    {
+      "name": "NpmDetector"
+    },
+    {
+      "name": "OpenAIDetector"
+    },
+    {
+      "name": "PrivateKeyDetector"
+    },
+    {
+      "name": "PypiTokenDetector"
+    },
+    {
+      "name": "SendGridDetector"
+    },
+    {
+      "name": "SlackDetector"
+    },
+    {
+      "name": "SoftlayerDetector"
+    },
+    {
+      "name": "SquareOAuthDetector"
+    },
+    {
+      "name": "StripeDetector"
+    },
+    {
+      "name": "TelegramBotTokenDetector"
+    },
+    {
+      "name": "TwilioKeyDetector"
+    }
+  ],
+  "filters_used": [
+    {
+      "path": "detect_secrets.filters.allowlist.is_line_allowlisted"
+    },
+    {
+      "path": "detect_secrets.filters.common.is_ignored_due_to_verification_policies",
+      "min_level": 2
+    },
+    {
+      "path": "detect_secrets.filters.heuristic.is_indirect_reference"
+    },
+    {
+      "path": "detect_secrets.filters.heuristic.is_likely_id_string"
+    },
+    {
+      "path": "detect_secrets.filters.heuristic.is_lock_file"
+    },
+    {
+      "path": "detect_secrets.filters.heuristic.is_not_alphanumeric_string"
+    },
+    {
+      "path": "detect_secrets.filters.heuristic.is_potential_uuid"
+    },
+    {
+      "path": "detect_secrets.filters.heuristic.is_prefixed_with_dollar_sign"
+    },
+    {
+      "path": "detect_secrets.filters.heuristic.is_sequential_string"
+    },
+    {
+      "path": "detect_secrets.filters.heuristic.is_swagger_file"
+    },
+    {
+      "path": "detect_secrets.filters.heuristic.is_templated_secret"
+    },
+    {
+      "path": "detect_secrets.filters.regex.should_exclude_file",
+      "pattern": [
+        "pyproject\\.toml|\\.github/workflows/config/\\.secrets\\.baseline"
+      ]
+    }
+  ],
+  "results": {
+    ".github/workflows/cicd-main.yml": [
+      {
+        "type": "Hex High Entropy String",
+        "filename": ".github/workflows/cicd-main.yml",
+        "hashed_secret": "9b048088dfb38c9b23ea87ab0226e4b80c3fce4f",
+        "is_verified": false,
+        "line_number": 95
+      }
+    ],
+    "docs/_extensions/ai_assistant/README.md": [
+      {
+        "type": "Secret Keyword",
+        "filename": "docs/_extensions/ai_assistant/README.md",
+        "hashed_secret": "aecdccc1cf64595b34e0cc152d238daabb32183a",
+        "is_verified": false,
+        "line_number": 76
+      },
+      {
+        "type": "Secret Keyword",
+        "filename": "docs/_extensions/ai_assistant/README.md",
+        "hashed_secret": "6d9c68c603e465077bdd49c62347fe54717f83a3",
+        "is_verified": false,
+        "line_number": 146
+      }
+    ],
+    "docs/_extensions/ai_assistant/core/AIClient.js": [
+      {
+        "type": "Base64 High Entropy String",
+        "filename": "docs/_extensions/ai_assistant/core/AIClient.js",
+        "hashed_secret": "54f034f9ba6c3f65758ffe3bb1264c5a836ce826",
+        "is_verified": false,
+        "line_number": 10
+      }
+    ],
+    "docs/_extensions/ai_assistant/core/main.js": [
+      {
+        "type": "Base64 High Entropy String",
+        "filename": "docs/_extensions/ai_assistant/core/main.js",
+        "hashed_secret": "54f034f9ba6c3f65758ffe3bb1264c5a836ce826",
+        "is_verified": false,
+        "line_number": 15
+      }
+    ],
+    "tests/models/client/test_openai_client.py": [
+      {
+        "type": "Secret Keyword",
+        "filename": "tests/models/client/test_openai_client.py",
+        "hashed_secret": "3acfb2c2b433c0ea7ff107e33df91b18e52f960f",
+        "is_verified": false,
+        "line_number": 41
+      }
+    ],
+    "tutorials/synthetic/README.md": [
+      {
+        "type": "Secret Keyword",
+        "filename": "tutorials/synthetic/README.md",
+        "hashed_secret": "aecdccc1cf64595b34e0cc152d238daabb32183a",
+        "is_verified": false,
+        "line_number": 18
+      }
+    ],
+    "tutorials/text/deduplication/semantic/semantic_e2e.ipynb": [
+      {
+        "type": "Base64 High Entropy String",
+        "filename": "tutorials/text/deduplication/semantic/semantic_e2e.ipynb",
+        "hashed_secret": "344b84ad013e4ab6518c0dc457416855b7b86904",
+        "is_verified": false,
+        "line_number": 736
+      }
+    ],
+    "tutorials/text/deduplication/semantic/semantic_step_by_step.ipynb": [
+      {
+        "type": "Base64 High Entropy String",
+        "filename": "tutorials/text/deduplication/semantic/semantic_step_by_step.ipynb",
+        "hashed_secret": "344b84ad013e4ab6518c0dc457416855b7b86904",
+        "is_verified": false,
+        "line_number": 728
+      }
+    ]
+  },
+  "generated_at": "2026-01-30T23:03:31Z"
+}
--- a/.github/workflows/detect-secrets.yml
+++ b/.github/workflows/detect-secrets.yml
@@ -1,0 +1,21 @@
+# Copyright (c) 2026 NVIDIA CORPORATION.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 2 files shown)


benchmarking/local-cc-e2e.yaml
```diff
@@ -0,0 +1,35 @@
...
+      --snapshot=2024-30
+      --url_limit=1
+      --record_limit=100
+      --executor=ray_data
+    timeout_s: 3600
+    ray:
+      num_cpus: 4
+      num_gpus: 0
+      enable_object_spilling: false
+    object_store_size_bytes: 11474836480
```
praateekmahajan: - Instead of a new file can you add it to nightly and then run with `--entries cc_e2e_pipeline_local`
- we probably should have more CPUs to test benchmarking
- for url limit = 1, that's only downloading one file.. and if download takes 10 minutes other N-1 cpus would just be idle.. so should probably have `url_limit = K * num_cpus` to test parallelism and load balancing
- for `record_limit` wdyt of removing it? because we know there is a backpressure issue in Ray Data if the stages are actors, so we can test for that later.. and IIRC when record_limit is higher iterate is slower than download.. i'm hoping in future when we combine iterate and extract we probably see perf benefits


for `fasttext-model-path` and `hf-home` we could probably use something in datasets path itself? wdyt?

benchmarking/scripts/cc_e2e_pipeline_benchmark.py
```diff
@@ -0,0 +1,403 @@
...
+        "--classifier_batch_size", type=int, default=256, help="Batch size for classifier model inference"
+    )
+
+    # Executor selection
+    p.add_argument("--executor", type=str, default="ray_data", choices=["xenna", "ray_data"])
+
+    # Ray cluster configuration
+    p.add_argument("--ray_temp_dir", type=str, default="/raid/vjawa/ray_tmp", help="Ray temporary directory")
+    p.add_argument("--num_gpus", type=int, default=None, help="Number of GPUs for Ray cluster")
+    p.add_argument("--num_cpus", type=int, default=None, help="Number of CPUs for Ray cluster")
```
praateekmahajan: Curious why is this here? Is this to logs stuff in `params.json`? If so I think `run.py` should do it since that's the one that that starts the cluster. I believe logging this and object store details is on @rlratzel todo already

benchmarking/scripts/cc_e2e_pipeline_benchmark.py
```diff
@@ -0,0 +1,403 @@
...
+    PunctuationFilter,
+    RepeatedLinesFilter,
+    RepeatingTopNGramsFilter,
+    UrlsFilter,
+    WordCountFilter,
+)
+from nemo_curator.stages.text.io.writer import JsonlWriter, ParquetWriter
+from nemo_curator.stages.text.modules.add_id import AddId
+from nemo_curator.stages.text.modules.score_filter import ScoreFilter
+
```
praateekmahajan: Probably pull main and use utils which has setup_executor and write_benchmark_results utilities

benchmarking/local-cc-e2e.yaml
praateekmahajan: Can we add more metrics to track such as number of rows in the beginning and number of rows in the end (exact_value).. 
and then probably something for throughput (min_value_

benchmarking/scripts/cc_e2e_pipeline_benchmark.py
```diff
@@ -0,0 +1,403 @@
...
+    start = time.perf_counter()
+
+    try:
+        results = pipeline.run(executor, initial_tasks=None)
+        success = True
+    except Exception as e:  # noqa: BLE001
+        logger.error(f"Pipeline failed: {e}")
+        results = []
+        success = False
```
rlratzel: This isn't a showstopper, but in a recent PR we've updated the pattern we're using in benchmark scripts to remove the catch-all exception handlers in favor of a try-finally block in `main()` (see [here](https://github.com/NVIDIA-NeMo/Curator/blob/058fa93d79a42e937bae350fdb8a5908c4c58d79/benchmarking/scripts/fuzzy_dedup_identification_benchmark.py#L136-L141)). This cleans up code and lets us see detailed tracebacks in benchmark `stdouterr` logs while still writing out results.

benchmarking/scripts/cc_e2e_pipeline_benchmark.py
rlratzel: +1, as seen [here](https://github.com/NVIDIA-NeMo/Curator/blob/058fa93d79a42e937bae350fdb8a5908c4c58d79/benchmarking/scripts/score_filter_benchmark.py#L31)
Standardizing these utils will help us if (when) we need to update common operations such as what/how data is exported to the framework from scripts, env setup and capture, etc.

VibhuJawa: Thanks for the input, standardized on using write_benchmark_results, and setup_executor

benchmarking/scripts/cc_e2e_pipeline_benchmark.py
VibhuJawa: moved the logic to try/finally logic

benchmarking/scripts/cc_e2e_pipeline_benchmark.py
VibhuJawa: Removed them, they were left around for my local testing. Cleaned it up.

benchmarking/local-cc-e2e.yaml
VibhuJawa: Added more metrics like `num_tar_files`, `num_input_documents`, `num_output_documents`,  first, i have not added `throughput ` as a requirment now.  Please take a look (because i have not run on the benchmarking machine)

VibhuJawa: Addressed the reivew: 
- Added something like arxiv_e2e_pipeline_*
- Moved to a local downloaded tar dump (to simulate the workflow better) 
- We run on downloaded 45 tar files now, can increase as a PR followup for more larger scale testing as needed
- Removed record_limti
- Added paths, let me know if that works.

benchmarking/scripts/arxiv_e2e_pipeline_benchmark.py
```diff
@@ -0,0 +1,514 @@
...
+    # Log configuration
+    logger.info("Starting ArXiv E2E pipeline execution...")
+    if args.download_from_s3:
+        logger.info(f"Input mode: S3 download (url_limit={args.url_limit}, record_limit={args.record_limit})")
+    else:
+        logger.info(f"Input mode: Local tar files from {args.tar_input_path}")
+
+    start = time.perf_counter()
+    results = pipeline.run(executor, initial_tasks=None)
+    elapsed = time.perf_counter() - start
```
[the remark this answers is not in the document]
VibhuJawa: Do not think i want to log bare exceptions.

benchmarking/scripts/arxiv_e2e_pipeline_benchmark.py
```diff
@@ -0,0 +1,503 @@
...
+    UrlsFilter,
+    WordCountFilter,
+)
+from nemo_curator.stages.text.io.writer import JsonlWriter, ParquetWriter
+from nemo_curator.stages.text.modules.add_id import AddId
+from nemo_curator.stages.text.modules.score_filter import ScoreFilter
+from nemo_curator.tasks import DocumentBatch, _EmptyTask
+
+# Default paths
+DEFAULT_TAR_INPUT_PATH = "/datasets/prospector-lm/arxiv_downloads"
```
sarahyurick: Let's not hardcode this?

sarahyurick (commented):
Very nice PR. Added a minor comment.

benchmarking/scripts/arxiv_e2e_pipeline_benchmark.py
VibhuJawa: yup, removed it. Thanks

benchmarking/scripts/arxiv_e2e_pipeline_benchmark.py
```diff
@@ -0,0 +1,499 @@
...
+    else:
+        logger.info(f"Input mode: Local tar files from {args.tar_input_path}")
+
+    start = time.perf_counter()
+    results = pipeline.run(executor, initial_tasks=None)
+    elapsed = time.perf_counter() - start
+
+    # Calculate metrics from stage performance data
+    num_tar_files = len(results) if results else 0
+    num_input_documents = get_aggregated_stage_stats(results, "extract_", "num_items_processed")
```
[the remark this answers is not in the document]
VibhuJawa: This is probably incorrect as it works today.

benchmarking/scripts/arxiv_e2e_pipeline_benchmark.py
```diff
@@ -0,0 +1,499 @@
...
+        super().__init__()
+
+    def decompose(self) -> list[ProcessingStage]:
+        """Decompose into constituent stages."""
+        return self.stages
+
+
+def create_e2e_pipeline(  # noqa: PLR0913
+    # Input source options
+    tar_input_path: str | None,
```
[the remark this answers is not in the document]
VibhuJawa: Fixed this by actually forcing us  to either provide`download_from_s3`  or `tar_input_path`

benchmarking/scripts/utils.py
```diff
@@ -94,3 +94,20 @@ def convert_paths_to_strings(obj: object) -> object:
     else:
         retval = obj
     return retval
+
+
+def get_aggregated_stage_stats(results: list | None, stage_prefix: str, stat: str) -> float | int:
```
praateekmahajan: this is a great utility! Definitely useful across benchmarks.

nit : a better place for this would be `nemo_curator/tasks/utils.`py where we already have other functions.. ideally there it'll also work with `list[Task] | WorkflowResult`

praateekmahajan: Or idk if this works for your use case but something I've done is https://github.com/NVIDIA-NeMo/Curator/blob/6348b44a5db4f285132d17ba4897f3ee8974f0bd/benchmarking/scripts/dedup_removal_benchmark.py#L122C5-L126

Basically use `TaskPerfUtils.aggregate_task_metrics` and then use the key to get `process_num_items_sum`

VibhuJawa: Added the util: get_aggregated_stage_stat using `aggregate_task_metrics `

benchmarking/scripts/arxiv_e2e_pipeline_benchmark.py
```diff
@@ -0,0 +1,499 @@
...
+        )
+    else:
+        logger.success("Output documents: 0")
+    logger.success(
+        f"Throughput: {throughput_tar_files_per_sec:.2f} tar files/sec, {throughput_docs_per_sec:.1f} docs/sec"
+    )
+
+    return {
+        "params": {
+            "tar_input_path": args.tar_input_path,
```
praateekmahajan, line 365: Can we dump `vars(args)` in case there are any that we unintentionally missed?

benchmarking/scripts/arxiv_e2e_pipeline_benchmark.py
```diff
@@ -0,0 +1,499 @@
...
+    )
+    input_group.add_argument(
+        "--download_from_s3",
+        action="store_true",
+        help="Download tar files from S3 instead of using local files",
+    )
+
+    # ========== DOWNLOAD/PROCESSING OPTIONS ==========
+    download_group = p.add_argument_group("Download/Processing Options")
+    download_group.add_argument("--download_path", type=str, default="./arxiv_e2e_downloads")
```
praateekmahajan: Super nit, but since argparse is used for cli, and cli commands are kebab-case can we make all these kebab-case?

e.g. `download_path -> download-path` 

e.g. of cli commands when multiworded being kebab case are `ls --human-readable` or `grep --ignore-case`

benchmarking/nightly-benchmark.yaml
```diff
@@ -382,3 +390,61 @@ entries:
...
+      num_gpus: 4
+      enable_object_spilling: false
+    requirements:
+      # Data integrity checks
+      - metric: num_tar_files
+        exact_value: 45
+      - metric: num_input_documents
+        exact_value: 116383
+      - metric: num_output_documents
+        exact_value: 61474
```
praateekmahajan, line 513: awesome!

nemo_curator/tasks/utils.py
```diff
@@ -136,3 +136,30 @@ def aggregate_task_metrics(
                         else:
                             metrics[metric_key] = 0.0
         return metrics
+
+    @staticmethod
+    def get_aggregated_stage_stat(
+        tasks: list[Task] | Mapping[str, list[Task]] | None,
```
praateekmahajan: thanks for moving here

nit, can this be list[Task] or `WorkflowResult` and then we can check `if isinstance(x, WorkflowResult): x.whatever`

benchmarking/nightly-benchmark.yaml
```diff
@@ -382,3 +390,61 @@ entries:
...
+      --executor=xenna
+    timeout_s: 3600
+    sink_data:
+      - name: slack
+        additional_metrics:
+          - throughput_docs_per_sec
+          - num_output_documents
+          - num_input_documents
+    ray:
+      num_cpus: 64
```
praateekmahajan: let's reduce number of CPUs as discussed offline since number of files in offline setting are only 47 to maybe 16.. i'd guess we'll have to change the timeout too

benchmarking/nightly-benchmark.yaml
```diff
@@ -36,6 +36,14 @@ datasets:
     formats:
     - type: "files"
       path: "{datasets_path}/mscoco/model_weights"
+  - name: "arxiv_downloads"
```
praateekmahajan, line 54: upto you if you think additional clarity would be helpful, but it'll be nice to document explicitly or internally which batches of arxiv are used
something like `pdf/arXiv_pdf_1001_001` to `pdf/arXiv_pdf_1001_002` or maybe it's all of `pdf/arXiv_pdf_1001_*` (if it is then maybe we can rename the name to `arxiv_downloads_1001` or to `arxiv_downloads_1001_to_1002`

benchmarking/scripts/arxiv_e2e_pipeline_benchmark.py
VibhuJawa: Added.

benchmarking/nightly-benchmark.yaml
VibhuJawa: Added path for now, lets get this in and clean it up later on

benchmarking/nightly-benchmark.yaml
```diff
@@ -42,6 +43,14 @@ datasets:
...
     - type: "parquet"
       path: "{datasets_path}/rpv2_2023-14_en"
+  - name: "arxiv_downloads"
+    formats:
+    - type: "tar"
+      path: "{datasets_path}/arxiv_downloads"
+  - name: "fasttext_model"
+    formats:
+    - type: "bin"
+      path: "{model_weights_path}/fasttext/lid.176.bin"
```
sarahyurick, line 61: For the image and video models we do it like:
```
  - name: "mscoco_model_weights"
    formats:
    - type: "files"
      path: "{datasets_path}/mscoco/model_weights"
```
and
```
  - name: "videos_model_weights"
    formats:
    - type: "files"
      path: "{datasets_path}/video_model_weights"
```
so do you think we should just do `path: "{datasets_path}/fasttext/lid.176.bin"` to match?

PR #1407, opened 2026-01-20
Add benchmarking for modifiers

sarahyurick:


benchmarking/nightly-benchmark.yaml
```diff
@@ -318,6 +318,42 @@ entries:
...
 
+  - name: modifier_raydata
+    enabled: true
+    script: modifier_benchmark.py
+    args: >-
+      --benchmark-results-path={session_entry_dir}
+      --output-path={session_entry_dir}/scratch/output
+      --executor=ray_data
+      --input-path={dataset:tinystories,parquet}
+    timeout_s: 20000
```
praateekmahajan: do we want to change this timeout?

sarahyurick: Yes, I can update.

PR #1409, opened 2026-01-21
[benchmarking] Bug fixes and UX improvements

rlratzel:
- Allows disabled blocks to be completely ignored. This fixes the issue of a disabled slack sink requiring an env var to be set, and also allows for incomplete blocks or blocks with known errors to remain in the file as disabled for later attention.
- No longer prints stderr output when checking ray status command to clean up output and prevent the appearance of an actual error.
- Allows users to use Ray's default value for object store size by accepting `default` as a value. This can be used to override a session-wide default setting (see below)
- Exposes default `object_store_size` as a session-level param instead of a baked-in default to improve understanding. This can be overridden by entries as before.

### DIFF
--- a/benchmarking/nightly-benchmark.yaml
+++ b/benchmarking/nightly-benchmark.yaml
@@ -57,10 +57,10 @@ sinks:
 # Whether to delete scratch dirs after each run
 delete_scratch: true
 
-# The size of the object store used by Ray, which can be either a value in bytes (int) or a
-# fraction of total system memory (float), or the value "default" (string) which represents
-# the default value for the object store size as used by "ray start".
-object_store_size: 0.5
+# The size of the object store used by Ray which can be either a value in bytes (int), or
+# a fraction of total system memory (float), or the value "default" (string) which allows
+# for "ray start" to determine object store size.
+object_store_size: 536870912000  # 500GB
 
 entries:
   - name: domain_classification_raydata
@@ -95,8 +95,6 @@ entries:
         exact_value: 149816
       - metric: domain_label_news_count
         exact_value: 2817
-    # override the session-level object_store_size setting for this entry
-    object_store_size: 214748364800
 
   - name: domain_classification_xenna
     enabled: true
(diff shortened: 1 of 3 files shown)


benchmarking/nightly-benchmark.yaml
```diff
@@ -90,8 +95,9 @@ entries:
         exact_value: 149816
       - metric: domain_label_news_count
         exact_value: 2817
+    # override the session-level object_store_size setting for this entry
+    object_store_size: 214748364800
```
praateekmahajan: This was unintentional most likely. We just need to use the default. 

I do think we should have a default of some reasonable number like 200 or 500gb since this number affects the performance, and we'd rather be prescriptive than having to second guess what object store ended up being used

praateekmahajan: I'm okay to have default of 0.5 as long as we log details from runtime.. see comment below

benchmarking/run.py
```diff
@@ -172,15 +172,15 @@ def run_entry(
...
             ray_log_path=logs_path / "ray.log",
-            object_store_size_bytes=entry.object_store_size_bytes,
+            object_store_size=None if entry.object_store_size == "default" else entry.object_store_size,
         )
 
         # Prepopulate <session_entry_path>/params.json with entry params.
         # These will be appended with the benchmark params by the benchmark script.
         (session_entry_path / "params.json").write_text(
             json.dumps(
                 {
```
praateekmahajan, line 183: WDYT of logging stuff from runtime as @ayushdg recommended

> `ray.cluster_resources()` exposes a nice dictionary we can log/use
```
{'accelerator_type:A100': 1.0,
 'node:__internal_head__': 1.0,
 'node:10.184.206.10': 1.0,
 'CPU': 128.0,
 'object_store_memory': 644245094400.0,
 'memory': 1225750720512.0,
 'GPU': 4.0}

```

Maybe new set of keys?

benchmarking/nightly-benchmark.yaml
rlratzel: I set it to 500GB based on offline discussion but I can certainly change it back to `0.5` since we log the object store size.

benchmarking/run.py
rlratzel: There was a placeholder for ray data in `results.json` which I used for this. Would we rather this data go in `params.json`? Since it's actual data returned from the cluster, it seems less like params to the benchmark and more like results, but I don't feel strongly.

PR #1424, opened 2026-01-23
Bump pyasn1 due to cve GHSA-63vm-454h-vhhq for r1.1.0

ayushdg:
## Usage

```python
```
## Checklist


ayushdg:
> shouldn't we add this bump in the constraints like the other CVEs?


Since this isn't a direct dependency but an indirect one coming from other packages I figured just bumping the uv lock file is sufficient. I'm not sure if that's the criteria. @thomasdhc Do you have thoughts?

praateekmahajan:
IIUC that's what `constraint-dependencies` solves for https://docs.astral.sh/uv/reference/settings/#constraint-dependencies

thomasdhc:
pyasn1 is already constrained https://github.com/NVIDIA-NeMo/Curator/blob/r1.1.0/pyproject.toml#L186

ayushdg:
Closing since it's already been updated.

PR #1444, opened 2026-01-30
ci: Optimize docker layer and uv with no cache

thomasdhc:
## Usage

```python
```
## Checklist

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -75,6 +75,8 @@ RUN git clone https://github.com/OpenGVLab/InternVideo.git && \
 FROM nemo_curator_dep AS nemo_curator
 
 WORKDIR /opt/Curator
+
+# Selective file copying before dependency installation for better layer caching
 COPY pyproject.toml uv.lock /opt/Curator/
 COPY nemo_curator/__init__.py nemo_curator/package_info.py /opt/Curator/nemo_curator/
 


docker/Dockerfile
```diff
@@ -73,14 +72,17 @@ RUN git clone https://github.com/OpenGVLab/InternVideo.git && \
...
+
+WORKDIR /opt/Curator
+COPY pyproject.toml uv.lock /opt/Curator/
+COPY nemo_curator/__init__.py nemo_curator/package_info.py /opt/Curator/nemo_curator
+
 # Install Curator
+RUN uv sync --link-mode copy --locked --extra all --all-groups --no-cache && \
+  uv add /opt/InternVideo/InternVideo2/multi_modality
+
 COPY . /opt/Curator
```
rlratzel, line 87: I'm not very familiar with `uv`, but I'm curious why we need to copy individual files from the repo on lines 78-79 if we just end up copying the entire repo on line 85? Can we just move line 85 up before the install and remove the individual file copy calls? I'm assuming 'no', but I'm curious why.

thomasdhc: This optimizes docker re-build. If we copy over the entire Curator directory, every time Curator gets updated the cache will break leading to re-install of all dependencies. By isolating files needed for dependency vs changes to the code base, re-install does not need to happen if dependencies have not changed.

rlratzel: Thank you for for explaining that. That's an excellent reason. 

One minor request, feel free to ignore if you think it's obvious: would you mind also adding a comment to the Dockerfile mentioning that? I'm afraid a future maintainer might not pick up on that and refactor it out.

thomasdhc: Yeah let me add that note

PR #1470, opened 2026-02-07
(fix): refactor audio stage names to be shown after running benchmark

### LINKED ISSUE #1464: Audio stage names aren't being propogated as noticed in benchmark script
**Describe the bug**

In the tasks.pkl outputted by audio benchmark (https://github.com/NVIDIA-NeMo/Curator/blob/main/benchmarking/scripts/audio_fleurs_benchmark.py) if we look at StagePerfStats we'll notice none of the stage names are being propogated therefore its hard to see which stages took how much time

SwekeR-463:
## Description

Fixes #1464 

- Preserve `_stage_perf` when stages return new task instances.
- Define explicit name fields for audio stages to populate `StagePerfStats` stage names.

## Snippet 

After re running `python benchmarking/run.py --config benchmarking/nightly-benchmark.yaml --entries audio_fleurs` got this output.

image

## Checklist

### DIFF
--- a/nemo_curator/stages/audio/common.py
+++ b/nemo_curator/stages/audio/common.py
@@ -92,7 +92,6 @@ class PreserveByValueStage(LegacySpeechStage):
         input_value_key: str,
         target_value: int | str,
         operator: str = "eq",
-        **kwargs,
     ):
         self.input_value_key = input_value_key
         self.target_value = target_value


praateekmahajan:
Hey @SwekeR-463 thanks for the PR, since few of the classes where you added name extend from `LegacySpeechStage` I wonder if you can just make the change in the base class.

Can you try making the same changes as https://github.com/NVIDIA-NeMo/Curator/pull/1433/changes and seeing if it works

Also I'm not sure if the changes in `stages/base.py` are needed, can you make the changes as shown in #1433 and try running the pipeline without the changes in `stages/base.py` I'm hoping it works

SwekeR-463:
> Can you try making the same changes as https://github.com/NVIDIA-NeMo/Curator/pull/1433/changes and seeing if it works
> 
> Also I'm not sure if the changes in `stages/base.py` are needed, can you make the changes as shown in #1433 and try running the pipeline without the changes in `stages/base.py` I'm hoping it works

No it didnt worked after making the changes you said for `stages/base.py`.

image

SwekeR-463:
> Can you try making the same changes as https://github.com/NVIDIA-NeMo/Curator/pull/1433/changes and seeing if it works

Worked with the having the same changes from #1433 and keeping my changes for `stages/base.py`.

Screenshot from 2026-02-08 08-42-23

nemo_curator/stages/audio/common.py
```diff
@@ -29,6 +29,10 @@ class LegacySpeechStage(ProcessingStage[Task, Task]):
 
     """
 
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+        self.name = self.__class__.__name__
```
sarahyurick: This seems to only work as intended for `PreserveByValueStage` since that is the only child that does a `super().__init__`.

nemo_curator/stages/base.py
```diff
@@ -180,9 +180,16 @@ def process_batch(self, tasks: list[X]) -> list[Y]:
...
+            if result is None:
+                continue
             if isinstance(result, list):
+                for r in result:
+                    if r is not task and hasattr(r, "_stage_perf") and not r._stage_perf:
+                        r._stage_perf = list(task._stage_perf)
                 results.extend(result)
             else:
+                if result is not task and hasattr(result, "_stage_perf") and not result._stage_perf:
+                    result._stage_perf = list(task._stage_perf)
```
sarahyurick: I am able to reproduce what @SwekeR-463 has seen with vs without this change. Although I would like to investigate if there is something that can be changed on the audio side instead of here...

nemo_curator/stages/audio/io/convert.py
```diff
@@ -24,6 +24,8 @@ class AudioToDocumentStage(ProcessingStage[AudioBatch, DocumentBatch]):
 
     """
 
+    name = "AudioToDocumentStage"
```
sarahyurick, line 27: Ah I think I found the bug! The `process` function below does not explicitly forward the `_stage_perf` list when returning the `DocumentBatch`.

We may need to update this elsewhere too... then we should be able to remove the changes from `nemo_curator/stages/base.py`.

@SwekeR-463 let me know what you think?

sarahyurick: Fix: make sure `_stage_perf=task._stage_perf` is being propagated everywhere when an `AudioBatch`/`DocumentBatch` is being returned.

praateekmahajan: Thanks a ton @sarahyurick for debugging this 🙏 

@SwekeR-463 here is an example of how other stages propogate it https://github.com/NVIDIA-NeMo/Curator/blob/90ce791c40473758e1e83ece0e1114b3c28ddf66/nemo_curator/stages/text/modules/add_id.py#L80-L81

SwekeR-463: > Fix: make sure `_stage_perf=task._stage_perf` is being propagated everywhere when an `AudioBatch`/`DocumentBatch` is being returned.

made changes as per mentioned. if any other changes required please tell. 😄

nemo_curator/stages/base.py
```diff
@@ -180,6 +180,8 @@ def process_batch(self, tasks: list[X]) -> list[Y]:
                 raise ValueError(msg)
 
             result = self.process(task)
+            if result is None:
+                continue
```
praateekmahajan: Why do we need this? Is there a stage that's returning `None`? Technically that's not allowed so can you share which one is that?

SwekeR-463: My bad sir forgot to remove this part after I made the required changes

nemo_curator/stages/audio/common.py
```diff
@@ -32,7 +32,11 @@ class LegacySpeechStage(ProcessingStage[Task, Task]):
     def process(self, task: AudioBatch) -> list[Task]:
         result = []
         for entry in task.data:
-            result.extend(self.process_dataset_entry(entry))
+            entries = self.process_dataset_entry(entry)
```
sarahyurick, line 35: This PR seems to work, based on my minimal testing. Since `AudioToDocumentStage` and `LegacySpeechStage` are so fast (~0 seconds) on the tiny audio benchmark that we support, I added `time.sleep(5)` to the former's `process` function and `time.sleep(9)` to the latter's `process` function. Here is the result I got:

Image

nemo_curator/stages/audio/common.py
```diff
@@ -32,7 +32,11 @@ class LegacySpeechStage(ProcessingStage[Task, Task]):
...
+            for r in entries:
+                if r is not task and not r._stage_perf:
```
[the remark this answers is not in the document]
sarahyurick: Another implementation I tried that worked for me was:
```
def process(self, task: AudioBatch) -> list[Task]:
        incoming_perf = list(getattr(task, "_stage_perf", None) or [])
        result = []
        for entry in task.data:
            entries = self.process_dataset_entry(entry)
            for r in entries:
                if r is not task:
                    r._stage_perf = list(incoming_perf)
            result.extend(entries)
        return result
```

It produced the same results as the current state of the PR.

nemo_curator/stages/audio/common.py
```diff
@@ -80,14 +85,15 @@ class PreserveByValueStage(LegacySpeechStage):
...
     """
 
+    name = "PreserveByValueStage"
+
     def __init__(
         self,
         input_value_key: str,
         target_value: int | str,
         operator: str = "eq",
         **kwargs,
```
sarahyurick:
```suggestion
```

sarahyurick:
Thank you for your contribution @SwekeR-463 !

PR #1474, opened 2026-02-09
Add ADV (Audio DataVerse) dependencies to audio extras

shubhamNvidia:
## Summary
- Added Audio DataVerse (ADV) dependencies to `audio_cpu` extras: Cython, packaging, torchaudio, seaborn, onnx, onnxruntime, torchcodec, silero-vad
- Added GPU-specific dependencies to `audio_cuda12` extras: nvidia-cudnn-cu12, onnxruntime-gpu

## Description
These dependencies are required to support the Audio DataVerse (ADV) pipeline in NeMo Curator. The additions include libraries for audio processing, VAD (Voice Activity Detection), and ONNX runtime inference.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -84,7 +84,7 @@ deduplication_cuda12 = [
 ]
 
 
-audio_cpu = [
+audio_common = [
     "nemo_toolkit[asr]==2.4.0",
     # ADV (Audio DataVerse) Dependencies
     "Cython",
@@ -92,15 +92,18 @@ audio_cpu = [
     "torchaudio",
     "seaborn==0.11.2",
     "onnx==1.19.0",
-    "onnxruntime==1.20.1",
     "torchcodec",
     "silero-vad",
 ]
+audio_cpu = [
+    "nemo_curator[audio_common]",
+    "onnxruntime==1.20.1",
+]
 audio_cuda12 = [
-    "nemo_curator[audio_cpu]",
+    "nemo_curator[audio_common]",
     "nemo_curator[cuda12]",
     "nvidia-cudnn-cu12",
-    "onnxruntime-gpu==1.20.1" ,
+    "onnxruntime-gpu==1.20.1",
 ]
 
 image_cpu = [


pyproject.toml
```diff
@@ -83,12 +83,24 @@ deduplication_cuda12 = [
...
+    "onnx==1.19.0",
+    "onnxruntime==1.20.1",
+    "torchcodec",
+    "silero-vad",
 ]
 audio_cuda12 = [
     "nemo_curator[audio_cpu]",
     "nemo_curator[cuda12]",
+    "nvidia-cudnn-cu12",
+    "onnxruntime-gpu==1.20.1" ,
```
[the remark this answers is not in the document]
shubhamNvidia: resolved

pyproject.toml
```diff
@@ -83,12 +83,27 @@ deduplication_cuda12 = [
...
-audio_cpu = [
+
+audio_common = [
     "nemo_toolkit[asr]==2.4.0",
+    # ADV (Audio DataVerse) Dependencies
+    "Cython",
+    "packaging",
+    "torchaudio",
+    "seaborn==0.11.2",
+    "onnx==1.19.0",
```
praateekmahajan, line 94: any reasons they're pinned rather than `>=`

pyproject.toml
```diff
@@ -83,12 +83,27 @@ deduplication_cuda12 = [
     "rapidsmpf-cu12==25.10.*",
 ]
 
-audio_cpu = [
+
+audio_common = [
```
praateekmahajan, line 87: Are some of the dependencies here that don't make sense in the CPU world and are GPU only?
E.g. onnx or torch?

shubhamNvidia: Not quite, onnx and torch are not GPU-only; both provide full CPU support and are widely used in CPU-only environments.
In this configuration, onnxruntime==1.20.1 is the CPU runtime, while onnxruntime-gpu, nvidia-cudnn-cu12, and nemo_curator[cuda12] are the components that specifically target GPU usage.
So the dependencies listed under audio_common are appropriate for CPU.

pyproject.toml
shubhamNvidia: yes you are right i will change it to >=

shubhamNvidia:
Verified locally with uv venv -p 3.12 and uv sync --extra audio_cpu --extra audio_cuda12 --all-groups — both audio extras resolve and install cleanly with no dependency conflicts. The only --all-extras failure is flash-attn (from video_cuda12) which needs nvcc at build time and is unrelated to the audio changes. Audio dependencies are conflict-free on Python 3.12.

PR #1476, opened 2026-02-09
Add feedback to tutorials

sarahyurick:


ayushdg (commented):
Quick question: What's the goal of importing gpustat? Is it to check if text_cuda12 was installed or not?

sarahyurick:
@ayushdg yes I just chose it because it is a GPU dependency in https://github.com/NVIDIA-NeMo/Curator/blob/main/pyproject.toml#L72. Happy to swap it out or remove it though.

thomasdhc:
Bypassing CI check as tutorials only change

PR #1480, opened 2026-02-10
[benchmarking] Updates Slack sink to address table size limitation

rlratzel:
This PR refactors the Slack sink used by the benchmark runner, motivated by recently hitting the Slack table size limit.

The changes allow the Slack sink to post each benchmark entry result as an individual message in a threaded reply to a main "summary" message. Each message has its own table and is much less likely to hit the 100 row table size limit.
By using the Slack SDK and a Slack app bot, the sink can now also update existing messages. This allows the sink to post the benchmarks to run at the start of the session and update the individual status as they are running. The old Slack sink only posted a single message at the end of the run.

- Updates other sinks to match new sink API changes motivated by changes to the Slack sink.
- Updates Slack sink to use Slack SDK instead of webhook URL and posts as bot to create threads and live updates.
- Adds Message classes to encapsulate message creation code unique to each type and to maintain the data and Slack post responses needed by the API.
- Adds a new "live_updates" config option: True posts updates to benchmarks as they are run, False posts all message at the end of the run (similar to the old behavior).
- Replaces the "webhook_url" config option with "channel_id", since the webhook URL is no longer used.
- Updates the module `__main__` to alow the module to act as a standalone script that will post a Slack report from pre-existing benchmark results.


benchmarking/nightly-benchmark.yaml
```diff
@@ -70,7 +70,8 @@ sinks:
 #    experiment: ray-curator-common-crawl
   - name: slack
     enabled: true
-    webhook_url: ${SLACK_WEBHOOK_URL}
+    live_updates: true
+    channel_id: ${SLACK_CHANNEL_ID}
```
ayushdg, line 77: As someone who may not want to set these for test runs or write to a slack sink will run.py error if these are not set?

rlratzel: Sorry I should have emphasized that the latest version of the framework will handle the "enabled" option for the sink more thoroughly.  If you simply set `enabled: false` for the Slack sink, won't attempt to load the sink at all and you can then leave those two config vars as-is and it shouldn't complain.

rlratzel: ...although setting `enabled: false` requires an annoying edit to the nightly YAML file.  Maybe I should have the sink print a warning if it's enabled but the env vars aren't set and have it disable itself?  Although, it might also be annoying to get an hour into a long run only to find out the env vars weren't set and you won't get a report. WDYT?

rlratzel: ...another option could be to promote devs having their own override YAML file so they don't have to edit the nightly YAML (and accidentally commit it, etc.)

ayushdg: I'm not too concerned about needing to modify the yaml file and setting it to false for now. We can figure out if we need to enhance this further based on dev/usage feedback. For the time being my only concern was that even with `enabled: false` the framework would still require these vars to be set which doesn't seem to be the case anymore!

ayushdg:
/ok to test [2848b10](https://github.com/NVIDIA-NeMo/Curator/pull/1480/commits/2848b108e4478ba4628b2aab8eb32d8d7d729f56)

praateekmahajan (approved):

PR #1490, opened 2026-02-11
Add check to make sure all reserved keywords are attributes and not properties

### LINKED ISSUE #1481: Make sure all reserved keywords are attributes and not properties [name / resources / batch_size]
`ProcessingStage.with_()` (and `CompositeStage’s` use of it) works only when subclasses define `name`, `resources`, and `batch_size` as normal class or instance attributes. If a stage overrides them with read-only `@property`, reading (e.g. _name, _resources, get_config()) works, but `with_()` fails with “property has no setter” because it assigns to new_instance.name / new_instance.resources / new_instance.batch_size on the deep-copied instance. The docs already say to use attributes rather than properties; this issue is to record that using @property is unsupported and breaks with_(), and to consider either documenting the limitation clearly or changing the implementation (e.g. avoid assigning to these names, or support property-based overrides) so that property-based stages work or are explicitly rejected with a clear error.

One known file where this is an issue - https://github.com/NVIDIA-NeMo/Curator/blob/90ce791c40473758e1e83ece0e1114b3c28ddf66/nemo_curator/stages/synthetic/nemotron_cc/base.py#L45-L47

omkar-334:
closes #1481


nemo_curator/stages/base.py
```diff
@@ -112,6 +112,15 @@ def __init_subclass__(cls, **kwargs):
             msg = f"{cls.__name__} must not override '_batch_size'"
             raise TypeError(msg)
 
+        for attr in ("name", "resources", "batch_size"):
```
sarahyurick, line 115: Thanks @omkar-334 ! Can you add some tests in https://github.com/NVIDIA-NeMo/Curator/blob/main/tests/stages/common/test_base.py too?

omkar-334: Done! I've added three tests for properties and one for attributes...

omkar-334:
The commits got messed up while rebasing and signing the commits I guess..

tests/stages/common/test_base.py
```diff
@@ -340,6 +340,68 @@ def inputs(self) -> tuple[list[str], list[str]]:
...
+                name = "MockStagePropertyBatchSize"
+
+                @property
+                def batch_size(self) -> int:
+                    return 1
+
+                def process(self, task: MockTask) -> MockTask:
+                    return task
+
+    def test_plain_attribute_override_allowed(self):
```
sarahyurick: I think this may not be needed and should already be covered by `TestProcessingStageWith` which uses `ConcreteProcessingStage`. What do you think @omkar-334 ?

omkar-334: You're right... `TestProcessingStageWith` already tests for attribute overrides. Thanks for this!. Ive removed it now

PR #1496, opened 2026-02-12
Add retry on 503 error

### LINKED ISSUE #1225: Add retry logic for common crawl download
Currently common crawl does `wget -O` we should also add retry on 503

**Additional context**

omkar-334:
Fixes #1225.  

Do we need to add this for wikipedia as well?


omkar-334:
cc @praateekmahajan

nemo_curator/stages/text/download/common_crawl/download.py
```diff
@@ -70,7 +70,7 @@ def _download_to_path(self, url: str, path: str) -> tuple[bool, str | None]:
         else:
             # We don't use -c (for continue resume) because we want to download file to temp path using -O
             # but -c and -O don't work well together
-            cmd = ["wget", url_to_download, "-O", path]
+            cmd = ["wget", url_to_download, "-O", path, "--retry-on-http-error=503", "--waitretry=5", "--tries=5"]
```
sarahyurick, line 73: Based on https://status.commoncrawl.org/ retrying on 503 makes sense. Not sure if we should add others like 429, etc.

Adding `-c` could be good too.

WDYT @omkar-334 @praateekmahajan @ayushdg ?

omkar-334: hi @sarahyurick 
i went through the [wget docs](https://www.gnu.org/software/wget/manual/wget.html).  
They say that
```
A combination with ‘-O’/‘--output-document’ is only accepted if the given output file does not exist.
```  
So the comment is true, i don't think we can/should add `-c`, since the output file would already exist.  
what do you think?

sarahyurick: Fair enough, I am fine with skipping it. Thanks!

PR #1503, opened 2026-02-12
[REVIEW] Add model model_weights_path

VibhuJawa:
## Description

Adding path to add model weights. Test is running currently locally , seems to be working .

### DIFF
--- a/benchmarking/runner/path_resolver.py
+++ b/benchmarking/runner/path_resolver.py
@@ -35,7 +35,7 @@ class PathResolver:
 
     def __init__(self, data: dict) -> None:
         """
-        data is a dictionary containing the paths for results and datasets
+        data is a dictionary containing the paths for results, datasets, and model weights.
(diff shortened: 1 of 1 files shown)


rlratzel:
> LGTM, one quick update though: greptile pointed out [this line](https://github.com/NVIDIA-NeMo/Curator/blob/53d943a28d8803b05318efe7c8b190e328f7e0fb/benchmarking/runner/session.py#L83) should probably be updated now too, which seems like a good idea.

on second thought - that change will force all users to define that var even if their YAML doesn't have model weights. My main motivation was to protect against a `KeyError`, but maybe there's a better way. I don't feel too strongly though either way.

rlratzel (approved):

benchmarking/runner/path_resolver.py
sarahyurick:
```suggestion
        data is a dictionary containing the paths for results, datasets, and model weights.
```

PR #1523, opened 2026-02-18
Fix bugs in prometheus / grafana setup and support clusters with multiple ray users

praateekmahajan:
### New features
1. Support `metrics_dir` to provide isolation of prometheus / grafana 
2. Allow RayClient started with a `temp_dir` to also work with prometheus
3. Automatically add all ray dashboards (core / data / serve)

### How to use

  1. `python start_prometheus_grafana.py --metrics_dir /abcd` must be run separately and beforehand (on the head node). This creates `prometheus.yml`, downloads binaries, starts the daemons, and writes PID/port files into the `metrics_dir`.
  4. When `RayClient.start()` runs inside the ray job submit command:
    - It checks `is_prometheus_running(metrics_dir)` and `is_grafana_running(metrics_dir) `— these read PID files from the metrics dir
    - If both are running, it calls `add_ray_prometheus_metrics_service_discovery(ray_temp_dir, metrics_dir)` which
  appends the Ray head node's `prom_metrics_service_discovery.json` path to `prometheus.yml` and sends a reload to Prometheus
   - **IMPORTANT for SLURM** : This means inside the head node we must do `RayClient.start(temp_dir=???, metrics_dir=???)` and that must match how we a) start prometheus/grafana b) started ray cluster and specified `temp_dir`

## Follow up
1. Add grafana / prometheus to benchmarking

## Deep Dive
If curious how all prometheus and grafana relate to Ray I took notes for myself here http://prtk.in/blog/ray-prometheus-grafana/

## From Greptile a nice mermaid chart
```mermaid
%%{init: {'theme': 'neutral'}}%%
flowchart TD
    A[User runs start_prometheus_grafana.py with --metrics_dir] --> B[Creates per-user metrics directory]
    B --> C[Downloads & extracts Prometheus/Grafana]
    C --> D[Writes prometheus.yml with empty file_sd_configs]
    C --> E[Writes PID and port files]
    E --> F[Starts Prometheus & Grafana daemons]
    
    G[RayClient.start called with metrics_dir] --> H{Check Prometheus & Grafana running?}
    H -->|PID files exist & valid| I[add_ray_prometheus_metrics_service_discovery]
    H -->|Not running| J[Log warning message]
    
    I --> K[Read prometheus.yml]
    K --> L[Append Ray metrics path to file_sd_configs]
    L --> M[Send reload signal to Prometheus]
    
    N[RayClient.stop called] --> O[remove_ray_prometheus_metrics_service_discovery]
    O --> P[Read prometheus.yml]
    P --> Q[Remove Ray metrics path from file_sd_configs]
    Q --> R[Send reload signal to Prometheus]
    R --> S[Cleanup Ray process]
```

## Checklist

### DIFF
--- a/nemo_curator/stages/text/download/common_crawl/download.py
+++ b/nemo_curator/stages/text/download/common_crawl/download.py
@@ -70,7 +70,7 @@ class CommonCrawlWARCDownloader(DocumentDownloader):
         else:
             # We don't use -c (for continue resume) because we want to download file to temp path using -O
             # but -c and -O don't work well together
-            cmd = ["wget", url_to_download, "-O", path]
+            cmd = ["wget", url_to_download, "-O", path, "--retry-on-http-error=503", "--waitretry=5", "--tries=5"]
 
         # Always capture stderr so we can provide meaningful error messages
         if self._verbose:
--- a/tests/stages/text/download/common_crawl/test_download.py
+++ b/tests/stages/text/download/common_crawl/test_download.py
@@ -37,7 +37,7 @@ class TestCommonCrawlWARCDownloader:
         assert success is True
         assert error_message is None
         mock_run.assert_called_once_with(
-            ["wget", url, "-O", temp_path],
+            ["wget", url, "-O", temp_path, "--retry-on-http-error=503", "--waitretry=5", "--tries=5"],
             stdout=subprocess.DEVNULL,
             stderr=subprocess.PIPE,
         )
@@ -76,7 +76,7 @@ class TestCommonCrawlWARCDownloader:
         assert success is True
         assert error_message is None
         mock_run.assert_called_once_with(
-            ["wget", url, "-O", temp_path],
+            ["wget", url, "-O", temp_path, "--retry-on-http-error=503", "--waitretry=5", "--tries=5"],
             stdout=None,
             stderr=None,
         )
@@ -94,7 +94,7 @@ class TestCommonCrawlWARCDownloader:
         assert success is True
         assert error_message is None
         mock_run.assert_called_once_with(
-            ["wget", url, "-O", temp_path],
+            ["wget", url, "-O", temp_path, "--retry-on-http-error=503", "--waitretry=5", "--tries=5"],
             stdout=subprocess.DEVNULL,
             stderr=subprocess.PIPE,
         )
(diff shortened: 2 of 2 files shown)


nemo_curator/core/client.py
```diff
@@ -95,18 +99,18 @@ def start(self) -> None:
 
         if self.include_dashboard:
```
ayushdg, line 100: Currently If `include_dashboard` we always attempt to check for prom/grafana and add it. Curious if we should include another flag `include_metrics/start prom/grafana` or something of that sort instead or is it okay to follow as long as `include_dashboard` is set to true. 

The reason I bring up is because it's confusing for externally started cluster to rely on a `include_dasboard` arg in our client API for enabling/disabling metrics stuff.

nemo_curator/core/client.py
```diff
@@ -161,6 +165,13 @@ def start(self) -> None:
                 raise RuntimeError(msg)
 
     def stop(self) -> None:
+        # Remove Ray metrics service discovery entry from prometheus config
+        if self.include_dashboard:
+            try:
+                remove_ray_prometheus_metrics_service_discovery(self.ray_temp_dir, self.metrics_dir)
```
ayushdg, line 171: What happens for externally started clusters? Do we want to remove this discovery or keep it untouched?
Or maybe we form an opinion/document that you must use the Client API for prom/grafana?

nemo_curator/metrics/constants.py
```diff
@@ -12,7 +12,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-DEFAULT_NEMO_CURATOR_METRICS_PATH = "/tmp/nemo_curator_metrics"  # noqa: S108
+import os
+
+DEFAULT_NEMO_CURATOR_METRICS_PATH = f"/tmp/nemo_curator_metrics_{os.getuid()}"  # noqa: S108
```
ayushdg, line 17: Nice! Thanks for this.

ayushdg (commented):
High level questions about behavior. No concerns with the implementation

nemo_curator/core/client.py
ayushdg: fixed: We don't start by default

nemo_curator/core/client.py
ayushdg: Can stay untouched.

Issue #1236, opened 2025-11-17
Allow the same tokenizer stage to be used for multiple classifiers


sarahyurick:
Several of our text classifiers use the same tokenizer. Instead of tokenizing for each classifier, we should allow the outputs of a single tokenization step to be used for multiple models in a pipeline.

This is almost already possible through Curator's use of composite stages, but in the model stages we currently drop the tokenized columns after classification is completed. We should create a `drop_tokens` boolean to control the behavior here:

- https://github.com/NVIDIA-NeMo/Curator/blob/main/nemo_curator/stages/text/classifiers/base.py#L142
- https://github.com/NVIDIA-NeMo/Curator/blob/main/nemo_curator/stages/text/classifiers/fineweb_edu.py#L124
- https://github.com/NVIDIA-NeMo/Curator/blob/main/nemo_curator/stages/text/classifiers/prompt_task_complexity.py#L266
- https://github.com/NVIDIA-NeMo/Curator/blob/main/nemo_curator/stages/text/classifiers/aegis.py#L221

Then, we can update our documentation for how to use the same tokens for multiple classifiers (i.e., how to construct a pipeline that is `TokenizerStage` -> `ModelStage1` -> `ModelStage2` ...). This documentation should include a list of which text classifiers use the same tokenizers.

As future work, we can consider handling this logic on the Curator side, instead of the user needing to directly interact with tokenizer and model stages. In other words, we can try to add logic to detect if generated tokens for a classifier can be used later within the same pipeline for another classifier.

Issue #213, opened 2024-08-20
[FEA] Align the character pruning vs sequence length based pruning for our models. 


VibhuJawa:
**Is your feature request related to a problem? Please describe.**

We need to align the character pruning vs sequence length based pruning for our models.  We need to ensure our max_char defaults are lines up and only do it for models we care about . 

**Additional context**
https://github.com/NVIDIA/NeMo-Curator/pull/168#discussion_r1723538420

sarahyurick:
Current values of `max_chars` (pre-tokenization) and `max_seq_length` (post-tokenization) per classifier:

| Classifier | `max_chars` | `max_seq_length` | Tokenizer | Updates |
| - | - | - | - | - | 
| [nvidia/domain-classifier](https://huggingface.co/nvidia/domain-classifier) | 2000 | 512 | `DebertaV2Tokenizer` | N/A |
| [nvidia/quality-classifier-deberta](https://huggingface.co/nvidia/quality-classifier-deberta) | 6000 | 1024 | `DebertaV2Tokenizer` | N/A |
| [HuggingFaceFW/fineweb-edu-classifier](https://huggingface.co/HuggingFaceFW/fineweb-edu-classifier) | None | 512 | `BertTokenizer` | N/A |
| [nvidia/Aegis-AI-Content-Safety-LlamaGuard-Defensive-1.0](https://huggingface.co/nvidia/Aegis-AI-Content-Safety-LlamaGuard-Defensive-1.0) | 6000 | 4096 | `meta-llama/LlamaGuard-7b` | N/A |
| [nvidia/Aegis-AI-Content-Safety-LlamaGuard-Permissive-1.0](https://huggingface.co/nvidia/Aegis-AI-Content-Safety-LlamaGuard-Permissive-1.0) | 6000 | 4096 | `meta-llama/LlamaGuard-7b` | N/A |
| [nvidia/instruction-data-guard](https://huggingface.co/nvidia/instruction-data-guard) | 6000 | 4096 | `meta-llama/LlamaGuard-7b` | `max_chars=None` |
| [nvidia/multilingual-domain-classifier](https://huggingface.co/nvidia/multilingual-domain-classifier) | 2000 | 512 | `DebertaV2Tokenizer` | N/A |
| [nvidia/content-type-classifier-deberta](https://huggingface.co/nvidia/content-type-classifier-deberta) | 5000 | 1024 | `DebertaV2Tokenizer` | `max_chars=6000` |
[...]

PR #1528, opened 2026-02-19
Update token handling for classifiers

sarahyurick:
Closes https://github.com/NVIDIA-NeMo/Curator/issues/213.
Closes https://github.com/NVIDIA-NeMo/Curator/issues/1236.

### DIFF
--- a/docs/curate-text/process-data/quality-assessment/distributed-classifier.md
+++ b/docs/curate-text/process-data/quality-assessment/distributed-classifier.md
@@ -457,6 +457,7 @@ The `AegisClassifier` variants ([nvidia/Aegis-AI-Content-Safety-LlamaGuard-Defen
 ```python
 aegis_defensive_classifier = AegisClassifier(
     aegis_variant="nvidia/Aegis-AI-Content-Safety-LlamaGuard-Defensive-1.0",
+    label_field="aegis_defensive_pred",
     use_existing_tokens=False,
     keep_tokens=True,
     keep_aegis_prompt_field=True,
@@ -466,6 +467,7 @@ pipeline.add_stage(aegis_defensive_classifier)
 
 aegis_permissive_classifier = AegisClassifier(
     aegis_variant="nvidia/Aegis-AI-Content-Safety-LlamaGuard-Permissive-1.0",
+    label_field="aegis_permissive_pred",
     use_existing_tokens=True,
     aegis_prompt_field="_curator_hidden_text",  # created by aegis_defensive_classifier
     keep_tokens=False,


nemo_curator/stages/text/classifiers/base.py
```diff
@@ -187,31 +218,47 @@ class DistributedDataClassifier(CompositeStage[DocumentBatch, DocumentBatch]):
...
-                autocast=self.autocast,
-            ),
-        ]
+            )
+            self.stages.append(tokenizer_stage)
+
+        if isinstance(self.use_existing_tokens, list):
+            if len(self.use_existing_tokens) != 2:  # noqa: PLR2004
+                msg = "use_existing_tokens must be a list of two strings representing the [input_ids, attention_mask] fields"
+                raise ValueError(msg)
```
VibhuJawa: Should we make this part of __init__/ post init  or something ?

nemo_curator/stages/text/models/model.py
```diff
@@ -61,7 +62,11 @@ def __init__(  # noqa: PLR0913
         padding_side: Literal["left", "right"] = "right",
         unpack_inference_batch: bool = False,
         autocast: bool = True,
+        token_fields: list[str] | None = None,
     ):
+        if token_fields is None:
+            token_fields = [INPUT_ID_FIELD, ATTENTION_MASK_FIELD]
```
VibhuJawa: Given this repeated across, should we move it to base or something ?

nemo_curator/stages/text/classifiers/base.py
sarahyurick: Yes, good point.

nemo_curator/stages/text/models/model.py
sarahyurick: Yes good point, we should be able to just do it once.

praateekmahajan:
Should we add unittests for this?

sarahyurick:
> Should we add unittests for this?

Yes, I will add unit tests and an example for how and when to use this (e.g., which classifiers use the same tokenizer).

nemo_curator/stages/text/classifiers/base.py
sarahyurick: Actually, I think this is used as minimally as possible right now. It is used in the `__post_init__` for each base `CompositeStage` we are implementing for classifiers (`DistributedDataClassifier`, `_FineWebBaseClassifier`, `PromptTaskComplexityClassifier`, `AegisClassifier`, and `InstructionDataGuardClassifier`).

VibhuJawa: Yup, you are right !

Issue #1532, opened 2026-02-20
Explore making stage_spec overridable during stage creation


ayushdg:
**Is your feature request related to a problem? Please describe.**
`with_` allows us to specify different batch sizes/resource requirements during stage creation that can override the default. However the same isn't as straightforward when you want to override stage_spec for a specific stage (either at runtime or as a part of a composite stage).

Context: I ran into this in the DocumentDownloadExtract stages where I want to override the stage spec just for common crawl extraction but since they all inherit from the same base class it's harder to do so without exposing it as an arg in the constructor of the Stage which isn't ideal. Maybe my issue is fixed with #1459 but the overarching issue of customizing StageSpec for CompositeStages or at runtime might be worth exploring.

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

PR #1534, opened 2026-02-20
Add supporting max_calls to the ray data executor

### LINKED ISSUE #1533: Justext extraction runs into OOM issues due to memory fragmentation or leaks
**Describe the bug**
When running a long running extraction job that uses justext, I run into memory usage growing across workers until I eventually OOM if the job is large enough. This is most likely due to a memory leak or fragmentation during the extraction process (most likely from lxml) where memory isn't returned to the OS. Over many warc files this overhead adds up leading to OOM errors.

```python
import gc
import glob
import os
import resource
import sys
import time

from nemo_curator.stages.text.download.base.iterator import DocumentIterateExtractStage
from nemo_curator.stages.text.download.common_crawl.extract import CommonCrawlHTMLExtractor
from nemo_curator.stages.text.download.common_crawl.warc_iterator import CommonCrawlWarcIterator
from nemo_curator.tasks import FileGroupTask


def get_rss_mb():
    return resource.getrusage(resource.RUSAGE_SELF).ru_maxrss / 1024


def main(warc_dir, num_files=5):
    warc_files = sorted(glob.glob(os.path.join(warc_dir, "*.warc.gz")))[:num_files]
    if not warc_files:
        print(f"No .warc.gz files found in {warc_dir}")
        sys.exit(1)

    stage = DocumentIterateExtractStage(
        iterator=CommonCrawlWarcIterator(),
        extractor=CommonCrawlHTMLExtractor(),
    )

    gc.collect()
    rss_start = get_rss_mb()

    print(f"Processing {len(warc_files)} WARC files in a single process\n")
    print(f"{'File':>5} | {'RAM Used':>9} | {'Growth':>8} | {'Records':>8} | {'Time':>6}")
    print("-" * 57)

    for i, warc_file in enumerate(warc_files):
        task = FileGroupTask(task_id=f"task-{i}", dataset_name="cc", data=[warc_file])
        t0 = time.time()
        result = stage.process(task)
        nrec = len(result.data) if result.data is not None and not result.data.empty else 0
        elapsed = time.time() - t0

        del result, task
        gc.collect()

        rss = get_rss_mb()
        print(f"{i+1:>5} | {rss:>7.0f}MB | {rss - rss_start:>+6.0f}MB | {nrec:>8} | {elapsed:>5.0f}s")

    print(f"\nTotal RAM growth: {get_rss_mb() - rss_start:+.0f}MB")
    print("Memory grows because lxml/libxml2 C-heap fragments across files.")
    print("Fix: use max_calls_per_worker=1 on DocumentIterateExtractStage")
    print("     to recycle Ray workers after each file.")


if __name__ == "__main__":
    warc_dir = sys.argv[1]
    num_files = int(sys.argv[2]) if len(sys.argv) > 2 else 5
    main(warc_dir, num_files)
```

**Expected behavior**

Ray supports a `max_calls` for [tasks](https://docs.ray.io/en/latest/ray-core/api/doc/ray.remote.html) to spawn a new process every `max_calls` number of tasks it's processed precisely to fix issues like these.

In some initial testing this has proven to be promising allowing me to use a lot more concurrent cores without running into OOM.

2 drawbacks:
1. Can only use it with Tasks not Actors - To support this for actors we might need to create a restarting actor every N tasks.
2. can only use this with Ray Data since xenna is actor based only (not sure if xenna supports max_calls).

**Additional context**

ayushdg:
## Description

This PR
Fixes #1533 
- [X] Adds support for stages to set max_calls that are propagated if it's a task based stage (not actor based).
- [X] Exposes it for the iterate extract stage so that we can override it for justext extraction. Currently exposed as a constructor arg but open to cleaner ideas. Discussed a bit in #1532.
- [X] Adds tests including some tests that ensure tasks with max_calls are not fused with other tasks and lose this information.

## Usage

```python
```
## Checklist

- [N/A ] The documentation is up to date with these changes.

### DIFF
--- a/.github/actions/test-template/action.yml
+++ b/.github/actions/test-template/action.yml
@@ -48,6 +48,9 @@ inputs:
     description: "Has Azure credentials"
     required: false
     default: "false"
+  HF_TOKEN:
+    description: "Hugging Face Token"
+    required: true
   PAT:
     description: "GitHub Personal Access Token"
     required: true
@@ -96,6 +99,8 @@ runs:
 
     - name: Start container
       shell: bash
+      env:
+        HF_TOKEN: ${{ inputs.HF_TOKEN }}
       run: |
         MNT_PATH=${{ steps.azure-fileshare.outputs.mnt_path }}
 
@@ -112,6 +117,7 @@ runs:
           -d \
           --name nemo_container_${{ github.run_id }} ${ARG[@]} \
           --shm-size=64g \
+          --env HF_TOKEN=${HF_TOKEN} \
           --env RUN_ID=${{ github.run_id }} \
           --volume $(pwd)/NeMo-Curator:/workspace \
           --workdir /workspace \
--- a/.github/workflows/cicd-main.yml
+++ b/.github/workflows/cicd-main.yml
@@ -66,6 +66,8 @@ jobs:
     needs: [pre-flight, cicd-wait-in-queue]
     runs-on: ${{ matrix.os }}
     name: Unit_Test_${{ matrix.folder}}_CPU_python-${{ matrix.python-version }}
+    env:
+      HF_TOKEN: ${{ secrets.HF_TOKEN }}
     environment: nemo-ci
     if: |
       (
@@ -148,6 +150,7 @@ jobs:
           azure-client-id: ${{ secrets.AZURE_CLIENT_ID }}
           azure-tenant-id: ${{ secrets.AZURE_TENANT_ID }}
           azure-subscription-id: ${{ secrets.AZURE_SUBSCRIPTION_ID }}
+          HF_TOKEN: ${{ secrets.HF_TOKEN }}
           PAT: ${{ secrets.PAT }}
           timeout: 20
           test-data-path: ${{ needs.pre-flight.outputs.test_data_path }}
--- a/tests/stages/text/classifiers/test_classifiers.py
+++ b/tests/stages/text/classifiers/test_classifiers.py
@@ -72,10 +72,7 @@ def run_and_assert_classifier_stages(
     # Check that the tokenizer stage inputs/output columns are correct
     tokenizer_stage = stages[0]
     assert all(col in dataset.data.columns for col in tokenizer_stage.inputs()[1])
-    try:
-        tokenizer_stage.setup_on_node()
-    except RuntimeError:
-        pytest.skip("Skipping test due to flaky Hugging Face download")
+    tokenizer_stage.setup_on_node()
     tokenizer_stage.setup()
     tokenized_batch = tokenizer_stage.process(dataset)
     assert all(col in tokenized_batch.data.columns for col in tokenizer_stage.outputs()[1])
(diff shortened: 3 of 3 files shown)


ayushdg:
/ok to test [d7cf34d](https://github.com/NVIDIA-NeMo/Curator/pull/1534/commits/d7cf34d3158a03d618006710456dc2e4c40f3487)

ayushdg:
/ok to test [ab01ccd](https://github.com/NVIDIA-NeMo/Curator/pull/1534/commits/ab01ccd42cb47959b40b52239c98b6a05e3ce7a8)

nemo_curator/stages/text/download/common_crawl/stage.py
```diff
@@ -71,6 +75,13 @@ def __init__(  # noqa: PLR0913
             algorithm_kwargs=html_extraction_kwargs,
             stop_lists=stop_lists,
         )
+        if extractor_max_calls_per_worker is None and isinstance(self.extractor.algorithm, JusTextExtractor):
+            extractor_max_calls_per_worker = 2
+            logger.info(
+                "justext extraction can cause memory fragmentation and lead to OOM errors."
```
sarahyurick:
```suggestion
                "jusText extraction can cause memory fragmentation and lead to OOM errors. "
```

tests/stages/text/download/base/test_iterator.py
```diff
@@ -520,3 +521,26 @@ def test_process_all_files_fail(self, tmp_path: Path, caplog: pytest.LogCaptureF
...
+
+    def test_ray_stage_spec_with_max_calls(self) -> None:
+        """Test that ray_stage_spec returns max_calls_per_worker when set."""
+        iterator = MockDocumentIterator()
+        stage = DocumentIterateExtractStage(iterator=iterator, max_calls_per_worker=1)
+
+        spec = stage.ray_stage_spec()
+        assert spec[RayStageSpecKeys.MAX_CALLS_PER_WORKER] == 1
+
+    def test_ray_stage_spec_with_max_calls_higher_value(self) -> None:
```
sarahyurick: Could use `@pytest.mark.parametrize` here instead?

nemo_curator/stages/text/download/common_crawl/stage.py
```diff
@@ -71,6 +75,13 @@ def __init__(  # noqa: PLR0913
             algorithm_kwargs=html_extraction_kwargs,
             stop_lists=stop_lists,
         )
+        if extractor_max_calls_per_worker is None and isinstance(self.extractor.algorithm, JusTextExtractor):
+            extractor_max_calls_per_worker = 2
```
sarahyurick, line 79: How did you decide on 2?

ayushdg: Somewhat arbitrary but verified it works a snapshot. 
On average the higher this value the higher a chance for an OOM through memory fragmentation. 
A lower value like `1` will rotate the PID every warc file which has some overhead to kill and spawn a new process.

PR #1541, opened 2026-02-24
Support Ray Serve through InferenceModelConfig and InferenceServer

praateekmahajan:
## Description

```python
  from openai import OpenAI
  from nemo_curator.core.client import RayClient
  from nemo_curator.core.serve import InferenceModelConfig, InferenceServer

  # 1. Start Ray cluster
  client = RayClient()
  client.start()

  # 2. Serve a model
  config = InferenceModelConfig(
      model_identifier="google/gemma-3-27b-it",
      engine_kwargs={"tensor_parallel_size": 4},
      deployment_config={
          "autoscaling_config": {
              "min_replicas": 1,
              "max_replicas": 1,
          },
      },
  )

  with InferenceServer(models=[config]) as server:
      # 3. Query via OpenAI SDK
      oai = OpenAI(base_url=server.endpoint, api_key="unused")
      response = oai.chat.completions.create(
          model="google/gemma-3-27b-it",
          messages=[{"role": "user", "content": "Hello!"}],
      )
      print(response.choices[0].message.content)
```

### Benchmark results (single node)
image

### Followups
1. https://github.com/ray-project/ray/issues/61269
1. https://github.com/ray-project/ray/issues/61608
## Usage

```python
```
## Checklist


nemo_curator/core/serve.py
```diff
@@ -0,0 +1,385 @@
...
+            Passed directly to LLMConfig.deployment_config.
+        engine_kwargs: vLLM engine keyword arguments (tensor_parallel_size, etc.).
+            Passed directly to LLMConfig.engine_kwargs.
+    """
+
+    model_identifier: str
+    model_name: str | None = None
+    deployment_config: dict[str, Any] = field(default_factory=dict)
+    engine_kwargs: dict[str, Any] = field(default_factory=dict)
+    runtime_env: dict[str, Any] = field(default_factory=dict)
```
ayushdg, line 62: nit: Runtime env is not documented above

nemo_curator/core/serve.py
```diff
@@ -0,0 +1,385 @@
...
+                "model_id": self.model_name or self.model_identifier,
+                "model_source": self.model_identifier,
+            },
+            deployment_config=self.deployment_config,
+            engine_kwargs=self.engine_kwargs,
+            runtime_env=merged_env or None,
+        )
+
+    @staticmethod
+    def _merge_runtime_envs(
```
ayushdg, line 88: Not blocking but I can see the dict merge being a common utility in a few places like how we use `merge_Executor_configs` in a Workflow or potentially when we add composite stage overrides with dicts

nemo_curator/core/serve.py
```diff
@@ -0,0 +1,385 @@
...
+    _started: bool = field(init=False, default=False, repr=False)
+
+    def start(self) -> None:
+        """Deploy all models and wait for them to become healthy.
+
+        Raises:
+            RuntimeError: If another ModelServer is already active in this
+                process.  Only one ModelServer can run at a time because all
+                instances share the same ``/v1`` routes on the Ray Serve HTTP
+                proxy.  Stop the existing server before starting a new one.
```
ayushdg: can you elaborate a bit on this? Does this mean that we can deploy multiple models but can only have a single instance of ModelServer at any given time?

nemo_curator/core/serve.py
```diff
@@ -0,0 +1,385 @@
...
+
+        _active_servers.discard(self.name)
+        self._started = False
+
+        logger.info("Ray Serve stopped")
+
+    @property
+    def endpoint(self) -> str:
+        """OpenAI-compatible base URL for the served models."""
+        return f"http://localhost:{self.port}/v1"
```
ayushdg, line 300: I know nothing about endpoints so apologies if this doesn't make sense. 
If I have multiple models are all exposed via the same endpoint? How do I request hitting a specific model at runtime?

nemo_curator/core/serve.py
```diff
@@ -0,0 +1,385 @@
...
+                    return
+            except Exception:  # noqa: BLE001
+                if self.verbose:
+                    logger.debug(f"Health check attempt {attempt} failed, retrying...")
+            time.sleep(1)
+        msg = f"Model server did not become ready within {self.health_check_timeout_s}s"
+        raise TimeoutError(msg)
+
+    def __enter__(self):
+        self.start()
```
ayushdg, line 394: Would it be helpful to "add" a model programmatically in my driver or do you think it's better for a user to specify all their models initially and let replicas be 0 if they are not hit?

nemo_curator/pipeline/pipeline.py
```diff
@@ -189,6 +189,19 @@ def run(self, executor: BaseExecutor | None = None, initial_tasks: list[Task] |
...
+
+        if is_ray_serve_active():
+            gpu_stages = [s for s in self.stages if s.resources.requires_gpu]
+            if gpu_stages:
+                names = ", ".join(s.name for s in gpu_stages)
+                logger.warning(
+                    f"Ray Serve is active and pipeline has GPU stages: [{names}]. "
+                    "GPU resource contention may occur between served models and pipeline stages. "
+                    "Consider using a CPU-only pipeline or stopping the ModelServer before running GPU stages."
+                )
```
ayushdg: Do you prefer adding this? This almost pushes people away from chaining their other stages with a ray serve stage though ray might handle the allocation naively?

nemo_curator/pipeline/pipeline.py
```diff
@@ -189,6 +189,19 @@ def run(self, executor: BaseExecutor | None = None, initial_tasks: list[Task] |
...
         self.build()
+
+        from nemo_curator.core.serve import is_ray_serve_active
+
+        if is_ray_serve_active():
+            gpu_stages = [s for s in self.stages if s.resources.requires_gpu]
+            if gpu_stages:
+                names = ", ".join(s.name for s in gpu_stages)
+                logger.warning(
+                    f"Ray Serve is active and pipeline has GPU stages: [{names}]. "
```
ayushdg, line 211: Do we know if Xenna handles contention since it has its own allocation logic? If not for Xenna jobs we might want to disallow GPU stages completely.

nemo_curator/core/serve.py
```diff
@@ -0,0 +1,385 @@
...
+    """Check whether any ModelServer is currently running in this process."""
+    return bool(_active_servers)
+
+
+@dataclass
+class ModelConfig:
+    """Configuration for a single model to be served via Ray Serve.
+
+    Args:
+        model_identifier: HuggingFace model ID or local path (maps to model_source in LLMConfig).
```
ayushdg, line 48: I vaguely recall some of our classifier needing to support something like huggingface offline only mode. is that something we need to support here?

nemo_curator/pipeline/pipeline.py
praateekmahajan: Good point, I know xenna==0.1.2 definitely just allocates the stage to all available GPUs and in this case it gets stuck in a deadlock. I'll add an exception in that case.

tests/core/test_serve.py
```diff
@@ -0,0 +1,270 @@
...
+
+from nemo_curator.core.constants import DEFAULT_SERVE_HEALTH_TIMEOUT_S, DEFAULT_SERVE_PORT
+from nemo_curator.core.serve import ModelConfig, ModelServer, is_ray_serve_active
+
+INTEGRATION_TEST_MODEL = "HuggingFaceTB/SmolLM2-135M-Instruct"  # pragma: allowlist secret
+INTEGRATION_TEST_MODEL_2 = "HuggingFaceTB/SmolLM-135M-Instruct"  # pragma: allowlist secret
+
+
+class TestModelConfig:
+    def test_model_config_defaults(self) -> None:
```
sarahyurick: I don't think we need tests for checking that parameters are being set.

nemo_curator/core/serve.py
sarahyurick: The HF offline thing was that it was still trying to access HF to download the domain model, etc. on a job with no internet access (even though the user already had the model in cache), so an extra environment variable needed to be set.

nemo_curator/core/serve.py
praateekmahajan: I agree, let me know if you have a path where I can move this.. Probably in utils but none of the current files serve the need exactly..

nemo_curator/core/serve.py
praateekmahajan: The model arg is what'll make the difference (see PR description above)

```python
      oai = OpenAI(base_url=server.endpoint, api_key="unused")
      response = oai.chat.completions.create(
          model="google/gemma-3-27b-it", # this line
          messages=[{"role": "user", "content": "Hello!"}],
      )

```

nemo_curator/core/serve.py
praateekmahajan: Hmmm good question, I haven't though of being able to add a  model programatically. To my knowledge min_replicas=0 will serve that use case. And I might be wrong, but I think the way to do would be to stop and redeploy

nemo_curator/pipeline/pipeline.py
praateekmahajan: I did test with Ray Data and noticed that it does a fine job of scaling up things. Xenna failed. However I'd still prefer a warning but maybe I can word it less alerty

nemo_curator/core/serve.py
praateekmahajan: Let me try this too, because I tried local path and it failed maybe I was missing the extra env variable

tests/core/test_serve.py
praateekmahajan: Good catch, got rid of that and still checking the LLM config test

nemo_curator/core/serve.py
```diff
@@ -0,0 +1,408 @@
...
+
+    @property
+    def endpoint(self) -> str:
+        """OpenAI-compatible base URL for the served models.
+
+        When multiple models are deployed, clients select a model by passing
+        ``model="<model_name>"`` in the request body (standard OpenAI API
+        convention).  The ``/v1/models`` endpoint lists all available models.
+        """
+        return f"http://localhost:{self.port}/v1"
```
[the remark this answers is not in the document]
sarahyurick: +1 should this be a parameter?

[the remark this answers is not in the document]
praateekmahajan: I don't think this is true and that localhost is needed for multi-node instances.
I'm not where adding `host` would be beneficial.. Context being that Ray Serve by default uses `EveryNode` in `ProxyLocation` (see https://docs.ray.io/en/latest/serve/api/doc/ray.serve.config.ProxyLocation.html) which states
> EveryNode: run a proxy on every node in the cluster that has at least one replica actor. This is the default.


This means this endpoint is available on all the nodes.

Let me know what use case did you have in mind for host @sarahyurick maybe I'm missing something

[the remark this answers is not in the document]
sarahyurick: ok got it, thanks!

oyilmaz-nvidia (commented):
This looks good to me. The tests give an idea how the user's will use this new class but is there an end2end example/tutorial that we can check to see how the RayServe will look like?

Also, does it make sense to merge this before checking the other alternatives?

nemo_curator/core/serve.py
praateekmahajan: Update the comment

nemo_curator/core/serve.py
praateekmahajan: TL;DR: local paths do work (I tried without env-vars and they did work).. I was unable to reproduce the error that I ran into but next time I run into it I'll try with the env var.

```python
  from nemo_curator.core.serve import InferenceModelConfig, InferenceServer

  config = InferenceModelConfig(
      model_identifier="/path/to/model/snapshots/<hash>",  # local path, not HF repo ID
      model_name="my-model",  # API-facing name clients use
      deployment_config={"autoscaling_config": {"min_replicas": 1, "max_replicas": 1}},
      engine_kwargs={"tensor_parallel_size": 1, "max_model_len": 4096, "enforce_eager": True},
  )

  with InferenceServer(models=[config]) as server:
      # server.endpoint -> http://localhost:<port>/v1
      ...
```

benchmarking/nightly-benchmark.yaml
```diff
@@ -669,30 +669,33 @@ entries:
...
+  - name: ndd_ray_serve_tp4
     enabled: true
     script: ndd_benchmark.py
     args: >-
       --benchmark-results-path={session_entry_dir}
-      --input-path={dataset:gretel_symptoms,jsonl}
+      --input-path={dataset:ndd_gretel_symptoms,jsonl}
       --output-path={session_entry_dir}/scratch/output
-      --model-type=nvidia-nim
+      --inference-server-type=ray-serve
```
ayushdg, line 753: Should we also keep a super tiny entry around for the nvidia-nim path. I don't think that'll be tested on GitHub CI

benchmarking/nightly-benchmark.yaml
```diff
@@ -669,30 +669,33 @@ entries:
...
       - name: slack
         additional_metrics:
           - input_row_count
           - output_row_count
           - throughput_rows_per_sec
+          - serve_startup_s
     requirements:
       - metric: output_row_count
-        exact_value: 820
+        exact_value: 853
```
ayushdg, line 774: Any idea why this changed if the model is the same?

nemo_curator/core/serve.py
```diff
@@ -0,0 +1,397 @@
...
+           and remove the ``VLLM_LOGGING_LEVEL`` env var workaround.
+
+        2. **Ray Serve access logs** (``POST /v1/... 200 Xms``):
+           ``configure_component_logger()`` only adds the access-log filter
+           to the *file* handler, not the stderr stream handler, so
+           ``LoggingConfig(enable_access_log=False)`` has no effect on
+           console output.  Workaround: ``RAY_SERVE_LOG_TO_STDERR=0``
+           (logs still go to files under the Ray session log directory).
+        """
+        return {
```
ayushdg, line 350: nit: TODO to remove once fixed upstream

tests/core/test_serve.py
```diff
@@ -0,0 +1,255 @@
...
+
+        response = client.chat.completions.create(
+            model=INTEGRATION_TEST_MODEL,
+            messages=[{"role": "user", "content": "Say hello in one word."}],
+            max_tokens=16,
+            temperature=0.0,
+        )
+        assert len(response.choices[0].message.content) > 0
+
+        server2.stop()
```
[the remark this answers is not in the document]
praateekmahajan: If the assertion fails we have other problems

nemo_curator/pipeline/pipeline.py
```diff
@@ -189,9 +189,29 @@ def run(self, executor: BaseExecutor | None = None, initial_tasks: list[Task] |
...
+        if is_ray_serve_active():
+            gpu_stages = [s for s in self.stages if s.resources.requires_gpu]
+            if gpu_stages:
+                names = ", ".join(s.name for s in gpu_stages)
+                if isinstance(executor, XennaExecutor):
+                    msg = (
+                        f"Cannot run XennaExecutor with GPU stages [{names}] while Ray Serve is active. "
+                        "Xenna manages GPU assignment independently of Ray's resource scheduler, "
+                        "which causes GPU contention with served models. "
+                        "Use RayDataExecutor or RayActorPoolExecutor instead."
```
ayushdg: nit: While technically support I'm almost hesitant to mention RayActorPool unless you think it's still valuable to mention.

tutorials/synthetic/nemo_data_designer/ndd_data_generation_example.py
```diff
@@ -233,53 +271,101 @@ def main() -> None:
...
-
-    # Add reader stage to read the seed dataset
-    pipeline.add_stage(
-        JsonlReader(
-            file_paths=seed_dir + "/*.jsonl",
-            fields=["diagnosis", "patient_summary"],
+    from nemo_curator.backends.experimental.ray_data import RayDataExecutor
+    from nemo_curator.core.client import RayClient
+
+    client = RayClient(num_cpus=16, num_gpus=4)
```
ayushdg: Similar to other dedup tutorials can you raise here if users don't actually have 4 GPUs on their system? This help with issues seen in the past with dedup tutorials.

pyproject.toml
```diff
@@ -258,6 +271,8 @@ torchvision = [
     { index = "pytorch", marker = "platform_machine == 'x86_64' and sys_platform != 'darwin'" },
     { index = "pypi", marker = "platform_machine != 'x86_64' or sys_platform == 'darwin'" },
 ]
+nixl = { index = "pypi" }
```
ayushdg, line 280: Curious if this is needed.

praateekmahajan:
Hey @oyilmaz-nvidia thanks for the review. Update our tutorial and benchmark in this commit [`69d07b4` (this PR)](https://github.com/NVIDIA-NeMo/Curator/pull/1541/commits/69d07b4f7a3740529f4bfb601d6832bc43230805)
 
> does it make sense to merge this before checking the other alternatives

Yes — I'll explore alternatives (vLLM direct, Dynamo, SGLang) in a fast-follow PR. I already have functional code for running vLLM directly and see ~20-30% speed improvements, so the next step is making that work on multi-node with a similarly clean programmatic UX. 

That said, I want to highlight a fundamental architectural reason Ray Serve is valuable beyond just being a serving layer: it's the only option that gives us unified GPU resource management. 

If a curator pipelines with GPU stage (e.g. DomainClassifier) and an InferenceServer runs on the same Ray cluster, Ray's scheduler sees all GPUs actors - both Serve replicas and pipeline actors. Ray's scheduler will allocates GPUs without contention. For example, with `min_replicas=2` on 8 GPUs, Serve holds 2 GPUs and the remaining 6 are available for pipeline GPU stages. This just works because both systems use Ray for scheduling. 

Any non-Ray alternative (vLLM direct, SGLang, Dynamo) breaks this. The external process claims GPUs that Ray doesn't know about, so when the Curator pipeline (Ray Data / Xenna) tries to schedule GPU stages, it will over-allocate and OOM. 

There's no clean way to solve this without manual `CUDA_VISIBLE_DEVICES` partitioning, which might be slightly more complex. So while I'll continue benchmarking alternatives for pure serving throughput, Ray Serve is the only solution today that lets users run mixed CPU+GPU pipelines alongside model inference without resource conflicts. The alternatives I'll explore will target the "inference-only" use case (no GPU pipeline stages) where the resource management concern doesn't apply.

benchmarking/nightly-benchmark.yaml
praateekmahajan: Dataset was never copied over properly I believe.. dataset has 853 rows

benchmarking/nightly-benchmark.yaml
praateekmahajan: Any benefit of testing the nvidia-nim path? Since that's just HTTP calling it's also covered here.. I'm hoping this test is more meaningful for us

pyproject.toml
praateekmahajan: [pypi.nvidia.com](http://pypi.nvidia.com/) for nixl-cu12 v0.10.0 only has:
  - cp310-cp310-manylinux_2_28_x86_64
  - cp311-cp311-manylinux_2_28_x86_64
  - cp314-cp314-manylinux_2_28_x86_64

benchmarking/nightly-benchmark.yaml
```diff
@@ -743,32 +743,35 @@ entries:
...
     args: >-
       --benchmark-results-path={session_entry_dir}
-      --input-path={dataset:gretel_symptoms,jsonl}
+      --input-path={dataset:ndd_gretel_symptoms,jsonl}
       --output-path={session_entry_dir}/scratch/output
-      --model-type=nvidia-nim
+      --inference-server-type=ray-serve
       --model-id=openai/gpt-oss-20b
       --executor=ray_data
+      --engine-kwargs='{"tensor_parallel_size": 4}'
```
VibhuJawa: Why are we testing `"tensor_parallel` for `gpt-oss-20b` , that should be data parallel no ?

tutorials/synthetic/nemo_data_designer/ndd_data_generation_example.py
```diff
@@ -218,68 +255,125 @@ def _print_sample_documents(output_files: list, all_data_frames: list) -> None:
...
     else:
-        config_builder = _build_config_manually()
+        config_builder = _build_config(
+            model_id=args.model,
+            provider_name=provider_name,
+            model_alias=model_alias,
+        )
 
+    pipeline = Pipeline(
+        name="ndd_data_generation", description="Generate synthetic text data using Nemo Data Designer"
```
sarahyurick: Super duper nit:
```suggestion
        name="ndd_data_generation", description="Generate synthetic text data using NeMo Data Designer"
```

benchmarking/nightly-benchmark.yaml
praateekmahajan: Glad you asked, because I had it here for the wrong reasons.. In my mind dp=4 would take longer to instantiate than tp=4.. but I was wrong and both have same startup time of ~100s.. 
However when I ran the benchmark TP=4 comes out ahead atleast at this scale (only 853 rows) and concurrency

image


At a larger scale DP4 is faster than TP4 (benchmark results in description)

sarahyurick (approved):

benchmarking/nightly-benchmark.yaml
VibhuJawa: Could we switch to dp=4, like we dont care about small rows  and folks (like me and agents) tend to copy without seeing .

PR #1542, opened 2026-02-24
Bump ray data to 2.54 and fix an issue related to the new version of the Ray

oyilmaz-nvidia:
## Description
Bump ray data to 2.54 and fix an issue related to the new version of the Ray.

### DIFF
--- a/.github/workflows/cicd-main.yml
+++ b/.github/workflows/cicd-main.yml
@@ -93,6 +93,7 @@ jobs:
         timeout-minutes: 40
         run: |
           uv sync --link-mode copy --locked --extra audio_cpu --extra sdg_cpu --extra text_cpu --extra video_cpu --group test
+          source .venv/bin/activate
           FOLDER="${{ matrix.folder }}"
           FOLDER="${FOLDER/stages-/stages/}"
           coverage run --branch --source=nemo_curator -m pytest -v "tests/$FOLDER" -m "not gpu"


tests/conftest.py
```diff
@@ -167,39 +171,41 @@ def pytest_ignore_collect(collection_path: Path, config: pytest.Config) -> bool:
...
+            str(ray_client_server_port),
+            "--dashboard-host",
+            "0.0.0.0",  # noqa: S104
+            "--num-cpus",
+            str(num_cpus),
+            "--num-gpus",
+            str(num_gpus),
+            "--object-store-memory",
+            str(object_store_memory),
+            "--block",
```
praateekmahajan, line 225: What was the reason to remove temp-dir? 

The reason we had it was to support clusters where > 1 users are running ray would write always in `/tmp/ray` (and therefore fail to start).

tests/conftest.py
```diff
@@ -236,21 +267,46 @@ def shared_ray_cluster(tmp_path_factory: pytest.TempPathFactory, pytestconfig: p
...
+        time.sleep(1)
+    else:
+        msg = f"Ray cluster at {ray_address} did not become ready within 60 seconds"
+        raise RuntimeError(msg)
+
+    _wait_for_head_node_alive(ray_address)
+
     try:
-        yield ray_address
+        yield ray_client_address
```
ayushdg, line 309: Can we go back to using the gcs address or is the client address needed?

For context: Connecting via the ray client server port is a bit fragile and might break in a few cases that use ray data. Connecting to the GCS port if possible is recommended.

pyproject.toml
```diff
@@ -198,7 +198,7 @@ constraint-dependencies = [
     "protobuf>=5.29.6", # Address CVE GHSA-8qvm-5x2c-j2w7
     "pyasn1>=0.6.2", # Address CVE GHSA-63vm-454h-vhhq
     "python-multipart>=0.0.22", # Address CVE GHSA-wp53-j4wj-2cfg
-    "ray[default,data]>=2.52", # Address CVE GHSA-q279-jhrf-cc6v
+    "ray[default,data]>=2.54", # Address CVE GHSA-q279-jhrf-cc6v
```
ayushdg, line 201: I guess we can remove this if we're bumping the main dependency to >=2.54

oyilmaz-nvidia:
@ayushdg @praateekmahajan things in this draft PR and will close it. Created this clean PR https://github.com/NVIDIA-NeMo/Curator/pull/1557. Let's review this one.

Issue #1099, opened 2025-09-24
rayClient stop should unset ray_address


ayushdg:
**Describe the bug**
Currently when using the RayClient.start method the method sets the `ray_address` environment variable that is never unset when stopping this cluster. 
This leads to cases where a subsequent RayClient.start is a no op because it thinks that a cluster already exists.

**Steps/Code to reproduce bug**
```python
ray_client = RayClient()
ray_client.start()
ray_client.stop()

import os
print(os.environ["RAY_ADDRESS"]) # '127.0.1.1:6379'

ray_client = RayClient()
ray_client.start() # no op

# some pipeline
pipeline.run() # hangs at Connecting to existing Ray cluster at address: 127.0.1.1:6379...
```

**Expected behavior**

Client.stop cleans up on exit.

**Additional context**

Open question: Is it possible to use a different env variable like `CURATOR_RAY_ADDRESS` or do we need to use `RAY_ADDRESS`? If it's the latter, the stop method should also track if the env was set before calling cluster.start and if so it should not unset it.

jrbourbeau:
I think https://github.com/NVIDIA-NeMo/Curator/pull/1101 resolved this issue

ayushdg:
Thanks for catching that. Closing as completed via #1101

PR #1556, opened 2026-02-26
Fix benchmarking failures caused by Ray cluster cleanup and responsiveness checks

praateekmahajan:
## Description
  6de7ccc added `check_ray_responsive()` inside `RayClient.start()` with a 300s timeout (replacing the benchmarking-local version), but when `check_ray_responsive()` failed, `client.start()` raised without calling `self.stop()`, leaving the Ray GCS process alive as a zombie. 
This meant any further `ray status` calls or `RayClient.start()` calls would act on the older `RAY_ADDRESS` and also the defunct processes would never be killed.

Apart from that also fixed
1. Removed the redundant responsiveness check inside becnharmk retry loop (client.start() already verifies it internally; therefore a failed attempt would now take double the time of 600s)
2. We sleep for 10s between two entries in case some ray processes need to be cleaned.
3. `get_ray_cluster_data()` is called before the entry is run rather than after 
4. Improved code for inferring `shm` information
 
## Usage

```python
```
## Checklist

### DIFF
--- a/benchmarking/runner/ray_cluster.py
+++ b/benchmarking/runner/ray_cluster.py
@@ -95,7 +95,7 @@ def setup_ray_cluster_and_env(  # noqa: PLR0913
         try:
             client.start()
             _ensure_ray_client_process_started(client, ray_client_start_timeout_s, ray_client_start_poll_interval_s)
-            responsive = check_ray_responsive()
+            responsive = True
         except Exception:
             logger.exception(f"Ray cluster start failed on attempt {retries + 1}")
             responsive = False


benchmarking/runner/ray_cluster.py
```diff
@@ -73,14 +91,23 @@ def setup_ray_cluster_and_env(  # noqa: PLR0913
...
+
         if not responsive:
-            logger.info("Ray cluster did not become responsive in time, stopping client and retrying...")
-            client.stop()
+            logger.info("Ray cluster did not become responsive, cleaning up before retry...")
+            try:
+                client.stop()
+            except Exception:
+                logger.exception("Failed to stop client during retry cleanup")
+            os.environ.pop("RAY_ADDRESS", None)
```
ayushdg, line 109: client.stop should also unset `RAY_ADDRESS` #1099  but I guess this protects against it for cases where stop doesn't trigger?

benchmarking/runner/ray_cluster.py
```diff
@@ -73,14 +91,23 @@ def setup_ray_cluster_and_env(  # noqa: PLR0913
...
             object_store_memory=object_store_size,
         )
-        client.start()
 
-        _ensure_ray_client_process_started(client, ray_client_start_timeout_s, ray_client_start_poll_interval_s)
-        responsive = check_ray_responsive()
-        run_shm_size_check(human_readable=True)
+        try:
+            client.start()
+            _ensure_ray_client_process_started(client, ray_client_start_timeout_s, ray_client_start_poll_interval_s)
```
rlratzel, line 97: Abhinav brought this up earlier as well but I wonder if we no longer need this since `client.start()` now does more to ensure ray is running and responsive before returning. I don't feel strongly either way though.

benchmarking/runner/ray_cluster.py
```diff
@@ -73,14 +91,23 @@ def setup_ray_cluster_and_env(  # noqa: PLR0913
...
+            responsive = True
```
rlratzel, line 98: Very minor, feel free to ignore: the word "responsive" was more meaningful back when the responsive check was being used here but now I think the word "started" would be better.

PR #1557, opened 2026-02-26
Update ray to 2.54

oyilmaz-nvidia:
## Description
This PR updates Ray to 2.54.

### DIFF
--- a/benchmarking/scripts/multimodal_mint1t_benchmark.py
+++ b/benchmarking/scripts/multimodal_mint1t_benchmark.py
@@ -54,7 +54,9 @@ def create_pipeline(args: argparse.Namespace) -> Pipeline:
             per_text_fields=tuple(args.per_text_fields) if args.per_text_fields else (),
         )
     )
-    pipeline.add_stage(InterleavedAspectRatioFilterStage(drop_invalid_rows=True, min_aspect_ratio=1.0, max_aspect_ratio=2.0))
+    pipeline.add_stage(
+        InterleavedAspectRatioFilterStage(drop_invalid_rows=True, min_aspect_ratio=1.0, max_aspect_ratio=2.0)
+    )
     pipeline.add_stage(
         InterleavedParquetWriterStage(
             path=args.output_path,
--- a/nemo_curator/stages/interleaved/io/readers/webdataset.py
+++ b/nemo_curator/stages/interleaved/io/readers/webdataset.py
@@ -90,7 +90,11 @@ class WebdatasetReaderStage(BaseInterleavedReader):
     # -- source_ref construction --
 
     def _build_source_ref(
-        self, ctx: _SampleContext, content_key: str | None, *, frame_index: int | None = None,
+        self,
+        ctx: _SampleContext,
+        content_key: str | None,
+        *,
+        frame_index: int | None = None,
     ) -> str:
         if content_key is None:
             return InterleavedBatch.build_source_ref(path=None, member=None)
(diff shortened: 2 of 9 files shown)


tests/conftest.py
```diff
@@ -241,6 +264,8 @@ def shared_ray_cluster(tmp_path_factory: pytest.TempPathFactory, pytestconfig: p
     logger.info(f"Starting Ray cluster with {num_gpus} GPUs")
     logger.info(f"Running Ray command: {' '.join(cmd_to_run)}")
 
+    os.environ["RAY_ACCEL_ENV_VAR_OVERRIDE_ON_ZERO"] = "0"
```
praateekmahajan: I'm guessing we added this here for Xenna unittests to pass?
If so this might mean that outside unittests the Xenna pipelines will fail. To fix that we could
1. Make changes in RayClient.start()
2. Alternatively make changes inside xenna/executor.py where we set the NOSET flag to 1

cc @abhinavg4 can you opine here

abhinavg4: I would say this is the best way forward:
Alternatively make changes inside xenna/executor.py where we set the NOSET flag to 1

oyilmaz-nvidia: Yeah, will run the benchmarks and see what's failing. Based on that, we can add what's needed.

tests/conftest.py
```diff
@@ -225,37 +188,32 @@ def shared_ray_cluster(tmp_path_factory: pytest.TempPathFactory, pytestconfig: p
...
-
-    # Use explicit path to ray command for security
-    ray_process = subprocess.Popen(cmd_to_run, shell=False)  # noqa: S603
-    logger.info(f"Started Ray process: {ray_process.pid}")
-
-    ray_address = f"localhost:{ray_port}"
-    os.environ["RAY_ADDRESS"] = ray_address
-    logger.info(f"Set RAY_ADDRESS for tests to: {ray_address}")
+    ray_address = os.environ["RAY_ADDRESS"]
+    logger.info(f"Ray cluster started at: {ray_address}")
```
praateekmahajan, line 209: Thank you for this 🙏

nemo_curator/backends/xenna/executor.py
```diff
@@ -136,11 +136,14 @@ def execute(self, stages: list[ProcessingStage], initial_tasks: list[Task] | Non
 
         try:
             register_loguru_serializer()
+            # Prevent Ray from overriding accelerator env vars when num_gpus=0, letting Xenna manage them instead.
             ray.init(
                 ignore_reinit_error=True,
                 runtime_env={
                     # We need to set this env var to avoid ray from setting CUDA_VISIBLE_DEVICES and let xenna do it
-                    "env_vars": {"RAY_EXPERIMENTAL_NOSET_CUDA_VISIBLE_DEVICES": "0"}
+                    "env_vars": {"RAY_EXPERIMENTAL_NOSET_CUDA_VISIBLE_DEVICES": "0",
```
praateekmahajan, line 144: @abhinavg4 in your PR for Xenna bump can you see if we need to get rid of `RAY_EXPERIMENTAL_NOSET_CUDA_VISIBLE_DEVICES`

tests/backends/experimental/ray_data/test_max_calls_pid.py
```diff
@@ -44,18 +42,21 @@ def single_cpu_ray_cluster():
     original_ray_address = os.environ.pop("RAY_ADDRESS", None)
 
     temp_dir = tempfile.mkdtemp(prefix="ray1cpu_")
-    cmd, ray_port = build_ray_command(str(temp_dir), num_cpus=1, num_gpus=0, object_store_memory=2 * (1024**3))
-    ray_process = subprocess.Popen(cmd, shell=False)  # noqa: S603
+    ray_client = RayClient(
```
ayushdg, line 45: Thanks for this change!

PR #1561, opened 2026-02-27
Expose a few more args to the exact dedup workflow

ayushdg:
## Description

For larger runs on clusters I often find myself setting `total_nparts/rmm_pool_size/spill_limit` manually for fine tuning. Longer term it makes sense to revisit and optimize better defaults but for now it would be nice to expose them to the workflow class rather than having to create a custom stage.

## Usage

```python
workflow = ExactDeduplicationWorfklow(total_nparts=512, rmm_pool_size=72*1024*1024*1024,...)
```
## Checklist

### DIFF
--- a/nemo_curator/stages/deduplication/exact/workflow.py
+++ b/nemo_curator/stages/deduplication/exact/workflow.py
@@ -166,7 +166,7 @@ class ExactDeduplicationWorkflow(WorkflowBase):
                     assign_id=self.assign_id,
                     id_field=self.id_field,
                     # Matches previous implementation to write out to 1/3 the number of input tasks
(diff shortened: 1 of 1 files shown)


benchmarking/scripts/exact_dedup_identification_benchmark.py
```diff
@@ -126,7 +144,21 @@ def main() -> int:
...
     )
-
+    parser.add_argument(
+        "--total-nparts",
+        type=int,
+        default=None,
+        help="Total number of output partitions",
+    )
+    parser.add_argument(
+        "--rmm-pool-size", type=_parse_memory_arg, default="auto", help="Size of the RMM GPU memory pool in bytes"
```
sarahyurick, line 154: Huh, I never realized `type` is used as a function here. Nice!

PR #1583, opened 2026-03-06
Add Interleaved filters

meatybobby:
## Description

This pull request introduces new filtering capabilities for interleaved datasets. 

The primary additions are the following four filters:
* **Blur Detection Filter:** Uses OpenCV to calculate the Laplacian variance (`cv2.Laplacian`) of an image, which acts as a measure of sharpness. If the variance falls below a specified threshold, the image is classified as blurry and filtered out.
* **QR Code Filter:** Uses OpenCV to detect the presence of QR codes and calculate the bounding box area they occupy within an image. If the ratio of the QR code area to the total image area exceeds a given threshold, the image is filtered out. 
* **CLIP Score Filter:** Uses a vision-language model (CLIP) to evaluate the semantic alignment between images and accompanying text. The algorithm works by computing both the image embeddings and the text embeddings, and then calculating their cosine similarity to ensure the content matches well and meets a quality threshold.
* **Image-To-Text Ratio Filter:** Computes the sample-level ratio of the number of images to the text word count, filtering out samples that don't meet the desired image-to-text balance.

## Usage
```python
pipeline.add_stage(
    InterleavedBlurFilterStage(
        score_threshold=blur_score_threshold,
    )
)
pipeline.add_stage(
    InterleavedQRCodeFilterStage(
        score_threshold=qrcode_score_threshold,
    )
)
pipeline.add_stage(
    InterleavedCLIPScoreFilterStage(
        model_dir=clip_model_dir,
        min_score=clip_min_score,
    )
)
pipeline.add_stage(
    InterleavedImageToTextRatioFilterStage(
        min_ratio=min_ratio,
        max_ratio=max_ratio,
    )
)
```
## Checklist


nemo_curator/stages/interleaved/filter/blur_filter.py
```diff
@@ -0,0 +1,119 @@
...
+    return float(cv2.Laplacian(image, cv2.CV_64F).var())
+
+
+def _image_bytes_to_array(image_bytes: bytes) -> np.ndarray | None:
+    """Decode image bytes to RGB numpy array for OpenCV, or None on failure."""
+    try:
+        arr = np.frombuffer(image_bytes, dtype=np.uint8)
+        image = cv2.imdecode(arr, cv2.IMREAD_COLOR)
+    except Exception:  # noqa: BLE001
+        return None
```
VibhuJawa: This seems problematic . We should not capture bare exceptions. 

Also, _image_bytes_to_array should be moved to a utils that we all use (and not live in `blur_filter.py`.  Also, is it faster to do decode if we provide extenstions. We have that information

nemo_curator/stages/interleaved/filter/blur_filter.py
```diff
@@ -0,0 +1,119 @@
...
+    return cv2.cvtColor(image, cv2.COLOR_BGR2RGB)
+
+
+def _process_one_blur(item: tuple, score_threshold: float) -> tuple:
+    """Process one (idx, image_bytes); return (idx, keep)."""
+    idx, image_bytes = item
+    if image_bytes is None:
+        return (idx, False)
+    image = _image_bytes_to_array(image_bytes)
```
VibhuJawa: This means that we always decode bytes to numpy arrays . We should add a buffer to the structure and reuse where possible

nemo_curator/stages/interleaved/filter/blur_filter.py
```diff
@@ -0,0 +1,119 @@
...
+        return (idx, False)
+    return (idx, _sharpness_score(image) >= score_threshold)
+
+
+def _worker_blur(
+    work_queue: Queue,
+    result_queue: Queue,
+    score_threshold: float,
+) -> None:
+    """Module-level worker for multiprocessing; runs until sentinel."""
```
VibhuJawa: Why are we using multi processing here, ray allready uses cores. What is the benefit expected here ?

nemo_curator/stages/interleaved/filter/blur_filter.py
```diff
@@ -0,0 +1,119 @@
...
+        for p in processes:
+            p.start()
+        for idx, image_bytes in self.iter_materialized_bytes(task=task, df=df, row_mask=image_mask):
+            work_queue.put((idx, image_bytes))
+        for _ in range(num_items):
+            i, keep = result_queue.get()
+            keep_mask.loc[i] = keep
+        for p in processes:
+            p.terminate()
+            p.join()
```
VibhuJawa: Unsure on benefit of doing multi processing here , given this is ray based already !!

nemo_curator/stages/interleaved/filter/clip_score_filter.py
```diff
@@ -0,0 +1,93 @@
...
+    For each image row, all text rows with the same sample_id form (image, text)
+    pairs. CLIP similarity is computed for each pair. An image is kept only if at
+    least one pair has score >= min_score; otherwise it is dropped.
+    """
+
+    model_dir: str | None = None
+    min_score: float = 0.15
+    image_content_types: tuple[str, ...] = ("image/jpeg", "image/jpg", "image/png")
+    name: str = "interleaved_clip_score_filter"
+    resources: Resources = field(default_factory=lambda: Resources(gpus=0.25))
```
VibhuJawa: Why are gpus=0.25 set here as default ?

nemo_curator/stages/interleaved/filter/clip_score_filter.py
```diff
@@ -0,0 +1,93 @@
...
+        self._model = CLIPImageEmbeddings(self.model_dir)
+        self._model.setup()
+
+    def content_keep_mask(self, task: InterleavedBatch, df: pd.DataFrame) -> pd.Series:
+        keep_mask = pd.Series(True, index=df.index, dtype=bool)
+        image_mask = (df["modality"] == "image") & (df["content_type"].isin(self.image_content_types))
+        if not image_mask.any():
+            return keep_mask
+
+        self._ensure_model()
```
VibhuJawa: Model should be loaded after init not during run !

nemo_curator/stages/interleaved/filter/clip_score_filter.py
```diff
@@ -0,0 +1,93 @@
...
+            if image is None:
+                keep_mask.loc[idx] = False
+                continue
+            sample_id = df.loc[idx, "sample_id"]
+            texts = _sample_texts_list_from_df(df, sample_id)
+            if not texts:
+                keep_mask.loc[idx] = False
+                continue
+            img_emb = self._model([image])
+            text_emb = self._model.encode_text(texts)
```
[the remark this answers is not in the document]
VibhuJawa: I agree with this, maybe worth doing

nemo_curator/stages/interleaved/filter/blur_filter.py
meatybobby: @VibhuJawa I didn't find opencv has faster decode with extension. It seems only works when we read a file, but we already loaded it into the memory.

nemo_curator/stages/interleaved/filter/blur_filter.py
meatybobby: It looks no improvement with testing. Remove this part.

nemo_curator/stages/interleaved/filter/clip_score_filter.py
```diff
@@ -0,0 +1,90 @@
...
+            image = image_bytes_to_array(image_bytes)
+            sample_id = df.loc[idx, "sample_id"]
+            texts = _sample_texts_list_from_df(df, sample_id)
+            if not texts:
+                keep_mask.loc[idx] = False
+                continue
+            img_emb = self._model([image])
+            if sample_id != cached_sample_id:
+                cached_sample_id = sample_id
+                text_emb = self._model.encode_text(texts)
```
VibhuJawa: We should do batching here and not send batch size=1 (do fwd pass for multiple samples right)

nemo_curator/stages/interleaved/filter/blur_filter.py
```diff
@@ -0,0 +1,57 @@
...
+    def content_keep_mask(self, task: InterleavedBatch, df: pd.DataFrame) -> pd.Series:
+        keep_mask = pd.Series(True, index=df.index, dtype=bool)
+        image_mask = (df["modality"] == "image") & (df["content_type"].isin(self.image_content_types))
+        if not image_mask.any():
+            return keep_mask
+        for idx, image_bytes in self.iter_materialized_bytes(task=task, df=df, row_mask=image_mask):
+            if image_bytes is None:
+                keep_mask.loc[idx] = False
+                continue
+            image = image_bytes_to_array(image_bytes)
```
VibhuJawa: We are still always creating numpy arrays for each image, we should cache/add it as a column to the frame and do a look up instead so that buffer can be reused again . 

Also add a way to clean up buffer if needed or something .

benchmarking/scripts/interleaved_filter_benchmark.py
```diff
@@ -0,0 +1,232 @@
...
+        write_kwargs["compression"] = args.parquet_compression
+    pipeline = Pipeline(
+        name="interleaved_filter_benchmark",
+        description=(
+            "Benchmark: WebDataset through blur, QR code, CLIP score, and image/text ratio filters to parquet"
+        ),
+    )
+    pipeline.add_stage(
+        WebdatasetReader(
+            source_id_field="pdf_name",
```
VibhuJawa: This will no longer be needed https://github.com/NVIDIA-NeMo/Curator/pull/1657 after is merged

benchmarking/scripts/interleaved_filter_benchmark.py
```diff
@@ -0,0 +1,232 @@
...
+    if success:
+        parquet_files = sorted(output_path.glob("*.parquet"))
+        if parquet_files:
+            result = validate_parquet_ordering(parquet_files[0])
+            ordering_valid = result["valid"]
+            if not ordering_valid:
+                logger.error("Ordering validation failed on {}: {}", parquet_files[0].name, result["errors"])
+            else:
+                logger.info("Ordering validation passed on {}", parquet_files[0].name)
+
```
VibhuJawa: You dont need these checks, these are only meant for readers/writers

nemo_curator/stages/interleaved/filter/blur_filter.py
```diff
@@ -0,0 +1,59 @@
...
+        for idx, image_bytes in self.iter_materialized_bytes(task=task, df=df, row_mask=image_mask):
+            if image_bytes is None:
+                keep_mask.loc[idx] = False
+                continue
+            try:
+                image = image_bytes_to_array(image_bytes)
+                sharpness = _sharpness_score(image)
+                keep_mask.loc[idx] = sharpness >= self.score_threshold
+            except OSError:
+                keep_mask.loc[idx] = False
```
VibhuJawa: Too broad of try/excpet. 

We probably want this on `image_bytes_to_array(image_bytes)` . This function is called repeatedely across filters we should save the image_np_array as a column  so that we can re use it

nemo_curator/stages/interleaved/filter/clip_score_filter.py
```diff
@@ -0,0 +1,93 @@
...
+    model_dir: str | None = None
+    min_score: float = 0.15
+    name: str = "interleaved_clip_score_filter"
+    resources: Resources = field(default_factory=lambda: Resources(gpu_memory_gb=20.0))
+
+    def __post_init__(self) -> None:
+        if self.model_dir is None:
+            msg = "InterleavedCLIPScoreFilterStage requires model_dir to be set"
+            raise RuntimeError(msg)
+        CLIPImageEmbeddings.download_weights_on_node(self.model_dir)
```
VibhuJawa, line 85: This should be  callled inside `setup_on_node` so that it sonly  once !

nemo_curator/stages/interleaved/filter/clip_score_filter.py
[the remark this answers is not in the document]
VibhuJawa: +1 agaion, please address @meatybobby

nemo_curator/stages/interleaved/filter/qrcode_filter.py
```diff
@@ -0,0 +1,74 @@
...
+        keep_mask = pd.Series(True, index=df.index, dtype=bool)
+        image_mask = df["modality"] == "image"
+        if not image_mask.any():
+            return keep_mask
+        for idx, image_bytes in self.iter_materialized_bytes(task=task, df=df, row_mask=image_mask):
+            if image_bytes is None:
+                keep_mask.loc[idx] = False
+                continue
+            try:
+                image = image_bytes_to_array(image_bytes)
```
VibhuJawa: Same review here (as mentioned in the earlier reviews)

VibhuJawa (changes requested):
I think a bunch of earlier reviews are still not addressed. Please address them

VibhuJawa:
@claude review

nemo_curator/stages/interleaved/filter/clip_score_filter.py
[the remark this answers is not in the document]
meatybobby: This is already resolved. We change it to merge all images and all texts with same sample_ids and just do one forward.

nemo_curator/stages/interleaved/filter/blur_filter.py
meatybobby: I think we should create another PR for this. It requires changes on InterleavedBatch

nemo_curator/stages/interleaved/filter/blur_filter.py
meatybobby: Same above

nemo_curator/stages/interleaved/filter/blur_filter.py
meatybobby: Add one error catch on image_bytes_to_array but we still need to keep this since cv2.Laplacian may cause error

nemo_curator/stages/interleaved/filter/qrcode_filter.py
meatybobby: same above, cv2.QRCodeDetector may cause error

nemo_curator/stages/interleaved/filter/blur_filter.py
VibhuJawa: How about we do this then (we essentially need to know if some image broke or its actually filtered. Above code does it silently which is problematic at scale when we need to debug these things.


```
def image_bytes_to_array(image_bytes: bytes, idx: int | None = None) -> np.ndarray | None:
    """Decode raw image bytes to a numpy array.

    Returns None if the image bytes cannot be decoded (corrupt, truncated, etc.).
    Infrastructure errors (permissions, connectivity) are re-raised.

    Args:
        image_bytes: Raw bytes of the image.
        idx: Optional row/sample index for log traceability.
    """
    try:
        buf = np.frombuffer(image_bytes, dtype=np.uint8)
        image = cv2.imdecode(buf, cv2.IMREAD_COLOR)
    except Exception as e:
        if isinstance(e, (PermissionError, FileNotFoundError, ConnectionError, TimeoutError)):
            raise
        logger.debug("Failed to decode image bytes at idx=%s: %s: %s", idx, type(e).__name__, e)
        return None

    if image is None:
        logger.debug("cv2.imdecode returned None at idx=%s (corrupt or unsupported format, %d raw bytes)", idx, len(image_bytes))
        return None

    return image


def _sharpness_score(image: np.ndarray, idx: int | None = None) -> float | None:
    """Compute Laplacian variance as a sharpness proxy.

    Returns None if scoring fails (e.g., zero-dim image, unexpected channels).

    Args:
        image: Decoded image array (BGR, uint8).
        idx: Optional row/sample index for log traceability.
    """
    try:
        return float(cv2.Laplacian(image, cv2.CV_64F).var())
    except cv2.error as e:
        logger.debug(
            "Laplacian scoring failed at idx=%s (shape=%s, dtype=%s): %s",
            idx, image.shape, image.dtype, e,
        )
        return None
```
Caller: 
```
image = image_bytes_to_array(image_bytes, idx=idx)
if image is None:
    keep_mask.loc[idx] = False
    continue

sharpness = _sharpness_score(image, idx=idx)
if sharpness is None:
    keep_mask.loc[idx] = False
    continue

keep_mask.loc[idx] = sharpness >= self.score_threshold
```

nemo_curator/stages/interleaved/filter/qrcode_filter.py
VibhuJawa: Unresolved again. LMK if above comment makes sense

nemo_curator/stages/interleaved/filter/clip_score_filter.py
```diff
@@ -0,0 +1,114 @@
...
+class InterleavedCLIPScoreFilterStage(BaseInterleavedFilterStage):
+    """Filter interleaved image rows by CLIP image-text relevance score.
+
+    For each image row, all text rows with the same sample_id form (image, text)
+    pairs. CLIP similarity is computed for each pair. An image is kept only if at
+    least one pair has score >= min_score; otherwise it is dropped.
+    """
+
+    model_dir: str | None = None
+    min_score: float = 0.15
```
VibhuJawa: Can we move default to use a constant like DEFAULT_MIN_SCORE (at top of this file or something)

nemo_curator/stages/interleaved/filter/blur_filter.py
VibhuJawa: Okay, please file an issue please so that we can track it . We can do this in the next release but we should track the issue.

meatybobby: create an issue #1759  for this

abhinavg4:
/ok to test [dc69766](https://github.com/NVIDIA-NeMo/Curator/pull/1583/commits/dc69766985e782fe3d43646388020e421e1c827f)

PR #1593, opened 2026-03-10
Add ray to each benchmarking entry and make sure no benchmarking/script has RayClient

praateekmahajan:
1. Added missing `ray: config (num_cpus: 64, num_gpus: 4)` to:
	- `domain_classification_xenna`
	- `embedding_generation_xenna`
	- `audio_fleurs`
2.  Removed RayClient import, ray_client.start(), and ray_client.stop(). The Ray cluster is already started by run.py.
	- `image_pipeline_benchmark.py`
	- `multimodal_mint1t_benchmark.py`
3.  Fixed `success_code = 0` (should be 1) in `arxiv_e2e_pipeline_benchmark.py` which masked benchmark failures as successes
4. Replaced hardcoded `XennaExecutor()` in `audio_fleurs_benchmark.py` with `setup_executor()` + --executor CLI arg, consistent with all other benchmark scripts
5.  Fixed potential `UnboundLocalError` crashes across multiple scripts (`exact_dedup_identification`, `score_filter`, `modifier`, `video_pipeline`, `image_pipeline`, `fasttext_filter`, `semdedup_identification`, `dedup_removal`) by initializing variables before conditional/try blocks
## Usage

```python
```
## Checklist

### DIFF
--- a/benchmarking/scripts/audio_fleurs_benchmark.py
+++ b/benchmarking/scripts/audio_fleurs_benchmark.py
@@ -23,9 +23,8 @@ from pathlib import Path
 from typing import Any
 
 from loguru import logger
-from utils import write_benchmark_results
+from utils import setup_executor, write_benchmark_results
 
-from nemo_curator.backends.xenna import XennaExecutor
 from nemo_curator.pipeline import Pipeline
 from nemo_curator.stages.audio.common import GetAudioDurationStage, PreserveByValueStage
 from nemo_curator.stages.audio.datasets.fleurs.create_initial_manifest import CreateInitialManifestFleursStage
@@ -44,6 +43,7 @@ def run_audio_fleurs_benchmark(  # noqa: PLR0913
     split: str,
     wer_threshold: float,
     gpus: int,
+    executor: str = "xenna",
     **kwargs,  # noqa: ARG001
 ) -> dict[str, Any]:
     """Run the audio fleurs benchmark and collect comprehensive metrics."""
@@ -65,7 +65,7 @@ def run_audio_fleurs_benchmark(  # noqa: PLR0913
     logger.info(f"WER threshold: {wer_threshold}")
     logger.info(f"GPUs: {gpus}")
 
-    executor = XennaExecutor()
+    executor_obj = setup_executor(executor)
     pipeline = Pipeline(name="audio_inference", description="Inference audio and filter by WER threshold.")
 
     # Add stages
@@ -106,7 +106,7 @@ def run_audio_fleurs_benchmark(  # noqa: PLR0913
         )
     )
 
-    results = pipeline.run(executor)
+    results = pipeline.run(executor_obj)
 
     logger.success("Benchmark completed successfully")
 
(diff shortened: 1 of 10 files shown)


benchmarking/scripts/image_pipeline_benchmark.py
```diff
@@ -172,9 +171,6 @@ def run_image_pipeline_benchmark(args: argparse.Namespace) -> dict[str, Any]:
 
 def main() -> int:
     """Main entry point for image pipeline benchmark."""
-    ray_client = RayClient()
-    ray_client.start()
```
ayushdg, line 176: Isn't this a noop if cluster is started already?

benchmarking/scripts/dedup_removal_benchmark.py
```diff
@@ -124,9 +124,11 @@ def run_removal_benchmark(  # noqa: PLR0913
         for k, v in TaskPerfUtils.aggregate_task_metrics(workflow_run_result).items()
         if k.endswith("_process_time_mean")
     }
-    io_percentage = round(
-        (task_metrics["jsonl_reader"] + task_metrics["parquet_writer"]) * 100 / sum(task_metrics.values()), 2
-    )
+    reader_key = f"{input_filetype}_reader"
```
ayushdg, line 127: Thanks!

benchmarking/scripts/image_pipeline_benchmark.py
praateekmahajan: > Having RayClient inside benchmarking/scripts/*.py should be a no-op however for sake of consistency I would recommend users to use benchmarking/run.py to do the heavy lifting since it'll log details related to env / ray cluster / object store as well, which otherwise would be missed and wouldn't ensure if a run is comparable to another

benchmarking/nightly-benchmark.yaml
```diff
@@ -153,6 +153,10 @@ entries:
           - number_of_domains_predicted
         ping_on_failure:
           - U022P8CDX40  # Sarah Yurick
+    ray:
+      num_cpus: 64
+      num_gpus: 4
+      enable_object_spilling: false
```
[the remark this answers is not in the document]
praateekmahajan: Currently it anyway uses 4 gpus we are just now being explicit it.

benchmarking/nightly-benchmark.yaml
ayushdg, line 163: One nice thing about throughput requirements is that we don't need to update with numbers GPUs.

benchmarking/scripts/audio_fleurs_benchmark.py
```diff
@@ -44,6 +43,7 @@ def run_audio_fleurs_benchmark(  # noqa: PLR0913
     split: str,
     wer_threshold: float,
     gpus: int,
```
ayushdg, line 45: unrelated to the PR but this arg is confusing IMO. The Cli claims its the number of GPUs to use but actually its the resource requirement for the inference stage

PR #1597, opened 2026-03-12
ci: Adding automation config for benchmark test

thomasdhc:
## Usage

```python
```
## Checklist

### DIFF
--- a/.github/workflows/cicd-main.yml
+++ b/.github/workflows/cicd-main.yml
@@ -154,7 +154,7 @@ jobs:
           azure-subscription-id: ${{ secrets.AZURE_SUBSCRIPTION_ID }}
           HF_TOKEN: ${{ secrets.HF_TOKEN }}
           PAT: ${{ secrets.PAT }}
-          timeout: 20
+          timeout: 30
           test-data-path: ${{ needs.pre-flight.outputs.test_data_path }}
           runner: ${{ matrix.runner }}
 


benchmarking/nightly-benchmark.yaml
sarahyurick, line 1: Do you have a ticket open somewhere for tracking adding the rest of the benchmarks to CI too?

benchmarking/nightly-benchmark.yaml
```diff
@@ -111,6 +111,9 @@ entries:
       --dataset-size-gb=10
       --model-inference-batch-size=1024
     timeout_s: 1000
+    ci:
+      scope: nightly
+      time: "00:20:00"
```
praateekmahajan: Would prefer to not modify this yaml.. Here is an alternative suggestion


```python
# run.py (most of the logic already exists)

@dataclass
Entry:
  script : str
  args : str
  requirements :...
  timeout : int 

def filter_entries(...):
   # logic to return entries by parsing the yaml
   # some of this code might exist already
	return list[Entry]
```

```python
# generate_ci_tests.py

ci_entry_names = ["domain_classification_raydata", "domain_classification_xenna" .....]
ci_entries_to_run = filter_entries(ci_entries_to_run)

for entry in ci_entries_to_run:
    generate_config_for_ci(entry)
```


Most of the code already should exist in `run.py` because we already do some filtering to support `--entries`. Maybe some slight refactor is needed. 

This helps us maintain all parsing of nightly.yaml code in one place, while any post-processing can be done based on the data class.

Allows us to make more changes to nightly.yaml structure without thinking how that might've affected your setup.

benchmarking/tools/ci_benchmark_launcher.sh
```diff
@@ -0,0 +1,27 @@
+# Copyright (c) 2026, NVIDIA CORPORATION.  All rights reserved.
```
praateekmahajan, line 2: Can we name this and other files something more specific such as `ci_eos_benchmark_launcher.sh`?

benchmarking/test-paths.yaml
```diff
@@ -0,0 +1,9 @@
+results_path: /tmp/curator/results/${BRANCH_NAME}
```
praateekmahajan, line 1: I believe we have blessings from Arham to push the the actual paths on EOS here, and then we can call this file `eos-ci-paths.yaml` or something

benchmarking/tools/ci_benchmark_launcher.sh
thomasdhc: Yeah I can update it, this shouldn't have any ties to specific clusters

benchmarking/test-paths.yaml
thomasdhc: I've updated the implementation at least on the ci side so that all test trigger off this path. So no matter what the cluster is this path will be used

benchmarking/tools/ci_benchmark_launcher.sh
praateekmahajan: Makes sense, in which case `ci_benchmark_launcher.sh` since we have `run.sh` which has been the `becnhmark_launcher` on dev machines

benchmarking/nightly-benchmark.yaml
thomasdhc: I feel like that would be maintaining two different config files, one nightly.yaml and one ci-nightly.yaml. The ci configs will have to live somewhere even with the filter function.

I'm ok with going that route for now though.

benchmarking/test-paths.yaml
praateekmahajan: That makes sense, what I was suggesting was that the final path list that the CI will use, can live in this repo directly, that way when developers add a new dataset they don't have to do cross-repo merges to make things work.

benchmarking/nightly-benchmark.yaml
thomasdhc: Let me follow up on that

benchmarking/test-paths.yaml
thomasdhc: Hmmmm I see, I guess the base dataset path will always be mounted though so even when developers add new datasets, the path can just be added here and nowhere else. Was there a different case you had in mind?

benchmarking/nightly-benchmark.yaml
praateekmahajan: Two different config files? Sorry maybe what I wanted to say isn't clear. I'm saying we still have one YAML file (but without any changes), and one `generate_ci_tests` (like you do) however the logic of reading the YAML, filtering entries should be reused from `run.py`.. 

Currently since CI will support only subset of entries we'll have `ci_entry_names` but once CI is enabled to support all entries we won't even need that.

benchmarking/test-paths.yaml
```diff
@@ -0,0 +1,9 @@
...
+datasets_path: /tmp/curator/datasets
+model_weights_path: /tmp/curator/datasets
+sinks: []
+datasets:
+  - name: "tinystories"
+    formats:
+      - type: "parquet"
+        path: /tmp/curator/datasets/tinystories_train_parquet
```
praateekmahajan:
```suggestion
        path:  {datasets_path}/tinystories_train_parquet
```

benchmarking/test-paths.yaml
praateekmahajan: Hmm I see what you mean, I guess that'll work.. So you're saying in CI there is a file that is called paths.yaml that'll just have 

```yaml
dataset_path : /lustre/......
results_path : /liustre/...
```

but nothing more.. while we have a paths.yaml in Curator (main) which is. 
```yaml
tinystories:
   paruqet:
      {dataset_path}/tinystories_train
```

If so that makes sense... We'll have to ensure all the other machines also follow similar tradition

benchmarking/nightly-benchmark.yaml
thomasdhc: Right but then where will CI specific configurations live? Maintaining a python dictionary in generate_ci_tests won't be ideal.

benchmarking/test-paths.yaml
thomasdhc: Yeah, I've configured the CI runner so that the mount path is always consistent. But if ever we feel the need to surface that path explicitly we can do so.

benchmarking/nightly-benchmark.yaml
thomasdhc: For example the variables that we currently have `scope` and `time` need to be associated with specific tests right, these values need to live somewhere aside from being needed for filtering.

thomasdhc:
@oyilmaz-nvidia Currently working on the cli to enable triggering that I'll share the details for that once that's available

benchmarking/nightly-benchmark.yaml
thomasdhc: I guess the ci specific configurations are small now but from other projects I've seen this expand and grow as we further development the testing infrastructure.

thomasdhc: I reread your suggestion and I understand now. Filtered entries will extrapolate values needed from the existing nightly-benchmark. And I guess technically all these tests under nightly-benchmark is suppose to run nightly.

The need to maintain ci_entry_name comes from the gap that we are missing some datasets, but once that is in place  `enabled` should control if the tests run on ci or not.

PR #1604, opened 2026-03-13
Add math modality benchmarking support

ronjer30:
Benchmark changes:
- Add `math_pipeline_benchmark.py` to run benchmark
- Add `prepare_math_benchmark_data.py` to enrich FineMath-4+ with Common Crawl binary content for offline benchmarking
- Add math_preprocess, math_preprocess_classifier, and math_preprocess_llm_cleanup entries to nightly-benchmark.yaml
- Install lynx in Dockerfiles for HTML-to-text extraction

Stage fixes:
- Fix deepcopy/pickle crash in MathContentExtractor by adding __getstate__/__setstate__ to strip unpickleable threading.Lock and magic.Magic before serialization 
- Add S3 transport to CommonCrawlWARCReader via boto3 as an alternative to HTTPS, activated by use_s3=True or CC_USE_S3 env var, this addresses latency and throttling issues revealed during benchmarking.
- Add boto3 to math_cpu dependency 

## Usage

```python
```
## Checklist

### DIFF
--- a/benchmarking/data_prep/prepare_math_benchmark_data.py
+++ b/benchmarking/data_prep/prepare_math_benchmark_data.py
@@ -59,6 +59,7 @@ from huggingface_hub import hf_hub_download, list_repo_files
 from huggingface_hub.utils import HfHubHTTPError, RepositoryNotFoundError
 from loguru import logger
 
+from benchmarking.scripts.utils import setup_executor
 from nemo_curator.pipeline import Pipeline
 from nemo_curator.stages.resources import Resources
 from nemo_curator.stages.text.download.common_crawl.download import CommonCrawlWARCReader
@@ -381,8 +382,6 @@ def run_enrichment(args: argparse.Namespace) -> int:
     except ValueError as e:
         logger.error(str(e))
         return 1
-
-    from utils import setup_executor
 
     logger.info("Building enrichment pipeline (ParquetReader -> CommonCrawlWARCReader -> ParquetWriter)...")
     enriched_dir = str(output_path / "enriched")
--- a/benchmarking/scripts/cc_index_benchmark.py
+++ b/benchmarking/scripts/cc_index_benchmark.py
@@ -32,13 +32,13 @@ import argparse
 import importlib.util
 import os
 import time
+import traceback
 from pathlib import Path
 
 import ray
 from loguru import logger
 from utils import setup_executor, write_benchmark_results
 
-from nemo_curator.core.client import RayClient
 from nemo_curator.pipeline import Pipeline
 from nemo_curator.stages.file_partitioning import FilePartitioningStage
 from nemo_curator.tasks.utils import TaskPerfUtils
@@ -68,7 +68,9 @@ def compute_output_metrics(output_path: str) -> dict:
         total_bytes = sum(os.path.getsize(f) for f in output_files)
         metrics["output_total_mb"] = total_bytes / (1024 * 1024)
     except Exception as e:
+        error_traceback = traceback.format_exc()
         logger.warning(f"Could not compute output metrics: {e}")
+        logger.debug(f"Full traceback:\n{error_traceback}")
     return metrics
 
 
@@ -79,17 +81,11 @@ def run_benchmark(args: argparse.Namespace) -> dict:
     output_path = Path(args.output_path).absolute()
     output_path.mkdir(parents=True, exist_ok=True)
 
-    # We need ray.put() below to broadcast query URLs, which triggers an
-    # implicit ray.init() if Ray isn't initialized yet.  That bare init
-    # would lack the runtime_env that XennaExecutor normally sets (see
-    # XennaExecutor.run()), causing Xenna's GPU probe to fail.  Mirror
-    # what XennaExecutor does: explicit ray.init with the env var BEFORE
-    # any other Ray calls.
+    # Ray must be initialized before ray.put(); replicate the executor's
+    # runtime_env so GPU visibility is configured for Xenna workers.
     os.environ.setdefault("RAY_EXPERIMENTAL_NOSET_CUDA_VISIBLE_DEVICES", "1")
-    ray_client = RayClient()
-    ray_client.start()
-
     ray.init(
+        address="auto",
         ignore_reinit_error=True,
         runtime_env={
             "env_vars": {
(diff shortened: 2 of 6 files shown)


benchmarking/scripts/math_pipeline_benchmark.py
```diff
@@ -0,0 +1,367 @@
...
+    llm_group.add_argument("--model", type=str, default="microsoft/phi-4", help="LLM model identifier")
+    llm_group.add_argument(
+        "--prompt", type=str, default="HTML_TO_TEXT_PROMPT", help="Prompt name from prompts module or literal string"
+    )
+    llm_group.add_argument("--chunk-data", action="store_true", help="Enable token-based chunking before LLM")
+    llm_group.add_argument("--chunk-length", type=int, default=4096, help="Max tokens per chunk")
+    llm_group.add_argument("--max-model-len", type=int, default=None, help="Max model context length for vLLM")
+
+    # Executor
+    p.add_argument("--executor", type=str, default="xenna", choices=["xenna", "ray_data", "ray_actors"])
```
sarahyurick: The `RayActorPoolExecutor` should not be applicable here:
```suggestion
    p.add_argument("--executor", type=str, default="xenna", choices=["xenna", "ray_data"])
```

benchmarking/nightly-benchmark.yaml
```diff
@@ -888,3 +802,69 @@ entries:
...
+    enabled: true
+    script: math_pipeline_benchmark.py
+    args: >-
+      --benchmark-results-path={session_entry_dir}
+      --input-path={dataset:finemath4plus,parquet}
+      --output-path={session_entry_dir}/scratch/output
+      --executor=xenna
+    timeout_s: 600
+    sink_data:
+      - name: slack
```
sarahyurick, line 1006: Who can we add as the PIC on Slack? @ronjer30 can it be you?

nemo_curator/stages/math/download/extract.py
```diff
@@ -105,6 +105,19 @@ def __post_init__(self):
...
+        state["_lynx"] = None
+        state["_magic"] = None
+        state["_lock"] = None
+        return state
+
+    def __setstate__(self, state: dict[str, Any]) -> None:
+        self.__dict__.update(state)
+        self._lynx = None
+        self._magic = None
+        self._lock = threading.Lock()
```
sarahyurick, line 119: Can you share more context about this? I saw in your PR description:
```
Fix deepcopy/pickle crash in MathContentExtractor by adding getstate/setstate to strip unpickleable threading.Lock and magic.Magic before serialization
```

ronjer30: Sure, the `threading.Lock` was introduced earlier for thread-safe lazy init of `_magic` and `_lynx`, but would fail with `TypeError: cannot pickle '_thread.lock' object`, since Lock isn't picklable when the executors pickle stages to send them to workers.   `__getstate__/__setstate__` fixes this by stripping the lock during serialization and recreating it on the worker.  `_magic` and `_lynx` are also stripped to avoid serializing unnecessary state and ensure they're initialized fresh on the actual worker node, where they're then cached and reused across subsequent tasks processed by that worker

benchmarking/data_prep/prepare_math_benchmark_data.py
sarahyurick, line 1: Thanks for this. Maybe instead of having it in `scripts/`, we can create a new subdirectory within https://github.com/NVIDIA-NeMo/Curator/tree/main/benchmarking called "data"/"data_prep"/something similar. cc @rlratzel WDYT?

benchmarking/nightly-benchmark.yaml
```diff
@@ -980,3 +992,102 @@ entries:
...
+          - U0849HBGCLD  # Sukrit Rao
+    ray:
+      num_cpus: 64
+      num_gpus: 4
+      enable_object_spilling: false
+
+  - name: math_cc_index_lookup
+    enabled: true
+    script: cc_index_benchmark.py
+    args: >-
```
sarahyurick, line 1089: It looks like there is no option for "xenna" vs "ray_data" for this one?

ronjer30: Thanks, explicitly added "xenna" as the default, not tested on ray_data.

benchmarking/scripts/cc_index_benchmark.py
```diff
@@ -0,0 +1,271 @@
...
+    ray_client.start()
+
+    ray.init(
+        ignore_reinit_error=True,
+        runtime_env={
+            "env_vars": {
+                "RAY_EXPERIMENTAL_NOSET_CUDA_VISIBLE_DEVICES": "1",
+            },
+        },
+    )
```
sarahyurick: Normally we don't want individual benchmarking scripts to initialize the Ray client on their own (https://github.com/NVIDIA-NeMo/Curator/pull/1593). I am wondering if there is a better way to do this.

benchmarking/scripts/math_pipeline_benchmark.py
```diff
@@ -0,0 +1,395 @@
...
+from nemo_curator.stages.text.modifiers import Modify
+from nemo_curator.tasks.utils import TaskPerfUtils
+from nemo_curator.utils import prompts
+from nemo_curator.utils.file_utils import get_all_file_paths_under
+
+MIN_HIGH_QUALITY_SCORE = 3
+
+
+def _fill_null_text(text: str | None) -> str:
+    import pandas as pd
```
sarahyurick: It can be a top-level import?

benchmarking/scripts/cc_index_benchmark.py
rlratzel, line 95: This suggestion is probably out-of-scope for this PR, but a possibly better longer-term solution would be to take advantage of the `ray:` block in the benchmark YAML that each entry can set. We could put entry-specific env var settings that are passed to Ray when the benchmark runner initializes it for that entry.  I can help with this if you'd like to do that here, or we can do that in a separate PR.

Another suggestion that's in-scope for this PR is to also pass the `address="auto"` option to `ray.init()` here so the script will error if it cannot connect to the existing cluster set up by the runner, otherwise it will start one itself if it can't connect, which could be confusing.

benchmarking/scripts/cc_index_benchmark.py
```diff
@@ -0,0 +1,271 @@
...
+        # Step 5: Run pipeline
+        executor = setup_executor(args.executor)
+        pipeline_start = time.perf_counter()
+        try:
+            results = pipeline.run(executor)
+            success = True
+        except Exception:
+            logger.exception("CC Index lookup pipeline failed")
+            results = []
+            success = False
```
rlratzel: I wonder if many of these catch-all exception handlers will be hard to debug on error.  Perhaps we should also log the caught exception and traceback, like we do [here](https://github.com/NVIDIA-NeMo/Curator/blob/705e7807547c7500d86073508e3c196220bfb214/benchmarking/run.py#L355-L366).

benchmarking/data_prep/prepare_math_benchmark_data.py
rlratzel: I agree with @sarahyurick that it might be better to keep the `scripts` subdir exclusive to the benchmark scripts that are referenced in the `entries` block in the YAML.  A better option could be to move this script to the `tools` dir?  `data_prep` or similar could work too.

nemo_curator/stages/text/download/common_crawl/download.py
```diff
@@ -162,17 +207,99 @@ def outputs(self) -> tuple[list[str], list[str]]:
...
+        Credentials, region, and endpoint are resolved entirely by boto3's
+        standard chain (``AWS_*`` env vars, ``~/.aws/config``, instance
+        profiles).  Only connection-pool and retry settings are overridden.
+        """
+        if self._s3_client is None:
+            with self._lock:
+                if self._s3_client is None:
+                    try:
+                        import boto3
+                    except ImportError as exc:
```
rlratzel: This is minor but I try to catch `ModuleNotFoundError` for these since `ImportError` also catches/masks problems related to corrupted installs, missing symbols in module extensions, etc. which the user may need to know about.

tutorials/math/1_cc_index_lookup.py
```diff
@@ -131,25 +131,31 @@ def process(self, task: FileGroupTask) -> FileGroupTask:
 
 
 def collect_unique_urls(input_path: str, url_col: str = "url") -> cudf.DataFrame:
-    """Collect unique URLs from input dataset using cuDF."""
+    """Collect unique URLs from input dataset.
+
+    Uses pandas for reading and dedup, then converts the small deduplicated result to cuDF
+    for the downstream GPU merge.
+    """
+    import pandas as pd
```
rlratzel: Would it be better to move this to the top scope with the other imports? I believe cudf depends on pandas, so it's probably already imported (in the cudf namespace) anyway and importing it here won't save startup time or reduce a dependency.

benchmarking/scripts/cc_index_benchmark.py
ronjer30: Agreed, removed this,`ray.init` is needed to set the env variables correctly before `ray.put` is called

ronjer30: Agreed on both points. I'll add `address="auto"` to the `ray.init()` call in this PR. The YAML ray: block enhancement for per-entry env vars makes sense as a follow-up, happy to pick that up in a separate PR.

benchmarking/scripts/cc_index_benchmark.py
ronjer30: I've added `traceback.format_exc()` logging

benchmarking/data_prep/prepare_math_benchmark_data.py
ronjer30: The `data_prep` folder makes sense, I've moved it to `benchmarking/data_prep/`

nemo_curator/stages/text/download/common_crawl/download.py
ronjer30: agreed, changed to `ModuleNotFoundError`

tutorials/math/1_cc_index_lookup.py
ronjer30: agreed, moved `import pandas` to the top

PR #1606, opened 2026-03-13
Make vLLM as default for text embedding generation in Semantic Dedup Workflow + vLLM <> Ray Data bug fix

praateekmahajan:
## Description
1. Switch the default embedding backend from SentenceTransformers to vLLM across the semantic deduplication workflow, benchmarks, and tutorials. 
2. Fix a CUDA fork error when running vLLM stages with `RayDataExecutor`.
   - vLLM's auto-detection (https://github.com/vllm-project/vllm/blob/3e6a1e1686958dcd7eff1438bc5418b8d56daa30/vllm/utils/system_utils.py#L126-L160) for `spawn` vs `fork` only triggers inside Ray actors, not Ray tasks. 
   - `RayDataExecutor.execute_setup_on_node` dispatches `setup_on_node` as a remote task, so vLLM defaults to `fork` and hits `RuntimeError: Cannot re-initialize CUDA in forked subprocess`.
   -  Fixed by setting `VLLM_WORKER_MULTIPROC_METHOD=spawn` in the remote task.

## Usage
```python
from nemo_curator.stages.text.deduplication.semantic import TextSemanticDeduplicationWorkflow

# Default: uses vLLM with google/embeddinggemma-300m
workflow = TextSemanticDeduplicationWorkflow(
    input_path="./data",
    output_path="./output",
    cache_path="./cache",
    perform_removal=True,
    n_clusters=100,
    eps=0.01,
)
workflow.run()

# Custom vLLM kwargs
workflow = TextSemanticDeduplicationWorkflow(
    input_path="./data",
    output_path="./output",
    cache_path="./cache",
    model_identifier="google/embeddinggemma-300m",
    embedding_pretokenize=True,
    embedding_vllm_init_kwargs={"enforce_eager": True, "max_model_len": 2048},
)
```

### Benchmark
Run name is `praateek-embed-test-2026-03-17T14:43:17-07:00` where Xenna is slower due to #1618 
image

## Followups
1. #1618 
2. #1617 
## Checklist

### DIFF
--- a/nemo_curator/stages/text/deduplication/semantic.py
+++ b/nemo_curator/stages/text/deduplication/semantic.py
@@ -75,6 +75,7 @@ class TextSemanticDeduplicationWorkflow:
     embedding_pretokenize: bool = False
     embedding_vllm_init_kwargs: dict[str, Any] | None = None
     hf_token: str | None = None
+    model_cache_dir: str | None = None
     # Semantic deduplication parameters
     n_clusters: int = 100
     id_field: str = CURATOR_DEDUP_ID_STR
@@ -129,6 +130,7 @@ class TextSemanticDeduplicationWorkflow:
         embedding_pretokenize: Whether to pre-tokenize input before passing to vLLM
         embedding_vllm_init_kwargs: Additional kwargs passed to vLLM's LLM initializer
         hf_token: HuggingFace token for private models
+        model_cache_dir: Directory to cache model weights
 
         # Semantic deduplication parameters
         n_clusters: Number of clusters for K-means
@@ -273,7 +275,9 @@ class TextSemanticDeduplicationWorkflow:
             max_chars=self.embedding_max_chars,
             pretokenize=self.embedding_pretokenize,
             vllm_init_kwargs=self.embedding_vllm_init_kwargs,
+            cache_dir=self.model_cache_dir,
             hf_token=self.hf_token,
+            verbose=self.verbose,
         )
         pipeline.add_stage(embedding_stage)
 
--- a/tutorials/text/deduplication/semantic/semantic_step_by_step.ipynb
+++ b/tutorials/text/deduplication/semantic/semantic_step_by_step.ipynb
@@ -192,7 +192,7 @@
     "    text_field=\"text\",\n",
     "    embedding_pooling=\"mean_pooling\",\n",
     "    model_inference_batch_size=256,\n",
-    "    use_sentence_transformer=False, # or false in which case embedding_pooling is ignored\n",
+    "    use_sentence_transformer=False, # or True in which case embedding_pooling is ignored\n",
     ")\n",
     "```\n"
    ]


benchmarking/scripts/embedding_generation_benchmark.py
```diff
@@ -12,79 +12,202 @@
...
+
+
+class EmbeddingModelVariation(Enum):
+    SENTENCE_TRANSFORMER = "sentence_transformer"
+    PYTORCH_MODEL = "pytorch_model"
+    VLLM_TEXT = "vllm_text"
+    VLLM_TEXT_PRETOKENIZED = "vllm_text_pretokenized"
+
+
+def _resolve_max_seq_length(model_identifier: str, max_seq_length: int | None) -> int:
```
sarahyurick: We could put this somewhere under https://github.com/NVIDIA-NeMo/Curator/tree/main/nemo_curator/utils and update https://github.com/NVIDIA-NeMo/Curator/blob/main/nemo_curator/stages/text/models/tokenizer.py#L127 to use it too if you like.

praateekmahajan: I'm okay for this to live here instead of utils primarily because this is for a specific vLLM where vLLM errors out on max_seq_length of 512 while HF / ST work for that, and that results to a non-apples-to-apples comparision (since one framework processes more data than the other)

sarahyurick: Not sure I completely understand. If we look at MiniLM it looks like the `max_seq_length` is 512 (from [here](https://huggingface.co/sentence-transformers/all-MiniLM-L6-v2/blob/main/tokenizer_config.json)). You're saying that vLLM errors with 512? This function will return 512.

benchmarking/scripts/embedding_generation_benchmark.py
```diff
@@ -115,22 +227,38 @@ def main() -> int:
...
+        "--model-variation",
+        default="vllm_text",
+        choices=[v.value for v in EmbeddingModelVariation],
+        help="Embedding model backend (default: vllm_text)",
+    )
+    parser.add_argument(
+        "--max-seq-length",
+        type=int,
+        default=None,
+        help="Max sequence length for tokenization (auto-detected from model if not set)",
```
sarahyurick: Why would we ever not auto-detect it?

nemo_curator/backends/experimental/utils.py
```diff
@@ -102,7 +103,14 @@ def get_available_cpu_gpu_resources(
...
 def _setup_stage_on_node(stage: ProcessingStage, node_info: NodeInfo, worker_metadata: WorkerMetadata) -> None:
-    """Ray remote function to execute setup_on_node for a stage."""
+    """Ray remote function to execute setup_on_node for a stage.
+
+    This runs as a Ray remote task (not an actor).
+    If the stage initializes vLLM, vLLM's auto-detection will force spawn
+    only forces spawn inside Ray actors, otherwise it will use fork which will result in CUDA fork errors.
+    We explicitly set the environment variable to spawn to prevent this.
+    """
+    os.environ.setdefault("VLLM_WORKER_MULTIPROC_METHOD", "spawn")
```
sarahyurick, line 114: Curious how did you encounter this?

nemo_curator/stages/text/deduplication/semantic.py
```diff
@@ -270,15 +266,13 @@ def _run_embedding_generation(self, executor: BaseExecutor) -> list[Task]:
         pipeline.add_stage(reader)
 
         # Embedding generation stage
-        embedding_stage = EmbeddingCreatorStage(
+        embedding_stage = VLLMEmbeddingModelStage(
```
sarahyurick, line 271: We can pass `cache_dir` here too right?

tests/stages/text/deduplication/test_semantic.py
```diff
@@ -118,7 +120,8 @@ def test_config(
             output_path=str(request.cls.output_dir),
             cache_path=str(request.cls.cache_dir),
             perform_removal=True,
-            model_identifier="sentence-transformers/all-MiniLM-L6-v2",
+            model_identifier=MODEL_ID,
+            embedding_vllm_init_kwargs={"enforce_eager": True},
```
sarahyurick, line 124: Curious why do you use this for the test? Since by default we are setting it to be False.

tutorials/text/deduplication/semantic/semantic_e2e.ipynb
```diff
@@ -264,7 +167,7 @@
     "    error_msg = \"The number of GPUs on this machine are lesser than the default this tutorial was tested with, please update `num_gpus` passed into `RayClient`\"\n",
     "    raise ValueError(error_msg)\n",
     "\n",
-    "client = RayClient(num_cpus=64, num_gpus=NUM_GPUS)\n",
+    "client = RayClient(num_cpus=64, num_gpus=NUM_GPUS, ray_temp_dir=\"/raid/praateekm/ray_temp\")\n",
```
sarahyurick:
```suggestion
    "client = RayClient(num_cpus=64, num_gpus=NUM_GPUS)\n",
```

tutorials/text/deduplication/semantic/semantic_step_by_step.ipynb
```diff
@@ -133,7 +174,7 @@
     "    error_msg = \"The number of GPUs on this machine are lesser than the default this tutorial was tested with, please update `num_gpus` passed into `RayClient`\"\n",
     "    raise ValueError(error_msg)\n",
     "\n",
-    "client = RayClient(num_cpus=64, num_gpus=NUM_GPUS)\n",
+    "client = RayClient(num_cpus=64, num_gpus=NUM_GPUS, ray_temp_dir=\"/raid/praateekm/ray_temp\")\n",
```
sarahyurick:
```suggestion
    "client = RayClient(num_cpus=64, num_gpus=NUM_GPUS)\n",
```

tutorials/text/deduplication/semantic/semantic_step_by_step.ipynb
```diff
@@ -217,33 +223,29 @@
     "\n",
     "We output the embeddings as Parquet files so that we can read more smartly during our K-Means step. This is the recommended file format before you run K-Means.\n",
     "\n",
-    "See the [API Reference](https://docs.nvidia.com/nemo/curator/latest/apidocs/stages/stages.text.embedders.base.html#stages.text.embedders.base.EmbeddingCreatorStage) for more information about the `EmbeddingCreatorStage` class."
+    "By default, we use `VLLMEmbeddingModelStage` which runs embedding models via vLLM for better throughput. If you prefer to use PyTorch / HuggingFace models directly (e.g., with `sentence-transformers`), you can use the `EmbeddingCreatorStage` instead:\n",
```
sarahyurick:
```suggestion
    "By default, we use `VLLMEmbeddingModelStage` which runs embedding models via vLLM for better throughput. If you prefer to use PyTorch or Hugging Face's Sentence Transformers, you can use the `EmbeddingCreatorStage` instead:\n",
```

tutorials/text/deduplication/semantic/semantic_step_by_step.ipynb
```diff
@@ -217,33 +223,29 @@
...
+    "\n",
+    "```python\n",
+    "from nemo_curator.stages.text.embedders import EmbeddingCreatorStage\n",
+    "\n",
+    "EmbeddingCreatorStage(\n",
```
sarahyurick, line 190: Mention `use_sentence_transformer` parameter too?

tutorials/text/deduplication/semantic/semantic_step_by_step.ipynb
```diff
@@ -254,13 +256,11 @@
...
-    "        EmbeddingCreatorStage(\n",
-    "            model_identifier=\"sentence-transformers/all-MiniLM-L6-v2\",\n",
+    "        VLLMEmbeddingModelStage(\n",
+    "            model_identifier=\"google/embeddinggemma-300m\",\n",
     "            text_field=\"text\",\n",
-    "            max_seq_length=None,\n",
     "            max_chars=None,\n",
-    "            embedding_pooling=\"mean_pooling\",\n",
-    "            model_inference_batch_size=256,\n",
+    "            pretokenize=False,\n",
```
sarahyurick, line 223: We always recommend `pretokenize=False` right?

nemo_curator/backends/experimental/utils.py
praateekmahajan: Just running the pipeline with Ray Data results in a failure. From Slack

> 1. Ray Data performs setup_on_node using Ray's tasks.. Xenna does everything using actors. ([Slide 31](https://docs.google.com/presentation/d/15lJHyAoRTbNFDWq8UJ6czMaUFck3WYPCRaMYPfO_qew/edit?slide=id.g3be823d0e3d_0_327#slide=id.g3be823d0e3d_0_327) in Developer Gudie)
> 2. vLLM checks when spawning the subprocess for the engine, whether its in a Ray Actor or Task ([code here](https://github.com/vllm-project/vllm/blob/c88ea8338b9ad2f01bfb24c7bbf8ae6140866afd/vllm/utils/system_utils.py#L126-L160))
> 3. Since setup_on_node also starts vLLM.LLM the subprocess errors out since CUDA process can't be forked

tests/stages/text/deduplication/test_semantic.py
praateekmahajan: `enforce_eager=True` disables capturing cuda graph and does a `torch.compile` which adds a minute or two, therefore we don't do that for our tests

benchmarking/scripts/embedding_generation_benchmark.py
praateekmahajan: Good catch, that was a bug (which would've resulted in an error) which I didn't catch because I ran benchmarks with embeddinggemma.
As a fix I'm just calling the same method which is used by vLLM to resolve the seq length which is the same as ST config.. HF models use a different config though..

So TLDR `(vLLM == ST) != HF` in case of miniLM

benchmarking/scripts/embedding_generation_benchmark.py
praateekmahajan: Yea good question TBH. Will remove it (it probably was there before I added the auto-detection and I was running miniLM benchmarks where I was manually setting it to 256)

nemo_curator/stages/text/deduplication/semantic.py
praateekmahajan: Good question, went into a long rabbit hole to see how it behaves when we are in an air-gapped environment. 
Yes we can pass `cache_dir`

tutorials/text/deduplication/semantic/semantic_step_by_step.ipynb
praateekmahajan: Its' very benchmark dependent, but good for default to be False

tests/stages/text/deduplication/test_semantic.py
weijiac0619: got the same issue when I worked on [integration test for video pipeline](https://github.com/NVIDIA-NeMo/Curator/pull/1614) haha. setting it into true disables CUDA Graphs and saved time. It runs operators one-by-one. This makes the model start instantly but run slightly slower, better for unit test but not good for large scale inference

benchmarking/scripts/embedding_generation_benchmark.py
```diff
@@ -12,86 +12,188 @@
...
+    model_path = snapshot_download(model_identifier, cache_dir=cache_dir, local_files_only=True)
+
+    st_config = get_sentence_transformer_tokenizer_config(model_path)
+    if st_config is None:
+        msg = f"No sentence-transformers config found for {model_identifier}"
+        raise ValueError(msg)
+
+    model_limit = st_config["max_seq_length"]
+    logger.info(f"Resolved max_seq_length={model_limit} from sentence-transformers config")
+    return model_limit
```
sarahyurick, line 72: Cool, I didn't know about `get_sentence_transformer_tokenizer_config`.

TBH I am not a huge fan of `snapshot_download` being here, but I guess it is okay since you are using `cache_dir` and the model needs downloading eventually anyways.

nemo_curator/stages/text/deduplication/semantic.py
sarahyurick: Doesn't it still need to be explicitly added here?

tests/stages/text/deduplication/test_semantic.py
sarahyurick: Thanks, good to know.

tutorials/text/deduplication/semantic/semantic_step_by_step.ipynb
```diff
@@ -217,33 +182,30 @@
...
+    "\n",
+    "```python\n",
+    "from nemo_curator.stages.text.embedders import EmbeddingCreatorStage\n",
+    "\n",
+    "EmbeddingCreatorStage(\n",
+    "    model_identifier=\"sentence-transformers/all-MiniLM-L6-v2\",\n",
+    "    text_field=\"text\",\n",
+    "    embedding_pooling=\"mean_pooling\",\n",
+    "    model_inference_batch_size=256,\n",
+    "    use_sentence_transformer=False, # or false in which case embedding_pooling is ignored\n",
```
sarahyurick:
```suggestion
    "    use_sentence_transformer=False, # or True in which case embedding_pooling is ignored\n",
```

nemo_curator/stages/text/deduplication/semantic.py
weijiac0619: yes let's pass `cache_dir` here

praateekmahajan: Will add it here, looks like we never had it in the past too.

benchmarking/scripts/embedding_generation_benchmark.py
praateekmahajan: Yea, so the reason for having `cache_dir` in the first place is the fact that, `embeddinggemma` needs an `hf_token` and our nightly's don't have it. So for now I've copied over the model weights to `model_weights` path.  

If I don't pass `model_path` to  `get_sentence_transformer_tokenizer_config ` it would try to resolve at default i.e. `~/.cache/hf_cache` and the `json` file doesn't exist there.

And to know where exactly we need to load from we need `snapshot_download(..)` which doesn't re-download if the model already exists at that path.

PR #1610, opened 2026-03-16
Add ray data for image

oyilmaz-nvidia:
## Description
This PR adds a few changes to test and benchmark Ray Data for image workflows.

## Fanouts
Why IS_FANOUT_STAGE on ImageReaderStage:

ImageReaderStage.process() returns list[ImageBatch] — for each .tar file it reads, DALI may produce multiple batches. In Ray Data, all those batches from one tar end up in the same block after map_batches. Without IS_FANOUT_STAGE, all of them get sent to the same downstream embedding actor, killing parallelism. The flag triggers repartition(target_num_rows_per_block=1) in the adapter, splitting them into individual blocks so each ImageBatch can be picked up by any available ImageEmbeddingStage actor independently.

It's the same reason FilePartitioningStage has it — it also returns a list[FileGroupTask].

## Should You Add More Fanouts?
For the standard pipeline, no.

Every other image stage returns a single task — confirmed:

Stage	                                                 process() return type	Fanout needed?
ImageReaderStage	                                 list[ImageBatch]	        ✅ added
ImageEmbeddingStage	                         ImageBatch	                no
ImageAestheticFilterStage	                 ImageBatch	                no
ImageNSFWFilterStage	                         ImageBatch	                no
ImageWriterStage	                                 FileGroupTask	                no
ConvertImageBatchToDocumentBatch	DocumentBatch	        no
DeduplicationRemovalStage	                 ImageBatch	                no

ImageReaderStage is the only image stage that fans out (1 tar → N batches), so it's the only one that needs the flag.

## Benchmarking:
The runtimes of Xenna and Ray Data for image curation benchmarking are almost the same.

## Checklist

### DIFF
--- a/benchmarking/nightly-benchmark.yaml
+++ b/benchmarking/nightly-benchmark.yaml
@@ -584,7 +584,7 @@ entries:
       - metric: num_images_processed
         exact_value: 3800
       - metric: throughput_images_per_sec
-        min_value: 2.5
+        min_value: 3.0
 
   - name: image_curation_xenna
     enabled: true


nemo_curator/stages/image/io/image_reader.py
```diff
@@ -51,6 +53,12 @@ def __post_init__(self) -> None:
         else:
             self.resources = Resources()
 
+    def ray_stage_spec(self) -> dict[str, Any]:
+        """Ray stage specification for this stage."""
+        return {
+            RayStageSpecKeys.IS_FANOUT_STAGE: True,
+        }
+
```
VibhuJawa, line 61: This makes sense, thanks for adding it .

oyilmaz-nvidia:
@suiyoubi Change the min val to default and the run time is similar to xenna.

Issue #1555, opened 2026-02-25
Fail tests if the test data is not found


oyilmaz-nvidia:
**Describe the bug**

A suggestion regarding tests: I am seeing some of the video.filtering tests such as tests/stages/video/filtering/test_clip_aesthetic_filter.py are printing test_video.mp4 not found but the tests pass even though they were not run. I think we should fail them.

sarahyurick:
We are also skipping an image test because the data is not there yet: https://github.com/NVIDIA-NeMo/Curator/blob/main/tests/stages/image/io/test_image_reader.py#L203

weijiac0619:
The "error message" seen in the test output (est_video.mp4 not found) was coming from VideoTask.validate(), which called os.path.exists and printed when the file wasn't found. However, Task.__post_init__ calls validate() but ignores the return value, so construction succeeded and all test logic ran normally. changing print into log.warning here in this PR.https://github.com/NVIDIA-NeMo/Curator/pull/1614

nightcityblade:
Hi, I'd like to work on this. I'll submit a PR shortly.

nightcityblade:
I took a closer look and I’m going to step back from this one for now rather than submit a half-baked fix. The right change likely needs a clearer decision about whether task validation should raise on missing files or whether the affected tests should use explicit test fixtures.

nightcityblade:
Hi, I’m picking this back up and will submit a focused PR shortly.

PR #1614, opened 2026-03-16
[Curator video caption pipeline] integration tests(vLLM + HF transformers) for Qwen 

weijiac0619:
## Description

1. Refer to the [design doc here](https://docs.google.com/document/d/1XmI1wUYaxXysTBlWifgas-J1w9yFAP0DEsMm0qC-sc0/edit?usp=sharing). Adds three end-to-end integration tests for the Qwen2.5-VL caption pipeline, exercising the real vLLM and HF transformers stack. 
The tests are scoped to the Qwen2.5-VL-7B model and require a GPU + model weights to run, gated by @pytest.mark.integration

2. For the https://github.com/NVIDIA-NeMo/Curator/issues/1555, we changed print into log.warning to remove user confusion. The "error message" seen in the test output (est_video.mp4 not found) was coming from VideoTask.validate(), which called os.path.exists and printed when the file wasn't found. However, Task.__post_init__ calls validate() but ignores the return value, so construction succeeded and all test logic ran normally.

## Usage

```
 pytest -m integration -s \
    --model-dir /path/to/models \
    tests/stages/video/caption/test_caption_integration.py
```

  Requires: GPU, Qwen2.5-VL-7B-Instruct weights at <model-dir>/Qwen/Qwen2.5-VL-7B-Instruct/. Model dir can also be set via QWEN_VL_MODEL_DIR env var; tests skip cleanly if neither is provided.

## Checklist

### DIFF
--- a/benchmarking/test-paths.yaml
+++ b/benchmarking/test-paths.yaml
@@ -1,9 +1,78 @@
 results_path: /tmp/curator/results/${BRANCH_NAME}
 datasets_path: /tmp/curator/datasets
-model_weights_path: /tmp/curator/datasets
-sinks: []
+model_weights_path: /tmp/curator/model_weights
 datasets:
   - name: "tinystories"
     formats:
-      - type: "parquet"
-        path: "{datasets_path}/tinystories_train_parquet"
+    - type: "parquet"
+      path: "{datasets_path}/tinystories_train_parquet"
+  - name: "commoncrawl"
+    formats:
+    - type: "jsonl"
+      path: "{datasets_path}/cleaned_exact_dedup_all_cc"
+    - type: "parquet_gemma_embeddings"
+      path: "{datasets_path}/cleaned_exact_dedup_all_cc_gemma_embeddings"
+  - name: "commoncrawl_id_map"
+    formats:
+    - type: "json"
+      path: "{datasets_path}/cleaned_exact_dedup_all_cc_fuzzy_output_nightly_container_paths/fuzzy_id_generator.json"
+  - name: "commoncrawl_ids"
+    formats:
+    - type: "parquet"
+      path: "{datasets_path}/cleaned_exact_dedup_all_cc_fuzzy_output_nightly_container_paths/FuzzyDuplicateIds"
+  - name: "mscoco"
+    formats:
+    - type: "wds"
+      path: "{datasets_path}/mscoco/wds/truncated_100K_mscoco_benchmarking"
+  - name: "mscoco_model_weights"
+    formats:
+    - type: "files"
+      path: "{datasets_path}/mscoco/model_weights"
+  - name: "videos"
+    formats:
+    - type: "mp4"
+      path: "{datasets_path}/openvid-1m/video"
+  - name: "videos_model_weights"
+    formats:
+    - type: "files"
+      path: "{model_weights_path}/video_model_weights"
+  - name: "rpv2-2023-14-en"
+    formats:
+    - type: "parquet"
+      path: "{datasets_path}/rpv2_2023-14_en"
+  - name: "arxiv_downloads"
+    formats:
+    - type: "tar"
+      path: "{datasets_path}/arxiv_downloads"
+  - name: "fasttext_langid_model"
+    formats:
+    - type: "bin"
+      path: "{model_weights_path}/text_model_weights/lid.176.bin"
+  - name: "fasttext_quality_model"
+    formats:
+    - type: "bin"
+      path: "{model_weights_path}/text_model_weights/fasttext-oh-eli5/openhermes_reddit_eli5_vs_rw_v2_bigram_200k_train.bin"
+  - name: "ndd_gretel_symptoms"
+    formats:
+    - type: "jsonl"
+      path: "{datasets_path}/ndd_gretel_symptoms"
+  - name: "text_models_hf_cache"
+    formats:
+    - type: "files"
+      path: "{model_weights_path}/text_model_weights/hf_cache"
+  - name: "multimodal_mint1t"
+    formats:
+    - type: "wds_tar_dir"
+      path: "{datasets_path}/multimodal/mint1t"
+  - name: "finemath4plus"
+    formats:
+    - type: "parquet"
+      path: "{datasets_path}/finemath4plus/enriched"
+  - name: "cc_index_query_dataset"
+    formats:
+    - type: "parquet"
+      path: "{datasets_path}/cc_index_benchmark/dataset/megamath_web/megamath-web"
+  - name: "cc_index_partitions"
+    formats:
+    - type: "parquet"
+      path: "{datasets_path}/cc_index_benchmark/cc_index"
--- a/benchmarking/tools/generate_ci_tests.py
+++ b/benchmarking/tools/generate_ci_tests.py
@@ -13,7 +13,6 @@
 # limitations under the License.
 
 import argparse
-import re
 from pathlib import Path
 
 from ruamel.yaml import YAML
@@ -40,45 +39,6 @@ def seconds_to_time(seconds: int) -> str:
     minutes, secs = divmod(remainder, 60)
     return f"{hours:02d}:{minutes:02d}:{secs:02d}"
 
-
-def get_required_datasets(entry: dict) -> set:
-    """
-    Parse {dataset:NAME,FORMAT} patterns from entry args.
-
-    Args:
-        entry: Dictionary from nightly-benchmark.yaml entries list
-
-    Returns:
-        datasets: Set of (name, format) tuples required by the entry
-    """
-    args = entry.get("args", "")
-    # Find all {dataset:NAME,FORMAT} references in args
-    all_datasets = set(re.findall(r"\{dataset:([^,}]+),([^}]+)\}", args))
-    # Cache dirs are downloaded at runtime, so they don't need to be pre-available
-    cache_datasets = set(re.findall(r"--cache-dir=\{dataset:([^,}]+),([^}]+)\}", args))
-    # Only require datasets that are not cache dirs
-    return all_datasets - cache_datasets
-
-
-def load_available_datasets(test_paths_file: str) -> set:
-    """
-    Load available (name, format) pairs from a test-paths YAML file.
-
-    Args:
-        test_paths_file: Path to the test-paths YAML file
-
-    Returns:
-        available: Set of (name, format) tuples available for testing
-    """
-    with open(test_paths_file, encoding="utf-8") as f:
-        config = yaml.load(f)
-
-    available = set()
-    for ds in config.get("datasets", []):
-        name = ds["name"]
-        for fmt in ds.get("formats", []):
-            available.add((name, fmt["type"]))
-    return available
 
 
 def generate_job(entry: dict, scope: str) -> dict:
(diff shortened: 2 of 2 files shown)


tests/stages/video/caption/conftest.py
```diff
@@ -0,0 +1,124 @@
...
+    else:
+        os.environ["TMPDIR"] = old
+
+
+@pytest.fixture(scope="session")
+def video_fixture_path() -> Path:
+    """Return the path to the small video fixture used for integration tests."""
+    path = _DEFAULT_VIDEO_FIXTURE
+    if not path.exists():
+        pytest.skip(f"Video fixture not found: {path}")
```
suiyoubi: I think we should raise error when we cannot find the test video.

tests/stages/video/caption/test_caption_integration.py
```diff
@@ -0,0 +1,245 @@
...
+
+
+# ---------------------------------------------------------------------------
+# Tests
+# ---------------------------------------------------------------------------
+
+
+@pytest.mark.integration
+@pytest.mark.gpu
+class TestQwenCaptionPipelineIntegration:
```
suiyoubi: This is great. In addition to the VL model, I suggest maybe we also adds a test for text model as well ? (CaptionEnhancementStage)

.gitignore
```diff
@@ -157,3 +157,4 @@ data/
 .DS_Store
 AGENTS.md
 alm_output/
+.tmp/
```
suiyoubi: Could we do cleanup in the test instead of adding to gitignore

suiyoubi:
/ok to test [2cdd471](https://github.com/NVIDIA-NeMo/Curator/pull/1614/commits/2cdd47104d88c476b1dc8038bcac0538dcd25a9f)

tests/stages/video/caption/conftest.py
```diff
@@ -0,0 +1,157 @@
+# Copyright (c) 2025, NVIDIA CORPORATION.  All rights reserved.
```
sarahyurick:
```suggestion
# Copyright (c) 2026, NVIDIA CORPORATION.  All rights reserved.
```

tests/stages/video/caption/test_caption_integration.py
```diff
@@ -0,0 +1,301 @@
+# modality: video
+
+# Copyright (c) 2025, NVIDIA CORPORATION.  All rights reserved.
```
sarahyurick:
```suggestion
# Copyright (c) 2026, NVIDIA CORPORATION.  All rights reserved.
```

nemo_curator/models/qwen_lm.py
```diff
@@ -62,7 +65,7 @@ def setup(self) -> None:
         self.llm = LLM(
             model=self.weight_file,
             quantization="fp8" if self.fp8 else None,
-            enforce_eager=False,
+            enforce_eager=self.enforce_eager,
```
sarahyurick: Should https://github.com/NVIDIA-NeMo/Curator/blob/main/nemo_curator/models/vllm_model.py#L113 be updated too?

pyproject.toml
```diff
@@ -342,4 +342,5 @@ testpaths = ["tests"]
 markers = [
     "asyncio",
     "gpu: marks tests as GPU tests (deselect with '-m \"not gpu\"')",
+    "integration: marks tests as integration tests requiring real GPU + model weights (select with '-m integration')",
```
sarahyurick: If I am understanding correctly, should https://github.com/NVIDIA-NeMo/Curator/blob/main/tests/stages/text/classifiers/test_classifiers.py be marked as integration tests too, since all of those tests download text classification models from Hugging Face?

What about https://github.com/NVIDIA-NeMo/Curator/blob/main/tests/stages/text/embedders/test_vllm.py too? And maybe some of the tests under https://github.com/NVIDIA-NeMo/Curator/blob/main/tests/stages/text/embedders/test_base.py and https://github.com/NVIDIA-NeMo/Curator/blob/main/tests/stages/text/deduplication/test_semantic.py too?

nemo_curator/models/qwen_lm.py
weijiac0619: The integration tests in this PR don’t go through VLLMModel; they only exercise QwenVL and QwenLM.

pyproject.toml
weijiac0619: I'm using that tag here narrowly. if there is no --model-dir / QWEN_VL_MODEL_DIR set for QwenVL / QwenLM weights then the tests will be skipped. That keeps CI fast and avoids requiring 7B+ VL weights on every GPU run. the tests you mentioned here seems to pull smaller HF models into cache.

sarahyurick: Sounds good, thanks for clarifying.

nemo_curator/models/qwen_lm.py
```diff
@@ -47,11 +47,14 @@ class QwenLM(ModelInterface):
...
 
-    def __init__(self, model_dir: str, caption_batch_size: int, fp8: bool, max_output_tokens: int):
+    def __init__(
+        self, model_dir: str, caption_batch_size: int, fp8: bool, max_output_tokens: int, enforce_eager: bool = False
+    ):
         self.model_dir = model_dir
         self.caption_batch_size = caption_batch_size
         self.fp8 = fp8
         self.max_output_tokens = max_output_tokens
+        self.enforce_eager = enforce_eager
```
praateekmahajan: Is there a preference to do expose many vLLM args as first class arguments rather than args?

tests/stages/video/caption/conftest.py
```diff
@@ -0,0 +1,157 @@
...
+# ---------------------------------------------------------------------------
+
+# Repo root: tests/stages/video/caption/conftest.py is 4 levels deep.
+_REPO_ROOT = Path(__file__).resolve().parents[4]
+
+
+@pytest.fixture(scope="session", autouse=True)
+def shared_ray_cluster() -> None:  # type: ignore[override]
+    """No-op override: caption integration tests run stages directly, not via Ray."""
+    return
```
praateekmahajan, line 35: Hmm what's the reason for this? In many of our tests, we do run bunch of stages directly, and ray cluster is not used. I don't think this should be necessary

tests/stages/video/caption/conftest.py
```diff
@@ -0,0 +1,157 @@
...
+# ---------------------------------------------------------------------------
+
+_ENV_VAR = "QWEN_MODEL_DIR"
+
+# Small video committed to the repo so tests are self-contained and portable.
+# 3s, 240x136, 2fps, mpeg4 — 3.8 KB
+_DEFAULT_VIDEO_FIXTURE = Path(__file__).parent / "fixtures" / "test_video.mp4"
+
+
+def pytest_addoption(parser: pytest.Parser) -> None:
```
praateekmahajan: I'm not too aware of this, but what does this do? What's the expected behavior in CI?

tests/stages/video/caption/conftest.py
```diff
@@ -0,0 +1,157 @@
...
+
+    env_value = os.environ.get(_ENV_VAR)
+    if env_value:
+        return env_value
+
+    pytest.skip(f"Model directory not specified. Pass --model-dir /path/to/models or set ${_ENV_VAR}.")
+
+
+@pytest.fixture(scope="session", autouse=True)
+def pipeline_tmpdir() -> Generator[Path, None, None]:
```
praateekmahajan: Why is this needed? You should have access to `tmp_path_factory: pytest.TempPathFactory` which is basically

I understand the 107 character limit (even Ray runs into it), however typically `tmp_path_factory` resolves to something like
```
/tmp/pytest-of-$USER/pytest-$ID/test_name
```
where $ID is an int that keeps increasing as you run more tests.. so at some point if you `rm -rf /tmp/pytest-of-$USER/` your $ID will reset to 0..

tests/stages/video/caption/test_caption_integration.py
```diff
@@ -0,0 +1,301 @@
...
+
+
+# ---------------------------------------------------------------------------
+# Tests
+# ---------------------------------------------------------------------------
+
+
+@pytest.mark.integration
+@pytest.mark.gpu
+class TestQwenCaptionPipelineIntegration:
```
praateekmahajan: Any reason to not add this to `test_caption_generation.py`... 

So far we've tried to follow the design principle of source file has a one-to-one mapping of test file, therefore its very easy to trace if I make a change in file X then I need to change file Y.

I'd recommend moving the `TestQwenCaptionGenerationPipelineIntegration` there and testing it in one file. You can create a utils.py inside the `tests/stages/video/captions` where you have mock code.

tests/stages/video/caption/test_caption_integration.py
```diff
@@ -0,0 +1,301 @@
...
+    """End-to-end integration tests for the Qwen caption pipeline.
+
+    Each test asserts *pipeline integrity*, not caption content:
+    - No unhandled exceptions
+    - Data flows from one stage to the next without loss
+    - Output fields are populated and have the expected types / shapes
+    """
+
+    def test_preparation_stage_populates_windows(
```
praateekmahajan: I see that the three tests test on the same `video_fixture_path` and run through a `preparation_stage.process`.. 

You can actually restructure the tests to look something like this (see `test_kmeans.py`)... this should make your expensive part (i.e. fwd pass through vLLM run once) while you perform assertions on the output


```python
class TestIntegration:
    output_of_stage = None

    @pytest.fixture(scope="class", autouse=True)
    def run_on_clip(self, request: pytest.FixtureRequest, preparation_stage: CaptionPreparationStage, video_fixture_path: Path):
        
        video_bytes = video_fixture_path.read_bytes()
        task = _make_task(video_bytes, task_id="prep-stage-test")

        self.output_of_stage =  preparation_stage.process(task)

    def test_returns_captions(self):

           assert self.output_of_stage has captions

    def test_has_windows(...): 

           assert bla bla
````

File: tests/stages/video/caption/test_caption_integration.py [hunk 14]
```diff
@@ -0,0 +1,301 @@
...
+
+
+# ---------------------------------------------------------------------------
+# Caption Enhancement tests
+# ---------------------------------------------------------------------------
+
+
+@pytest.mark.integration
+@pytest.mark.gpu
+class TestQwenLMCaptionEnhancementIntegration:
```
praateekmahajan: Similarly I think this can live inside test_caption_enhancement.py

pyproject.toml
praateekmahajan: Dumb question why do we need a new tag? If we do maybe `--gpu-model-tests` or something, since an integration test need not have anything to do with models / weights at all.. e.g. kmeans is an integration tests or test_pipeline.py has an integration test

nemo_curator/models/qwen_lm.py
weijiac0619: Replaced enforce_eager with **vllm_kwargs

tests/stages/video/caption/conftest.py
weijiac0619: The override is needed here specifically because shared_ray_cluster is autouse=True and unconditionally starts a Ray cluster + sets RAY_ADDRESS for every test session even if no test uses Ray

tests/stages/video/caption/conftest.py
weijiac0619: good catch. updated

pyproject.toml
weijiac0619: does the GPU CI runner have a persistent HF model cache? Qwen2.5-VL-7B (~15GB) + Qwen2.5-14B (~28GB) is ~43GB total. it's fine if models are cached, but would likely blow the CI timeout if downloaded fresh every run

sarahyurick: It does not, but we are planning to add an HF cache soon.

cc @thomasdhc

tests/stages/video/caption/test_caption_integration.py
weijiac0619: refactored

tests/stages/video/caption/test_caption_generation.py
```diff
@@ -12,11 +12,18 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from __future__ import annotations
```
praateekmahajan, line 15: do we need these?

PR #1626, opened 2026-03-18
Server call with  OpenAI API

karpnv:
## Description
Start Qwen3 server and call it in Curator Stage via OpenAI API.

### Run server using 2 GPU 

```bash
docker pull qwenllm/qwen3-omni
export NUM_GPU=2
docker run -itd --restart unless-stopped --gpus='all' --ipc=host \
 --shm-size=8g --ulimit memlock=-1 --name qwen3-1 -p 8201:8201 \
 -v /home:/home -v /mnt:/mnt qwenllm/qwen3-omni \
 bash -lc "vllm serve Qwen/Qwen3-Omni-30B-A3B-Instruct --port 8201 --host 0.0.0.0 --dtype bfloat16 --max-model-len 65536 --allowed-local-media-path / -tp $NUM_GPU"
```

### Run the pipeline that uses the server
```bash
conda activate curator3

git clone https://github.com/NVIDIA-NeMo/Curator.git
cd Curator
pip install -e .
pip install -r tutorials/audio/qwen_omni/requirements.txt
python tutorials/audio/qwen_omni/run_pipeline.py \
  --input_manifest --input_manifest s3://granary/manifest_0.json \
  --input_tar s3://granary/audio_0.tar \
  --lhotse_mode nemo_tarred \
  --host localhost --port 8201
```

### One-line example to run server and pipeline in one container
```python
docker run --gpus all --ipc=host --shm-size=8g \
    -v /path/to/data:/data \
    -e HEALTH_TIMEOUT=1200 -e NUM_GPU=2 \
    -e VLLM_MODEL=Qwen/Qwen3-Omni-30B-A3B-Instruct \
    -e VLLM_PORT=8213 \
    -e VLLM_MAX_MODEL_LEN=4096 \
    nvcr.io/nvidian/curator-qwen3-omni:latest \
    --input_manifest /data/input.jsonl \
    --input_tar /path/to/backet_0.tar \
    --lhotse_mode nemo_tarred \
    --output_path /data/output/
```

### DIFF
--- a/tutorials/audio/qwen_omni/run_serve_pipeline.py
+++ b/tutorials/audio/qwen_omni/run_serve_pipeline.py
@@ -16,7 +16,8 @@ import argparse
 from nemo_curator.core.client import RayClient
 from nemo_curator.backends.experimental.ray_data import RayDataExecutor
 from nemo_curator.pipeline import Pipeline
-from nemo_curator.stages.audio.request.omni_llm import OpenAiRequestStage
+from nemo_curator.models.client import AsyncOpenAIClient
+from nemo_curator.stages.audio.request.omni_llm_request import OmniLLMRequestStage
 from nemo_curator.stages.text.io.writer.jsonl import JsonlWriter
 from nemo_curator.stages.audio.request.prepare_omni_lhotse import PrepareOmniLhotseStage
 from nemo_curator.core.serve import InferenceModelConfig, InferenceServer
@@ -102,6 +103,13 @@ def main(args: argparse.Namespace) -> None:
         server.start()
 
     base_url = f"http://{args.host}:{args.port}/v1"
+    llm_client = AsyncOpenAIClient(base_url=base_url, api_key=args.api_key)
+    generation_config = {
+        "max_tokens": args.max_tokens,
+        "temperature": args.temperature,
+        "top_p": args.top_p,
+    }
+
     pipeline = Pipeline(name="qwen3_omni")
     pipeline.add_stage(
         PrepareOmniLhotseStage(
@@ -114,12 +122,10 @@ def main(args: argparse.Namespace) -> None:
         )
     )
     pipeline.add_stage(
-        OpenAiRequestStage(
+        OmniLLMRequestStage(
+            client=llm_client,
             model_name=args.model_name,
-            server_endpoint=base_url,
-            api_key=args.api_key,
-            predicted_text_key="predicted_text",
-            fields_to_drop=["messages"],
+            generation_config=generation_config,
         )
     )
     pipeline.add_stage(JsonlWriter(path=args.output_path, write_kwargs={"force_ascii": False}).with_(batch_size=1))


Jorjeous:
Will add s3cfg reading, + untarred s3 processing with boto3. + need to add new dependencies to docker image

nemo_curator/stages/audio/request/omni_llm_request.py
Jorjeous, line 1: Fix name typo

nemo_curator/stages/audio/request/prepare_omni_lhotse.py
```diff
@@ -147,11 +147,14 @@ def _build_cutset(self) -> CutSet:
...
+        """Images always use ``image_url``; audio depends on ``format`` setting."""
+        if image:
+            return "image_url"
         if self.format == "data_url":
-            return "image_url" if image else "audio_url"
+            return "audio_url"
         if self.format == "input_data":
-            return "input_image" if image else "input_audio"
-        msg = f"Invalid format: {self.format}. Use data_url or input_data."
+            return "input_audio"
```
karpnv: Is input_image not supported?

nemo_curator/stages/audio/request/prepare_omni_lhotse.py
```diff
@@ -0,0 +1,221 @@
...
+        content_parts: list[dict] = [part]
+        if str(text).strip():
+            content_parts.append({"type": "text", "text": text})
+
+        messages.append({"role": "user", "content": content_parts})
+        return messages
+
+    def process(self, _: _EmptyTask) -> DocumentBatch:
+        cuts = self._build_cutset()
+        rows: list[dict] = []
```
melllinia, line 447: It looks like `PrepareOmniLhotseStage.process()` loads and base64-encodes all audio before returning anything.

For small datasets the issue is not visible, but for large datasets (~3M samples, in my example I used YODAS), this blocks the pipeline for a long time, keeps GPUs idle, and uses a lot of memory.
In my run (8×A100), it spent ~34 minutes in this stage, processed 0 samples, and got killed due to idle GPUs.

Maybe we can make this more incremental instead of loading everything at once. For example:

- First stage: just emit shard info (e.g., manifest + tar paths), no audio loading
- Second stage: read one shard at a time, load audio, and emit results

This way inference can start as soon as the first shard is ready, and memory will not be overloaded, CPUs can load data while GPUs do inference.

@karpnv @Jorjeous

tutorials/audio/qwen_omni/run_pipeline.py
```diff
@@ -0,0 +1,153 @@
...
+            ],
+        )
+        server.start()
+
+    try:
+        if not args.no_ray_local:
+            ray.init(address="local", ignore_reinit_error=True)
+
+        base_url = f"http://{args.host}:{args.port}/v1"
+        llm_client = OpenAIClient(base_url=base_url, api_key=args.api_key)
```
melllinia: This is using the sync `OpenAIClient`, so it sends one request at a time.
Even with `--data-parallel-size 4`, only one vLLM replica is actually being used because the client waits for each response before sending the next request.

We could switch to `AsyncOpenAIClient` to send multiple requests in parallel and keep all replicas busy. For example:

```
llm_client = AsyncOpenAIClient(
    base_url=base_url,
    api_key=args.api_key,
    max_concurrent_requests=32,
)
```

Could we either default to the async client here, or add a flag to enable it? The async version is already available in the codebase.

nemo_curator/stages/audio/request/prepare_omni_lhotse.py
```diff
@@ -0,0 +1,221 @@
...
+        for idx, cut in enumerate(cuts, start=1):
+            row = {
+                "cut_id": cut.id,
+                "messages": self._cut_to_messages(cut),
+            }
```
melllinia, line 454: The Lhotse cut has useful metadata like `cut.duration`, `cut.supervisions[0].text` (reference transcript), and the audio filepath, but only `cut_id` and `messages` are kept here.

Could we preserve at least duration and text for later WER calculation and throughput analysis? Something like:

```suggestion
            row = {
                "cut_id": cut.id,
                "messages": self._cut_to_messages(cut),
                "duration": cut.duration,
                "text": cut.supervisions[0].text if cut.supervisions else None,
            }
```

tutorials/audio/qwen_omni/run_pipeline.py
Jorjeous: Probably should work now!

nemo_curator/stages/audio/request/prepare_omni_lhotse.py
Jorjeous: Testing this new ver

PR #1576, opened 2026-03-05
Add BandFilterStage for audio bandwidth classification

shubhamNvidia:
Adds BandFilterStage which classifies audio as full_band or narrow_band using a scikit-learn model on spectral features. Items not matching the configured band_value are filtered out. Supports standalone usage (loads audio from filepath) and pipeline usage (accepts waveform from upstream stages). Includes BandFilterConfig, the band_filter_module (feature extraction + prediction), and unit tests with mocked predictor.

PR #1575, opened 2026-03-05
Add audio preprocessing stages (MonoConversion, SegmentConcatenation, TimestampMapper)

shubhamNvidia:
Adds foundational audio preprocessing stages for the ADV pipeline:
MonoConversionStage: Converts multi-channel audio to mono and verifies sample rate, with strict/non-strict sample rate enforcement.
SegmentConcatenationStage: Concatenates multiple audio segments with configurable silence gaps between them.
Config dataclasses (MonoConversionConfig, SegmentConcatenationConfig) with from_dict/to_dict support.
Unit tests with mocked I/O covering config, stage properties, process logic, and edge cases.

PR #1640, opened 2026-03-22
Add AudioDataFilterStage composite pipeline for end-to-end audio curation

shubhamNvidia:
## Summary

- Add `AudioDataFilterStage`, a `CompositeStage` that decomposes into a configurable sequence of independent audio processing stages for extracting clean single-speaker segments from raw audio files.
- Add `AudioDataFilterConfig` dataclass with unified configuration for all pipeline stages (VAD, band filter, UTMOS, SIGMOS, speaker separation, timestamp mapping).
- The composite stage leverages the executor's cross-file parallelism -- each sub-stage has its own resource allocation (CPU for band/concat, GPU for VAD/UTMOS/SIGMOS/speaker-sep).

## Pipeline stages (when all enabled)

1. **MonoConversion** (CPU, 1:1) -- normalize to mono at target sample rate
2. **VAD** (GPU, batch mode) -- segment audio into speech chunks
3. **BandFilter** (CPU, 1:1) -- filter by bandwidth (full_band/narrow_band)
4. **UTMOS** (GPU, 1:1) -- filter by MOS quality score
5. **SIGMOS** (GPU, 1:1) -- filter by multi-dimensional quality metrics
6. **SegmentConcatenation** (CPU, 1:1) -- merge filtered segments with silence gaps
7. **SpeakerSeparation** (GPU, 1:N fan-out) -- diarize and split by speaker
8-11. **Per-speaker filters** -- VAD + Band + UTMOS + SIGMOS on each speaker's audio
12. **TimestampMapper** (CPU, 1:1) -- resolve final positions back to original file

## Configuration

All stages are individually toggleable via `AudioDataFilterConfig`:

- `enable_vad`, `enable_band_filter`, `enable_utmos`, `enable_sigmos`, `enable_speaker_separation`
- Quality thresholds: `utmos_mos_threshold`, `sigmos_noise_threshold`, `sigmos_ovrl_threshold`, etc.
- General: `sample_rate`, `band_value`, `vad_min_duration_sec`, `vad_max_duration_sec`
- Supports `to_dict()` / `from_dict()` for serialization

## Usage

from nemo_curator.stages.audio.advance_pipelines import AudioDataFilterStage, AudioDataFilterConfig

config = AudioDataFilterConfig(
    enable_utmos=True,
    enable_sigmos=True,
    enable_speaker_separation=True,
    utmos_mos_threshold=3.5,
)
pipeline.add_stage(AudioDataFilterStage(config=config))

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,6 +1,7 @@
 # Checklist
 
 We are glad you are contributing to NeMo Curator! Before you make a PR, be sure to read over this guide in detail.
+This project will only accept contributions under the project's Apache2 license.
 This checklist ensures that NeMo Curator stays easy-to-use by both users and developers.
 Not all steps are necessary for some contributions, so read the linked sections for more information about each item.
 
--- a/LICENSE
+++ b/LICENSE
@@ -173,6 +173,7 @@
       incurred by, or claims asserted against, such Contributor by reason
       of your accepting any such warranty or additional liability.
 
+
    END OF TERMS AND CONDITIONS
 
    APPENDIX: How to apply the Apache License to your work.
@@ -199,3 +200,25 @@
    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
    See the License for the specific language governing permissions and
    limitations under the License.
+
+MIT License
+
+Copyright (c) Microsoft Corporation.
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE
--- a/uv.lock
+++ b/uv.lock
@@ -9284,12 +9284,12 @@ dependencies = [
     { name = "typing-extensions", marker = "platform_machine == 'x86_64' and sys_platform != 'darwin'" },
 ]
 wheels = [
-    { url = "https://download.pytorch.org/whl/cu128/torch-2.9.1%2Bcu128-cp310-cp310-manylinux_2_28_x86_64.whl", hash = "sha256:c8d670aa0be6fbecd2b0e7b7d514a104dbdefcc3786ca446cf0c3415043ea40a" },
-    { url = "https://download.pytorch.org/whl/cu128/torch-2.9.1%2Bcu128-cp310-cp310-win_amd64.whl", hash = "sha256:64399adaa8ea0896d02cf844cba3c5dd77e769520a1af73572599e0eaa2cf551" },
-    { url = "https://download.pytorch.org/whl/cu128/torch-2.9.1%2Bcu128-cp311-cp311-manylinux_2_28_x86_64.whl", hash = "sha256:2a1da940f0757621d098c9755f7504d791a72a40920ec85a4fd98b20253fca4e" },
-    { url = "https://download.pytorch.org/whl/cu128/torch-2.9.1%2Bcu128-cp311-cp311-win_amd64.whl", hash = "sha256:633005a3700e81b5be0df2a7d3c1d48aced23ed927653797a3bd2b144a3aeeb6" },
-    { url = "https://download.pytorch.org/whl/cu128/torch-2.9.1%2Bcu128-cp312-cp312-manylinux_2_28_x86_64.whl", hash = "sha256:7cb4018f4ce68b61fd3ef87dc1c4ca520731c7b5b200e360ad47b612d7844063" },
-    { url = "https://download.pytorch.org/whl/cu128/torch-2.9.1%2Bcu128-cp312-cp312-win_amd64.whl", hash = "sha256:3a01f0b64c10a82d444d9fd06b3e8c567b1158b76b2764b8f51bfd8f535064b0" },
+    { url = "https://download-r2.pytorch.org/whl/cu128/torch-2.9.1%2Bcu128-cp310-cp310-manylinux_2_28_x86_64.whl", hash = "sha256:c8d670aa0be6fbecd2b0e7b7d514a104dbdefcc3786ca446cf0c3415043ea40a" },
+    { url = "https://download-r2.pytorch.org/whl/cu128/torch-2.9.1%2Bcu128-cp310-cp310-win_amd64.whl", hash = "sha256:64399adaa8ea0896d02cf844cba3c5dd77e769520a1af73572599e0eaa2cf551" },
+    { url = "https://download-r2.pytorch.org/whl/cu128/torch-2.9.1%2Bcu128-cp311-cp311-manylinux_2_28_x86_64.whl", hash = "sha256:2a1da940f0757621d098c9755f7504d791a72a40920ec85a4fd98b20253fca4e" },
+    { url = "https://download-r2.pytorch.org/whl/cu128/torch-2.9.1%2Bcu128-cp311-cp311-win_amd64.whl", hash = "sha256:633005a3700e81b5be0df2a7d3c1d48aced23ed927653797a3bd2b144a3aeeb6" },
+    { url = "https://download-r2.pytorch.org/whl/cu128/torch-2.9.1%2Bcu128-cp312-cp312-manylinux_2_28_x86_64.whl", hash = "sha256:7cb4018f4ce68b61fd3ef87dc1c4ca520731c7b5b200e360ad47b612d7844063" },
+    { url = "https://download-r2.pytorch.org/whl/cu128/torch-2.9.1%2Bcu128-cp312-cp312-win_amd64.whl", hash = "sha256:3a01f0b64c10a82d444d9fd06b3e8c567b1158b76b2764b8f51bfd8f535064b0" },
 ]
 
 [[package]]


nemo_curator/stages/audio/advance_pipelines/Audio_data_filter/audio_data_filter.py
```diff
@@ -0,0 +1,178 @@
...
+    VADSegmentationStage,
+    SIGMOSFilterStage,
+    UTMOSFilterStage,
+    BandFilterStage,
+    SpeakerSeparationStage,
+    SegmentConcatenationStage,
+    TimestampMapperStage,
+)
+from nemo_curator.stages.audio.configs import (
+    VADConfig,
```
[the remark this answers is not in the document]
shubhamNvidia: This pr will be merged with other PR there all this stages are present

nemo_curator/stages/audio/advance_pipelines/Audio_data_filter/audio_data_filter.py
```diff
@@ -0,0 +1,178 @@
...
+                stages.append(BandFilterStage(
+                    config=BandFilterConfig(band_value=cfg.band_value),
+                    name="BandFilter_Speaker").with_(resources=band_res))
+
+            if cfg.enable_utmos:
+                stages.append(UTMOSFilterStage(
+                    config=UTMOSConfig(mos_threshold=cfg.utmos_mos_threshold),
+                    name="UTMOS_Speaker").with_(resources=gpu_res))
+
+            if cfg.enable_sigmos:
```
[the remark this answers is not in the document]
shubhamNvidia: Even without speaker separation, the TimestampMapper serves a valuable purpose:
Strips waveform from final output (metadata-only output, saves memory)
Normalizes output format — every item gets the same standard fields regardless of pipeline path
Applies passthrough_keys filtering — controls which metadata fields appear in final output
If we guarded it behind enable_speaker_separation, the non-speaker-sep path would output raw items with waveform tensors still attached and no normalized timestamp fields.
Verdict: Not a bug. The unconditional append is intentional and correct. The _build_output_item_no_mapping path exists specifically for this case.

shubhamNvidia:
@sarahyurick I’ve merged with main and resolved the conflicts—you can go ahead and run the Ruff linter now

nemo_curator/stages/audio/advanced_pipelines/audio_data_filter/default_config.yaml
sarahyurick, line 1: The stages themselves can be initialized when reading the YAML file by providing the import path. See https://github.com/NVIDIA-NeMo/Curator/blob/main/tutorials/audio/fleurs/pipeline.yaml#L47 as an example.

sarahyurick:
Hi @shubhamNvidia the Ruff list is here: https://github.com/NVIDIA-NeMo/Curator/actions/runs/23553029249/job/68572060454?pr=1640

nemo_curator/stages/audio/advanced_pipelines/audio_data_filter/default_config.yaml
shubhamNvidia: That functionality was also added

nemo_curator/stages/audio/advance_pipelines/Audio_data_filter/audio_data_filter.py
```diff
@@ -0,0 +1,230 @@
...
+        self,
+        config_path: str | Path | None = None,
+        config: dict[str, Any] | None = None,
+        name: str = "AudioDataFilter",
+    ) -> None:
+        super().__init__()
+        self.name = name
+        self._cfg = load_config(config_path)
+        if config:
+            from .config import _deep_merge
```
sarahyurick: This can be a top-level import.

shubhamNvidia: Currently, there is no test for this PR. Will add tests once the dependent PRs (filtering, segmentation, preprocessing) are merged; currently can't test due to cross-PR import dependencies.

nemo_curator/stages/audio/advanced_pipelines/audio_data_filter/audio_data_filter.py
```diff
@@ -0,0 +1,228 @@
...
+
+"""
+Audio Data Filter Stage -- CompositeStage that decomposes into independent
+pipeline stages for extracting clean single-speaker segments.
+
+Pipeline (when all filters + speaker separation enabled)::
+
+    1. MonoConversion (1:1)
+    2. VAD batch mode (1:1, items = N segments)
+    3. BandFilter (1:1, filter items)
```
sarahyurick, line 23: Hi @shubhamNvidia can you remove the unnecessary `_initialize_predictor` call in the `BandFilterStage`'s `process` method here: https://github.com/NVIDIA-NeMo/Curator/blob/main/nemo_curator/stages/audio/filtering/band.py#L171 ?

nemo_curator/stages/audio/advanced_pipelines/audio_data_filter/audio_data_filter.py
```diff
@@ -0,0 +1,228 @@
+# Copyright (c) 2026, NVIDIA CORPORATION.  All rights reserved.
```
mohammadaaftabv, line 1: directory name violates PEP 8 — package names should be all-lowercase with underscores. This makes the import path look wrong:

`from nemo_curator.stages.audio.advance_pipelines.Audio_data_filter import AudioDataFilterStage`

Every other package in the codebase uses lowercase: `audio/alm/`, `audio/filtering/`, `audio/preprocessing/`, `audio/segmentation/`.

Rename to `audio_data_filter/`.

(greptile already flagged `advance_pipelines` → `advanced_pipelines`, but the subdirectory naming also needs fixing.)

nemo_curator/stages/audio/advanced_pipelines/audio_data_filter/audio_data_filter.py
```diff
@@ -0,0 +1,228 @@
...
+                    buffer_time=speaker.get("buffer_time", 0.5),
+                    name="SpeakerSeparation",
+                )
+            )
+
+            self._append_filter_stages(
+                stages,
+                vad,
+                band,
+                utmos,
```
mohammadaaftabv, line 155: When speaker separation is enabled (the default), the full filter chain (VAD → BandFilter → UTMOS → SIGMOS) runs TWICE:

1. First pass: filters on the raw audio segments
2. After SpeakerSeparation: same filters again on per-speaker segments

UTMOS and SIGMOS are GPU-inference stages. Running them twice doubles the GPU compute cost. This is especially wasteful because audio quality metrics (MOS scores, noise, bandwidth) don't fundamentally change between pre-speaker and post-speaker segments — the waveform content is the same, just attributed to different speakers.

With defaults, `AudioDataFilterStage()` decomposes into 12 stages including redundant GPU inference.

Consider making the second filter pass optional via a config flag (`post_speaker_refilter: true/false`), or document why double filtering is architecturally necessary.

nemo_curator/stages/audio/advanced_pipelines/audio_data_filter/default_config.yaml
```diff
@@ -0,0 +1,53 @@
...
+
+band_filter:
+  enable: true
+  band_value: full_band
+
+utmos:
+  enable: true
+  mos_threshold: 3.5
+
+sigmos:
```
mohammadaaftabv, line 36: Every stage has `enable: true` by default, including speaker separation, SIGMOS, and UTMOS. A user writing `AudioDataFilterStage()` with no args gets the full 12-stage pipeline including GPU inference stages that require:
- Silero VAD model (torch.hub download)
- SIGMOS ONNX model + onnxruntime + librosa + scipy
- UTMOS model
- SpeakerSeparation NeMo model

First-time usage will likely crash if users don't have all these dependencies and models available.

Consider either:
(a) Enabling only basic stages by default (mono conversion, VAD) and requiring explicit opt-in for GPU-heavy stages, or
(b) Documenting the full dependency set prominently in the class docstring.

nemo_curator/stages/audio/advanced_pipelines/audio_data_filter/audio_data_filter.py
```diff
@@ -0,0 +1,228 @@
...
+        if config:
+            self._cfg = _deep_merge(self._cfg, config)
+
+    def decompose(self) -> list[ProcessingStage]:
+        cfg = self._cfg
+        stages: list[ProcessingStage] = []
+
+        mc = cfg.get("mono_conversion", {})
+        stages.append(
+            MonoConversionStage(
```
mohammadaaftabv, line 92: `MonoConversionStage` is always added unconditionally — there's no `enable` flag for it, unlike every other stage (VAD, Band, UTMOS, SIGMOS, SpeakerSep all have `enable` toggles in the YAML config).

If audio is already mono at the correct sample rate, this is wasted processing. For consistency, add `enable: true` to the `mono_conversion` section in `default_config.yaml` and guard the append in `decompose()`.

nemo_curator/stages/audio/advanced_pipelines/audio_data_filter/config.py
```diff
@@ -0,0 +1,145 @@
...
+        logger.warning(f"Unknown config sections (ignored): {unknown_sections}")
+
+    merged = _deep_merge(defaults, user_cfg)
+
+    _validate(merged)
+
+    return merged
+
+
+def _validate(cfg: dict[str, Any]) -> None:
```
mohammadaaftabv: `_validate()` only checks two constraints:
  1. `vad.min_duration_sec < vad.max_duration_sec`
  2. `concatenation.silence_duration_sec >= 0`
  
  Missing validations for threshold parameters:
  - UTMOS `mos_threshold` should be in [1, 5] (MOS scale)
  - SIGMOS thresholds (`noise_threshold`, `ovrl_threshold`, etc.) should be in [1, 5]
  - VAD `threshold` should be in [0, 1]
  - `speaker_separation.min_duration` should be positive
  - `mono_conversion.output_sample_rate` should be a positive integer
  
  A user passing `utmos: { mos_threshold: 50 }` gets no validation error but filters out 100% of audio silently.

nemo_curator/stages/audio/advanced_pipelines/audio_data_filter/audio_data_filter.py
```diff
@@ -0,0 +1,228 @@
...
+        """Append VAD + quality filter stages to *stages* list."""
+        if enable_vad:
+            stages.append(
+                VADSegmentationStage(
+                    min_duration_sec=vad.get("min_duration_sec", 2.0),
+                    max_duration_sec=vad.get("max_duration_sec", 60.0),
+                    threshold=vad.get("threshold", 0.5),
+                    min_interval_ms=vad.get("min_interval_ms", 500),
+                    speech_pad_ms=vad.get("speech_pad_ms", 300),
+                    nested=(suffix == ""),
```
mohammadaaftabv, line 207: `nested=(suffix == "")` passes `True` to `VADSegmentationStage` for the first (pre-speaker) pass and `False` for the post-speaker pass. This controls whether VAD stores segments as a nested list inside a single task vs. fanning out into separate tasks.

This is an implicit coupling between `AudioDataFilterStage` and `VADSegmentationStage` that isn't documented. A comment explaining the intent would help future maintainers:

```python
# Pre-speaker pass: nested=True so VAD stores segments inside the task
# (SegmentConcatenation needs all segments in one task to merge them).
# Post-speaker pass: nested=False so VAD fans out into separate tasks.

File: nemo_curator/stages/audio/__init__.py [hunk 11]
```diff
@@ -18,9 +18,11 @@
 This module provides stages for processing and curating audio data,
 including ASR inference, quality assessment, ALM data preparation,
 and bandwidth classification filtering,
-and audio preprocessing (mono conversion, segment concatenation, timestamp mapping).
+and audio preprocessing (mono conversion, segment concatenation, timestamp mapping),
+and advanced audio processing pipelines.
 """
 
+from nemo_curator.stages.audio.advance_pipelines import AudioDataFilterStage
```
mohammadaaftabv: Adding `AudioDataFilterStage` to the top-level `audio/__init__.py` means `import nemo_curator.stages.audio` triggers the entire import chain:

- `AudioDataFilterStage` → `SIGMOSFilterStage` → third-party `sigmos.py` → `import librosa`, `import scipy`, `import onnxruntime`
- `AudioDataFilterStage` → `VADSegmentationStage` → `import torch`, `import torchaudio`, `from silero_vad import ...`
- `AudioDataFilterStage` → `UTMOSFilterStage` → `import torch`

Users who only need basic audio stages (e.g., `GetAudioDurationStage`, `PreserveByValueStage`) pay the cost of loading all these heavy ML libraries.

Fix with lazy-importing `AudioDataFilterStage`.

mohammadaaftabv (changes requested):

mohammadaaftabv:
Adding comments for adv specific changes already on main below:

1.`SpeakerSeparationStage` imports `pydub` at module level in both `speaker_separation.py` and `speaker_sep.py`, but `pydub` is not declared in the `audio_common`, `audio_cpu`, or `audio_cuda12` extras in `pyproject.toml`.

Users installing with `pip install nemo_curator[audio_cpu]` will get error 
when importing `SpeakerSeparationStage`.

Fix: add `"pydub>=0.25.1"` to the `audio_common` extra (same category as the missing `librosa`/`scipy` for SIGMOS).

2. Use ffspec instead of os everywhere. Multiple locations use `os.path.exists()` or `os.path.isfile()` to check audio file paths:

- `common.py` → `resolve_waveform_from_item()` (line 102, 115)
- `mono_conversion.py` → `process()` (line 731)
- `utmos.py` → `_load_waveform_tensor()` uses `os.path.isfile(path)`

These silently return `False` for cloud paths (`s3://`, `gs://`, etc.), causing tasks to be dropped without error. The codebase already has `fsspec` available. Consider:

```python
from fsspec.core import url_to_fs
fs, path = url_to_fs(audio_filepath)
if fs.exists(path):
```

3. main — nemo_curator/stages/audio/segmentation/speaker_separation_module/speaker_sep.py lines 332–342 (diarize_audio) and lines 478–487 (get_speaker_audio_data)
When called with a waveform tensor (not a file path), `speaker_sep.py` creates temp files twice for the same input:

1. `diarize_audio()` writes waveform → temp `.wav`, passes to `SortformerEncLabelModel.diarize(audio=temp_path)`
2. `get_speaker_audio_data()` writes the same waveform → another temp `.wav`, reads it as `AudioSegment.from_file(temp_path)`

The second write could be avoided by constructing `AudioSegment` directly from the numpy array:
```python
from pydub import AudioSegment
import numpy as np

samples = (waveform.squeeze().cpu().numpy() * 32767).astype(np.int16)
original_audio = AudioSegment(
    data=samples.tobytes(),
    sample_width=2,
    frame_rate=sample_rate,
    channels=1,
)
```
3. In `speaker_sep.py` `get_speaker_audio_data()`:

```python
silent_audio = AudioSegment.silent(duration=duration_ms)
for start_time, end_time in segments:
    segment_audio = original_audio[start_ms:end_ms]
    silent_audio = silent_audio.overlay(segment_audio, position=start_ms)
    
```
For a 1-hour audio file, this creates a full-hour silent AudioSegment per speaker, then overlays small segments onto it. Memory usage scales with (num_speakers × file_duration) even if each speaker only has a few seconds of speech.

A more efficient approach would be to concatenate only the speaker's segments directly (like SegmentConcatenationStage)

4. `utmos.py` defines a standalone `_load_waveform_tensor()` function that duplicates most of the logic in `resolve_waveform_from_item()` from `common.py`. Both:

- Check for `waveform` + `sample_rate` in the item dict
- Fall back to loading from `audio_filepath`
- Handle missing sample rate
- Convert to correct tensor format

`UTMOSFilterStage` should use the shared utility from `common.py` instead. `BandFilterStage` (#1576) already uses `resolve_waveform_from_item` correctly.

5. `band_filter_module/predict.py` `predict_audio()` returns error strings on failure:

```python
return f"Error: Could not load model: {e}"
return f"Error during prediction: {e}"
```
The caller in band.py checks with isinstance(pred, str) and not pred.startswith("Error"), which is fragile — if the error message format changes, the check silently breaks.

Other stages (UTMOS, SIGMOS) use the standard pattern: raise an exception or return None. predict_audio() should do the same for consistency.

6. `torchcodec` is listed only under `audio_cuda12` but it has CPU wheels available on PyPI. If any stage uses `torchcodec` for audio decoding on CPU, it won't be available.

If `torchcodec` is intentionally CUDA-only, that should be documented. Otherwise, consider adding it to `audio_cpu` as well, or at least to `audio_common`.

7. `MonoConversionStage` with `strict_sample_rate=True` (default) silently drops any audio not at `output_sample_rate` (default 48kHz). With `strict_sample_rate=False`, it accepts as-is but doesn't resample.

There's no option to actually resample to the target rate, which limits the stage's usefulness for heterogeneous datasets. Consider adding a `resample=True` option using `torchaudio.transforms.Resample`, similar to what `UTMOSFilterStage` does internally with its `_resamplers` cache.

8. `SegmentConcatenationStage._concatenate()` logs a warning when segments have different sample rates but concatenates them anyway:

```python
if parts and sr != sample_rate:
    logger.warning(f"Sample rate mismatch: expected {sample_rate}Hz, got {sr}Hz. Output audio may be corrupted.")
```
Concatenating waveforms with different sample rates produces corrupted audio (pitch-shifted, time-stretched). This should either:

Resample the segment to match the first segment's rate, or Skip the mismatched segment with a warning or explain why this pipeline allows for corrupted audio output in the first place.

9. `speaker_sep.py` `_get_param()` supports 4 different config access patterns: direct attribute, dict access, nested dict `["speaker_separation"][param]`, and `.get()` method. But `SpeakerSeparationStage` always passes a plain `dict`.

This adds ~30 lines of complexity for unused flexibility. Consider simplifying to plain dict access:

```python
def _get_param(self, param_name: str, default_value):
    return self.config.get(param_name, default_value)
```    
10. `speaker_sep.py` suppresses: `C901` (complexity), `PLR0912` (too many branches), `PLR0913` (too many arguments), `BLE001` (broad except), `S110` (pass in except).

The timeline-based overlap processing in `clean_cut_overlapping_segments()` and `exclude_overlapping_segments()` could be extracted into a separate `TimelineProcessor` class with smaller, focused methods, reducing complexity and eliminating the need for `noqa` suppressions.

11. Tests only cover the stage wrapper (`process()`) with mocked separators. There are no unit tests for the core timeline algorithms in `SpeakerSeparator`:

- `clean_cut_overlapping_segments()`
- `exclude_overlapping_segments()`
- `merge_adjacent_segments()`
- `filter_short_segments()`

These are complex timeline algorithms with edge cases (adjacent segments, fully-overlapping segments, buffer time exceeding segment duration). PR #1575 (preprocessing) demonstrates excellent test coverage — the same standard should apply here.

nemo_curator/stages/audio/advanced_pipelines/audio_data_filter/audio_data_filter.py
shubhamNvidia: sure

nemo_curator/stages/audio/advanced_pipelines/audio_data_filter/audio_data_filter.py
shubhamNvidia: The double filter pass is intentional — after speaker separation, the per-speaker waveforms are different from the original mixed audio. VAD on isolated single-speaker audio produces different segment boundaries, and quality scores (UTMOS, SIGMOS) on separated speech are more representative of actual per-speaker signal quality. This was validated through extensive analysis on our test sets.

nemo_curator/stages/audio/advanced_pipelines/audio_data_filter/default_config.yaml
shubhamNvidia: The default config with all stages enabled represents our best-validated pipeline derived from extensive analysis, optimized for extracting high-quality audio data. Users who want our recommended accuracy and results can use it as-is. Any stage can be individually disabled via enable: false in the YAML config to match specific requirements or reduce dependency overhead. Will add prominent documentation listing the full dependency set so users know what's needed upfront.

nemo_curator/stages/audio/advanced_pipelines/audio_data_filter/audio_data_filter.py
shubhamNvidia: MonoConversion is intentionally unconditional because it also serves as the audio loader — it reads the file, populates waveform, sample_rate, duration, and num_samples into the task data. Disabling it would force each downstream stage to independently load from audio_filepath, causing redundant file I/O and risking failures on multi-channel files. The strict_sample_rate check acts as a guard against corrupted or mismatched files in large-scale datasets that would otherwise produce silent errors downstream. The operation is lightweight (CPU-only, no GPU), so keeping it always-on prevents failures at negligible overhead.

nemo_curator/stages/audio/advanced_pipelines/audio_data_filter/audio_data_filter.py
shubhamNvidia: done

nemo_curator/stages/audio/advanced_pipelines/audio_data_filter/audio_data_filter.py
shubhamNvidia: done

tests/stages/audio/advanced_pipelines/test_audio_data_filter.py
```diff
@@ -0,0 +1,330 @@
...
+        assert cfg["utmos"]["mos_threshold"] == 3.4
+
+
+# ---------------------------------------------------------------------------
+# 3. Config: _validate()
+# ---------------------------------------------------------------------------
+
+
+class TestValidate:
+    def test_validate_valid_defaults(self) -> None:
```
sarahyurick: I don't think this test is needed?

tests/stages/audio/advanced_pipelines/test_audio_data_filter.py
```diff
@@ -0,0 +1,330 @@
...
+class TestDecomposeConfig:
+    def test_decompose_vad_nested_flag(self) -> None:
+        stage = AudioDataFilterStage()
+        stages = stage.decompose()
+        vad_stages = [s for s in stages if isinstance(s, VADSegmentationStage)]
+        assert len(vad_stages) == 2
+        assert vad_stages[0].nested is True
+        assert vad_stages[1].nested is False
+
+    def test_decompose_stage_names_have_suffix(self) -> None:
```
sarahyurick: I don't think we need this one either.

tests/stages/audio/advanced_pipelines/test_audio_data_filter.py
```diff
@@ -0,0 +1,330 @@
...
+
+
+# ---------------------------------------------------------------------------
+# 8. Default YAML Consistency
+# ---------------------------------------------------------------------------
+
+
+class TestPickling:
+    def test_audio_data_filter_stage_pickling(self) -> None:
+        import pickle
```
sarahyurick: This can be a top-level import.

tests/stages/audio/advanced_pipelines/test_audio_data_filter.py
```diff
@@ -0,0 +1,330 @@
...
+            "sigmos", "concatenation", "speaker_separation", "timestamp_mapper",
+        ]
+        for stage_name in stages_with_cpus:
+            assert "cpus" in cfg[stage_name], f"{stage_name} missing 'cpus' in config"
+
+        stages_with_gpus = ["vad", "band_filter", "utmos", "sigmos", "speaker_separation"]
+        for stage_name in stages_with_gpus:
+            assert "gpus" in cfg[stage_name], f"{stage_name} missing 'gpus' in config"
+
+    def test_default_yaml_loads_without_error(self) -> None:
```
sarahyurick: Same here, I think this can be removed.

sarahyurick (commented):
Mostly LGTM, adding some comments about tests.

tests/stages/audio/advanced_pipelines/test_audio_data_filter.py
shubhamNvidia: done

tests/stages/audio/advanced_pipelines/test_audio_data_filter.py
shubhamNvidia: done

tests/stages/audio/advanced_pipelines/test_audio_data_filter.py
shubhamNvidia: done

PR #1570, opened 2026-03-04
[REVIEW] Feat/interleaved io readers writers

VibhuJawa:
## Summary

Adds two new interleaved IO stages and centralizes schema alignment for all arrow-based readers/writers:

- **InterleavedParquetReader / InterleavedParquetReaderStage**: Native pyarrow reading (no pandas round-trip). Gracefully null-fills columns absent from the file when fields are requested. Supports `schema` / `schema_overrides` for schema alignment.
- **InterleavedWebdatasetWriterStage**: Reconstructs WebDataset tar shards from interleaved rows. Uses `fsspec.open()` for cloud path support. Groups by `sample_id`, writes JSON metadata + binary image members per sample. Keys are percent-encoded (injective — `%` escaped first as `%25`) to prevent collisions.

> **Note:** `InterleavedLanceFragmentWriterStage` has been removed from this PR. It will be re-added in a follow-up with a proper `InterleavedLanceWriter` (CompositeStage) + `LanceCommitStage` (barrier via `num_workers=1` + `teardown()`) design.

### Centralized schema alignment (utils/schema.py)

All schema logic moved to a single module shared by all readers and writers:

| Function | Purpose |
|----------|---------|
| `reconcile_schema(inferred)` | Map core columns to INTERLEAVED_SCHEMA types, preserve large_string/large_binary, propagate field metadata |
| `align_table(table, target)` | Null-fill missing columns, drop extras, reorder, cast to target schema. `safe=False` only for reserved large↔small downcasts; `safe=True` for passthrough columns to surface overflow errors |

Both `BaseInterleavedReader` and `BaseInterleavedWriter` gain `schema` / `schema_overrides` parameters and `_align_output()`.

### Other changes

- **BaseInterleavedWriter**: `on_materialize_error` parameter (`error`/`warn`/`drop_row`/`drop_sample`); policy now applies to both fetch errors and errors set by upstream stages (e.g. `ImageValidationStage`). `_write_dataframe` changed from `@abstractmethod` to a concrete `NotImplementedError`-raising default so subclasses overriding `write_data()` directly don't need to implement it, while still failing fast if neither is overridden.
- **`InterleavedParquetReader.schema_overrides`**: tightened annotation from `dict[str, Any]` to `dict[str, pa.DataType]` to match `_resolve_schema` expectations.
[...]

PR #1652, opened 2026-03-24
feat(interleaved): Foundation IO improvements — schema utilities, base-class fixes, and bug fixes

VibhuJawa:
## Summary

This is **Part 1 of 2** splitting PR #1570. It contains only the foundation layer improvements (base classes, schema utilities, bug fixes) with no new reader/writer stages. Part 2 (new \`InterleavedParquetReader\` + \`InterleavedWebdatasetWriter\`) will follow as a separate PR building on top of this.

---

### New: \`utils/schema.py\`
Centralized schema utilities shared by all arrow-based readers and writers:
- \`reconcile_schema()\` — canonical types for reserved columns; passthrough columns preserved as-is; avoids unsafe large↔small downcasts; unwraps Parquet dictionary encoding from passthrough columns
- \`align_table()\` — pad, reorder, and cast an Arrow table to a target schema; reserved columns use \`safe=False\` for large↔small, passthrough columns use \`safe=True\` to surface overflow rather than silently corrupt data; does **not** re-reconcile the user-provided target (which would silently override intentional type choices)
- \`resolve_schema()\` — merges \`schema=\` and \`schema_overrides=\` with \`INTERLEAVED_SCHEMA\`; warns when both are provided; returns \`None\` when neither is set

### Reader base (\`readers/base.py\`)
- Add \`schema=\` and \`schema_overrides=\` parameters for strict output alignment
- Warn when both \`schema=\` and \`schema_overrides=\` are provided (instead of silently ignoring overrides)
- Preserve \`nullable=False\` from \`INTERLEAVED_SCHEMA\` when applying \`schema_overrides\`

### Writer base (\`writers/base.py\`)
- Same \`schema=\` / \`schema_overrides=\` API as the reader
- New \`on_materialize_error=\` parameter: \`"error"\` | \`"warn"\` | \`"drop_row"\` | \`"drop_sample"\`
- \`_materialize_dataframe()\` applies the error policy after fetch, covering errors set by both the fetch step and upstream stages
- \`_write_dataframe()\` changed from \`@abstractmethod\` to a default \`NotImplementedError\` body, allowing WDS-style subclasses that override \`write_data()\` directly without a dead stub
- Pre-compute \`_effective_write_kwargs\` in \`__post_init__\` (filters \`storage_options\`, forces \`index=False\`) instead of copying the dict on every task

### Materialization (\`utils/materialization.py\`)
- Extract \`_build_global_range_index()\`: group paths by filesystem object so mixed-backend batches (e.g. S3 + local) work correctly instead of failing silently; restore \`list[tuple[str, int, int]]\` type hints on \`_scatter_range_blobs\`
- Fix \`_extract_tiff_frame()\`: copy frame before closing the PIL context to avoid use-after-close; do not reuse source JPEG compression which corrupts RGBA frames

### WebDataset reader (\`readers/webdataset.py\`)
- Remove \`source_id_field\` — it only excluded a JSON key from passthrough columns, achievable via \`fields=\` already. Having both \`source_id_field\` and \`sample_id_field\` with overlapping names was confusing with no benefit
- Rename class from \`WebdatasetReaderStage\` → \`InterleavedWebdatasetReaderStage\` for consistency
- Rename tiff result variable from \`extracted\` to \`tiff_frame\` to eliminate variable shadowing
- Fix \`source_files\` metadata per split: when \`max_batch_bytes\` causes multiple output batches, each batch's \`source_files\` now lists only the tar files that actually contributed rows to that split (instead of all source tars). Tracks \`sample_id → tar\` during reading; deduplicates with \`unique()\` in Arrow before resolving.

### \`pyproject.toml\`
- Add \`multimodal_cpu\` extras group: \`Pillow\` and \`s3fs>=2024.12.0\`

---

## Tests added / updated
- \`test_base_writer.py\` (new, 9 tests): \`on_materialize_error\` modes, schema alignment, write orchestration
- \`test_materialization.py\`: 4 new tests for multi-backend path grouping
- \`test_multimodal_core.py\`: 4 new \`align_table\` tests for safe-cast behaviour
- \`test_multimodal_reader.py\`: \`schema_overrides\` and \`nullable\` preservation tests; remove all \`source_id_field=\` usage; new test \`test_reader_source_files_per_split_only_contributing_tars\` verifying per-split source file accuracy

**149 / 149 tests pass. Ruff clean.**

---

## What's NOT included (Part 2)

- \`InterleavedParquetReader\` / \`InterleavedParquetReaderStage\` — new composite + stage for reading Parquet
- \`InterleavedWebdatasetWriter\` / \`InterleavedWebdatasetWriterStage\` — new composite + stage for writing WebDataset tar archives

---

## Test plan

- [x] \`python -m pytest tests/stages/interleaved/ -v\` → 149 passed
- [x] \`python -m ruff check nemo_curator/stages/interleaved/ tests/stages/interleaved/\` → all checks passed
- [x] Pre-commit hooks pass (ruff, signed-off-by, large-file, private-key)

### DIFF
--- a/.github/workflows/config/.secrets.baseline
+++ b/.github/workflows/config/.secrets.baseline
@@ -350,6 +350,29 @@
         "is_verified": false,
         "line_number": 670
       }
+    ],
+    "nemo_curator/models/nemotron_h_vl.py": [
+      {
+        "type": "Hex High Entropy String",
+        "filename": "nemo_curator/models/nemotron_h_vl.py",
+        "hashed_secret": "7e44041c8ec081054af6ccf0758efa3771e99123",
+        "is_verified": false,
+        "line_number": 52
+      },
+      {
+        "type": "Hex High Entropy String",
+        "filename": "nemo_curator/models/nemotron_h_vl.py",
+        "hashed_secret": "f1a89dd2b19e1d30ac4fc1a8d1f4993796c9d1e6",
+        "is_verified": false,
+        "line_number": 54
+      },
+      {
+        "type": "Hex High Entropy String",
+        "filename": "nemo_curator/models/nemotron_h_vl.py",
+        "hashed_secret": "d7f11e2581dd9d25522592e7cfb9346a749f51f3",
+        "is_verified": false,
+        "line_number": 55
+      }
     ]
   },
   "generated_at": "2026-03-24T16:41:24Z"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -256,12 +256,12 @@ override-dependencies = [
     "apex; sys_platform == 'never'",
     "distance; sys_platform == 'never'",
     "huggingface-hub>=0.34,<1.0", # Override huggingface-hub, transformers and data-designer require two different versions of hugging-face hub
+    "transformers>=4.56.0,<5.0", # Override nemo-toolkit[asr]'s <=4.52.0 constraint; required for Nemotron VL support
     "kaldiio;  sys_platform == 'never'",
     "levenshtein;  sys_platform == 'never'",
     "numpy>=2.0.0,<=2.2.0", # Override nemo-toolkits constraint of <2.0.0, upperbounds for Numba compatibility
     "protobuf>=5.29.5",  # Override nemo-toolkits constraint of ~=5.29.5
     "setuptools>=80.10.1", # Override setuptools range in other dependencies to address CVE GHSA-58pv-8j8x-9vj2
-    "transformers<=4.55.2", # Else Cosmos Embed imports fail
     "nixl-cu12>=0.10.0; (platform_machine == 'x86_64' and platform_system != 'Darwin')",  # Override ray[llm]'s unconditional nixl dep for ARM
 ]
 
--- a/nemo_curator/models/cosmos_embed1.py
+++ b/nemo_curator/models/cosmos_embed1.py
@@ -32,9 +32,9 @@ _COSMOS_EMBED1_VARIANTS_INFO: Final = {
 }
 
 COSMOS_EMBED1_MODEL_REVISION_INFO: Final = {
-    "224p": "85f5627",
-    "336p": "5d8309d",
-    "448p": "9f4ff4d",
+    "224p": "787e0b9",
+    "336p": "0e8a28f",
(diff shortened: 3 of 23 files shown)


.github/workflows/config/.secrets.baseline
```diff
@@ -90,10 +90,6 @@
     {
       "path": "detect_secrets.filters.allowlist.is_line_allowlisted"
```
abhinavg4, line 91: Are all changes to this file intended?

benchmarking/scripts/multimodal_mint1t_benchmark.py
```diff
@@ -42,7 +42,6 @@ def create_pipeline(args: argparse.Namespace) -> Pipeline:
     )
     pipeline.add_stage(
         WebdatasetReader(
-            source_id_field="pdf_name",
```
abhinavg4, line 45: I think this change will affect pdf PR too. Why was this removed? Is the reading happening via schema now ?

.github/workflows/config/.secrets.baseline
VibhuJawa: yes. This is because of the tiff issue we fixed in this PR , these are all base64 jpegs (that i internally kept around), so that tutorial actually shows the issue visually too.

benchmarking/scripts/multimodal_mint1t_benchmark.py
VibhuJawa: I just removed it , because it was excessive to have around.  It now is just a passthrough column and not given special treatment.

VibhuJawa: I think if you remove it and its in the dataset, it will just work. Happy to help with your PR too.

VibhuJawa: We will revisit this discussion in 1657

pyproject.toml
thomasdhc, line 210: Do you want interleaved_cpu to be installed as apart of `all`?

VibhuJawa: yes, it should be . let me edit that .

VibhuJawa: added it !

PR #1657, opened 2026-03-24
feat(interleaved): add InterleavedParquetReader and InterleavedWebdatasetWriter

VibhuJawa:
# feat(interleaved): add InterleavedParquetReader and InterleavedWebdatasetWriter
 
## Overview
 
Completes the interleaved IO round-trip by adding the two missing directions:
 
```
Parquet  ──read──►  InterleavedBatch  ──write──►  WDS tar
```
 
With the existing WDS reader and Parquet writer on `main`, all four IO combinations are now supported:
 
```
WDS tar  ⇄  InterleavedBatch  ⇄  Parquet
```
 
## New classes
 
| Class | File | Description |
| --- | --- | --- |
| `InterleavedParquetReaderStage` | `io/readers/parquet.py` | Reads Parquet → `InterleavedBatch` |
| `InterleavedParquetReader` | `io/reader.py` | Composite stage (FilePartitioning + ReaderStage) |
| `InterleavedWebdatasetWriterStage` | `io/writers/webdataset.py` | Writes `InterleavedBatch` → MINT-1T-style tar shards |
 
### `InterleavedParquetReaderStage`
 
- `fields=` tuple selects passthrough columns — consistent with WDS reader; reserved columns always read; missing columns null-filled in a single pass
- `pq.read_schema()` per file for push-down column projection (avoids reading unnecessary data from disk/S3)
- `max_batch_bytes` splitting via `split_table_by_group_max_bytes`; each split's `source_files` lists only its contributing parquet files
- `_source_files_for_split()` moved to `BaseInterleavedReader` so both WDS and Parquet readers share per-split lineage tracking
 
### `InterleavedWebdatasetWriterStage`
 
- Overrides `_write_dataframe()`; base class handles deterministic naming, materialization, schema alignment, and `process()` orchestration
- `urllib.parse.quote(sample_id, safe="")` key escaping — injective, roundtrip-safe via `sample_id_field="sample_id"` on read
- `df.groupby("sample_id", sort=False)` — O(n) single pass instead of O(n×m) per-sample filter
- Positions may have gaps; output arrays sized `max_pos + 1` with `None` at gaps (WDS reader skips `None`)
- Supported modalities: `metadata`, `text`, `image` — any other raises `ValueError` at write time
 
## Benchmarks
 
All four IO paths benchmarked end-to-end on 80 local NVMe shards (6,818 samples, MINT-1T PDF data) with an aspect-ratio filter applied. Full results and reproduction instructions: [benchmark gist](https://gist.github.com/VibhuJawa/d743b9b122dc85342c9eef347cfa0b93).
 
| Path | Wall-clock | Samples/sec | Input | Output |
| --- | --- | --- | --- | --- |
| WDS → Parquet | 76.8 s | 88.8 | 9.9 GB (80 tars) | 3.84 GB (57,713 rows) |
| WDS → WDS | 75.4 s | 90.4 | 9.9 GB (80 tars) | 4.06 GB (6,818 samples) |
| **PQ → Parquet** | **15.7 s** | **435.0** | 6.0 GB (80 pq) | 3.83 GB (57,713 rows) |
| **PQ → WDS** | **18.5 s** | **368.4** | 6.0 GB (80 pq) | 4.06 GB (6,818 samples) |
 
**Key takeaway:** Parquet-sourced paths are **~5× faster** than WDS-sourced paths. The bottleneck is entirely in the WDS reader (396 s cumulative stage time across 80 shards vs 14 s for PQ) due to sequential tar scanning and network image fetches. Filter and writer costs are near-identical across formats.
 
### Stage-level breakdown (cumulative across 80 tasks, 16 workers)
 
| Stage | WDS→PQ | WDS→WDS | PQ→PQ | PQ→WDS |
| --- | --- | --- | --- | --- |
| Reader | 396.7 s | 395.3 s | 13.4 s | 14.1 s |
| Aspect ratio filter | 35.1 s | 34.2 s | 40.5 s | 40.6 s |
| Writer | 12.0 s | 45.4 s | 13.1 s | 48.1 s |
 
**Environment:** Python 3.10.12 · Ray 2.54.0 · PyArrow 19.0.1 · 16-core CPU · NVMe · `materialize_on_read=True`
 
### Parquet reader (`test_multimodal_reader.py`)
 
- `test_parquet_reader_roundtrip` — write then read back; sample_id, modality, text_content match
- `test_parquet_reader_missing_columns_filled_with_null` — 3-column file; all reserved cols null-filled
- `test_parquet_reader_fields_subset` — `fields=` selects extra cols; reserved always present
- `test_parquet_reader_fields_null_fill_missing` — `fields=("nonexistent",)` → null column, no error
- `test_parquet_reader_max_batch_bytes_splits` — 2 files → 2 splits; each lists only its contributing file in `source_files`
- `test_parquet_reader_empty_file` — 0-row parquet → empty batch, correct schema, no crash
- `test_parquet_reader_composite_decompose` — `decompose()` → `[FilePartitioningStage, InterleavedParquetReaderStage]`
 
### WDS writer (`test_multimodal_writer.py`)
 
- `test_escape_key_encodes_special_chars` — `a/b:c` → `a%2Fb%3Ac`
- `test_ext_from_content_type_known` / `_fallback` — MIME → extension mapping
- `test_wds_writer_roundtrip` — write then read back via `InterleavedWebdatasetReaderStage`
- `test_wds_writer_text_only_sample` — no image rows; JSON has all-None images list
- `test_wds_writer_unsupported_modality_raises` — `modality="video"` → `ValueError`
- `test_wds_writer_key_escaping` — special-char sample_id; roundtrip recovers original id
- `test_wds_writer_passthrough_columns_in_json` — extra `url` col in metadata row → appears in JSON payload
- `test_wds_writer_null_binary_skips_member` — null `binary_content` → no image tar member written
- `test_wds_writer_deterministic_filename` — same `source_files` + `task_id` → same output filename

### DIFF
--- a/nemo_curator/stages/interleaved/io/readers/parquet.py
+++ b/nemo_curator/stages/interleaved/io/readers/parquet.py
@@ -49,6 +49,10 @@ class InterleavedParquetReaderStage(BaseInterleavedReader):
     max_batch_bytes: int | None = None
     name: str = "interleaved_parquet_reader"
 
+    def __post_init__(self) -> None:
+        super().__post_init__()
+        self._storage_options = resolve_storage_options(io_kwargs=self.read_kwargs)
+
     def _columns_to_read(self, file_schema: pa.Schema) -> list[str] | None:
         """Return the column list to pass to ``pq.read_table``.
 
@@ -88,12 +92,11 @@ class InterleavedParquetReaderStage(BaseInterleavedReader):
         return table
 
     def process(self, task: FileGroupTask) -> InterleavedBatch | list[InterleavedBatch]:
-        storage_options = resolve_storage_options(io_kwargs=self.read_kwargs)
         tables: list[pa.Table] = []
         sample_id_to_path: dict[str, str] = {}
 
         for path in task.data:
-            fs, _ = url_to_fs(path, **(storage_options or {}))
+            fs, _ = url_to_fs(path, **(self._storage_options or {}))
             pa_fs = PyFileSystem(FSSpecHandler(fs))
             file_schema = pq.read_schema(path, filesystem=pa_fs)
             columns = self._columns_to_read(file_schema)
@@ -119,8 +122,8 @@ class InterleavedParquetReaderStage(BaseInterleavedReader):
                 metadata["source_files"] = list(task.data)
             else:
                 metadata["source_files"] = self._source_files_for_split(split, idx, sample_id_to_path, task.data)
-            if storage_options:
-                metadata["source_storage_options"] = storage_options
+            if self._storage_options:
+                metadata["source_storage_options"] = self._storage_options
             batches.append(
                 InterleavedBatch(
                     task_id=task_id,
--- a/nemo_curator/stages/interleaved/io/writers/webdataset.py
+++ b/nemo_curator/stages/interleaved/io/writers/webdataset.py
@@ -130,15 +130,15 @@ def _write_sample(
     texts: list[str | None] = [None] * (max_pos + 1)
     images: list[str | None] = [None] * (max_pos + 1)
 
-    for _, row in content_rows.iterrows():
-        pos = int(row["position"])
-        if row["modality"] == "text":
-            texts[pos] = row["text_content"]
-        elif row["modality"] == "image":
-            ext = _ext_from_content_type(row.get("content_type"))
+    for row in content_rows.itertuples(index=False):
+        pos = int(row.position)
+        if row.modality == "text":
+            texts[pos] = row.text_content
+        elif row.modality == "image":
+            ext = _ext_from_content_type(row.content_type)
             member_name = f"{escaped}.{pos}.{ext}"
             images[pos] = member_name
-            raw = row.get("binary_content")
+            raw = row.binary_content
             if not _is_null(raw):
                 img_bytes = bytes(raw)
                 info = tarfile.TarInfo(name=member_name)
--- a/nemo_curator/stages/interleaved/io/readers/webdataset.py
+++ b/nemo_curator/stages/interleaved/io/readers/webdataset.py
@@ -84,6 +84,7 @@ class InterleavedWebdatasetReaderStage(BaseInterleavedReader):
 
     def __post_init__(self) -> None:
         super().__post_init__()
+        self._storage_options = resolve_storage_options(io_kwargs=self.read_kwargs)
 
     # -- source_ref construction --
 
@@ -422,11 +423,10 @@ class InterleavedWebdatasetReaderStage(BaseInterleavedReader):
     def process(self, task: FileGroupTask) -> InterleavedBatch | list[InterleavedBatch]:
         rows: list[dict[str, Any]] = []
         sample_id_to_tar: dict[str, str] = {}
-        storage_options = resolve_storage_options(io_kwargs=self.read_kwargs)
 
         for tar_path in task.data:
             with (
-                fsspec.open(tar_path, mode="rb", **storage_options) as fobj,
+                fsspec.open(tar_path, mode="rb", **self._storage_options) as fobj,
                 tarfile.open(fileobj=fobj, mode="r:*") as tf,
             ):
                 members = [m for m in tf.getmembers() if m.isfile()]
@@ -435,7 +435,7 @@ class InterleavedWebdatasetReaderStage(BaseInterleavedReader):
                     tar_path=tar_path,
                     member_names=member_names,
                     member_info={m.name: m for m in members},
-                    storage_options=storage_options,
+                    storage_options=self._storage_options,
                     byte_cache={},
                 )
                 for member in members:
@@ -462,8 +462,8 @@ class InterleavedWebdatasetReaderStage(BaseInterleavedReader):
                 metadata["source_files"] = list(task.data)
             else:
                 metadata["source_files"] = self._source_files_for_split(split, idx, sample_id_to_tar, task.data)
-            if storage_options:
-                metadata["source_storage_options"] = storage_options
+            if self._storage_options:
+                metadata["source_storage_options"] = self._storage_options
             batches.append(
                 InterleavedBatch(
                     task_id=task_id,


nemo_curator/stages/interleaved/io/writers/webdataset.py
```diff
@@ -0,0 +1,169 @@
...
+def _ext_from_content_type(content_type: str | None) -> str:
+    """Return a file extension for *content_type*, falling back to ``"bin"``."""
+    if content_type:
+        ext = _CONTENT_TYPE_TO_EXT.get(content_type)
+        if ext:
+            return ext
+        guessed = mimetypes.guess_extension(content_type, strict=False)
+        if guessed:
+            return guessed.lstrip(".")
+    return "bin"
```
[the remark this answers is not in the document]
VibhuJawa: Resolved

nemo_curator/stages/interleaved/README.md
```diff
@@ -141,7 +139,7 @@ stages/interleaved/
 │   ├── reader.py                   # WebdatasetReader (CompositeStage)
 │   ├── readers/
 │   │   ├── base.py                 # BaseInterleavedReader
-│   │   └── webdataset.py           # WebdatasetReaderStage (ProcessingStage)
+│   │   └── webdataset.py           # InterleavedWebdatasetReaderStage (ProcessingStage)
 │   └── writers/
```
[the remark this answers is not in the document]
VibhuJawa: Resolved

nemo_curator/stages/interleaved/io/readers/parquet.py
```diff
@@ -0,0 +1,133 @@
...
+                table = table.append_column(schema_field, pa.nulls(len(table), type=schema_field.type))
+                existing.add(schema_field.name)
+        if self.fields:
+            for f in self.fields:
+                if f not in existing:
+                    table = table.append_column(pa.field(f, pa.null()), pa.nulls(len(table), type=pa.null()))
+        return table
+
+    def process(self, task: FileGroupTask) -> InterleavedBatch | list[InterleavedBatch]:
+        storage_options = resolve_storage_options(io_kwargs=self.read_kwargs)
```
abhinavg4: **Perf nit:** `resolve_storage_options` is called on every `process()` invocation, but `read_kwargs` is immutable after `__post_init__`. This means storage options are re-parsed once per task, which is unnecessary overhead — especially on S3/GCS paths where storage options construction isn't free.

Consider caching it in `__post_init__`, consistent with how `InterleavedWebdatasetReader` handles `storage_options`:
```python
def __post_init__(self):
    super().__init__()
    self._storage_options = resolve_storage_options(io_kwargs=self.read_kwargs)
```
Then reference `self._storage_options` in `process()`.

nemo_curator/stages/interleaved/io/writers/webdataset.py
```diff
@@ -0,0 +1,217 @@
...
+    passthrough_cols: list[str],
+) -> None:
+    """Write one sample (JSON + image binaries) into *tf*."""
+    escaped = _escape_key(sample_id)
+    content_rows = sample_df[sample_df["position"] >= 0].sort_values("position")
+    max_pos = int(content_rows["position"].max()) if not content_rows.empty else -1
+    texts: list[str | None] = [None] * (max_pos + 1)
+    images: list[str | None] = [None] * (max_pos + 1)
+
+    for _, row in content_rows.iterrows():
```
abhinavg4: **Perf nit:** `iterrows()` boxes every value into Python objects and is typically 5–10× slower than `itertuples()`. This is the inner loop over every content row in every sample in every task, so it's the hot path in the WDS writer.

Consider switching to `itertuples(name=None)`:
```python
for row in content_rows.itertuples(name=None):
    pos = int(row[content_rows.columns.get_loc('position') + 1])
    modality = row[content_rows.columns.get_loc('modality') + 1]
    ...
```
Or pre-extracting the needed columns as numpy arrays before the loop to avoid per-row overhead entirely.

benchmarking/scripts/utils.py
```diff
@@ -117,25 +185,37 @@ def collect_parquet_output_metrics(output_path: Path) -> dict[str, Any]:
...
+    """Collect output metrics for WebDataset tar archives (files, size, and sample count)."""
+    tar_paths, num_files, total_size_bytes = _collect_file_size_metrics(output_path, [".tar"])
+    num_samples = _count_wds_samples(tar_paths)
+    return {
+        "num_output_files": num_files,
+        "output_total_bytes": total_size_bytes,
+        "output_total_mb": total_size_bytes / 1e6,
+        "num_rows": None,
+        "num_samples": num_samples,
+        "modality_counts": {"metadata": num_samples} if num_samples else {},
```
abhinavg4: **Bug:** `modality_counts` is hardcoded to `{"metadata": num_samples}` even though WDS tars contain metadata, text, and image members. This is because `_count_wds_samples` only counts `.json` members (one per sample), so text and image are never tallied.

The downstream effect is misleading: when comparing input vs. output modality counts, text and image rows appear to have been dropped entirely, even when the pipeline ran correctly. For example on a PQ→WDS run:
```
input_modality_counts:  {"metadata": 1288, "text": 5853, "image": 6929}
output_modality_counts: {"metadata": 1288}   ← looks like data loss
```

There's also a secondary effect: `num_rows` is `None` for WDS output, so `throughput_rows_per_sec` evaluates to `0.0` in the metrics JSON (confirmed when running all 4 paths — WDS output paths always show `throughput_rows_per_sec: 0` regardless of actual speed).

Suggested fix: either count image-extension and text-extension tar members separately to report a real per-modality breakdown, or drop `modality_counts` from WDS output and document it as unavailable. The current approach silently gives false negatives on correctness checks.

VibhuJawa: fixed it

nemo_curator/stages/interleaved/io/writers/webdataset.py
VibhuJawa: fixed.

nemo_curator/stages/interleaved/io/readers/parquet.py
VibhuJawa: fixed

abhinavg4 (approved):
Thanks

PR #1427, opened 2026-01-26
[Tutorial] Merge multiple file prefixes generated by `MegatronTokenizerWriter`

asolergi-nv:
## Description

Addresses #1399. Adds `tutorials/text/megatron-tokenizer/merge_datasets.py` script to merge multiple file prefixes generated by `MegatronTokenizerWriter` into a single one . It is a simplified version of [`tools/merge_datasets.py`](https://github.com/NVIDIA/Megatron-LM/blob/main/tools/merge_datasets.py) script from the Megatron-LM library. Tested with the 2 different vocab sizes (≤65535, >65535).

## Usage

```bash
python tutorials/text/megatron-tokenizer/merge_datasets.py \
    --input-dir /path/to/tokenized/files \
    --output-prefix /path/to/output/merged
```
## Checklist

- [x] I am familiar with the [Contributing Guide](https://github.com/NVIDIA-NeMo/Curator/blob/main/CONTRIBUTING.md).
- [ ] New or Existing tests cover these changes.
- [x] The documentation is up to date with these changes.

Issue #1296, opened 2025-12-12
Unable to use models which need `trust_remote_code=True` in `EmbeddingCreatorStage`


kipraveen:
**Describe the bug**

When using models like `nvidia/llama-3.2-nv-embedqa-1b-v2` in `EmbeddingCreatorStage`, the pipeline fails with
```
ValueError: The repository nvidia/llama-3.2-nv-embedqa-1b-v2 contains custom code which must be executed to correctly load the model. You can inspect the repository content at https://hf.co/nvidia/llama-3.2-nv-embedqa-1b-v2 .
 You can inspect the repository content at https://hf.co/nvidia/llama-3.2-nv-embedqa-1b-v2.
Please pass the argument `trust_remote_code=True` to allow custom code to be run.
```

**Steps/Code to reproduce bug**

1. Use `EmbeddingCreatorStage` in a pipeline
2. Set `model_identifier='nvidia/llama-3.2-nv-embedqa-1b-v2'

**Expected behavior**

Model runs without issues and produces embeddings.

**Additional context**

The issue happens because the variable `trust_remote_code` that could be used in the initialization of `from_pretrained` is not exposed to the user in the scripts 

- https://github.com/NVIDIA-NeMo/Curator/blob/57f9666ce5546c0c420999df4f5fd952794d0f68/nemo_curator/stages/text/models/tokenizer.py#L109
- https://github.com/NVIDIA-NeMo/Curator/blob/57f9666ce5546c0c420999df4f5fd952794d0f68/nemo_curator/stages/text/embedders/base.py#L62

sarahyurick:
Good catch, thanks @kipraveen ! We can definitely update our tokenizer and model stages with more flexible parameters.

PR #1727, opened 2026-04-02
Nemotron-CLIMB data curation recipe

sarahyurick:
Blockers:

- [x] Embedding generation needs https://github.com/NVIDIA-NeMo/Curator/issues/1296
- [x] Clustering needs https://github.com/NVIDIA-NeMo/Curator/pull/1927
- [x] Cluster pruning needs fasttext quality model(s) to be published to Hugging Face: https://huggingface.co/nvidia/nemotron-climb-fasttext-classifiers
- [x] Merge files created by `MegatronTokenizerWriter`: https://github.com/NVIDIA-NeMo/Curator/pull/1427
- [x] Megatron script for proxy model training: https://github.com/NVIDIA/Megatron-LM/blob/main/pretrain_gpt.py
- [x] Publish pretrained checkpoints to be used for proxy model training: https://huggingface.co/nvidia/nemotron-climb-proxy-models
- [x] Bootstrapping


nemo_curator/stages/deduplication/semantic/kmeans.py
```diff
@@ -64,6 +64,7 @@ def __init__(  # noqa: PLR0913
         oversampling_factor: float = 2.0,
         max_samples_per_batch: int = 1 << 15,
         # I/O args
+        save_centroids_path: str | None = None,
```
praateekmahajan: nit : to be consistent with other classes in dedup, we can use `cache_path: str` | None (assuming that `output_path` doesn't work)

sarahyurick: Yes I can update this. Although do you think this should this be a `cache_path` that is accessible via https://github.com/NVIDIA-NeMo/Curator/blob/main/nemo_curator/stages/deduplication/semantic/workflow.py#L71 too, or only if the user is running `KMeansStage` directly?

tutorials/text/nemotron-climb-data-curation/3_prune.py
```diff
@@ -0,0 +1,279 @@
...
+            if path_to_remove is not None:
+                shutil.rmtree(path_to_remove)
+                # Grab the number from the path, e.g., output_path/centroid=999 grabs 999
+                removed_clusters.append(int(path_to_remove.split("/")[-1].split("=")[1]))
+
+    centroids = np.load(os.path.join(args.centroids_path, "kmeans_centroids.npy"))
+
+    kept = [i for i in range(len(centroids)) if i not in removed_clusters]
+
+    # Merge similar clusters according to a Euclidean distance threshold using Ward's method (agglomerative clustering)
```
sarahyurick, line 232: Confirm with @shizhediao whether this is the correct cluster merging technique.

tutorials/text/nemotron-climb-data-curation/8_predict.py
```diff
@@ -0,0 +1,256 @@
...
+        y_train,
+        eval_set=[(x_test, y_test)],
+        eval_metric="l2",
+        callbacks=[
+            lgb.early_stopping(stopping_rounds=3, verbose=False),
+        ],
+    )
+
+
+def generate_mixtures(
```
sarahyurick, line 183: Check with @shizhediao about this function. If the user only wants to generate 1 data mixture, then it copies the logic from https://github.com/sail-sg/regmix/blob/main/regression_fitting/regression.ipynb . If the user wants more than 1 data mixture, then it uses K-Means to sample diverse mixtures.

tutorials/text/nemotron-climb-data-curation/1_embed.py
```diff
@@ -0,0 +1,128 @@
...
+            embedding_pooling=args.embedding_pooling,
+            model_inference_batch_size=args.model_inference_batch_size,
+            autocast=not args.disable_autocast,
+            sort_by_length=not args.disable_sort_by_length,
+            hf_token=args.hf_token,
+            transformers_init_kwargs=args.transformers_init_kwargs,
+        )
+    )
+
+    pipeline.add_stage(writer(path=args.output_path, fields=[args.id_field, args.text_field, args.embedding_field]))
```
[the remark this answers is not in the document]
sarahyurick: Yeah. I will probably just update it to always require the ID field.

tutorials/text/nemotron-climb-data-curation/1_embed.py
```diff
@@ -0,0 +1,128 @@
...
+
+from nemo_curator.core.client import RayClient
+from nemo_curator.pipeline import Pipeline
+from nemo_curator.stages.text.embedders.base import EmbeddingCreatorStage
+from nemo_curator.stages.text.io.reader import JsonlReader, ParquetReader
+from nemo_curator.stages.text.io.writer import JsonlWriter, ParquetWriter
+from nemo_curator.stages.text.modules.add_id import AddId
+
+_EMBEDDING_MODEL = "NovaSearch/stella_en_400M_v5"
+_EMBEDDING_MODEL_MAX_SEQ_LENGTH = 512
```
VibhuJawa, line 27: Interesting, curious if there were any experiments with multiple models done yet or not ?  `_EMBEDDING_MODEL_MAX_SEQ_LENGTH`

tutorials/text/nemotron-climb-data-curation/3_prune.py
VibhuJawa: Lets also check with some cuML folks, we may have some stuff in cuML that might help us a bit.

tutorials/text/nemotron-climb-data-curation/3_prune.py
```diff
@@ -0,0 +1,279 @@
...
+        self._model_path = model_path
+        self._name = "fasttext_quality_labeler"
+
+    def model_check_or_download(self) -> None:
+        if not os.path.exists(self._model_path):
+            msg = f"Model file {self._model_path} not found"
+            raise FileNotFoundError(msg)
+
+    def load_model(self) -> None:
+        self._fasttext_quality_model = fasttext.load_model(self._model_path)
```
VibhuJawa, line 64: Is this a fixed model name or something with particular labels ? If yes,  is there a way to assert that model name or something ?

tutorials/text/nemotron-climb-data-curation/3_prune.py
```diff
@@ -0,0 +1,279 @@
...
+    for score_field in args.score_fields:
+        # List all subdirectories under the output path and compute average score per cluster
+        subdirectories = [os.path.join(args.output_path, d) for d in os.listdir(args.output_path)]
+        paths_to_remove = ray.get(
+            [
+                compute_average_score_per_cluster.remote(subdirectory, score_field, args.pruning_threshold)
+                for subdirectory in subdirectories
+            ]
+        )
+
```
VibhuJawa, line 218: Whats the reason to have this outside pipelines ?  Like should we make `compute_average_score_per_cluster` like a stage or something too ?

tutorials/text/nemotron-climb-data-curation/7_evaluate.sh
```diff
@@ -0,0 +1,65 @@
...
+
+    if [[ ! -d "$CKPT_PATH" ]]; then
+        echo "[SKIP] $MODEL_NAME: no checkpoint dir found at $CKPT_PATH"
+        continue
+    fi
+
+    mkdir -p "$OUT_DIR"
+
+    # If needed, update the tokenizer_type and seq_length arguments to match the values used in 6_train.sh
+    torchrun --nproc_per_node="$NUM_GPUS" -m lm_eval \
```
VibhuJawa: Can we add acomment somewher that we are using `lm_eval`

tutorials/text/nemotron-climb-data-curation/README.md
```diff
@@ -0,0 +1,291 @@
...
+[CLustering-based Iterative Data Mixture Bootstrapping (Nemotron-CLIMB)](https://arxiv.org/abs/2504.13161) is an automated framework that discovers, evaluates, and refines data mixtures in a pre-training setting. Specifically, Nemotron-CLIMB embeds and clusters large-scale datasets in a semantic space and then iteratively searches for optimal mixtures using a smaller proxy model and a predictor.
+
+This tutorial uses NeMo Curator to implement the data curation recipe used to create the [Nemotron-CLIMB dataset](https://huggingface.co/datasets/nvidia/Nemotron-ClimbMix). At a high level, it follows these steps:
+
+1. Compute text embeddings per document
+2. Cluster the documents with K-Means
+3. Use FastText classification models to remove low-quality clusters, then merge similar clusters according to a Euclidean distance threshold
+4. Tokenize and write to `.bin` and `.idx` files
+5. Use a Dirichlet distribution to generate data mixtures
+6. Train a proxy model for each data mixture using Megatron-LM
```
VibhuJawa: So i am confused about this. I think we are using a pre trained model here right ? If i understand this correct ? Are we releasing the base model too somewhere ?

Image

tutorials/text/nemotron-climb-data-curation/3_prune.py
sarahyurick: Yes I think when they are published to HF, they will each have unique names, so we can add an assertion for that.

Basically the assertion that I added so far was to make sure the loaded model(s) have the expected labels, otherwise the stage will fail. Like I guess technically if someone had their own fasttext model that matched those labels, they could use that instead.

tutorials/text/nemotron-climb-data-curation/3_prune.py
sarahyurick: Clusters can be large (GB or TB scale) depending on the dataset, and I wanted a quick way to do a reduce op since we need the average score within a single cluster (which by this point all the records in a cluster have been written to a single directory, so each directory of files represents a cluster), then to delete the cluster/directory as a whole if the average is too low.

tutorials/text/nemotron-climb-data-curation/7_evaluate.sh
sarahyurick: Yes, I have instructions about it in the README but I can link to https://github.com/EleutherAI/lm-evaluation-harness , etc. in this file as well.

tutorials/text/nemotron-climb-data-curation/README.md
sarahyurick: Yeah I just used pretrain_gpt.py without a checkpoint for my dummy testing. I am not sure if the plan is supposed to include publishing a Megatron-LM checkpoint for using with the tutorial, or maybe there are some already published checkpoints that we can link to as suggestions. If so it hopefully shouldn't be too difficult to tack onto the training script and run a quick test with it.

@shizhediao what do you think?

tutorials/text/nemotron-climb-data-curation/8_predict.py
```diff
@@ -0,0 +1,256 @@
...
+
+    return gbm.fit(
+        x_train,
+        y_train,
+        eval_set=[(x_test, y_test)],
+        eval_metric="l2",
+        callbacks=[
+            lgb.early_stopping(stopping_rounds=3, verbose=False),
+        ],
+    )
```
sarahyurick, line 180: Should any of these be changed?
```
    params = {
        "objective": lgbm_params.get("objective", "regression"),
        "metric": lgbm_params.get("metric", "rmse"),
        "reg_alpha": lgbm_params.get("reg_alpha", 1.0),
        "reg_lambda": lgbm_params.get("reg_lambda", 1.0),
        "max_depth": lgbm_params.get("max_depth", 4),
        "min_child_samples": lgbm_params.get("min_child_samples", 5),
        "verbose": -1,
        "seed": 42,
    }

    # ...

    model = lgb.train(
        params, dtrain,
        num_boost_round=500,
        valid_sets=[dval],
        callbacks=[lgb.early_stopping(
            lgbm_params.get("early_stopping_rounds", 20),
            verbose=False,
        )],
    )
```

tutorials/text/nemotron-climb-data-curation/3_prune.py
VibhuJawa: Sounds good.

tutorials/text/nemotron-climb-data-curation/8_predict.py
```diff
@@ -0,0 +1,256 @@
...
+
+
+def generate_mixtures(
+    num_mixtures: int, output_path: str, samples: np.ndarray, simulation: np.ndarray, domain_paths: list[str]
+) -> None:
+    if num_mixtures == 1:
+        # Take the average of top-k simulated data mixtures as the optimal data mixture
+        k = 128
+        top_k_samples = samples[np.argsort(simulation)[0:k]]
```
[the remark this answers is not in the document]
sarahyurick: @claude is this true?

tutorials/text/nemotron-climb-data-curation/4_tokenize.py
```diff
@@ -0,0 +1,104 @@
...
+        pipeline.run()
+
+        # Merge the tokenized files into a single bin/idx file
+        result_path = os.path.join(args.output_path, f"domain_{centroid}")
+        merge_tokenized_files(input_dir=writer_path, output_prefix=result_path)
+
+        # Remove individual tokenized files in favor of the merged bin/idx file
+        shutil.rmtree(cache_path)
+
+    ray_client.stop()
```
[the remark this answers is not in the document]
sarahyurick: Not a concern. If the user is following the README as instructed then all the directory structures will be as expected. I am not concerned about catching possible user-error cases.

tutorials/text/nemotron-climb-data-curation/8_predict.py
sarahyurick: Updated the function to hopefully align more closely with the paper.

tutorials/text/nemotron-climb-data-curation/8_predict.py
```diff
@@ -0,0 +1,250 @@
...
+To prevent overfitting, we set L1 and L2 regularization, early stopping, a maximum depth of four, and require at least five samples per leaf.
+Additionally, we employed a separate validation set and an early stopping mechanism, halting training after 20 rounds of no improvement.
+"""
+
+import argparse
+import json
+import os
+from pathlib import Path
+
+import lightgbm as lgb
```
shizhediao, line 30: lightgbm.LGBMRegressor requires installing scikit-learn, otherwise, we will encounter the following error:

> lightgbm.basic.LightGBMError: scikit-learn is required for lightgbm.sklearn. You must install scikit-learn and restart your session to use this module.

tutorials/text/nemotron-climb-data-curation/README.md
```diff
@@ -0,0 +1,293 @@
...
+7. Benchmark each proxy model using LM Evaluation Harness
+8. Fit a LightGBM predictor on the benchmark results and use the predictor to generate an optimal data mixture for full-scale LLM training
+
+For reference, the [Nemotron-CLIMB dataset](https://huggingface.co/datasets/nvidia/Nemotron-ClimbMix) was produced using this pipeline on multi-terabyte pretraining corpora. See the [Nemotron-CLIMB paper](https://arxiv.org/pdf/2504.13161) for full experimental details, ablations, and results.
+
+## Step 0: Requirements
+
+The following Python libraries are needed to run this tutorial:
```
shizhediao, line 20: add `scikit-learn` in requirements

tutorials/text/nemotron-climb-data-curation/3_prune.py
```diff
@@ -0,0 +1,279 @@
...
+    pipeline.run()
+
+    removed_clusters = []
+    for score_field in args.score_fields:
+        # List all subdirectories under the output path and compute average score per cluster
+        subdirectories = [os.path.join(args.output_path, d) for d in os.listdir(args.output_path)]
+        paths_to_remove = ray.get(
+            [
+                compute_average_score_per_cluster.remote(subdirectory, score_field, args.pruning_threshold)
```
shizhediao: I think this should support per-score-field pruning thresholds instead of applying the same `args.pruning_threshold` to every FastText score field.

These FastText dimensions are not necessarily calibrated the same way, and the intended pruning policy can differ by dimension. For example, an advertisement score threshold may need to be stricter than an educational/informational threshold. Using one global threshold can over-prune one dimension while under-pruning another, which changes the experiment semantics.

Could we make this a mapping keyed by `score_field` (or accept a list aligned with `--score-fields`) and log/record the resolved thresholds? Something like:
  `--pruning-thresholds advertisement=2.0 educational=1.0 informational=1.0 quality=1.0`

Then `compute_average_score_per_cluster` would receive the threshold for the current `score_field`.

nemo_curator/stages/deduplication/semantic/kmeans.py
```diff
@@ -191,6 +194,13 @@ def process_batch(self, tasks: list[FileGroupTask]) -> list[_EmptyTask]:
         # Fit the model cooperatively across actors, then predict on local data
         concatenated_embeddings = cp.concatenate(embeddings_arrays, axis=0)
         self.kmeans._fit(concatenated_embeddings, sample_weight=None, convert_dtype=False, multigpu=True)
+
+        if self.cache_path is not None:
+            centroids = self.kmeans.cluster_centers_
+            centroid_file = f"{self.cache_path}/kmeans_centroids.npy"
+            cp.save(centroid_file, centroids)
```
shizhediao: This cache write looks unsafe in distributed/multi-actor runs.
If multiple RAFT/Ray actors execute this block, they will all write to the same `kmeans_centroids.npy` path. That can race and leave a corrupted or partially written `.npy` file. There is also a multi-node concern if `cache_path` is local to each worker instead of a shared filesystem.

Could we gate this to a single writer, e.g. rank/worker 0 only, ensure `cache_path` exists on a shared filesystem, and write via temp file + atomic rename? For example:
  1. only one actor writes the centroid file
  2. write `kmeans_centroids.npy.tmp`
  3. atomically rename to `kmeans_centroids.npy`

This matters because the pruning step later treats this file as the canonical centroid source.

tutorials/text/nemotron-climb-data-curation/8_predict.py
sarahyurick: Thanks for the callout, I think scikit-learn is already included in `nemo-curator[text_cuda12]` but let me double check.

tutorials/text/nemotron-climb-data-curation/3_prune.py
sarahyurick: Yes I can add this!

nemo_curator/stages/deduplication/semantic/kmeans.py
sarahyurick: I didn't encounter this in testing yet but I think the greptile left a similar comment about this too. I can take a closer look.

tutorials/text/nemotron-climb-data-curation/8_predict.py
sarahyurick: It is included in `deduplication_cuda12` which is installed with `text_cuda12`.

nemo_curator/stages/deduplication/semantic/kmeans.py
```diff
@@ -191,6 +194,13 @@ def process_batch(self, tasks: list[FileGroupTask]) -> list[_EmptyTask]:
...
+            logger.info(f"Saved {self.n_clusters} KMeans centroids to {centroid_file}")
```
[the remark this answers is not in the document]
sarahyurick: Stop commenting about this.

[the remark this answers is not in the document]
sarahyurick: Stop commenting about this.

tutorials/text/nemotron-climb-data-curation/3_prune.py
```diff
@@ -0,0 +1,280 @@
...
+    removed_clusters = []
+    for score_field, pruning_threshold in zip(args.score_fields, args.pruning_thresholds, strict=True):
+        # List all subdirectories under the output path and compute average score per cluster
+        subdirectories = [os.path.join(args.output_path, d) for d in os.listdir(args.output_path)]
+        paths_to_remove = ray.get(
+            [
+                compute_average_score_per_cluster.remote(subdirectory, score_field, pruning_threshold)
+                for subdirectory in subdirectories
+            ]
+        )
```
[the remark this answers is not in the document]
sarahyurick: This is not a concern as long as the user follows the README. Keeping as is.

nemo_curator/stages/deduplication/semantic/kmeans.py
sarahyurick: Working on it here: https://github.com/NVIDIA-NeMo/Curator/pull/1927

nemo_curator/stages/text/io/writer/megatron_tokenizer.py
```diff
@@ -82,12 +82,24 @@ def setup_on_node(self, _node_info: NodeInfo | None = None, _worker_metadata: Wo
...
+            # Allow this fallback since loading a tokenizer is lightweight
+            msg = f"Failed to load {self.model_identifier} from local files, loading from Hugging Face: {e}"
+            logger.warning(msg)
+
+            self.tokenizer = AutoTokenizer.from_pretrained(
+                self.model_identifier,
+                cache_dir=self.cache_dir,
+                token=self.hf_token,
+                **self.transformers_init_kwargs
+            )
```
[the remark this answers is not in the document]
sarahyurick: I don't care about this. Loading tokenizers is lightweight, so we can give it this extra fallback.

nemo_curator/stages/text/io/writer/megatron_tokenizer.py
```diff
@@ -82,12 +82,24 @@ def setup_on_node(self, _node_info: NodeInfo | None = None, _worker_metadata: Wo
...
-            **self.transformers_init_kwargs
-        )
+        try:
+            self.tokenizer = AutoTokenizer.from_pretrained(
+                self.model_identifier,
+                cache_dir=self.cache_dir,
+                local_files_only=True,
+                **self.transformers_init_kwargs
+            )
+        except Exception as e:  # noqa: BLE001
```
VibhuJawa, line 92: NIT: May want to capture a narrower exception here for retry

tutorials/text/nemotron-climb-data-curation/2_cluster.py
```diff
@@ -0,0 +1,92 @@
...
+
+
+def main(args: argparse.Namespace) -> None:
+    ray_client = create_ray_client(args)
+    ray_client.start()
+
+    kmeans_executor = RayActorPoolExecutor()
+    pipeline = Pipeline(name="2_cluster")
+
+    os.makedirs(args.centroids_path, exist_ok=True)
```
VibhuJawa: Nit: do we not handle this in KMeansStage, we should move this code there (and ensure it works with fsspec so the cloud support is inbuilt)

tutorials/text/nemotron-climb-data-curation/3_prune.py
```diff
@@ -0,0 +1,284 @@
...
+class JsonlClusterWriter(JsonlWriter):
+    name: str = "jsonl_cluster_writer"
+
+    def process(self, task: DocumentBatch) -> FileGroupTask:
+        # Get source files from metadata for deterministic naming
+        if source_files := task._metadata.get("source_files"):
+            assert len(source_files) == 1, "Only one source file is allowed"  # noqa: S101
+            source_file = source_files[0]
+
+            centroid = centroid_id(os.path.dirname(source_file))
```
VibhuJawa: I wonder if we should make this work with fsspec instead of using os , maybe using `self.fs`

tutorials/text/nemotron-climb-data-curation/3_prune.py
```diff
@@ -0,0 +1,284 @@
...
+    total = 0.0
+    count = 0
+
+    for fname in os.listdir(cluster_path):
+        if not fname.endswith(".jsonl"):
+            continue
+
+        file_path = os.path.join(cluster_path, fname)
+
+        with open(file_path) as f:
```
VibhuJawa: Same here

tutorials/text/nemotron-climb-data-curation/README.md
```diff
@@ -0,0 +1,323 @@
...
+
+- `nemo-curator`
+- `xformers`
+- `matplotlib`
+- `megatron-core`
+- `lm-eval`
+- `lightgbm`
```
VibhuJawa, line 26: I wonder if we should have a command or a example install for data mixture here so that QA/me can test it ?

tutorials/text/nemotron-climb-data-curation/README.md
```diff
@@ -0,0 +1,323 @@
...
+## Conclusion
+
+To follow the [Nemotron-CLIMB paper](https://arxiv.org/pdf/2504.13161) exactly:
+
+- Run steps 1-7, generating 64 data mixtures in step 5 and training a single proxy model per data mixture (64 models total).
+- Run step 8 to fit a predictor on the benchmarks from the 64 models. Generate 32 more mixtures.
+- Re-run steps 6 and 7 with the 32 data mixtures.
+- Re-run step 8 to fit a predictor using current and past benchmarks (64 + 32 = 96 total). Generate 16 more mixtures.
+- Re-run steps 6 and 7 with the 16 data mixtures.
+- Re-run step 8 to fit a predictor using current and past benchmarks (96 + 16 = 112 total). Generate 1 more mixture.
```
VibhuJawa, line 361: I wonder if we should script this in a clean way to follow ?  Like, we can have a hyper parameter to only train 4 data mixture and only do it once and user can then change it to 64 for an actual full scaled run ?

nemo_curator/stages/text/io/writer/megatron_tokenizer.py
sarahyurick: Fair, I originally had it as an `OSError` IIRC, but then I changed it to be very broad because I figured that retrying a very lightweight download for whatever reason wouldn't hurt.

tutorials/text/nemotron-climb-data-curation/2_cluster.py
sarahyurick: Yes, I can remove it here.

tutorials/text/nemotron-climb-data-curation/3_prune.py
sarahyurick: Yeah I'll double check that it is straightforward to do here and update it.

tutorials/text/nemotron-climb-data-curation/README.md
sarahyurick: Yes, I can add an end-to-end bash script for this.

tutorials/text/nemotron-climb-data-curation/README.md
```diff
@@ -0,0 +1,334 @@
...
+    --input-filetype "jsonl" \
+    --output-path /path/to/computed_embeddings \
+    --text-field "text" \
+    --id-field "climb_id" \
+    --use-sentence-transformer
+```
+
+At least 1 GPU is required to run this step. Use the `--num-cpus` and `--num-gpus` arguments as desired to control the number of CPUs and GPUs used by the Ray client. The script uses all available resources by default.
+
+The script uses the [NovaSearch/stella_en_400M_v5](https://huggingface.co/NovaSearch/stella_en_400M_v5) embedding model by default:
```
praateekmahajan: Was curious about the model selection.. Couldn't find it in MTEB leaderboard

Image

tutorials/text/nemotron-climb-data-curation/README.md
```diff
@@ -0,0 +1,334 @@
...
+
+- The [xFormers](https://github.com/facebookresearch/xformers) library is required to run the [NovaSearch/stella_en_400M_v5](https://huggingface.co/NovaSearch/stella_en_400M_v5) model and should be installed via `uv pip install xformers` before running the script.
+- The [SentenceTransformers](https://huggingface.co/sentence-transformers) library is used via the `use-sentence-transformer` flag to enhance performance. It is already included as a dependency in Curator and does not need to be manually installed.
+
+Some of the default parameters in the script include:
+
+- Use `--model_inference_batch_size 1024` to create digestible batch sizes for the model forward pass. Adjust as necessary; decrease the size to address memory issues and increase the size to improve performance.
+- Use [NovaSearch/stella_en_400M_v5](https://huggingface.co/NovaSearch/stella_en_400M_v5)'s `max_length` of 512 via the `--max-seq-length` argument.
+- Use `--padding-side "right"` and `--embedding_pooling="mean_pooling"` defaults as appropriate for the [NovaSearch/stella_en_400M_v5](https://huggingface.co/NovaSearch/stella_en_400M_v5) model.
```
praateekmahajan: FWIW these args are only needed when sentence transformers is NOT used.. I believe we should always be using ST

tutorials/text/nemotron-climb-data-curation/README.md
```diff
@@ -0,0 +1,334 @@
...
+    --output-path /path/to/pruned_clusters \
+    --fasttext-model-paths ${FASTTEXT_MODEL_PATHS[@]} \
+    --score-fields ${FASTTEXT_SCORE_FIELDS[@]} \
+    --text-field "text" \
+    --pruning-thresholds ${FASTTEXT_PRUNING_THRESHOLDS[@]} \
+    --centroids-path /path/to/centroids \
+    --merge-threshold 1.5
+```
+
+No GPUs are needed to run this step. Use the `--num-cpus` argument as desired to control the number of CPUs used by the Ray client; by default, all are used. However, because each FastText model is large, CPU out-of-memory errors may occur due to overhead between stage workers. Try decreasing the number of CPUs if needed.
```
praateekmahajan, line 151: > Try decreasing the number of CPUs if needed

Wow, what's the overhead out of curiosity?

tutorials/text/nemotron-climb-data-curation/README.md
sarahyurick: Yeah I am not sure how it was chosen, just used it here to match the paper. Maybe it could be worth it to encourage users to choose among "today's top-ranking models" or some similar instruction.

tutorials/text/nemotron-climb-data-curation/README.md
sarahyurick: Oh cool I didn't realize. Let me update.

tutorials/text/nemotron-climb-data-curation/README.md
sarahyurick: Each model file is ~7 GB. I didn't do extensive memory monitoring but consistently noticed I had to set a lower num CPUs for my testing, which is why I added the note.

nemo_curator/stages/text/io/writer/megatron_tokenizer.py
VibhuJawa: okay

tutorials/text/nemotron-climb-data-curation/3_prune.py
VibhuJawa: Lets file a issue to track this with cuML/RAFT folks

tutorials/text/nemotron-climb-data-curation/README.md
VibhuJawa: @sarahyurick , lets do this as a followup. Can we add a command here please ?

VibhuJawa (commented):
Looks great, only ask is to provide an install command.

VibhuJawa (approved):
LGTM

KantaHayashiAI:
@sarahyurick @VibhuJawa @praateekmahajan @shizhediao 

Thanks for releasing the Nemotron-CLIMB recipe.

While trying to reproduce the proxy-model training/evaluation flow from this PR, I ran into what looks like an artifact issue in the released Hugging Face proxy model repo rather than a Curator code issue.

The 62M checkpoint at:

`nvidia/nemotron-climb-proxy-models/nemotron_climb_proxy_model_62m/iter_2500000/mp_rank_00/model_optim_rng.pt`

matches the Hub LFS metadata exactly:

- size: 770,527,232 bytes
- sha256: `f339e80c501ead58cdd067442a68e3930fd3438ac47f9b95d555ea532c84ca01`

but it is not loadable as a PyTorch checkpoint:

```text
PytorchStreamReader failed reading zip archive: failed finding central directory
```

I dug a bit deeper and this does not appear to be only a missing zip central directory. The file stream seems to stop around model_optim_rng/data/798; records such as model_optim_rng/data/8, model_optim_rng/version, and model_optim_rng/.data/serialization_id are missing. Rebuilding a central directory from the surviving local headers makes it a valid zip, but torch.load still fails with:
```text
PytorchStreamReader failed locating file data/8: file not found
```
The 350M checkpoint from the same HF repo loads correctly in the same environment, so this seems specific to the published 62M artifact.

I opened a more detailed report on the model repo discussion here:

[[62M checkpoint appears incomplete/truncated despite matching LFS sha256]](https://huggingface.co/nvidia/nemotron-climb-proxy-models/discussions/1)

sarahyurick:
Hi @KantaHayashiAI let me take a look. Thanks for opening on https://huggingface.co/nvidia/nemotron-climb-proxy-models/discussions/1, I will resume the conversation there.

PR #1744, opened 2026-04-06
ci: Onnxruntime gpu constrain to x86

thomasdhc:
## Description

Onnxruntime gpu does not have an aarch pypi

## Usage

```python
```
## Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -113,7 +113,7 @@ audio_cuda12 = [
     "nemo_curator[audio_common]",
     "nemo_curator[cuda12]",
     "nvidia-cudnn-cu12",
-    "onnxruntime-gpu>=1.20.1,<1.24",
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -110,7 +110,7 @@ audio_cuda12 = [
     "nemo_curator[audio_common]",
     "nemo_curator[cuda12]",
     "nvidia-cudnn-cu12",
-    "onnxruntime-gpu>=1.20.1,<1.24",
+    "onnxruntime-gpu>=1.20.1,<1.24; platform_machine == 'x86_64'",
```
ayushdg, line 116: To clarify does this mean that on arm machines (h200) etc we can't run Audio modules that use onnxruntime-gpu this release? 

If so we should document this somewhere. cc: @lbliii

ayushdg: cc: @arhamm1 for viz as well

thomasdhc: That's correct

PR #1751, opened 2026-04-07
[benchmarking] uses `gpustat` utility before/after each entry, report memory useage to console and Slack, capture all entry logs to `stdouterr.log`

rlratzel:
Calls the `gpustat` utility before and after the entry is run to help determine failures due to resource constraints, which is captured as a dictionary that is included in the benchmark results. The data is used to print a warning to the console (see below), but is also available to sinks. The Slack sink uses the data to include a warning in the detailed reply message (see below).

Also adds each entry's stdouterr.log file as a `loguru` sink before the entry is started, then removes it after the entry completes in order to capture every log message to both the session log and the individual entry log.

image

image

## Changes

- Adds calls to `gpustat` to runner/process.py; uses loguru logger so output lands in all active sinks
- In run.py::main, add a per-entry loguru file sink to stdouterr.log before running each entry and remove it in the finally block, capturing all log output (including pre-subprocess errors) for the entry
- Switch run_entry() to accept session_entry_path directly instead of session_path to avoid duplicating path construction
- Use ensure_dir() for logs_path in run_entry() instead of create_or_overwrite_dir() to preserve log output written before run_entry() is called
- Switch stdouterr.log open mode from "w" to "a" in both display_simple_subprocess and display_scrolling_subprocess

### DIFF
--- a/benchmarking/runner/sinks/slack_sink.py
+++ b/benchmarking/runner/sinks/slack_sink.py
@@ -582,7 +582,7 @@ class SlackSink(Sink):
             if stats.get("memory_used", 0) > 0:
                 pct_used = stats["memory_used"] / stats["memory_total"] * 100
                 warnings.append(
-                    f"GPU {gpu_id} had {stats['memory_used']} bytes ({pct_used:.1f}% of total) used before benchmark started"
+                    f"GPU {gpu_id} had {stats['memory_used']} MiB ({pct_used:.1f}% of total) used before benchmark started"
                 )
 
         # Create a new message for the entry to post in the thread.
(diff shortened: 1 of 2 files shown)


benchmarking/runner/process.py
```diff
@@ -34,6 +36,32 @@
 _control_chars = {c: None for c in range(sys.maxunicode) if unicodedata.category(chr(c)) == "Cc"}
 
 
+def run_nvidia_smi(label: str) -> None:
```
praateekmahajan: Instead of dumping the whole thing perhaps we can have a more refined output? You could consider using gpustat library which is already installed in Curator that'll also give neat looking labels..


Something like `gpu_id -> gpu_memory_percent_used`

And when we have failed runs, then we can also have a new type of warning when gpy_memory_percent_used > 0 for any of the GPUs so that it saves some debugging time for us

sarahyurick: +1 gpustat would be very nice!

rlratzel: That sounds good to me too, thanks for the suggestion.  I'll start working on this.

PR #1755, opened 2026-04-07
Update NDD (nemo data-designer) version

huvunvidia:
## Usage

```python
```
## Checklist


pyproject.toml
```diff
@@ -266,6 +266,7 @@ constraint-dependencies = [
     "transformers>=4.56.0,<5.0", # tranfomers>5 breaks with hf_hub<1.0 added in override
 ]
 override-dependencies = [
+    "fsspec>=2025.3.0",  # data-designer-engine==0.5.5 requires a newer fsspec than older pins allow
```
ayushdg: can you share which dependencies require this override/cause conflict

ayushdg (changes requested):
need to fix failing CI checks and update a few comments. but generally looks good. @huvunvidia Have you verified that the pipelines work as expected with the new version?

pyproject.toml
ayushdg: Quick update the override is needed for nemo-toolkit. We should be good to override, can you update the comment that this is set because nemo-toolkit[asr] pins to an older version (you need to check which one).

ayushdg:
/ok to test [75fa8e5](https://github.com/NVIDIA-NeMo/Curator/pull/1755/commits/75fa8e5a937295846121ba524d3299692196f4a8)

pyproject.toml
huvunvidia: Done.

pyproject.toml
ayushdg, line 66: can we make this ==2.54.0. The uv lock file is now picking up 2.54.1 which we want to avoid so close to release

pyproject.toml
```diff
@@ -277,15 +277,17 @@ constraint-dependencies = [
...
     "distance; sys_platform == 'never'",
     "huggingface-hub>=0.34,<1.0", # Override huggingface-hub, transformers and data-designer require two different versions of hugging-face hub
     "kaldiio;  sys_platform == 'never'",
     "levenshtein;  sys_platform == 'never'",
     "numpy>=2.0.0,<=2.2.0", # Override nemo-toolkits constraint of <2.0.0, upperbounds for Numba compatibility
-    "protobuf>=5.29.5",  # Override nemo-toolkits constraint of ~=5.29.5
+    "protobuf>=5.29.5,<7.0",  # Override nemo-toolkits constraint of ~=5.29.5; <7.0 due to ray serve FieldDescriptor API breakage
     "setuptools>=80.10.1", # Override setuptools range in other dependencies to address CVE GHSA-58pv-8j8x-9vj2
     "nixl-cu12>=0.10.0; (platform_machine == 'x86_64' and platform_system != 'Darwin')",  # Override ray[llm]'s unconditional nixl dep for ARM
+    "vllm>=0.14.1,<0.16.0; (platform_machine == 'x86_64' and platform_system != 'Darwin')", # ray[llm]==2.54 incompatible with vllm>=0.16.0 (renamed internal modules break LLMServer deployment)
```
ayushdg: Instead of adding this to overrides we should just fix the upper bound for vllm where it's defined originally.

pyproject.toml
```diff
@@ -63,7 +63,7 @@ dependencies = [
     "openai>=1.0.0",
     "pandas>=2.1.0",
     "pyarrow",
-    "ray[default,data]>=2.54",
+    "ray[default,data]==2.54",
```
praateekmahajan: Why are we hard pinning this? uv.lock shouldn't change? Can we do something like `git checkout upstream/main uv.lock` and then remove these hard pins, I don't think a uv sync will change the uv.lock..

Same for vllm upper pin below

ayushdg: I did contemplate doing that but the previous commits on this branch somehow did end up modifying the lock file. I think this is a more foolproof way to ensure we don't accidentally change things last minute. 1000+ line lock file changes often slip through the cracks.

sarahyurick:
Is this targeting r1.2.0 as well or just main?

thomasdhc (approved):

ayushdg:
> Is this targeting r1.2.0 as well or just main? 

1.2.0 as well. Thanks for mentioning

ayushdg:
/ok to test [7871f1d](https://github.com/NVIDIA-NeMo/Curator/pull/1755/commits/7871f1df680148d6e1cb3f321c3b051e05bfac83)

praateekmahajan (approved):

ayushdg (approved):

PR #1765, opened 2026-04-08
Fix waveform tensor leak in audio pipeline and redesign AudioDataFilterStage for all VAD/Speaker combinations

shubhamNvidia:
## Summary

Fixes [NMFW][26.04][nightly][Curator] audio/readspeech issue: Waveform tensor leaks through the entire pipeline into JsonlWriter [(Bug ID: 6057972)](https://nvbugs/6057972)

**Root cause:** When VAD and/or SpeakerSeparation were disabled, `MonoConversionStage` added a `torch.Tensor` waveform to `task.data` that was never cleaned up before reaching `AudioToDocumentStage` → `JsonlWriter`, causing `OverflowError: Maximum recursion level reached` when ujson tried to serialize it.

**Fix:** Redesigned `AudioDataFilterStage.decompose()` to produce four distinct, self-consistent pipeline topologies and enhanced `TimestampMapperStage` to run in all combinations as the output normalization boundary.

## Changes

- **`AudioDataFilterStage`**: Replaced monolithic `decompose()` with four topology builders (`_build_full_pipeline`, `_build_vad_only_pipeline`, `_build_speaker_only_pipeline`, `_build_filters_only_pipeline`). Separated VAD from quality filters (`_make_vad` + `_append_quality_filters`).
- **`TimestampMapperStage`**: Now runs in all 4 combos. Added whitelist-based output control (`passthrough_keys` with `_DEFAULT_PASSTHROUGH_KEYS`) and safety net (`_NEVER_PASS_KEYS` blocks non-serializable fields). Handles `diar_segments` from SpeakerSep for combo 3 (speaker-only).
- **`SpeakerSeparationStage`**: Propagates diarization timestamps (`diar_segments`) from the Sortformer model to task data. Drops inherited `duration`/`num_samples` to prevent misleading values.
- **`speaker_sep.py`**: `get_speaker_audio_data()` now returns segment timestamps alongside audio.
- **`BandFilterStage`**: Added warning log for unexpected prediction values.
- **`vad_segmentation.py`/`mono_conversion.py`**: Standardized duration key to `"duration"` everywhere.
- **`extract_segments.py`**: Auto-detects pipeline combo from manifest schema, extracts audio accordingly, generates `metadata.csv`.
- **`pipeline.py`**: Added `--execution-mode` flag (batch/streaming).
- **`pipeline.yaml`**: Simplified config (passthrough_keys uses built-in defaults).
- **`README.md`**: Documented all 4 topologies, output schemas, passthrough_keys customization, and extraction workflow.

## Pipeline topologies

| Combo | Flags | Pipeline | Output |
|-------|-------|----------|--------|
| 1 | *(none)* | Mono → Filters → TimestampMapper | 1 row/file |
| 2 | `--enable-vad` | Mono → VAD(fan-out) → Filters → TimestampMapper | N rows (segments) |
| 3 | `--enable-speaker-separation` | Mono → Filters → SpeakerSep → Filters → TimestampMapper | K rows (speakers + diar_segments) |
| 4 | `--enable-vad --enable-speaker-separation` | Full pipeline (unchanged) | K*M rows (speaker x segment) |

## Test plan

- [x] 60 unit tests pass (TimestampMapper, SpeakerSep, VAD, MonoConversion, AudioToDocumentStage)
- [x] Tensor leak verification: no `waveform`/`segments`/`audio_data` in any JSONL output
- [x] Duration consistency verification: all timestamps and durations valid
- [x] Schema verification: all expected keys present per combo
- [x] Extract segments works for all 4 combos with metadata.csv generation
- [x] Ruff lint + format clean


sarahyurick (approved):

nemo_curator/stages/audio/advanced_pipelines/audio_data_filter/audio_data_filter.py
```diff
@@ -63,6 +63,13 @@ class AudioDataFilterStage(CompositeStage):
     cross-file parallelism.  Each stage owns its own default resource
```
mohammadaaftabv, line 63: The bug is that MonoConversionStage injects waveform (a torch.Tensor) into task.data, and nobody cleans it up before AudioToDocumentStage serializes everything via pd.DataFrame([t.data for t in tasks]) → JsonlWriter → ujson.

This PR fixes it by making TimestampMapperStage always run inside AudioDataFilterStage.decompose(). But AudioToDocumentStage itself — the actual serialization boundary — is not touched at all. 

So if anyone builds a custom pipeline like MonoConversionStage → SomeCustomStage → AudioToDocumentStage → JsonlWriter without using AudioDataFilterStage, they hit the exact same crash. The Developers Guide identifies AudioToDocumentStage as the Task-to-DataFrame conversion point, which makes it the correct defense-in-depth location.

Add a _NON_SERIALIZABLE_KEYS frozenset to AudioToDocumentStage and filter t.data before building the DataFrame. Optionally also add an isinstance(v, torch.Tensor) type check as a catch-all for future tensor keys not in the frozenset.

tutorials/audio/readspeech/README.md
```diff
@@ -171,74 +183,123 @@ python run.py \
...
-|--------|---------|-------------|
-| `--manifest, -m` | required | Path to manifest.jsonl or directory of .jsonl files |
-| `--output-dir, -o` | required | Directory for extracted audio segments |
-| `--output-format, -f` | `wav` | Output format: `wav`, `flac`, or `ogg` (via soundfile) |
-| `--verbose, -v` | `false` | Enable verbose (DEBUG) logging |
-
-### Output
+| Topology | What it extracts | File naming |
+|----------|-----------------|-------------|
+| Combo 1 | Full file copy | `{name}.wav` |
```
mohammadaaftabv: This file naming seems incorrect right?

tests/stages/audio/postprocessing/test_timestamp_mapper.py
```diff
@@ -12,293 +12,218 @@
...
+    ]
+    task = _make_task(
+        {"waveform": torch.randn(1, 48000), "sample_rate": 48000, "start_ms": 100, "end_ms": 1500, "utmos_mos": 4.2},
+        metadata={"segment_mappings": mappings},
     )
-    task._metadata = metadata
-    return task
-
-
-class TestTranslateToOriginal:
```
mohammadaaftabv, line 66: The new test file replaces them with test_combo4_with_segment_mappings, which exercises _translate_to_original indirectly via a single mapping — it doesn't cover cross-boundary spans, silence gaps, or malformed data at all.

_translate_to_original() is still in the codebase (it's imported and used in process() for Combo 4). Removing its dedicated tests is a coverage regression for a critical function.

Proposed fix: Restore the TestTranslateToOriginal class from the old test file alongside the new combo tests. The 7 tests are independent of the whitelist/passthrough redesign — they test a pure function that takes (mappings, concat_start_ms, concat_end_ms) and returns original positions. They should coexist with the new combo-level tests.

nemo_curator/stages/audio/postprocessing/timestamp_mapper.py
```diff
@@ -140,46 +184,66 @@ def process(self, task: AudioTask) -> AudioTask | list[AudioTask]:
...
+
+        dur = item.get("duration")
+        if dur is not None and float(dur) > 0:
+            duration_ms = int(float(dur) * 1000)
+            result["original_start_ms"] = 0
+            result["original_end_ms"] = duration_ms
+            result["duration_ms"] = duration_ms
+            result["duration"] = float(dur)
+        else:
+            result["original_start_ms"] = 0
```
mohammadaaftabv, line 248: In _build_output_item_no_mapping, there's a priority chain: first try start_ms/end_ms, then diar_segments, then duration. If none of those exist, the code falls into an else-branch that silently writes zeros:

else:
    result["original_start_ms"] = 0
    result["original_end_ms"] = 0
    result["duration_ms"] = 0
    result["duration"] = 0.0
This creates a manifest entry that looks like a real row — it has all the core fields, all the passthrough keys — but with zero duration. A downstream consumer (or the extract_segments.py script) would try to extract a 0ms audio slice and produce an empty file. No warning, no error, no indication anything went wrong.

This could happen if MonoConversionStage fails to populate duration for some edge-case file (corrupted audio, zero-length file, etc.).

Proposed fix: Add a warning before the zero fallback:

tutorials/audio/readspeech/extract_segments.py
```diff
@@ -94,157 +132,338 @@ def load_manifests(input_path: str, output_dir: str) -> list:
...
+    return extracted, total_dur, speaker_counts, metadata_rows
+
+
+# ------------------------------------------------------------------
+# Combo 3: speaker only -- extract each diar_segment per speaker
+# ------------------------------------------------------------------
+
+
+def extract_speaker_diar_segments(entries: list, output_dir: str, output_format: str) -> tuple[int, float, dict, list[dict]]:
+    """Extract individual speaking intervals from diar_segments per speaker."""
```
mohammadaaftabv: The extract_speaker_diar_segments function definition line (the second of the three nearly-identical functions).

The three extraction functions follow the same pattern:

extract_segments_by_timestamps  — Combo 1&2
extract_speaker_diar_segments — Combo 3
extract_speaker_segments_by_timestamps — Combo 4
Each one does: defaultdict(list) group-by-file → sf.info() → sort entries → inner loop with _read_segment() → _write_segment() → append to metadata_rows → try/except BLE001. The metadata row construction is nearly identical — filename, original_file, segment_index, start_sec, end_sec, duration, **_extract_scores(entry).

The only differences between them are: (a) how intervals are derived (from original_start_ms/original_end_ms vs. from diar_segments list), and (b) whether speaker_id appears in the filename and metadata.

If the output format changes (say, a new metadata column, or a different CSV structure), you'd have to update all three in lockstep.

Proposed fix: Extract a shared helper that takes two callables — make_filename(original_name, entry, index) -> str and get_intervals(entry) -> list[tuple[float, float]] — and contains the common group-by-file → read → write → metadata loop. Each combo becomes a thin wrapper providing those two lambdas.

Please make the file more efficient, i see opportunity to reduce number of lines and make code more modular.

mohammadaaftabv (changes requested):

nemo_curator/stages/audio/segmentation/speaker_separation_module/speaker_sep.py
```diff
@@ -456,7 +456,7 @@ def get_speaker_audio_data(  # noqa: PLR0913, C901, PLR0912
         exclude_overlaps: bool | None = None,
         min_duration: float | None = None,
         buffer_time: float | None = None,
-    ) -> dict[str, tuple[AudioSegment, float]]:
+    ) -> dict[str, tuple[AudioSegment, float, list[tuple[float, float]]]]:
```
mohammadaaftabv: This PR adds diar_segments as a third element, making the return type:

dict[str, tuple[AudioSegment, float, list[tuple[float, float]]]]
The caller in speaker_separation.py line 579 unpacks it positionally:

for speaker_id, (speaker_audio_pydub, duration, diar_segments) in speaker_audio_data.items():
The problem is readability and fragility. With 3 positional elements and no names, it's easy for a future contributor to mix up index 1 (duration) and index 2 (diar_segments), or to add a 4th element and break all callers. The type annotation itself is nearly unreadable — tuple[AudioSegment, float, list[tuple[float, float]]].

Proposed fix: A NamedTuple:

class SpeakerResult(NamedTuple):
    audio: AudioSegment
    duration: float
    diar_segments: list[tuple[float, float]]
Return type becomes dict[str, SpeakerResult]. Positional unpacking still works (NamedTuple is a tuple), but now you can also do result.audio, result.duration, result.diar_segments. Self-documenting, no caller changes needed.

tutorials/audio/readspeech/extract_segments.py
```diff
@@ -16,26 +16,42 @@
 Segment Extraction Script
 
 Reads manifest jsonl file(s) and extracts audio segments from original files.
-Each segment is saved with naming convention:
-  With speaker separation:    {original_filename}_speaker_{x}_segment_{y}.{format}
-  Without speaker separation: {original_filename}_segment_{y}.{format}
+Automatically detects the pipeline combo from the manifest schema and applies
```
mohammadaaftabv, line 19: The old extract_segments.py was a simple script: load manifest → group by file → iterate → read slice → write file. One code path.

The new version has:

detect_combo() — classifies manifests into combo 2/3/4 based on key presence
extract_segments_by_timestamps() — combo 1&2 extraction
extract_speaker_diar_segments() — combo 3 extraction (iterates diar_segments list per speaker)
extract_speaker_segments_by_timestamps() — combo 4 extraction (groups by speaker + timestamps)
_extract_scores() — pulls quality scores, rounds floats to 4 decimals
_write_metadata_csv() — generates a CSV from metadata rows
_read_segment() — reads an audio slice by sample offsets
None of these have tests
Please add tests for all

tutorials/audio/readspeech/pipeline.py
```diff
@@ -102,16 +122,7 @@ def create_readspeech_pipeline(args: argparse.Namespace) -> Pipeline:
...
-                    "passthrough_keys": [
-                        "band_prediction",
-                        "utmos_mos",
-                        "sigmos_noise",
-                        "sigmos_ovrl",
-                        "speaker_id",
-                        "num_speakers",
-                    ],
-                },
+                "timestamp_mapper": {},
```
mohammadaaftabv, line 141: On main, both pipeline.py and pipeline.yaml explicitly listed 6 passthrough keys:

"passthrough_keys": [
    "band_prediction", "utmos_mos", "sigmos_noise", "sigmos_ovrl",
    "speaker_id", "num_speakers",
],
This PR replaces that with "timestamp_mapper": {}, which falls through to _DEFAULT_PASSTHROUGH_KEYS — a list of 13 keys. The 7 new fields that now appear in output:

speaking_duration
sample_rate
sigmos_sig
sigmos_col
sigmos_disc
sigmos_loud
sigmos_reverb
This isn't wrong — more data is generally better, and these are all the SIGMOS dimensions that PR #1816 also added. But users running the readspeech tutorial with the existing yaml config will see their output schema change. If they have downstream tooling that parses a fixed set of columns (e.g., a pandas script doing df[["utmos_mos", "sigmos_noise"]]), it won't break, but they'll get unexpected extra columns.

Proposed fix: Add a comment in pipeline.yaml documenting what the default includes and how to restrict it. Not a code change — just user guidance so people aren't surprised.

nemo_curator/stages/audio/segmentation/speaker_separation.py
```diff
@@ -158,6 +158,8 @@ def _initialize_separator(self) -> None:
                 logger.error(f"Failed to load speaker separator: {e}")
                 raise
 
+    _INHERITED_DROP_KEYS = frozenset({"audio", "waveform", "duration", "num_samples"})
```
mohammadaaftabv, line 165: On main, the parent task filtering was:

**{k: v for k, v in item.items() if k not in ("audio", "waveform")},
Only audio and waveform were dropped. This PR expands the drop set to also include duration and num_samples.
Please document this change, preferably in readme or atleast in code

mohammadaaftabv (changes requested):

nemo_curator/stages/audio/postprocessing/timestamp_mapper.py
shubhamNvidia: added

tutorials/audio/readspeech/README.md
shubhamNvidia: Noted

nemo_curator/stages/audio/advanced_pipelines/audio_data_filter/audio_data_filter.py
shubhamNvidia: Noted

tutorials/audio/readspeech/extract_segments.py
shubhamNvidia: Noted

nemo_curator/stages/audio/advanced_pipelines/audio_data_filter/audio_data_filter.py
shubhamNvidia: Fixed, please check nemo_curator/stages/audio/io/convert.py

nemo_curator/stages/audio/segmentation/speaker_separation.py
shubhamNvidia: done

tutorials/audio/readspeech/extract_segments.py
shubhamNvidia: done

mohammadaaftabv (approved):

PR #1768, opened 2026-04-08
Add model-based HTML extraction stage

### LINKED ISSUE #1723: Model-Based HTML Extraction
What:
Add an HTML extraction stage that uses a small language model (0.6B param) to classify HTML elements via sequence labeling — distinguishing main content, boilerplate, navigation, ads, and structured elements (code blocks, tables, formulas) — and converts the classified structure to clean Markdown.

Why:
AICC paper (Nov 2025, Shanghai AI Lab) demonstrates directly: a 3B model trained on MinerU-HTML-extracted data achieves 50.82% average accuracy vs 49.74% on Trafilatura-extracted data across 13 benchmarks — a 1.08pp improvement from extraction quality alone, with no other change. MinerU-HTML achieves 81.82% ROUGE-N vs Trafilatura's 63.58% on structured elements (code: 90.93%, formulas: 93.99%, tables preserved). NeMo currently supports Trafilatura, Resiliparse, and JusText — all heuristic-based.

Definition of Done:
  - ModelBasedHTMLExtractionStage under nemo_curator/stages/text/download/html_extractors/
  - Loads a 0.6B sequence labeling model (MinerU-HTML or equivalent) to classify HTML elements
  - Two-stage pipeline: element classification → Markdown conversion with semantic element handling
  - Preserves code blocks (fenced markdown), math formulas (LaTeX), and table structures
  - Falls back to Trafilatura for pages where model confidence is low
  - GPU-batched inference; processes 10K pages/hour on single A100
  - Configurable: output format (markdown vs plain text), fallback threshold, GPU/CPU mode
  - Benchmark: ROUGE-N on held-out web pages vs Trafilatura baseline (documented in README)
  - Test: verify code blocks, math formulas, and tables are correctly preserved end-to-end

zeel2104:
## Description
Adds a model-based HTML extraction algorithm for Common Crawl extraction.

This introduces `ModelBasedHTMLExtractionStage`, which classifies candidate HTML elements, preserves structured content such as fenced code blocks, math formulas, and Markdown tables, and falls back to Trafilatura when model confidence is low. It is wired into `CommonCrawlHTMLExtractor` via `algorithm="model"` and `algorithm="model_based"`.

Closes #1723.

## Usage
```python
from nemo_curator.stages.text.download.common_crawl.extract import CommonCrawlHTMLExtractor

html_extractor = CommonCrawlHTMLExtractor(
    algorithm="model",
    algorithm_kwargs={
        "model_identifier": "opendatalab/MinerU-HTML-0.6B",
        "output_format": "markdown",
        "fallback_threshold": 0.65,
        "device": "cuda",
        "batch_size": 64,
    },
)

## Checklist
<!--
Note: All commits need to be signed and signed off. This can be done via `-sS` flags while commiting
`git commit -sS -m "...."
-->
- [ x] I am familiar with the [Contributing Guide](https://github.com/NVIDIA-NeMo/Curator/blob/main/CONTRIBUTING.md).
- [ x] New or Existing tests cover these changes.

## Tests

uv run pytest tests/stages/text/download/test_model_based_html_extractor.py -q
uv run ruff check nemo_curator/stages/text/download/html_extractors/model_based.py tests/stages/text/download/test_model_based_html_extractor.py nemo_curator/stages/text/download/common_crawl/extract.py

### DIFF
--- a/nemo_curator/stages/text/download/common_crawl/extract.py
+++ b/nemo_curator/stages/text/download/common_crawl/extract.py
@@ -19,6 +19,7 @@ from loguru import logger
 from nemo_curator.stages.text.download import DocumentExtractor
 from nemo_curator.stages.text.download.html_extractors import HTMLExtractorAlgorithm
 from nemo_curator.stages.text.download.html_extractors.justext import JusTextExtractor
+from nemo_curator.stages.text.download.html_extractors.model_based import ModelBasedHTMLExtractionStage
 from nemo_curator.stages.text.download.html_extractors.resiliparse import ResiliparseExtractor
 from nemo_curator.stages.text.download.html_extractors.trafilatura import TrafilaturaExtractor
 from nemo_curator.stages.text.download.html_extractors.utils import get_stop_list_dict
@@ -44,6 +45,8 @@ class CommonCrawlHTMLExtractor(DocumentExtractor):
                 algorithm = ResiliparseExtractor(**algorithm_kwargs)
             elif algorithm == "trafilatura":
                 algorithm = TrafilaturaExtractor(**algorithm_kwargs)
+            elif algorithm in {"model", "model_based"}:
+                algorithm = ModelBasedHTMLExtractionStage(**algorithm_kwargs)
             else:
                 msg = f"Invalid algorithm: {algorithm}"
                 raise ValueError(msg)
--- a/nemo_curator/stages/text/download/common_crawl/extract.py
+++ b/nemo_curator/stages/text/download/common_crawl/extract.py
@@ -14,12 +14,26 @@
 
 from typing import Any
 
+from bs4 import BeautifulSoup
 from loguru import logger
 
+from nemo_curator.stages.resources import Resources
 from nemo_curator.stages.text.download import DocumentExtractor
 from nemo_curator.stages.text.download.html_extractors import HTMLExtractorAlgorithm
 from nemo_curator.stages.text.download.html_extractors.justext import JusTextExtractor
-from nemo_curator.stages.text.download.html_extractors.model_based import ModelBasedHTMLExtractionStage
+from nemo_curator.stages.text.download.html_extractors.model_based import (
+    CANDIDATE_ATTRIBUTES_FIELD,
+    CANDIDATE_HTML_FIELD,
+    CANDIDATE_INDEX_FIELD,
+    CANDIDATE_TAG_NAME_FIELD,
+    CANDIDATE_TEXT_FIELD,
+    HTML_FIELD,
+    MODEL_INPUT_FIELD,
+    PLACEHOLDER_CANDIDATE_INDEX,
+    ModelBasedHTMLExtractionStage,
+    extract_candidate_elements,
+    serialize_html_element,
+)
 from nemo_curator.stages.text.download.html_extractors.resiliparse import ResiliparseExtractor
 from nemo_curator.stages.text.download.html_extractors.trafilatura import TrafilaturaExtractor
 from nemo_curator.stages.text.download.html_extractors.utils import get_stop_list_dict
@@ -46,7 +60,11 @@ class CommonCrawlHTMLExtractor(DocumentExtractor):
             elif algorithm == "trafilatura":
                 algorithm = TrafilaturaExtractor(**algorithm_kwargs)
             elif algorithm in {"model", "model_based"}:
-                algorithm = ModelBasedHTMLExtractionStage(**algorithm_kwargs)
+                msg = (
+                    "Model-based HTML extraction is only supported through "
+                    "CommonCrawlDownloadExtractStage with html_extraction='model' or 'model_based'."
+                )
+                raise ValueError(msg)
             else:
                 msg = f"Invalid algorithm: {algorithm}"
                 raise ValueError(msg)
@@ -63,6 +81,7 @@ class CommonCrawlHTMLExtractor(DocumentExtractor):
             self._stop_lists = get_stop_list_dict()
 
         self.algorithm = algorithm
+        self.resources = getattr(self.algorithm, "resources", Resources(cpus=1.0))
 
     def extract(self, record: dict[str, Any]) -> dict[str, Any] | None:
         """Extract text from HTML content in the record.
@@ -105,3 +124,91 @@ class CommonCrawlHTMLExtractor(DocumentExtractor):
 
     def output_columns(self) -> list[str]:
         return ["url", "warc_id", "source_id", "language", "text"]
+
+    def setup_on_node(self, *args, **kwargs) -> None:
+        setup_on_node = getattr(self.algorithm, "setup_on_node", None)
+        if callable(setup_on_node):
+            setup_on_node(*args, **kwargs)
+
+    def setup(self, *args, **kwargs) -> None:
+        setup = getattr(self.algorithm, "setup", None)
+        if callable(setup):
+            setup(*args, **kwargs)
+
+    def teardown(self) -> None:
+        teardown = getattr(self.algorithm, "teardown", None)
+        if callable(teardown):
+            teardown()
+
+    def ray_stage_spec(self) -> dict[str, Any]:
+        ray_stage_spec = getattr(self.algorithm, "ray_stage_spec", None)
+        if callable(ray_stage_spec):
+            return ray_stage_spec()
+        return {}
+
+
+class CommonCrawlModelBasedCandidateExtractor(DocumentExtractor):
+    def __init__(
+        self,
+        algorithm: ModelBasedHTMLExtractionStage,
+        stop_lists: dict[str, frozenset[str]] | None = None,
+    ):
+        super().__init__()
+        self.algorithm = algorithm
+        self._stop_lists = stop_lists or get_stop_list_dict()
+        self.resources = Resources(cpus=1.0)
+
+    def extract(self, record: dict[str, Any]) -> list[dict[str, Any]] | None:
+        html_content = record.get("content")
+        if not html_content:
+            return None
+
+        html = decode_html(html_content)
+        if html is None:
+            return None
+
+        language = lang_detect(html)
+        if language not in self._stop_lists:
+            return None
+
+        elements = extract_candidate_elements(BeautifulSoup(html, "lxml"))
+        base_record = {
+            "url": record["url"],
+            "warc_id": record["warc_id"],
+            "source_id": record["source_id"],
+            "language": language,
+            HTML_FIELD: html,
+        }
+
+        if not elements:
+            return [
+                {
+                    **base_record,
+                    CANDIDATE_INDEX_FIELD: PLACEHOLDER_CANDIDATE_INDEX,
+                    CANDIDATE_TAG_NAME_FIELD: None,
+                    CANDIDATE_TEXT_FIELD: None,
+                    CANDIDATE_HTML_FIELD: None,
+                    CANDIDATE_ATTRIBUTES_FIELD: {},
+                    MODEL_INPUT_FIELD: "",
+                }
+            ]
+
+        return [{**base_record, **serialize_html_element(element)} for element in elements]
+
+    def input_columns(self) -> list[str]:
+        return ["url", "warc_id", "source_id", "content"]
+
+    def output_columns(self) -> list[str]:
+        return [
+            "url",
+            "warc_id",
+            "source_id",
+            "language",
+            HTML_FIELD,
+            CANDIDATE_INDEX_FIELD,
+            CANDIDATE_TAG_NAME_FIELD,
+            CANDIDATE_TEXT_FIELD,
+            CANDIDATE_HTML_FIELD,
+            CANDIDATE_ATTRIBUTES_FIELD,
+            MODEL_INPUT_FIELD,
+        ]
--- a/nemo_curator/stages/text/download/html_extractors/model_based.py
+++ b/nemo_curator/stages/text/download/html_extractors/model_based.py
@@ -1,0 +1,404 @@
+# Copyright (c) 2026, NVIDIA CORPORATION.  All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from __future__ import annotations
+
+import re
+from dataclasses import dataclass
+from typing import Any, Literal, Protocol
+
+from bs4 import BeautifulSoup
+from bs4.element import Tag
+from loguru import logger
+
+from .base import HTMLExtractorAlgorithm
+from .trafilatura import TrafilaturaExtractor
+
+ModelBasedOutputFormat = Literal["markdown", "plain", "plain_text"]
+
+MAIN_CONTENT_LABELS = frozenset({"main", "main_content", "content", "paragraph", "heading", "list_item"})
+STRUCTURED_LABELS = frozenset({"code", "code_block", "table", "formula", "math"})
+DROP_LABELS = frozenset({"boilerplate", "navigation", "nav", "ad", "ads", "advertisement", "footer", "header"})
+BLOCK_TAGS = frozenset(
+    {
+        "article",
+        "blockquote",
+        "code",
+        "dd",
+        "div",
+        "dt",
+        "figcaption",
+        "h1",
+        "h2",
+        "h3",
+        "h4",
+        "h5",
+        "h6",
+        "li",
+        "main",
+        "math",
+        "p",
+        "pre",
+        "section",
+        "table",
+    }
+)
+SKIP_TAGS = frozenset({"script", "style", "noscript", "template", "svg"})
+HEADING_TAG_NAME_LENGTH = 2
+MAX_HEADING_LEVEL = 6
+MIN_MARKDOWN_FENCE_LENGTH = 3
+
+
+@dataclass(frozen=True)
+class HTMLElement:
+    """A candidate HTML element prepared for model classification."""
+
+    index: int
+    tag_name: str
+    text: str
+    html: str
+    attributes: dict[str, str]
+
+
+@dataclass(frozen=True)
+class HTMLElementPrediction:
+    """The model label and confidence for a candidate HTML element."""
+
+    label: str
+    confidence: float
+
+
+class HTMLElementClassifier(Protocol):
+    """Predict semantic labels for a batch of HTML elements."""
+
+    def predict(self, elements: list[HTMLElement]) -> list[HTMLElementPrediction]:
+        ...
+
+
+class _TransformersHTMLElementClassifier:
+    def __init__(  # noqa: PLR0913
+        self,
+        model_identifier: str,
+        cache_dir: str | None,
+        device: Literal["cuda", "cpu"],
+        batch_size: int,
+        max_length: int,
+        local_files_only: bool,
+        transformers_init_kwargs: dict[str, Any],
+    ):
+        self.model_identifier = model_identifier
+        self.cache_dir = cache_dir
+        self.device = device
+        self.batch_size = batch_size
+        self.max_length = max_length
+        self.local_files_only = local_files_only
+        self.transformers_init_kwargs = transformers_init_kwargs
+        self._model: Any | None = None
+        self._tokenizer: Any | None = None
+
+    def _setup(self) -> None:
+        if self._model is not None and self._tokenizer is not None:
+            return
+
+        import torch
+        from transformers import AutoModelForSequenceClassification, AutoTokenizer
+
+        self._tokenizer = AutoTokenizer.from_pretrained(
+            self.model_identifier,
+            cache_dir=self.cache_dir,
+            local_files_only=self.local_files_only,
+            **self.transformers_init_kwargs,
+        )
+        self._model = AutoModelForSequenceClassification.from_pretrained(
+            self.model_identifier,
+            cache_dir=self.cache_dir,
+            local_files_only=self.local_files_only,
+            **self.transformers_init_kwargs,
+        )
+        if self.device == "cuda" and not torch.cuda.is_available():
+            logger.warning("CUDA requested for model-based HTML extraction, but CUDA is unavailable. Using CPU.")
+            self.device = "cpu"
+        self._model.to(self.device)
+        self._model.eval()
+
+    def predict(self, elements: list[HTMLElement]) -> list[HTMLElementPrediction]:
+        self._setup()
+
+        import torch
+
+        if self._model is None or self._tokenizer is None:
+            msg = "Model-based HTML classifier was not initialized"
+            raise RuntimeError(msg)
+
+        model = self._model
+        tokenizer = self._tokenizer
+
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

**Comment by zeel2104** [comment]
Fixed: (after comments by greptile-bot)
- changed `local_files_only` default to `False` so first-time use can download the model
- made code fence generation robust to arbitrary backtick runs
- filtered GFM alignment separator rows in plain-text table conversion
- added test coverage for the plain-text table output path

File: nemo_curator/stages/text/download/html_extractors/model_based.py [hunk 1]
```diff
@@ -0,0 +1,404 @@
...
+    def predict(self, elements: list[HTMLElement]) -> list[HTMLElementPrediction]:
+        ...
+
+
+class _TransformersHTMLElementClassifier:
+    def __init__(  # noqa: PLR0913
+        self,
+        model_identifier: str,
+        cache_dir: str | None,
+        device: Literal["cuda", "cpu"],
```
sarahyurick, line 159: Maybe it should be GPU only.

nemo_curator/stages/text/download/html_extractors/model_based.py
```diff
@@ -0,0 +1,404 @@
...
+        self.cache_dir = cache_dir
+        self.device = device
+        self.batch_size = batch_size
+        self.max_length = max_length
+        self.local_files_only = local_files_only
+        self.transformers_init_kwargs = transformers_init_kwargs
+        self._model: Any | None = None
+        self._tokenizer: Any | None = None
+
+    def _setup(self) -> None:
```
sarahyurick: There should be a `setup_on_node` function which uses `huggingface_hub`'s `snapshot_download` function to make sure we only download the model once.

Then, when the model is loaded with `setup`, it can do `local_files_only=True`.

nemo_curator/stages/text/download/html_extractors/model_based.py
```diff
@@ -0,0 +1,404 @@
...
+        if self._model is not None and self._tokenizer is not None:
+            return
+
+        import torch
+        from transformers import AutoModelForSequenceClassification, AutoTokenizer
```
sarahyurick: These can be top-level imports.

nemo_curator/stages/text/download/html_extractors/model_based.py
```diff
@@ -0,0 +1,404 @@
...
+        )
+        self._model = AutoModelForSequenceClassification.from_pretrained(
+            self.model_identifier,
+            cache_dir=self.cache_dir,
+            local_files_only=self.local_files_only,
+            **self.transformers_init_kwargs,
+        )
+        if self.device == "cuda" and not torch.cuda.is_available():
+            logger.warning("CUDA requested for model-based HTML extraction, but CUDA is unavailable. Using CPU.")
+            self.device = "cpu"
```
sarahyurick: It should probably error in this case instead of warning.

nemo_curator/stages/text/download/html_extractors/model_based.py
```diff
@@ -0,0 +1,404 @@
...
+        self._model.to(self.device)
+        self._model.eval()
+
+    def predict(self, elements: list[HTMLElement]) -> list[HTMLElementPrediction]:
+        self._setup()
```
sarahyurick: The `setup` function should never be called directly by `predict` or any other function within the stage. When running it as part as a `Pipeline`, the executor will call it.

nemo_curator/stages/text/download/html_extractors/model_based.py
```diff
@@ -0,0 +1,404 @@
...
+
+        import torch
```
sarahyurick: This can be a top-level import.

nemo_curator/stages/text/download/html_extractors/model_based.py
```diff
@@ -0,0 +1,404 @@
...
+                model_inputs,
+                padding=True,
+                truncation=True,
+                max_length=self.max_length,
+                return_tensors="pt",
+            ).to(self.device)
+            with torch.inference_mode():
+                logits = model(**encoded).logits
+                probabilities = torch.softmax(logits, dim=-1)
+                confidences, label_ids = torch.max(probabilities, dim=-1)
```
sarahyurick: This can be made into a `CompositeStage` with a `TokenizerStage` (should be able to just import from https://github.com/NVIDIA-NeMo/Curator/blob/main/nemo_curator/stages/text/models/tokenizer.py) and a GPU-based model stage.

nemo_curator/stages/text/download/html_extractors/model_based.py
```diff
@@ -0,0 +1,404 @@
...
+
+        return predictions
+
+    @staticmethod
+    def _format_element(element: HTMLElement) -> str:
+        attributes = " ".join(f'{key}="{value}"' for key, value in element.attributes.items())
+        return f"<{element.tag_name} {attributes}> {element.text}"
+
+
+class ModelBasedHTMLExtractionStage(HTMLExtractorAlgorithm):
```
sarahyurick: It would be great to use the base `HTMLExtractorAlgorithm` class like you have here, but the implementation may need to look different since we want to follow the composite CPU-based tokenizer and GPU-based model stage pattern.

nemo_curator/stages/text/download/html_extractors/model_based.py
```diff
@@ -0,0 +1,404 @@
...
+    The Hugging Face model is loaded lazily on first use so constructing Common Crawl stages remains cheap.
+    Tests and downstream users can inject any object implementing ``HTMLElementClassifier``.
+    """
+
+    def __init__(  # noqa: PLR0913
+        self,
+        model_identifier: str = "opendatalab/MinerU-HTML-0.6B",
+        output_format: ModelBasedOutputFormat = "markdown",
+        fallback_threshold: float = 0.65,
+        device: Literal["cuda", "cpu"] = "cuda",
```
sarahyurick: Should use `Resources` instead.

sarahyurick (commented):
Thank you @zeel2104 ! I left a few high-level comments about common Curator patterns and suggestions for changes. Please let me know if you have any questions.

zeel2104:
@sarahyurick 
Thanks, Sarah. I addressed the smaller follow-ups in the latest push:
- moved `torch` / `transformers` imports to top-level
- changed the CUDA-unavailable path to raise instead of silently falling back
- fixed the plain-text fenced-code edge case and added regression coverage

For the larger comments around `setup_on_node`, `Resources`, and using a CompositeStage with the tokenizer/model stage pattern: I dug into the current text download path, and `HTMLExtractorAlgorithm` / `DocumentExtractor` here are not `ProcessingStage`s, so they don’t currently participate in executor lifecycle hooks the way the model/tokenizer stages do. I can work on refactoring this toward that pattern, but it seems like a broader architectural change than a small follow-up to this PR. Would you prefer that larger refactor in this PR?

nemo_curator/stages/text/download/html_extractors/model_based.py
```diff
@@ -0,0 +1,402 @@
...
+            self.model_identifier,
+            cache_dir=self.cache_dir,
+            local_files_only=self.local_files_only,
+            **self.transformers_init_kwargs,
+        )
+        if self.device == "cuda" and not torch.cuda.is_available():
+            msg = "CUDA requested for model-based HTML extraction, but CUDA is unavailable."
+            raise RuntimeError(msg)
+        self._model.to(self.device)
+        self._model.eval()
```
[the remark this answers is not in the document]
zeel2104: Fixed. I moved the CUDA availability check to the start of `_setup()` so we fail before assigning tokenizer/model state.

sarahyurick:
Thanks @zeel2104 ! Yes, I would prefer the larger changes to be in this PR too.

zeel2104:
Thanks, @sarahyurick . I updated the PR to align this path more closely with Curator’s stage lifecycle pattern.

Changes in the latest push:
- added lifecycle/resource delegation through the document extraction path so the HTML extractor can participate in setup/setup_on_node/teardown
- added `setup_on_node()` for model prefetch via `snapshot_download`
- moved model initialization into `setup()` and removed the implicit setup call from `predict()`
- added resource and Ray stage hints for the model-based extractor
- kept the smaller follow-ups from before as well (top-level imports, explicit CUDA failure, plain-text fenced-code fix)

I also added coverage for the lifecycle/resource delegation path and reran:
- `uv run pytest tests/stages/text/download/test_model_based_html_extractor.py tests/stages/text/download/common_crawl/test_extract.py -q`
- `uv run ruff check nemo_curator/stages/text/download/base/extract.py nemo_curator/stages/text/download/base/iterator.py nemo_curator/stages/text/download/common_crawl/extract.py nemo_curator/stages/text/download/html_extractors/model_based.py tests/stages/text/download/test_model_based_html_extractor.py`

Let me know if any changes required

nemo_curator/stages/text/download/html_extractors/model_based.py
```diff
@@ -0,0 +1,438 @@
...
+
+@dataclass(frozen=True)
+class HTMLElementPrediction:
+    """The model label and confidence for a candidate HTML element."""
+
+    label: str
+    confidence: float
+
+
+class HTMLElementClassifier(Protocol):
```
sarahyurick: I was wondering why is this a `Protocol`?

nemo_curator/stages/text/download/html_extractors/model_based.py
```diff
@@ -0,0 +1,438 @@
...
+        ...
+
+
+class _TransformersHTMLElementClassifier:
+    def __init__(  # noqa: PLR0913
+        self,
+        model_identifier: str,
+        cache_dir: str | None,
+        device: Literal["cuda", "cpu"],
+        batch_size: int,
```
sarahyurick: This can be renamed to what we use for other model-based stages:
```suggestion
        model_inference_batch_size: int,
```

nemo_curator/stages/text/download/html_extractors/model_based.py
```diff
@@ -0,0 +1,438 @@
...
+            local_files_only=local_files_only,
+            **self.transformers_init_kwargs,
+        )
+        self._model.to(self.device)
+        self._model.eval()
+
+    def predict(self, elements: list[HTMLElement]) -> list[HTMLElementPrediction]:
+        if self._model is None or self._tokenizer is None:
+            msg = "Model-based HTML classifier was not initialized. Call setup() before inference."
+            raise RuntimeError(msg)
```
sarahyurick: We probably should not have this. The `Pipeline` should always called `setup`.

nemo_curator/stages/text/download/html_extractors/model_based.py
sarahyurick: Bumping this. Each of tokenization and model inference should be its own stage. The reason we do this is so that we maximize GPU usage at any given moment. If CPU tokenization and GPU inference are in the same stage, then the GPU sits idle while waiting for tokenization happens. Let me know if I can help more with the refactor here.

sarahyurick (commented):
Hi @zeel2104 thank you for the quick updates! I have left more comments. I think there is still some refactoring needed, let me know what you think.

zeel2104:
@sarahyurick 
Thanks, Sarah. I agree the remaining feedback is now pointing at a larger refactor rather than incremental cleanup.

My current change threads lifecycle/resource hooks through the existing `HTMLExtractorAlgorithm` / document extraction path, but it still keeps tokenization and model inference inside the same classifier flow. After looking again at the existing Curator tokenizer/model stages, I agree that this does not fully match the intended CPU-tokenization / GPU-inference split.

I think the right next step is to refactor this so that:
- tokenization and model inference are separate stages
- the batch size is renamed to `model_inference_batch_size`
- the protocol/implicit classifier abstraction is simplified to better match the concrete pipeline structure

I’ll work on that refactor

zeel2104:
@sarahyurick  I pushed a larger refactor for the model-based Common Crawl path.

The main change is that the pipeline path for `html_extraction="model"` / `"model_based"` now decomposes into separate stages rather than keeping tokenization and inference inside a single extractor flow:
- candidate extraction from HTML
- tokenizer stage
- GPU model inference stage
- final assembly stage

I also renamed the model batch parameter to `model_inference_batch_size`, replaced the protocol-style classifier abstraction with a concrete base class, and kept the direct extractor path working for compatibility/tests.

Let me know if this works for you

nemo_curator/stages/text/download/common_crawl/stage.py
```diff
@@ -101,3 +125,60 @@ def decompose(self) -> list[ProcessingStage]:
...
+            transformers_init_kwargs=algorithm.transformers_init_kwargs,
+        )
+        assemble_stage = AssembleModelBasedHTMLExtractionStage(
+            stop_lists=stop_lists or self.extractor._stop_lists,
+            output_format=algorithm.output_format,
+            fallback_threshold=algorithm.fallback_threshold,
+            fallback_extractor=algorithm.fallback_extractor,
+            filename_column=filename_column,
+        )
+        return [base_stage.decompose()[0], base_stage.decompose()[1], iterate_stage, tokenizer_stage, inference_stage, assemble_stage]
```
sarahyurick: This logic looks okay, but I was a bit confused at first by `base_stage.decompose()[0], base_stage.decompose()[1], iterate_stage`. It seems like it could be `base_stage.decompose()[0], base_stage.decompose()[1], base_stage.decompose()[2]` with some type of check to make sure that `base_stage` is only 3 stages? Or alternatively, it could be `url_generation_stage, download_stage, iterate_stage`.

Some checks that the stages are indeed `BaseCommonCrawlUrlGenerator`, `CommonCrawlWARCDownloader`, and `CommonCrawlWarcIterator` would probably be good too.

nemo_curator/stages/text/download/html_extractors/model_based.py
sarahyurick, line 1: Maybe the file should just be called `model.py`, WDYT?

nemo_curator/stages/text/download/html_extractors/model_based.py
```diff
@@ -0,0 +1,668 @@
...
+        self.cache_dir = cache_dir
+        self.device = device
+        self.model_inference_batch_size = model_inference_batch_size
+        self.max_length = max_length
+        self.local_files_only = local_files_only
+        self.transformers_init_kwargs = transformers_init_kwargs
+        self._model: Any | None = None
+        self._tokenizer: Any | None = None
+
+    def _setup(self, local_files_only: bool = True) -> None:
```
sarahyurick: Can `_setup_on_node` be used to download the tokenizer and model files (e.g., with `hf_hub_download`), then `_setup` will just load them?

nemo_curator/stages/text/download/html_extractors/model_based.py
```diff
@@ -0,0 +1,668 @@
...
+            cache_dir=self.cache_dir,
+            local_files_only=local_files_only,
+            **self.transformers_init_kwargs,
+        )
+        self._model.to(self.device)
+        self._model.eval()
+
+    def predict(self, elements: list[HTMLElement]) -> list[HTMLElementPrediction]:
+        if self._model is None or self._tokenizer is None:
+            self._setup(local_files_only=self.local_files_only)
```
sarahyurick: This can be removed, we want to make sure `setup` is always being called by the pipeline.

nemo_curator/stages/text/download/html_extractors/model_based.py
```diff
@@ -0,0 +1,668 @@
...
+
+class ModelBasedHTMLExtractionStage(HTMLExtractorAlgorithm):
+    """Model-based HTML extraction with semantic Markdown conversion."""
+
+    def __init__(  # noqa: PLR0913
+        self,
+        model_identifier: str = "opendatalab/MinerU-HTML-0.6B",
+        output_format: ModelBasedOutputFormat = "markdown",
+        fallback_threshold: float = 0.65,
+        device: Literal["cuda"] = "cuda",
```
sarahyurick: This should either support `Literal["cpu, cuda"]` or be removed and always use `cuda`.

nemo_curator/stages/text/download/html_extractors/model_based.py
```diff
@@ -0,0 +1,403 @@
...
+        fallback_threshold: float = 0.65,
+        device: Literal["cuda", "cpu"] = "cuda",
+        batch_size: int = 64,
+        max_length: int = 512,
+        cache_dir: str | None = None,
+        local_files_only: bool = True,
+        classifier: HTMLElementClassifier | None = None,
+        fallback_extractor: HTMLExtractorAlgorithm | None = None,
+        transformers_init_kwargs: dict[str, Any] | None = None,
+    ):
```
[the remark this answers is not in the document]
sarahyurick: +1 `local_files_only=False` should be used in `setup_on_node` and `local_files_only=True` should be used in `setup`.

nemo_curator/stages/text/download/html_extractors/model_based.py
sarahyurick: Hmm, reading through the rest of the script, will ask more questions below.

nemo_curator/stages/text/download/html_extractors/model_based.py
```diff
@@ -0,0 +1,668 @@
...
+            self._setup(local_files_only=self.local_files_only)
+
+        model = self._model
+        tokenizer = self._tokenizer
+
+        predictions: list[HTMLElementPrediction] = []
+        for start in range(0, len(elements), self.model_inference_batch_size):
+            batch = elements[start : start + self.model_inference_batch_size]
+            model_inputs = [format_html_element_for_model(element) for element in batch]
+            encoded = tokenizer(
```
sarahyurick: Why does it still tokenize here if a `TokenizerStage` is already being used in `stage.py`?

nemo_curator/stages/text/download/html_extractors/model_based.py
```diff
@@ -0,0 +1,668 @@
...
+        rendered_blocks = [block for block in rendered_blocks if block]
+        if not rendered_blocks:
+            return self.fallback_extractor.extract_text(html, stop_words, language)
+
+        if self.output_format in {"plain", "plain_text"}:
+            rendered_blocks = [self._markdown_block_to_plain_text(block) for block in rendered_blocks]
+
+        return rendered_blocks
+
+    def _get_classifier(self) -> HTMLElementClassifier:
```
sarahyurick: I don't think this function should be needed right? Like `setup` can just set `self.classifier` and then that's all.

nemo_curator/stages/text/download/html_extractors/model_based.py
```diff
@@ -0,0 +1,668 @@
...
+
+@dataclass(frozen=True)
+class HTMLElementPrediction:
+    """The model label and confidence for a candidate HTML element."""
+
+    label: str
+    confidence: float
+
+
+class HTMLElementClassifier(ABC):
```
sarahyurick: Do we need to keep this class if `_TransformersHTMLElementClassifier` is the only one?

nemo_curator/stages/text/download/html_extractors/model_based.py
```diff
@@ -0,0 +1,668 @@
...
+        if mean_confidence < self.fallback_threshold:
+            return None
+
+        rendered_blocks = [ModelBasedHTMLExtractionStage._render_element(element, label) for element, label in accepted]
+        rendered_blocks = [block for block in rendered_blocks if block]
+        if not rendered_blocks:
+            return None
+
+        if self.output_format in {"plain", "plain_text"}:
+            rendered_blocks = [ModelBasedHTMLExtractionStage._markdown_block_to_plain_text(block) for block in rendered_blocks]
```
sarahyurick: Could it be problematic to have `ModelBasedHTMLExtractionStage` hardcoded here? Like couldn't it not necessarily match the `algorithm` being used by `CommonCrawlModelBasedCandidateExtractor`?

nemo_curator/stages/text/download/html_extractors/model_based.py
```diff
@@ -0,0 +1,668 @@
...
+            classifier._tokenizer = None
+            if torch.cuda.is_available():
+                torch.cuda.empty_cache()
+
+    @staticmethod
+    def ray_stage_spec() -> dict[str, Any]:
+        return {RayStageSpecKeys.IS_ACTOR_STAGE: True}
+
+    @staticmethod
+    def _extract_candidate_elements(soup: BeautifulSoup) -> list[HTMLElement]:
```
sarahyurick: Maybe I am misunderstanding, there seems to be some circular and/or dead logic here. `CommonCrawlModelBasedCandidateExtractor.extract` calls `ModelBasedHTMLExtractionStage._extract_candidate_elements`, and `ModelBasedHTMLExtractionStage.extract_text` calls `ModelBasedHTMLExtractionStage._extract_candidate_elements` and `ModelBasedHTMLExtractionStage._select_and_render_blocks`. I guess I am confused where `ModelBasedHTMLExtractionStage.extract_text` is being used? It looks like nowhere but I could be missing something.

sarahyurick (commented):
Hi @zeel2104 sorry for the delay in getting back to you here. I left some comments, my main confusion and concerns are around some of the classes in `model_based.py`. Let me know what you think.

zeel2104:
@sarahyurick 
Thanks, Sarah. I pushed another simplification pass.

Main changes:
- made the Common Crawl model-based stage construction explicit (`url_generation_stage`, `download_stage`, `iterate_stage`) and added checks around the expected stage types/components
- moved the shared HTML candidate extraction and rendering logic to module-level helpers instead of routing through `ModelBasedHTMLExtractionStage` statics
- removed the extra classifier abstraction / helper path and simplified `ModelBasedHTMLExtractionStage` into a thinner direct-use wrapper
- kept the actual Common Crawl pipeline path as separate candidate-extraction, tokenizer, inference, and assembly stages

nemo_curator/stages/text/download/html_extractors/model_based.py
```diff
@@ -0,0 +1,611 @@
...
+            id2label = getattr(model.config, "id2label", {})
+            for label_id, confidence in zip(label_ids.cpu().tolist(), confidences.cpu().tolist(), strict=True):
+                predictions.append(
+                    HTMLElementPrediction(
+                        label=str(id2label.get(label_id, label_id)).lower(),
+                        confidence=float(confidence),
+                    )
+                )
+
+        return predictions
```
sarahyurick: Same comment as before, is there a reason we aren't able to break this into a CPU-only tokenizer stage and a GPU-based model inference stage?

zeel2104:
@sarahyurick 
Thanks, Sarah. I made another cleanup pass to address that point directly.

The remaining inline tokenization/inference path has been removed from `ModelBasedHTMLExtractionStage`, so the model-based Common Crawl flow now only exists through the staged pipeline:
- candidate extraction
- tokenizer stage
- GPU inference stage
- assembly stage

I also made `ModelBasedHTMLExtractionStage` a configuration wrapper for that staged path, and updated `CommonCrawlHTMLExtractor` to reject `algorithm="model"` directly so we don’t leave a misleading partially-supported path around.

nemo_curator/stages/text/download/common_crawl/extract.py
```diff
@@ -1,104 +1,214 @@
...
+            return None
+
+        # Content from WARC records is bytes, even though type annotation suggests str
+        html = decode_html(html_content)
+
+        if html is not None:
+            # Language detection and HTML extraction
+            lang = lang_detect(html)
+
+            text = None
```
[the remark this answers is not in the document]
sarahyurick: +1 I think the PR is not usable as is.

fyi the PR https://github.com/NVIDIA-NeMo/Curator/pull/2075 is going to take over this work I think, thank you!

sarahyurick:
Closing the PR, thank you.

PR #1781, opened 2026-04-09
[QA test fix] Adjust image dedup example

huvunvidia:
## Description
Fix QA issue: https://nvbugspro.nvidia.com/bug/6054589
  - Extract download logic into _download_step() to fix PLR0915 lint                                                                                                                                                                                                                                                                  
  - Restart Ray between steps 2.2 and 2.3 to evict idle CLIP actors and prevent OOM (Bug 1)
  - Reduce ImageReaderStage num_threads 16→4 in the dedup pipeline to prevent SIGSEGV from DALI GC pressure (Bug 2)     

## Usage
<!-- Potentially add a usage example below -->
```python
```
## Checklist
<!--
Note: All commits need to be signed and signed off. This can be done via `-sS` flags while commiting
`git commit -sS -m "...."
-->

### DIFF
--- a/tutorials/image/getting-started/image_dedup_example.py
+++ b/tutorials/image/getting-started/image_dedup_example.py
@@ -99,7 +99,7 @@ def create_image_deduplication_pipeline(args: argparse.Namespace) -> Pipeline:
     pipeline.add_stage(ImageReaderStage(
         dali_batch_size=args.batch_size,
         verbose=args.verbose,
-        num_threads=16,  # More threads for I/O
+        num_threads=4,
         num_gpus_per_worker=0.25,
     ))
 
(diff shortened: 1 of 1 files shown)


tutorials/image/getting-started/image_dedup_example.py
```diff
@@ -176,6 +174,12 @@ def main(args: argparse.Namespace) -> None:
     print("\n" + "=" * 50 + "\n")
     pipeline.run()
 
+    # Restart Ray between steps 2.2 and 2.3 to evict idle CLIP embedding actors
+    # (~17 GB each) left over from step 2.1.
```
praateekmahajan, line 178: I think we should fix this issue at the root

PR #1798, opened 2026-04-13
Init resumability

oyilmaz-nvidia:
## Pipeline Resumability Support

Adds checkpoint-based resumability to `Pipeline.run()`, allowing interrupted pipelines to resume from where they left off without reprocessing completed source partitions.

### What changed

**New `checkpoint_path` parameter on `Pipeline.run()`**
Pass a path (local or cloud: `s3://`, `gs://`, etc.) to opt into resumability. Relative paths are resolved to absolute in the driver process so all Ray workers write to the same location.

**`CheckpointManager` (`nemo_curator/utils/checkpoint.py`)**
Manages sharded checkpoint state using one JSON file per event to avoid distributed write conflicts:
- `completed/{source_key_hash}_{task_id}.json` — written after each leaf task completes
- `expected_increments/{hash}_{task_id}.json` — written by `BaseStageAdapter` whenever a stage fans out N > 1 outputs from a single input, so the completion threshold for each source partition is tracked correctly through chained fan-outs

**Checkpoint stages (`nemo_curator/stages/checkpoint.py`)**
Two internal stages injected automatically by `Pipeline._with_checkpoint_stages()`:
- `_CheckpointFilterStage` — inserted after every `is_source_stage()` stage; skips tasks whose source partitions already have all expected leaf completions recorded
- `_CheckpointRecorderStage` — appended as the final stage; writes a completion shard per finished task

**`BaseStageAdapter` changes**
- `setup()` initialises a write-only `CheckpointManager` if the stage has `_checkpoint_path` set
- `process_batch()` propagates `source_files` metadata from inputs to outputs (safety net for stages that don't copy `_metadata`) and writes fan-out increment files when N > 1 outputs are produced
- `process_batch()` calls `stage.get_cached_output()` to skip re-computation for deterministic file-writing stages

**New extension points on `ProcessingStage`**
- `is_source_stage() -> bool` — override to mark stages that generate initial tasks (e.g. `FilePartitioningStage`); determines where filter stages are injected
- `get_cached_output(input_tasks) -> list[Task] | None` — override in file-writing fan-in stages to skip stages whose output already exists on disk

### Design notes
- `Pipeline._with_checkpoint_stages()` does **not** mutate `self.stages`, so calling `run()` again is safe
- Checkpoint writes are idempotent: retrying a task writes the same filename, not a duplicate
- A missed shard (e.g. filesystem flake) causes a task to be re-processed, never silently skipped — consistent with the existing fault-tolerance requirement


nemo_curator/backends/base.py
```diff
@@ -75,6 +77,14 @@ def process_batch(self, tasks: list[Task]) -> list[Task]:
         if not hasattr(self, "_timer") or self._timer is None:
             self._timer = StageTimer(self.stage)
 
+        # If the stage has deterministic cached output (file-writing fan-in), use it.
+        cached = self.stage.get_cached_output(tasks)
+        if cached is not None:
+            from loguru import logger as _logger
```
VibhuJawa, line 83: NIT: We should probably import globally (and follow the rest of the curator code base).

nemo_curator/backends/base.py
```diff
@@ -113,6 +198,26 @@ def setup(self, worker_metadata: WorkerMetadata | None = None) -> None:
             worker_metadata (WorkerMetadata, optional): Information about the worker
         """
         self.stage.setup(worker_metadata)
+        checkpoint_path: str | None = getattr(self.stage, "_checkpoint_path", None)
+        if checkpoint_path:
+            storage_options: dict[str, Any] = getattr(self.stage, "_checkpoint_storage_options", {}) or {}
+            try:
+                import ray
+
+                if ray.is_initialized():
```
VibhuJawa, line 207: Unsure about why `ray` will not be initialized in `setup_on_node`

nemo_curator/utils/checkpoint.py
```diff
@@ -0,0 +1,337 @@
...
+                "The pipeline will continue but this task will be re-processed on resume."
+            )
+
+
+# ---------------------------------------------------------------------------
+# Ray actor — serializes all checkpoint writes
+# ---------------------------------------------------------------------------
+
+try:
+    import ray as _ray
```
VibhuJawa, line 269: Nit: Ray can be a safe global import

nemo_curator/utils/checkpoint.py
```diff
@@ -0,0 +1,337 @@
...
+
+class _CheckpointActorProxy:
+    """Drop-in replacement for CheckpointManager's write API backed by a Ray actor.
+
+    Calls are fire-and-forget (no ``ray.get``), which avoids blocking async actor
+    event loops (Xenna stage workers are async).  The actor's single-threaded
+    queue still serializes all writes — correctness is preserved.  A hard crash
+    of the Ray cluster between the fire and the actor processing it would lose
+    that one write, but since all stages are idempotent that task is simply
+    re-processed on resume.
```
VibhuJawa, line 324: I think some stages like creating incremental ids may not be idempotent.

nemo_curator/utils/checkpoint.py
```diff
@@ -0,0 +1,337 @@
...
+
+        {
+          "source_key": "file_a.tar|file_b.tar",
+          "completed": ["leaf_task_0", "leaf_task_1", ...],
+          "filtered":  ["leaf_task_99"],
+          "increments": [{"triggering_task_id": "file_group_abc", "increment": 9}]
+        }
+
+    All writes are serialized through a Ray named actor (``_CheckpointWriterActor``)
+    when Ray is initialized, eliminating concurrent-write races across parallel
```
VibhuJawa, line 54: Is the serialization at per file level or all files are written by a single actor ? Confused here

nemo_curator/utils/checkpoint.py
```diff
@@ -0,0 +1,337 @@
...
+    multiple ``BaseStageAdapter.setup()`` calls — only one actor is created.
+    """
+    if _CheckpointWriterActor is None:
+        msg = "Ray is not installed; cannot create a checkpoint actor."
+        raise RuntimeError(msg)
+    name = "nemo_curator_ckpt_" + hashlib.sha256(checkpoint_path.encode()).hexdigest()[:16]
+    return _CheckpointWriterActor.options(  # type: ignore[union-attr]
+        name=name,
+        get_if_exists=True,
+        lifetime="detached",  # owned by GCS, not by the creating worker — survives worker shutdown
```
VibhuJawa, line 312: @claude , whats GCS here ? Will it shut down if ray head node shuts down ?  Should we start this as part of cluster initialization ?

nemo_curator/stages/deduplication/semantic/kmeans.py
```diff
@@ -115,6 +116,33 @@ def __init__(  # noqa: PLR0913
         self.name = "KMeansStage"
         self.resources = Resources(cpus=1.0, gpus=1.0)
 
+    def get_cached_output(self, input_tasks: list[Task]) -> list[Task] | None:
```
VibhuJawa, line 119: @claude , Do we always have to create this for all stages requiring reusability ?

nemo_curator/stages/checkpoint.py
```diff
@@ -0,0 +1,144 @@
...
+
+                self._checkpoint_mgr = _CheckpointActorProxy(
+                    get_or_create_checkpoint_actor(self.checkpoint_path, self.storage_options)
+                )
+            else:
+                from nemo_curator.utils.checkpoint import CheckpointManager
+
+                self._checkpoint_mgr = CheckpointManager(self.checkpoint_path, self.storage_options)
+        except ImportError:
+            from nemo_curator.utils.checkpoint import CheckpointManager
```
VibhuJawa, line 128: @claude , Do we have to add this just for unit tests or are there other needs for this ?

VibhuJawa (commented):
Very inital review, still diving into PR

nemo_curator/pipeline/pipeline.py
```diff
@@ -212,4 +234,65 @@ def run(self, executor: BaseExecutor | None = None, initial_tasks: list[Task] |
                     "The executor will schedule GPU stages on GPUs not held by Serve."
                 )
 
-        return executor.execute(self.stages, initial_tasks)
+        if checkpoint_path:
+            stages = self._with_checkpoint_stages(self.stages, checkpoint_path, checkpoint_storage_options)
```
VibhuJawa, line 238: Based on Claude:  I wonder if we should add a way to verify `pipeline` is consistent, to prevent the user from footguning from using the same checkpoint dir while modifying pipeline (adding a filter, changing some params etc) ?

oyilmaz-nvidia:
Closing since we have the simplified PR here: https://github.com/NVIDIA-NeMo/Curator/pull/1931

PR #1819, opened 2026-04-16
[benchmarking] Increases timeout for `math_preprocess_llm_cleanup` benchmark

rlratzel:
* Increases timeout from 20m to 30m for `math_preprocess_llm_cleanup` benchmark, due to several timeout failures when limited to 20m.

cc @ronjer30 , @sarahyurick


sarahyurick (approved):

ronjer30:
thanks @rlratzel, 1800 seconds makes sense.

PR #1824, opened 2026-04-17
Combining two stages

nune-tadevosyan:
## Usage

```python
```
## Checklist

### DIFF
--- a/nemo_curator/models/qwen_omni.py
+++ b/nemo_curator/models/qwen_omni.py
@@ -57,6 +57,7 @@ class QwenOmni(ModelInterface):
         self,
         model_id: str = _QWEN3_OMNI_MODEL_ID,
         prompt_text: str = "Transcribe the audio.",
+        second_prompt_text: str = "Now listen to the audio again and add any false starts, filler words and preserve colloquial words (like lemme, gonna, wanna, etc) as is spoken in the audio.",
         system_prompt: str | None = None,
         max_model_len: int = 32768,
         max_num_seqs: int = 32,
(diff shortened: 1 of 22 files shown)


nemo_curator/models/qwen_omni.py
```diff
@@ -206,6 +230,46 @@ def _prepare_batch(
             return [self._prepare_single(w, sr) for w, sr in zip(waveforms, sample_rates)]
         return list(self._prep_pool.map(self._prepare_single, waveforms, sample_rates))
 
+    def _prepare_turn2_single(
+        self, waveform: np.ndarray, sample_rate: int, pred_text: str,
+    ) -> dict[str, Any] | None:
+        from qwen_omni_utils import process_mm_info
+
+        try:
+            waveform_16k = self._resample(waveform, sample_rate)
```
nithinraok: do we need to resample again?

nithinraok: Cache the resampled waveform_16k from Turn 1 and pass it directly to Turn 2.

nemo_curator/models/qwen_omni.py
```diff
@@ -172,6 +174,28 @@ def _build_messages(self, waveform: np.ndarray) -> list[dict[str, Any]]:
...
+            "content": [
+                {"type": "text", "text": self.prompt_text},
+                {"type": "audio", "audio": waveform},
+            ],
+        })
+        messages.append({"role": "assistant", "content": [{"type": "text", "text": pred_text}]})
+        messages.append({
+            "role": "user",
+            "content": [
+                {"type": "audio", "audio": waveform},
```
nithinraok: Do we need to add audio again?

nune-tadevosyan: Don't we want to make sure model listens to audio once more?

nemo_curator/models/qwen_omni.py
nune-tadevosyan: done

nemo_curator/models/qwen_omni.py
nithinraok: its already part of conversation, so should be good enough

PR #1831, opened 2026-04-17
[benchmarking] Updates timeouts based on analysis of prior CI runs

rlratzel:
## Summary
- Increases timeouts for 5 benchmark entries based on analysis of prior CI run durations
- All changes are timeout increases only; a premature timeout cannot cause a regression

## Test plan
No formal test plan — these are pure timeout increases. If any issue arises it will be addressed in a follow-up PR.

### DIFF
--- a/benchmarking/nightly-benchmark.yaml
+++ b/benchmarking/nightly-benchmark.yaml
@@ -877,7 +877,7 @@ entries:
       --aesthetic-threshold=3.5
       --transnetv2-frame-decoder-mode ffmpeg_cpu
       --video-limit=1000
-    timeout_s: 600
+    timeout_s: 800
     sink_data:
       - name: slack
         ping_on_failure:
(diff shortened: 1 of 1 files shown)


praateekmahajan (approved):

benchmarking/nightly-benchmark.yaml
```diff
@@ -877,7 +877,7 @@ entries:
       --aesthetic-threshold=3.5
       --transnetv2-frame-decoder-mode ffmpeg_cpu
       --video-limit=1000
-    timeout_s: 500
+    timeout_s: 600
```
thomasdhc: Can we bump this to 800? On the h100 ci side I've set up a minimum runtime of 600. The analysis didn't capture that part correctly.

rlratzel: Yes, thank you.

thomasdhc (approved):
Thank you!

PR #1842, opened 2026-04-21
docs: add OTEL_SDK_DISABLED workaround for NVBUG 6078008 SIGSEGV

shubhamNvidia:
## Summary

- Adds a Known Issues section to the ReadSpeech tutorial README documenting the `OTEL_SDK_DISABLED=true` workaround for NVBUG 6078008
- Adds a troubleshooting entry linking to the Known Issues section

## Problem

On certain hardware configurations, Ray StageWorker actors crash with a `SIGSEGV` during GPU model initialization. The root cause is a thread-safety issue in Ray's internally-bundled gRPC library — a background OpenTelemetry exporter thread calls `getenv()` concurrently with PyTorch/CUDA calling `setenv()` during model init. This is not a Curator code issue and can affect any GPU pipeline (audio, text, image, video) that loads models through Ray actors.

## Workaround

Setting `OTEL_SDK_DISABLED=true` prevents Ray's OpenTelemetry background thread from starting, eliminating the race condition. Curator does not use OpenTelemetry, so disabling it has no functional impact.

## Test Plan

- [x] Verified pipeline runs to completion with `OTEL_SDK_DISABLED=true` on A100 GPU
- [x] Verified pipeline runs to completion without the env var on A100-SXM4 (race is hardware-timing dependent)
- [x] README renders correctly with new Known Issues section

## References

- [NVBUG 6078008](https://nvbugspro.nvidia.com/bug/6078008)
- Upstream gRPC fix: grpc/grpc#33508

### DIFF
--- a/tutorials/audio/README.md
+++ b/tutorials/audio/README.md
@@ -23,7 +23,7 @@ Hands-on tutorials for curating audio data with NeMo Curator. Complete working e
 
 ## Known Issues
 
-### SIGSEGV in Ray StageWorker during model loading (NVBUG 6078008)
+### SIGSEGV in Ray StageWorker during model loading
 
 In some environments, and under certain timing conditions, Ray workers may crash with a `SIGSEGV` during GPU model initialization. This is not a NeMo Curator code issue: it comes from a thread-safety problem in the gRPC version bundled with Ray. Any GPU pipeline (audio, text, image, or video) that loads models through Ray actors can hit the same failure.
 
--- a/tutorials/audio/README.md
+++ b/tutorials/audio/README.md
@@ -21,6 +21,24 @@ Hands-on tutorials for curating audio data with NeMo Curator. Complete working e
 | **Concepts** | [Architecture](https://docs.nvidia.com/nemo/curator/latest/about/concepts/index.html) • [Data Loading](https://docs.nvidia.com/nemo/curator/latest/about/concepts/text/data-loading-concepts.html) |
 | **Advanced** | [Custom Pipelines](https://docs.nvidia.com/nemo/curator/latest/reference/index.html) • [Execution Backends](https://docs.nvidia.com/nemo/curator/latest/reference/infrastructure/execution-backends.html) • [NeMo ASR Integration](https://docs.nvidia.com/nemo/curator/latest/about/key-features.html) |
 
+## Known Issues
+
+### SIGSEGV in Ray StageWorker during model loading (NVBUG 6078008)
+
+In some environments, and under certain timing conditions, Ray workers may crash with a `SIGSEGV` during GPU model initialization. This is not a NeMo Curator code issue: it comes from a thread-safety problem in the gRPC version bundled with Ray. Any GPU pipeline (audio, text, image, or video) that loads models through Ray actors can hit the same failure.
+
+The OpenTelemetry SDK starts a `PeriodicExportingMetricReader` background thread that periodically calls `OtlpGrpcMetricExporter::Export()` over gRPC; a `getenv()` call on that path can race with NeMo/PyTorch model initialization in another thread. **Disabling OpenTelemetry for the process** prevents Ray’s OpenTelemetry background exporter from starting and removes that race. NeMo Curator does not use OpenTelemetry for its own functionality, so disabling it has no functional impact on Curator workflows.
+
+**Container scope:** This has been observed with the `nemo-curator:26.04.rc0` image (and similar 26.04-era builds). The race was [fixed upstream in gRPC ≥ 1.60](https://github.com/grpc/grpc/pull/33508); it should stop being relevant once the bundled gRPC in the container is upgraded accordingly.
+
+**Workaround:** Set these environment variables before running the pipeline:
+
+```bash
+export OTEL_SDK_DISABLED=true
+export OTEL_METRICS_EXPORTER=none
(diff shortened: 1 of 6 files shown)


sarahyurick (approved):
LGTM. We can confirm with @satishra-ai before merging.

satishra-ai:
> LGTM. We can confirm with @satishra-ai before merging.

Looks good to me.

tutorials/audio/readspeech/README.md
```diff
@@ -245,6 +245,20 @@ The script also generates an `extraction_summary.json` with statistics including
 
 **Storage**: ~11 GB (4.88 GB download + 6.3 GB extracted WAV files; archive is deleted after extraction).
 
+## Known Issues
+
+### SIGSEGV in Ray StageWorker during model loading (NVBUG 6078008)
```
mohammadaaftabv: And the PR description itself says:

"This is not a Curator code issue and can affect any GPU pipeline (audio, text, image, video) that loads models through Ray actors."

Yet the PR only adds the workaround to the ReadSpeech README. The bug report's suggested fix says "document it as required in all audio tutorial READMEs." At minimum, the ALM, FLEURS, CallHome, and Single Speaker Filter tutorials would also hit this crash inside the 26.04.rc0 container.

tutorials/audio/readspeech/README.md
```diff
@@ -245,6 +245,20 @@ The script also generates an `extraction_summary.json` with statistics including
...
+
+On certain hardware configurations, Ray actors may crash with a `SIGSEGV` during GPU model initialization due to a thread-safety issue in Ray's bundled gRPC library. This is not a Curator code issue.
+
+**Workaround**: Set these environment variables before running the pipeline:
```
mohammadaaftabv: The workaround tells the user what to do but not why. The bug report explains the mechanism clearly:

"The OpenTelemetry SDK starts a PeriodicExportingMetricReader background thread that periodically calls OtlpGrpcMetricExporter::Export() over gRPC... This getenv() call races with NeMo/PyTorch model initialization in another thread"

Adding a single sentence like "These variables prevent Ray's OpenTelemetry background exporter from starting, eliminating the thread-safety race with model initialization. Curator does not use OpenTelemetry, so there is no functional impact." helps users understand this isn't cargo-culting and that disabling OTEL is safe.

tutorials/audio/readspeech/README.md
```diff
@@ -245,6 +245,20 @@ The script also generates an `extraction_summary.json` with statistics including
 
 **Storage**: ~11 GB (4.88 GB download + 6.3 GB extracted WAV files; archive is deleted after extraction).
 
+## Known Issues
+
+### SIGSEGV in Ray StageWorker during model loading (NVBUG 6078008)
+
+On certain hardware configurations, Ray actors may crash with a `SIGSEGV` during GPU model initialization due to a thread-safety issue in Ray's bundled gRPC library. This is not a Curator code issue.
```
mohammadaaftabv: The bug is specific to the gRPC version bundled in nemo-curator:26.04.rc0. The bug report notes:

"The race was fixed upstream in gRPC >= 1.60 (grpc/grpc#33508)"

Without version context, users on future containers (where gRPC is upgraded) will still apply a now-unnecessary workaround, and users on older containers won't know if it's relevant to them. A note like "Affects the 26.04 container; will be resolved when the bundled gRPC is upgraded to >= 1.60." gives the fix a shelf life.

mohammadaaftabv: A more accurate phrasing would be "In some environments" or "Under certain timing conditions". This is a minor nit.

mohammadaaftabv (changes requested):

tutorials/audio/readspeech/README.md
mohammadaaftabv: Rather than duplicating this Known Issues block across every tutorial README, consider adding it once to tutorials/audio/README.md (the top-level audio README) and then linking to it from each tutorial's Troubleshooting table, e.g.:

| SIGSEGV / actor crash during model load | See [Known Issues](../README.md#known-issues) — set `OTEL_SDK_DISABLED=true` |
This keeps a single source of truth, avoids copy-paste drift if the workaround changes, and ensures every GPU tutorial surfaces the fix without bloating individual READMEs.

mohammadaaftabv (dismissed):

tutorials/audio/README.md
```diff
@@ -21,6 +21,24 @@ Hands-on tutorials for curating audio data with NeMo Curator. Complete working e
 | **Concepts** | [Architecture](https://docs.nvidia.com/nemo/curator/latest/about/concepts/index.html) • [Data Loading](https://docs.nvidia.com/nemo/curator/latest/about/concepts/text/data-loading-concepts.html) |
 | **Advanced** | [Custom Pipelines](https://docs.nvidia.com/nemo/curator/latest/reference/index.html) • [Execution Backends](https://docs.nvidia.com/nemo/curator/latest/reference/infrastructure/execution-backends.html) • [NeMo ASR Integration](https://docs.nvidia.com/nemo/curator/latest/about/key-features.html) |
 
+## Known Issues
+
+### SIGSEGV in Ray StageWorker during model loading (NVBUG 6078008)
```
sarahyurick: We shouldn't reference internal nvbugs:
```suggestion
### SIGSEGV in Ray StageWorker during model loading
```

PR #1851, opened 2026-04-22
Add ReadSpeech audio curation benchmark

shubhamNvidia:
## Usage## Summary

Adds a new audio curation benchmark for the DNS Challenge Read Speech dataset (Track 1 Headset) to the benchmarking framework.

- **New benchmark script** (`audio_readspeech_benchmark.py`): Runs the full AudioDataFilterStage pipeline (VAD, Band Filter, UTMOS, SIGMOS, Speaker Separation) with configurable filters via CLI args
- **Multi-part data prep** (`prepare_readspeech_data.py`): Supports downloading 1–21 archive parts of the DNS5 Read Speech dataset (~4.88 GB per part, ~102 GB total), with resume support for interrupted downloads
- **Nightly integration** (`nightly-benchmark.yaml`): Adds `audio_readspeech_xenna` and `audio_readspeech_raydata` entries to the nightly CI benchmark suite

### Benchmark Results (10-part dataset, 4x A100-SXM4-80GB)

| Metric | Value |
|--------|-------|
| Input files | 146,796 |
| Input duration | 193.2 hours |
| Output segments | 31,285 (21.3% pass rate) |
| Output duration | 47.1 hours |
| Pipeline time | 3.24 hours (59.6x realtime) |
| Filters | VAD + Band + UTMOS + SIGMOS + Speaker Separation |

### Files Changed

| File | Description |
|------|-------------|
| `benchmarking/scripts/audio_readspeech_benchmark.py` | New benchmark script |
| `benchmarking/data_prep/prepare_readspeech_data.py` | Multi-part dataset download tool |
| `benchmarking/nightly-benchmark.yaml` | Added ReadSpeech entries to nightly CI |

## Test plan

- [x] Ran benchmark with 1 part (14,279 files) — all filters enabled, completed in ~20 min
- [x] Ran benchmark with 10 parts (146,796 files) — all filters enabled, completed in ~3.2 hours
- [x] Verified multi-part data prep download with `--num-parts 10` and `--verify-only`
- [x] Confirmed ruff passes on all modified files


benchmarking/audio-readspeech-benchmark.yaml
sarahyurick, line 1: I don't think we typically include standalone files, the preference is for everything to be in nightly. Will let @rlratzel comment on it.

benchmarking/nightly-benchmark.yaml
```diff
@@ -690,6 +694,68 @@ entries:
       - metric: is_success
         exact_value: true
 
+  - name: audio_readspeech_xenna
+    enabled: true
+    script: audio_readspeech_benchmark.py
+    timeout_s: 14400  # 4 hours; full dataset with all filters takes ~3.24h on 4×A100
```
sarahyurick: cc @rlratzel in case we want to tweak this according to some other setup?

benchmarking/nightly-benchmark.yaml
```diff
@@ -690,6 +694,68 @@ entries:
...
+          - time_taken_s
+        ping_on_failure:
+          - U07EVDG9FBK  # Shubham Bhawsar
+    requirements:
+      - metric: is_success
+        exact_value: true
+
+  - name: audio_readspeech_raydata
+    enabled: true
+    script: audio_readspeech_benchmark.py
```
sarahyurick, line 720: Should this have a timeout too?

benchmarking/nightly-benchmark.yaml
```diff
@@ -690,6 +694,68 @@ entries:
...
+    args: >-
+      --benchmark-results-path={session_entry_dir}
+      --scratch-output-path={session_entry_dir}/scratch
+      --raw-data-dir={dataset:read_speech,wav}
+      --no-auto-download
+      --executor=ray_data
+      --max-samples=5000
+      --sample-rate=48000
+      --enable-utmos
+      --enable-band-filter
```
sarahyurick, line 735: Should the Xenna and Ray Data parameters match each other exactly like this:
```suggestion
      --benchmark-results-path={session_entry_dir}
      --scratch-output-path={session_entry_dir}/scratch
      --raw-data-dir={dataset:read_speech,wav}
      --no-auto-download
      --executor=ray_data
      --max-samples=-1
      --sample-rate=48000
      --enable-vad
      --enable-sigmos
      --enable-utmos
      --enable-band-filter
      --enable-speaker-separation
```
?

benchmarking/nightly-benchmark.yaml
rlratzel: yes we can easily override this on a case-by-case basis using one or more separate override files once [this PR](https://github.com/NVIDIA-NeMo/Curator/pull/1840) is merged. The idea is that if we have multiple nightly runs on different machines, they'd have their own machine-specific overrides, and this could be one of those.

Question: is the timeout purely something to prevent a hung benchmark from holding up the entire suite, or are we using it as a means to communicate performance regressions?  If the latter, then I think a `requirements` check for a maximum `exec_time_s` value would be better.  If it's just to prevent hangs from holding up the entire suite, then I think trying to use specific values on different platforms wouldn't be as important IMO.

benchmarking/nightly-benchmark.yaml
```diff
@@ -690,6 +694,68 @@ entries:
...
+      - name: slack
+        additional_metrics:
+          - throughput_segments_per_sec
+          - num_output_segments
+          - time_taken_s
+        ping_on_failure:
+          - U07EVDG9FBK  # Shubham Bhawsar
+    requirements:
+      - metric: is_success
+        exact_value: true
```
rlratzel, line 751: This is already enforced by the framework, and we've been just putting that requirement in the config as a placeholder/example.

Perhaps there's some additional metrics that have performance requirements that can be checked? Is there a deterministic value for `num_output_segments` that can be enforced, or a minimum throughput (when run on the slowest hardware)?

benchmarking/audio-readspeech-benchmark.yaml
rlratzel: Using standalone files like this is supported by the runner, but it requires updates to our nightly job so they can be included in the script and possibly other updates, and without that our nightly won't use this file.

This extra file can be included in our repo if it's useful for developers when running a standalone benchmark, but like @sarahyurick mentioned, it will need to be in `nightly-benchmark.yaml` if it's required for the nightly, RC runs, etc.

benchmarking/nightly-benchmark.yaml
rlratzel: > Slightly off topic, but should we add support for a global/default timeout you can specify at the top-level of the YAML (like `delete_scratch`, `object_store_size`, etc.)? Each entry can override it if the default isn't appropriate. cc @praateekmahajan

Nevermind, I forgot we already support that with `default_timeout_s` [here](https://github.com/NVIDIA-NeMo/Curator/blob/06b8388a21bf4981c458c17c9ecbdde2e5017692/benchmarking/nightly-benchmark.yaml#L92).

ayushdg (dismissed):
High level question - Will the benchmark download on every run or will it use the existing datasets if already downloaded?

benchmarking/nightly-benchmark.yaml
mohammadaaftabv: +1

benchmarking/data_prep/prepare_readspeech_data.py
```diff
@@ -0,0 +1,281 @@
...
+    logger.info(f"Combined archive: {combined_archive.stat().st_size / (1024**3):.2f} GB")
+    logger.info("Extracting...")
+
+    try:
+        result = subprocess.run(  # noqa: S603
+            ["tar", "-xzf", str(combined_archive), "-C", str(output_path), "--ignore-zeros"],  # noqa: S607
+            capture_output=True, text=True, check=False,
+        )
+        if result.returncode != 0:
+            logger.error(f"Extraction failed (exit code {result.returncode}): {result.stderr[:500]}")
```
mohammadaaftabv: GNU tar exit codes:

0 = success
1 = "some files differ" / files changed while reading — harmless for archive extraction
2 = fatal error (corrupt archive, disk full, decompression failure)
The current code rejects exit code 1 as a failure. For large multi-part archives, tar commonly returns 1 due to timestamp differences or files modified during read. This would cause the prep script to falsely report extraction failure and leave ~49 GB of orphaned data.

benchmarking/data_prep/prepare_readspeech_data.py
```diff
@@ -0,0 +1,275 @@
...
+        url = DNS_READSPEECH_BASE_URL + part
+        filename = f"read_speech.tgz.{part}"
+        filepath = output_path / filename
+
+        if filepath.exists() and filepath.stat().st_size > 0:
+            logger.info(f"  [{i}/{num_parts}] Already downloaded: {filename} ({filepath.stat().st_size / (1024**3):.2f} GB)")
+        else:
+            logger.info(f"  [{i}/{num_parts}] Downloading {filename}...")
+            download_file(url, str(output_path), verbose=True)
+            logger.info(f"  [{i}/{num_parts}] Downloaded: {filepath.stat().st_size / (1024**3):.2f} GB")
```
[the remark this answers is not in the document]
mohammadaaftabv: This skips re-download if the file exists and is non-empty. But a partially downloaded 2 GB file (out of ~4.88 GB expected) also satisfies st_size > 0. The user won't know the file is truncated until tar fails during extraction — potentially after downloading 9 other parts and waiting an hour.

benchmarking/data_prep/prepare_readspeech_data.py
```diff
@@ -0,0 +1,281 @@
...
+        result = subprocess.run(  # noqa: S603
+            ["tar", "-xzf", str(combined_archive), "-C", str(output_path), "--ignore-zeros"],  # noqa: S607
+            capture_output=True, text=True, check=False,
+        )
+        if result.returncode != 0:
+            logger.error(f"Extraction failed (exit code {result.returncode}): {result.stderr[:500]}")
+            return False
+
+        for fpath in downloaded_files:
+            os.remove(fpath)
```
mohammadaaftabv, line 237: On extraction failure:

The combined archive (read_speech.tgz) is cleaned up in the finally block — good
But the individual part files (read_speech.tgz.partaa, read_speech.tgz.partab, etc.) remain on disk
For a 10-part download, that's ~49 GB of orphaned files. The part cleanup (os.remove(fpath)) only runs on the success path.

However — keeping part files on failure is actually correct behavior, since _download_parts() skips existing files. If extraction fails (e.g., disk full), the user can free space and re-run without re-downloading ~49 GB. The problem is that the user isn't told about the retained files or their size.

Suggested fix: Add a log message on the failure path:

if result.returncode not in (0, 1):
    total_parts_gb = sum(os.path.getsize(f) for f in downloaded_files) / (1024**3)
    logger.error(f"Extraction failed (exit code {result.returncode}): {result.stderr[:500]}")
    logger.info(f"Part files retained for retry ({total_parts_gb:.1f} GB). Delete manually to free space.")
    return False

benchmarking/scripts/audio_readspeech_benchmark.py
```diff
@@ -0,0 +1,312 @@
...
+                        help="Path to pre-downloaded ReadSpeech WAV files (skips download to scratch)")
+    parser.add_argument("--executor", default="xenna", choices=["xenna", "ray_data"],
+                        help="Executor to use (default: xenna)")
+    parser.add_argument("--max-samples", type=int, default=5000,
+                        help="Maximum samples to process (-1 for all, default: 5000)")
+    parser.add_argument("--batch-size", type=int, default=1,
+                        help="Batch size for manifest creation (default: 1)")
+    parser.add_argument("--sample-rate", type=int, default=48000,
+                        help="Target sample rate (default: 48000)")
+    parser.add_argument("--auto-download", action="store_true", default=True,
```
mohammadaaftabv: parser.add_argument("--auto-download", action="store_true", default=True,
                    help="Auto-download dataset if missing (default: True)")
action="store_true" with default=True is a no-op flag. The value is True whether or not the user passes --auto-download. The --no-auto-download counterpart works correctly (action="store_false"), but the --auto-download flag misleads users into thinking they're enabling something.

Same issue on line ~283:

parser.add_argument("--speaker-exclude-overlaps", action="store_true", default=True,
                    help="Exclude overlapping speech (default: True)")
Greptile flagged this.

Suggested fix: Drop the --auto-download flag entirely. Keep only --no-auto-download since the default is already True:

parser.add_argument("--no-auto-download", dest="auto_download", action="store_false",
                    help="Disable automatic download (default: enabled)")
parser.set_defaults(auto_download=True)
Same pattern for --speaker-exclude-overlaps / --no-speaker-exclude-overlaps.

benchmarking/nightly-benchmark.yaml
mohammadaaftabv: +1

benchmarking/scripts/audio_readspeech_benchmark.py
```diff
@@ -0,0 +1,312 @@
...
+def _count_jsonl_lines(results_dir: Path) -> int:
+    """Count total lines across all JSONL files in a directory."""
+    total = 0
+    for jsonl_file in results_dir.glob("*.jsonl"):
+        with open(jsonl_file) as f:
+            total += sum(1 for _ in f)
+    return total
+
+
+def run_readspeech_benchmark(  # noqa: PLR0913, PLR0915
```
mohammadaaftabv, line 50: PLR0913 (too many arguments — 24 parameters) and PLR0915 (too many statements) are suppressed. For a benchmark script that's essentially a CLI wrapper, this is understandable. But 24 positional keyword arguments make it easy to pass the wrong value to the wrong parameter.

consider grouping into a BenchmarkConfig dataclass or a dict-based approach.

benchmarking/scripts/audio_readspeech_benchmark.py
```diff
@@ -0,0 +1,312 @@
...
+    band_value: str = "full_band",
+    enable_utmos: bool = False,
+    utmos_mos_threshold: float = 3.4,
+    enable_sigmos: bool = False,
+    sigmos_noise_threshold: float = 4.0,
+    sigmos_ovrl_threshold: float = 3.5,
+    enable_speaker_separation: bool = False,
+    speaker_exclude_overlaps: bool = True,
+    speaker_min_duration: float = 0.8,
+    **kwargs,  # noqa: ARG001
```
mohammadaaftabv: This silently swallows any unknown keyword arguments. Since the function is called via **vars(args) from main(), any new argparse argument that doesn't match a function parameter will be silently ignored instead of raising a TypeError.

For example, if someone adds --new-flag to argparse but forgets to add the corresponding parameter to run_readspeech_benchmark(), the flag is silently dropped — no error, no warning.

Suggested fix: Filter vars(args) to only known parameters instead:

import inspect
sig = inspect.signature(run_readspeech_benchmark)
known_params = {k: v for k, v in vars(args).items() if k in sig.parameters}
result_dict.update(run_readspeech_benchmark(**known_params))
Or simply remove **kwargs and let mismatches raise TypeError — that's the safer default for a benchmark where silent data loss is worse than a crash.

mohammadaaftabv (dismissed):

benchmarking/audio-readspeech-benchmark.yaml
shubhamNvidia: Removed audio-readspeech-benchmark.yaml. Both entries are already in nightly-benchmark.yaml, so nightly/RC are unaffected.

benchmarking/nightly-benchmark.yaml
shubhamNvidia: Hang guard only — clarified in a comment. Will follow up with a time_taken_s requirement once we have stable baseline numbers.

benchmarking/nightly-benchmark.yaml
shubhamNvidia: Matched the raydata config to xenna (full dataset, all five filters, same timeout_s) so the two executors are directly comparable. Will tune the timeout if raydata can't keep up

benchmarking/nightly-benchmark.yaml
shubhamNvidia: Agreed is_success is redundant. I'll replace it with num_output_segments (exact_value) and a throughput_segments_per_sec (min_value) floor in a follow-up once we have one clean nightly run on the target hardware to lock in the numbers

sarahyurick (approved):
LGTM!

PR #1863, opened 2026-04-23
Add filtration and PnCWithLLM stages for audio tagging pipeline

sushmitha-deva-09:
## Description

Add new audio tagging pipeline stages for text processing, quality metrics, and LLM-based Punctuation & Capitalization (PnC).

### New Stages

- PNCwithvLLMInferenceStage / CleanLLMOutputStage — LLM-powered PnC via vLLM with CER-based validation and BERT fallback
- VLLMInference — Reusable vLLM lifecycle helper (weight pre-download, engine setup, cleanup)
- ComputeWER/CER/SQUIM/BandwidthStage — Transcript and audio quality metrics
- PrepareModuleSegmentsStage — Punctuation-based segmentation for training-ready chunks
- ITN / ArabicRemoveDiacritics / ChineseConversion — Text normalization stages

### Other Changes

- Moved sortformer.py to inference/speaker_diarization/
- Added nemo_text_processing, opencc-python-reimplemented, pyarabic deps
- Force VLLM_USE_V1=0 to prevent Ray actor conflicts
- Expanded tutorials/audio/tagging/README.md with pipeline diagram and PnC guide

## Tests

- Unit tests for all new stages (51+ tests)
- GPU E2E tests: PnC LLM (simple, first-pass, second-pass ASR) and full ASR pipeline
- Reference manifests for deterministic validation

## Checklist


benchmarking/scripts/audio_tagging_benchmark.py
```diff
@@ -91,7 +95,7 @@ def run_audio_tagging_benchmark(  # noqa: PLR0913
             name="PyAnnoteDiarization",
             hf_token=hf_token,
             max_length=max_segment_length,
-        ).with_(resources=Resources(cpus=cpus, gpus=0.5))
+        ).with_(resources=Resources(cpus=cpus, gpus=0.25))
```
ayushdg: What's the motivation for these changes?

benchmarking/scripts/audio_tagging_benchmark.py
```diff
@@ -110,7 +114,7 @@ def run_audio_tagging_benchmark(  # noqa: PLR0913
             is_fastconformer=True,
             decoder_type="rnnt",
             batch_size=asr_batch_size,
-        ).with_(resources=Resources(cpus=cpus, gpus=0.45))
+        ).with_(resources=Resources(gpus=0.35))
```
ayushdg: Same question.

nemo_curator/stages/audio/inference/llm/vllm_base.py
```diff
@@ -0,0 +1,237 @@
...
+    # Force the classic (v0) engine before importing vllm, because vllm
+    # caches this decision at import time.
+    os.environ["VLLM_USE_V1"] = "0"
+    from vllm import LLM, SamplingParams
+except ImportError:
+    LLM = None
+    SamplingParams = None
+
+
+class VLLMInference:
```
ayushdg: The repo already has 2-3 different ways to do inference:
1. In process vllm where the stage spins up a vllm server and processes its own batches. Like video captioning
2. InferenceServer - VLLM server started that exposes an endpoint usable by any stage.

Which category does this fall under? Can we use existing features here rather than implementing it for audio?

nemo_curator/stages/audio/tagging/text/arabic_remove_diacritics.py
```diff
@@ -0,0 +1,73 @@
...
+from dataclasses import dataclass
+
+from pyarabic import araby
+
+from nemo_curator.stages.base import ProcessingStage
+from nemo_curator.tasks import AudioTask
+
+
+@dataclass
+class ArabicRemoveDiacriticsStage(ProcessingStage[AudioTask, AudioTask]):
```
ayushdg: This is fine for now but we should think about abstracting this into a common class with language options if there's overlap with other languages.

pyproject.toml
```diff
@@ -108,6 +108,9 @@ audio_common = [
     "pydub>=0.25.1",
     "transformers",
     "accelerate",
+    "nemo_text_processing",
+    "opencc-python-reimplemented",
+    "pyarabic",
```
ayushdg: note to review dependencies before adding

ayushdg (changes requested):

nemo_curator/stages/audio/inference/llm/vllm_base.py
sushmitha-deva-09: This falls under category 1 (in process vllm). Existing VLLMModel (nemo_curator/models/vllm_model.py) is the closest existing abstraction, but it lacks few features we need for audio such as chat template support, yaml based prompt file/field configuration, setup_on_node() to predownload model etc. This class is intentionally a plain helper class (not a stage) that can be composed by any stage needing LLM generation. Let me know if there's a better place to keep this in repo or or if I should update existing ones.

benchmarking/scripts/audio_tagging_benchmark.py
```diff
@@ -125,6 +129,49 @@ def run_audio_tagging_benchmark(  # noqa: PLR0913
         ).with_(resources=Resources(cpus=cpus))
     )
 
+    # Bandwidth estimation per segment
+    pipeline.add_stage(BandwidthEstimationStage(name="BandwidthEstimation").with_(resources=Resources(cpus=cpus)))
+
+    # Audio quality metrics (PESQ, STOI, SI-SDR)
+    pipeline.add_stage(TorchSquimQualityMetricsStage(name="SquimMetrics").with_(resources=Resources(gpus=0.05)))
```
sarahyurick, line 138: Do you know if there is an advantage to using 0.05 GPUs versus having this be just a CPU stage?

nemo_curator/stages/audio/inference/llm/vllm_base.py
```diff
@@ -0,0 +1,243 @@
...
+            msg = f"Generation failed: {e}"
+            raise RuntimeError(msg) from e
+
+    def process_entry_prompts(self, entry_prompts: list, batch_size: int = 10000) -> list:
+        """Generate in batches, then clean up.
+
+        If :meth:`setup` has not been called yet, it is called automatically.
+        """
+        if self.llm is None:
+            self.setup()
```
sarahyurick: Avoid having `setup` in a process function:
```suggestion
```

nemo_curator/stages/audio/tagging/text/pnc.py
```diff
@@ -0,0 +1,509 @@
...
+      ``segments`` list inside each manifest entry (or falls back to the
+      top-level ``text_key``) and applies PNC to each text.
+    * **segmented entry** (``is_audio_entry=False``) -- treats each manifest
+      row as an individual segment, reconstructs text from the ``alignment``
+      word list, runs PNC, and writes the punctuated words back into
+      ``alignment``.
+
+    .. note::
+       Requires ``nemo_toolkit < 2.4.1``.  When the installed version is
+       2.4.1 or later the stage silently passes tasks through unchanged.
```
sarahyurick: Should there be a hard check about this somewhere in the script?

nemo_curator/stages/audio/tagging/text/pnc.py
```diff
@@ -0,0 +1,509 @@
...
+    def inputs(self) -> tuple[list[str], list[str]]:
+        return [], []
+
+    def outputs(self) -> tuple[list[str], list[str]]:
+        return [], []
+
+    @property
+    def _device(self) -> str:
+        """Derive device from resources configuration."""
+        return "cuda" if self.resources.requires_gpu and torch.cuda.is_available() else "cpu"
```
sarahyurick: If the user does `Resources(gpus=1)` but `torch.cuda.is_available()` is false, I think it should fail loudly instead of silently falling back to CPU.

nemo_curator/stages/audio/tagging/text/pnc.py
```diff
@@ -0,0 +1,509 @@
...
+
+    def load_model(self) -> None:
+        try:
+            from nemo.collections.nlp.models import PunctuationCapitalizationModel
+        except (ImportError, ModuleNotFoundError):
+            logger.warning(
+                f"[{self.name}] Could not import PunctuationCapitalizationModel. "
+                f"This model is only available in nemo_toolkit <= 2.4.1. "
```
sarahyurick: I see. Should we even be adding this stage if it is completely deprecated in nemo toolkit?

Alternatively, should we ask them to add it back?

nemo_curator/stages/audio/tagging/text/pnc.py
```diff
@@ -0,0 +1,509 @@
...
+                if word.get("word", "") != "":
+                    if pnc_idx >= len(pnc_words):
+                        logger.warning(f"[{self.name}] PNC word count mismatch; stopping alignment update.")
+                        break
+                    word["word"] = pnc_words[pnc_idx]
+                    pnc_idx += 1
+
+    def process(self, task: AudioTask) -> AudioTask:
+        if self._skip:
+            return task
```
sarahyurick: I think if the user really wants to use the stage, then they will add it to their pipeline, else they will exclude it from their pipeline. It should fail loudly instead of skipping.

nemo_curator/stages/audio/tagging/text/pnc.py
```diff
@@ -0,0 +1,509 @@
...
+    def inputs(self) -> tuple[list[str], list[str]]:
+        return [], []
+
+    def outputs(self) -> tuple[list[str], list[str]]:
+        return [], [self.generation_field]
+
+    def process(self, task: AudioTask) -> AudioTask:
+        """Process a single AudioTask — collect prompts, run inference, write back."""
+        if self._vllm is None or self._vllm.llm is None:
+            return task
```
sarahyurick:
```suggestion
```

nemo_curator/stages/audio/tagging/text/pnc.py
```diff
@@ -0,0 +1,509 @@
...
+
+        for (_, key_path), output in zip(items, outputs, strict=True):
+            self._store_generation(data, key_path, output.outputs[0].text, self.generation_field)
+
+        return task
+
+    def process_batch(self, tasks: list[AudioTask]) -> list[AudioTask]:
+        """Batch-process multiple AudioTasks with a single vLLM call for efficiency."""
+        if not tasks or self._vllm is None or self._vllm.llm is None:
+            return tasks if tasks else []
```
sarahyurick:
```suggestion
```

nemo_curator/stages/audio/inference/llm/vllm_base.py
```diff
@@ -0,0 +1,243 @@
...
+        """Instantiate the vLLM engine on the target device.
+
+        Called once per worker.  Detects CUDA availability, then loads the
+        model.  If :meth:`setup_on_node` was not called first, the tokenizer
+        and sampling params are created here as well.
+        """
+        self.device = "cuda"
+        if not torch.cuda.is_available():
+            self.device = "cpu"
+            logger.warning("CUDA is not available, using CPU")
```
sarahyurick: Maybe let it fail instead.

nemo_curator/stages/audio/tagging/metrics/squim.py
```diff
@@ -0,0 +1,145 @@
...
+    def inputs(self) -> tuple[list[str], list[str]]:
+        return [], [self.audio_filepath_key, "segments"]
+
+    def outputs(self) -> tuple[list[str], list[str]]:
+        return [], [self.audio_filepath_key, "segments", "metrics"]
+
+    @property
+    def _device(self) -> str:
+        """Derive device from resources configuration."""
+        return "cuda" if self.resources.requires_gpu and torch.cuda.is_available() else "cpu"
```
sarahyurick: Same as other comments, let it error.

nemo_curator/stages/audio/tagging/metrics/squim.py
```diff
@@ -0,0 +1,145 @@
...
+            num_frames = end_frame - start_frame
+
+            if num_frames <= 0:
+                logger.warning(f"[{self.name}] Zero-length segment at {start}-{end}s in {audio_path}, skipping")
+                continue
+
+            y = audio[start_frame:end_frame]
+            y = torch.from_numpy(y).unsqueeze(0)
+
+            target_sr = 16000
```
sarahyurick: Should this not be hardcoded?

tests/stages/audio/tagging/e2e/test_pnc_llm_e2e.py
```diff
@@ -0,0 +1,174 @@
...
+    cfg.workspace_dir = str(tmp_path)
+    cfg.language_short = "en"
+
+    if prompt_file:
+        cfg.prompt_file = str(prompt_file)
+
+    if OmegaConf.is_missing(cfg, "resampled_audio_dir"):
+        cfg.resampled_audio_dir = str(tmp_path / "audio_resampled")
+    if hasattr(cfg, "hf_token"):
+        cfg.hf_token = os.getenv("HF_TOKEN", "")
```
sarahyurick: Has the HF token in our CI/CD been configured to work for this test yet?

tests/stages/audio/tagging/e2e/test_pnc_llm_e2e.py
```diff
@@ -0,0 +1,174 @@
...
+
+    pipeline = create_pipeline_from_yaml(cfg)
+    executor = XennaExecutor(config={"execution_mode": "batch"})
```
sarahyurick: Why batch? Usually it should be streaming, that is supposed to be more performant.
```suggestion
    executor = XennaExecutor(config={"execution_mode": "streaming"})
```

tests/stages/audio/tagging/e2e/test_pnc_llm_e2e.py
```diff
@@ -0,0 +1,174 @@
...
+        logger.warning("Reference manifest generated: %s", ref_path)
+        return
+
+    check_output(cfg.final_manifest, reference_manifest_file, text_key)
+
+
+@pytest.fixture
+def pnc_simple_manifest(tmp_path: Path) -> str:
+    """Input manifest with pre-populated text_2 for the simple PNC test (no ASR stage)."""
+    import json
```
sarahyurick: This can be a top-level import.

tests/stages/audio/tagging/e2e/test_pnc_llm_e2e.py
```diff
@@ -0,0 +1,174 @@
...
+        config_path,
+        final_manifest,
+        reference_manifest_file,
+        text_key="text_2",
+        prompt_file=PROMPT_FILE,
+    )
+
+
+@pytest.mark.gpu
+@pytest.mark.skipif(not os.getenv("HF_TOKEN"), reason="HF_TOKEN required for PyAnnote models")
```
sarahyurick: Let's remove this so we can make sure this test is actually running.

tests/stages/audio/tagging/e2e/test_tts_e2e.py
```diff
@@ -60,7 +60,7 @@ def test_tts_e2e(tmp_path: Path, get_input_manifest: str) -> None:
     pipeline = create_pipeline_from_yaml(cfg)
     executor = XennaExecutor(
         {
-            "cpu_allocation_percentage": 0.9,
+            "execution_mode": "batch",  # batch mode to satisfy cluster resource constraints
```
sarahyurick: Not sure I understand this comment?

tests/stages/audio/tagging/text/test_text.py
```diff
@@ -0,0 +1,90 @@
...
+
+try:
+    from nemo.collections.nlp.models import PunctuationCapitalizationModel  # noqa: F401
+
+    _pnc_available = True
+except (ImportError, ModuleNotFoundError):
+    _pnc_available = False
+
+
+@pytest.mark.skipif(not _pnc_available, reason="PunctuationCapitalizationModel requires nemo_toolkit <= 2.4.1")
```
sarahyurick: So this will always be skipped since this is the pin: https://github.com/NVIDIA-NeMo/Curator/blob/main/pyproject.toml#L102

tutorials/audio/tagging/asr_pipeline.yaml
```diff
@@ -0,0 +1,226 @@
...
+language_short: en
+hf_token: ""
+sample_rate: 16000
+max_segment_length: 40
+execution_mode: streaming
+batch_size: 4
+text_key1: text
+text_key2: text_2
+resources:
+  cpus: 2
```
sarahyurick: It seems like maybe the same `{resources}` should not be used for every stage as below? It seems like this makes it hard to configure individual resources per stage.

sarahyurick (changes requested):

nemo_curator/stages/audio/tagging/text/pnc.py
sushmitha-deva-09: We have already consulted with the NeMo team, and they confirmed that this was intentionally deprecated. They advised us to install earlier versions of nemo_toolkit if we need to use these model classes.

sarahyurick: Is there a reason we can't just add the model class to Curator ourselves?

benchmarking/scripts/audio_tagging_benchmark.py
sushmitha-deva-09: I added a new GPU stage and wanted to test benchmarking on a single GPU, so I adjusted the GPU split percentages across the various gpu stages.

tutorials/audio/tagging/asr_pipeline.yaml
sushmitha-deva-09: Do you have any suggestions on how to add resources then ?

tests/stages/audio/tagging/text/test_text.py
sushmitha-deva-09: yes

nemo_curator/stages/audio/tagging/text/pnc.py
sushmitha-deva-09: this would introduce a significant amount of NLP code into Curator, which may not be necessary at this point

sarahyurick: This is the file right? https://github.com/NVIDIA-NeMo/NeMo/blob/v2.4.1/nemo/collections/nlp/models/token_classification/punctuation_capitalization_model.py

Any reason it can't essentially be copy/pasted?

I'm having trouble understanding why we would even want to support something deprecated by NeMo with no plans for adding it back.

sushmitha-deva-09: Yes, that's the correct file, but it requires importing NLP modules, models, data, and other dependencies - essentially, we may need to copy some NLP code. The reason for the deprecation is that they want to focus exclusively on speech-related models going forward, so they've removed all text/NLP inference code. Surprisingly, they haven't added the removed code anywhere, and they simply advise using previous versions if needed. 
We still have a few BERT-based PNC models that were trained some time ago and are currently being used for PNC improvements on ASR text, so I included support for them in this PR.

sushmitha-deva-09: I removed the PnC using bert inference in this PR and will address this in a separate PR .

benchmarking/scripts/audio_tagging_benchmark.py
sushmitha-deva-09: yes, even with just 5% of GPU usage, this stage is faster with GPU by ~20 times than CPU for a 1 min audio.

benchmarking/scripts/audio_tagging_benchmark.py
```diff
@@ -125,6 +129,49 @@ def run_audio_tagging_benchmark(  # noqa: PLR0913
         ).with_(resources=Resources(cpus=cpus))
     )
 
+    # Bandwidth estimation per segment
+    pipeline.add_stage(BandwidthEstimationStage(name="BandwidthEstimation").with_(resources=Resources(cpus=cpus)))
```
sarahyurick: The `cpus` parameter for `run_audio_tagging_benchmark` is a bit confusing I think. I would have assumed it meant total CPUs to use.

Maybe let's remove it and replace with just `with_(resources=Resources(cpus=1))` everywhere

nemo_curator/stages/audio/inference/llm/vllm_base.py
sarahyurick: Since https://github.com/NVIDIA-NeMo/Curator/blob/main/nemo_curator/models/vllm_model.py is supposed to be as generic as possible, let's try to use it here.

nemo_curator/stages/audio/tagging/text/pnc.py
```diff
@@ -0,0 +1,363 @@
...
+    def process(self, task: AudioTask) -> AudioTask:
+        """Process a single AudioTask — collect prompts, run inference, write back."""
+        data = task.data
+        items = self._collect_prompts(data)
+
+        if not items:
+            return task
+
+        prompts = [p for p, _ in items]
+        outputs = self._vllm.process_batch(prompts)
```
sarahyurick: So `self._vllm.process_batch` is called either way whether we are in `PNCwithvLLMInferenceStage`'s `process` or `process_batch` function. I think maybe we should only have `PNCwithvLLMInferenceStage.process_batch`.

nemo_curator/stages/audio/tagging/text/pnc.py
```diff
@@ -0,0 +1,363 @@
...
+    alignment_key: str = "alignment"
+
+    name: str = "CleanLLMOutput"
+    resources: Resources = field(default_factory=lambda: Resources(cpus=1.0))
+
+    _cleaned_generation_field: str = field(default="", repr=False)
+    _full_vocab_set: set[str] = field(default_factory=set, repr=False)
+
+    def __post_init__(self) -> None:
+        vocab = load_vocab_file(self.vocab_file) if self.vocab_file is not None else set(self.vocab_set)
```
sarahyurick: Should this be in `setup` instead of `post_init`?

tests/stages/audio/tagging/e2e/test_pnc_llm_e2e.py
sarahyurick: Bumping this.

tutorials/audio/tagging/asr_pipeline.yaml
sarahyurick: It can be done like: https://github.com/NVIDIA-NeMo/Curator/blob/main/tutorials/audio/fleurs/pipeline.yaml

tutorials/audio/tagging/tts_pipeline.yaml
```diff
@@ -128,5 +133,23 @@ stages:
     words_key: "words"
     resources: ${resources}
 
+  - _target_: nemo_curator.stages.audio.tagging.metrics.bandwidth.BandwidthEstimationStage
+    name: "BandwidthEstimation"
+    resources: ${resources}
```
sarahyurick: Same comment as above.

tutorials/audio/tagging/README.md
```diff
@@ -68,6 +250,29 @@ python tutorials/audio/tagging/main.py \
   hf_token=<your_hf_token>
```
sarahyurick, line 121: Can you update the document with more specific instructions for what the user should be requesting HF access to, including a link?

sushmitha-deva-09: Done, added a separate section for which all models we would need access to run the pipeline

nemo_curator/stages/audio/inference/llm/vllm_base.py
sushmitha-deva-09: I moved the class to vllm_model.py

nemo_curator/stages/audio/tagging/text/pnc.py
```diff
@@ -0,0 +1,353 @@
...
+        if not all_prompts:
+            return tasks
+
+        outputs: list = []
+        for i in range(0, len(all_prompts), self.inference_batch_size):
+            batch = all_prompts[i : i + self.inference_batch_size]
+            outputs.extend(self._vllm.process_batch(batch))
+
+        for (task_idx, key_path), output in zip(result_map, outputs, strict=True):
+            self._store_generation(tasks[task_idx].data, key_path, output.outputs[0].text, self.generation_field)
```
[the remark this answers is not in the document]
mohammadaaftabv: +1
This assumes every vLLM result has at least one completion in `output.outputs`. `zip(..., strict=True)` catches a result-count mismatch, but it does not protect against an individual `RequestOutput(outputs=[])`, which would raise `IndexError` here.

Can we add a small helper to safely extract generated text and either fail loudly with task/audio id + prompt index, or mark the segment with something like `pnc_generation_error`? This should also have tests for empty completions and partial/mismatched vLLM outputs.

nemo_curator/models/vllm_model.py
```diff
@@ -204,3 +213,194 @@ def get_tokenizer(self) -> Any:  # noqa: ANN401
...
+
+    def setup_on_node(self) -> None:
+        """Download model weights and tokenizer to the local cache.
+
+        Called once per node before workers start.  Only downloads artifacts
+        (no GPU memory used) so that :meth:`setup` can load instantly.
+        """
+        model_name = self.model_params.get("model", "")
+        if model_name:
+            snapshot_download(repo_id=model_name)
```
mohammadaaftabv: `snapshot_download()` returns the local snapshot path, but we discard it here. Later `setup()` still calls `AutoTokenizer.from_pretrained(model_name)` and `load_model()` still passes the original HF model id into `LLM(**self.model_params)`.

That means worker setup can still depend on live Hugging Face model resolution even after `setup_on_node()` predownloaded the model. For distributed runs this is fragile and can cause repeated metadata/auth/network calls across workers.

Can we store the local snapshot path returned by `snapshot_download()` and replace `model_params["model"]` with that path before tokenizer/vLLM construction? 

local_model_path = snapshot_download(repo_id=model_name, ...)
self.model_params["model"] = local_model_path
self.tokenizer = AutoTokenizer.from_pretrained(local_model_path)


Ideally this should also support `cache_dir`, `revision`, and token if those are expected from config.

nemo_curator/stages/audio/tagging/text/pnc.py
```diff
@@ -0,0 +1,353 @@
...
+
+    @staticmethod
+    def _store_generation(data: dict, key_path: list, text: str, gen_field: str) -> None:
+        """Write *text* into *data* following *key_path*."""
+        target = data
+        for key in key_path:
+            target = target[key]
+        target[gen_field] = text
+
+    def inputs(self) -> tuple[list[str], list[str]]:
```
mohammadaaftabv: This stage currently declares no required inputs, but it needs either top-level text via `text_key` or segment text via `segments_key`. Returning `([], [])` bypasses Curator's normal input validation, so a misconfigured pipeline can silently no-op instead of failing early.

Can we make the input contract explicit? One option is to add an `input_scope: "entry" | "segments"` config and require `text_key` or `segments_key` accordingly. If auto-detection is intended, the stage should still validate that at least one expected input exists and fail or mark a clear skip reason when the user explicitly configured the stage.

nemo_curator/stages/audio/tagging/metrics/wer.py
```diff
@@ -0,0 +1,272 @@
...
+
+            metrics = segment.get("metrics", {})
+
+            hypothesis_pnc, hypothesis_clean = self.normalize_and_clean_text(segment[self.hypothesis_text_key])
+            reference_pnc, reference_clean = self.normalize_and_clean_text(segment[self.reference_text_key])
+
+            metrics["char_rate"] = self.get_char_rate(segment[self.hypothesis_text_key], duration)
+            metrics["word_rate"] = self.get_word_rate(segment[self.hypothesis_text_key], duration)
+
+            wer_val, tokens, ins_rate, del_rate, sub_rate = word_error_rate_detail(
```
mohammadaaftabv: This checks that the raw hypothesis/reference keys exist, but the normalized strings can still be empty. For example, empty text, whitespace-only text, punctuation-only text, or text stripped by the normalizer can produce `hypothesis_clean == ""` or `reference_clean == ""`.

Can we guard empty normalized hypothesis/reference before calling `word_error_rate_detail()`? I think the stage should either skip the metric with an explicit `metric_skip_reason` or write `None` for WER/CER, rather than relying on the WER helper's behavior for empty inputs. It would be good to add tests for empty, whitespace-only, punctuation-only, and normalized-to-empty cases.

Proposed Fix

Before each WER/CER call, explicitly check normalized strings.

For example:

if not reference_clean.strip() or not hypothesis_clean.strip():
    metrics["wer"] = None
    metrics["cer"] = None
    metrics["metric_skip_reason"] = "empty_normalized_text"
    segment["metrics"] = metrics
    continue
Or use a helper:

safe_word_error_rate_detail(...)
that returns None/skip metadata for empty refs/hyps.

nemo_curator/stages/audio/tagging/metrics/squim.py
```diff
@@ -0,0 +1,150 @@
...
+        """Load model. Called once per worker before processing."""
+        if self.model is None:
+            self.model = SQUIM_OBJECTIVE.get_model()
+
+        if self._device == "cuda":
+            self.model = self.model.cuda()
+
+        logger.info(f"[{self.name}] Initialized SQUIM model on {self._device}")
+
+    def process(self, task: AudioTask) -> AudioTask:
```
mohammadaaftabv: This stage requests GPU resources by default, but it only implements `process()` and calls the SQUIM model one segment at a time. That will underutilize the GPU and add repeated CPU/GPU synchronization overhead for long manifests with many segments.

Can this be implemented as a batched `process_batch()` stage instead? The stage could collect valid segments across the task batch, resample/pad/collate them into model batches, run SQUIM in fewer GPU calls, and scatter the metrics back to the original task/segment. This would also match the audio README guidance for GPU stages.

Please make this change and allow configurable batch size if not exists already, so we can allow users to decide whether or not they want to run with batch size 1 or more?

nemo_curator/stages/audio/tagging/metrics/squim.py
```diff
@@ -0,0 +1,150 @@
...
+                raise RuntimeError(msg)
+            return "cuda"
+        return "cpu"
+
+    def setup_on_node(
+        self, _node_info: NodeInfo | None = None, _worker_metadata: WorkerMetadata | None = None
+    ) -> None:
+        """Setup stage on node."""
+        if self.model is None:
+            self.model = SQUIM_OBJECTIVE.get_model()
```
mohammadaaftabv: I think `setup_on_node()` should not retain the SQUIM model object on the stage. Per the audio-stage lifecycle, `setup_on_node()` is best used for node-level preparation/cache warming, while worker-local model allocation should happen in `setup()`.

Can we move model construction fully into `setup()`? If we need to predownload/cache SQUIM weights once per node, `setup_on_node()` can warm the cache without storing the model on `self`.

Can we follow the same lifecycle pattern used by other audio stages? For example, `PnCRestorationStage.setup_on_node()` documents that GPU/vLLM engine creation belongs in `setup()` and only pre-downloads weights/validates prompts at node setup time (`text_filtering/pnc_restoration.py`). `NeMoASRAlignerStage.setup_on_node()` similarly downloads model weights without loading the model into memory, while `setup()` loads the model onto the worker/device.

Here, `setup_on_node()` stores `self.model = SQUIM_OBJECTIVE.get_model()`, which retains a heavyweight model object at node setup time. Can we move model construction to `setup()` and keep `setup_on_node()` only for cache warming/predownload if needed?


Ref:
references/Curator/nemo_curator/stages/audio/text_filtering/itn_restoration.py:243-270

nemo_curator/stages/audio/tagging/prepare_module_segments.py
```diff
@@ -0,0 +1,421 @@
...
+            if "segments" not in data_entry:
+                logger.info(f"[{self.name}] No segments in metadata for: {data_entry.get('audio_filepath', '')}")
+                return task
+
+            words = self.get_words_list_from_all_segments(data_entry)
+            if self.module == "asr":
+                self.prepare_asr_segments(words, data_entry)
+            elif self.module == "tts":
+                self.prepare_tts_segments(words, data_entry)
+        except Exception as e:  # noqa: BLE001
```
mohammadaaftabv: This catches all exceptions and then returns the task unchanged. Since this stage prepares the segment structure used downstream for ASR/TTS data creation, silently passing through the original task can make a failed preparation look like a successful output.

Can we make the failure policy explicit? For example, `on_error: "raise" | "drop" | "mark"` with `raise` as the default for tutorials/tests. If we choose to keep the task, it should be marked with something like `segment_preparation_error` so downstream stages do not treat it as successfully prepared.

tutorials/audio/tagging/asr_pipeline.yaml
```diff
@@ -0,0 +1,215 @@
...
+input_manifest: ???
+final_manifest: ???
+workspace_dir: /tmp/tagging_workspace
+resampled_audio_dir: ${workspace_dir}/audio_resampled
+language_short: en
+hf_token: ""
+sample_rate: 16000
+max_segment_length: 40
+execution_mode: streaming
+batch_size: 4
```
mohammadaaftabv: Is this top-level `batch_size` actually consumed anywhere? The stage configs below appear to use their own batch sizes, e.g. `ASRAlignment.batch_size: 32` and `ASRAlignment2.batch_size: 64`.

If this global value is not wired into stage construction, I think we should remove it or replace it with explicit per-stage variables like `asr_alignment_batch_size`, `second_pass_asr_batch_size`, and `pnc_batch_size`. Otherwise users may change this value expecting throughput changes while nothing actually changes.

nemo_curator/stages/audio/tagging/text/pnc.py
```diff
@@ -0,0 +1,353 @@
...
+        return [], []
+
+    def outputs(self) -> tuple[list[str], list[str]]:
+        return [], [self.generation_field]
+
+    def process(self, task: AudioTask) -> AudioTask:
```
mohammadaaftabv: Since this is a GPU/batched inference stage, should `process()` be unsupported here? Right now `process()` calls `process_batch([task])`, which makes single-task execution work but can hide accidental non-batched execution in tests or backends.

I think it would be safer to make `process_batch()` the only supported path and have `process()` raise `NotImplementedError`, similar to other batched audio stages. That way tests exercise the same batched path we expect in production.

benchmarking/scripts/audio_tagging_benchmark.py
mohammadaaftabv: Can we add a short comment that these fractional GPU settings are benchmark-specific empirical values, not general production defaults? They may be valid for this single-GPU benchmark, but values like `0.05` GPU for SQUIM are hardware/workload dependent and could be misleading if copied into other pipeline configs.

mohammadaaftabv (changes requested):

nemo_curator/stages/audio/tagging/metrics/wer.py
sushmitha-deva-09: Even if one text is empty, we'll need to call the wer helper method and save all insertion, deletion rate values.

nemo_curator/stages/audio/metrics/wer.py
sarahyurick, line 1: Can this be integrated with https://github.com/NVIDIA-NeMo/Curator/blob/main/nemo_curator/stages/audio/metrics/get_wer.py ?

nemo_curator/models/vllm_model.py
```diff
@@ -204,3 +213,200 @@ def get_tokenizer(self) -> Any:  # noqa: ANN401
             msg = "Model not initialized. Call setup() first."
             raise RuntimeError(msg)
         return self._llm.get_tokenizer()
+
+
+class VLLMInference:
```
sarahyurick: I guess I am wondering what the overlaps of `VLLMModel` and `VLLMInference` are? Like is there some parent class we should create for them to share?

sushmitha-deva-09: Added overlaps to a parent class and modified both the classes

tests/stages/audio/tagging/e2e/test_pnc_llm_e2e.py
sarahyurick: Bumping this comment. Is this test passing in our CI/CD or is it being skipped?

benchmarking/nightly-benchmark.yaml
```diff
@@ -1058,6 +1058,31 @@ entries:
...
+    script: audio_tagging_benchmark.py
+    args: >-
+      --benchmark-results-path={session_entry_dir}
+      --input-manifest={session_entry_dir}/input.jsonl
+      --repeat-factor=1
+      --hf-token=${HF_SECRET_KEY}
+      --max-segment-length=40
+      --asr-batch-size=50
+      --executor=xenna
+      --cpus=10
```
mohammadaaftabv: This enabled nightly entry will fail before the benchmark runs because it passes `--cpus=10`, but `audio_tagging_benchmark.py` does not register a `--cpus` argument. The parser only accepts `--input-manifest`, `--repeat-factor`, `--benchmark-results-path`, `--hf-token`, `--max-segment-length`, `--asr-batch-size`, and `--executor`, so argparse will exit with `unrecognized arguments: --cpus=10` and `is_success` can never be true.

Can we either add a real `--cpus` parser argument and thread it into executor/resource config, or remove `--cpus=10` from both `audio_tagging_tts` entries? Also, both entries use the same name, `audio_tagging_tts`; if both are intended, please give them unique names so benchmark reporting does not collide.

nemo_curator/stages/audio/inference/speaker_diarization/sortformer.py
mohammadaaftabv, line 1: This move removes the old public import path `nemo_curator.stages.audio.inference.sortformer` without a compatibility shim. Existing tutorials/docs and user code can still import `InferenceSortformerStage` from the old path, so this becomes a breaking change.

Please update all public docs/release examples need to be updated in the same PR.

nemo_curator/stages/audio/tagging/text/pnc.py
```diff
@@ -0,0 +1,370 @@
...
+
+        If ``segments`` are present, each segment with non-empty text produces
+        one prompt (key_path ``[segments_key, idx]``).  Otherwise a single
+        prompt is produced from the top-level text (key_path ``[]``).
+        """
+        items: list[tuple[str | list, list]] = []
+        if self.segments_key in data:
+            for idx, segment in enumerate(data[self.segments_key]):
+                if self.text_key in segment and segment[self.text_key] != "":
+                    prompt = self._vllm.get_entry_prompt(segment)
```
mohammadaaftabv: `get_entry_prompt()` returns `[]` when a prompt field is missing or the prompt/template cannot be formatted, but `_collect_prompts()` appends that value unconditionally. That means a malformed row still gets sent to vLLM as an empty prompt/list, even though the helper logged that it was “skipping”.

Can we skip falsy prompts here before appending to `items`? Ideally the stage should also mark the affected segment/entry with a clear prompt-formatting error or metric counter, so one bad row does not poison the whole vLLM batch.

nemo_curator/stages/audio/tagging/text/pnc.py
```diff
@@ -0,0 +1,370 @@
...
+
+    @staticmethod
+    def _store_generation(data: dict, key_path: list, text: str, gen_field: str) -> None:
+        """Write *text* into *data* following *key_path*."""
+        target = data
+        for key in key_path:
+            target = target[key]
+        target[gen_field] = text
+
+    def inputs(self) -> tuple[list[str], list[str]]:
```
mohammadaaftabv: This input contract is still too strict for the behavior below. The stage supports either segment-level text or top-level text, but `inputs()` currently requires both `segments_key` and `text_key` on every task. A valid top-level-only manifest can fail base validation for missing `segments`, and a valid segment-only manifest can fail for missing top-level `text`.

Can we make the scope explicit, for example `input_scope: "segments" | "entry"`, and declare inputs accordingly? If the stage intentionally supports both, it likely needs custom validation in `process_batch()` that accepts `segments_key` OR `text_key` rather than requiring both.

nemo_curator/stages/audio/metrics/squim.py
```diff
@@ -0,0 +1,222 @@
...
+            sorted_results = self._compute_metrics_batched(sorted_waveforms)
+            for rank, (pesq_val, stoi_val, sisdr_val) in enumerate(sorted_results):
+                orig_idx = sorted_indices[rank]
+                task_idx, seg_idx, _ = all_waveform_metadata[orig_idx]
+                if self.segments_key in tasks[task_idx].data:
+                    segment = tasks[task_idx].data[self.segments_key][seg_idx]
+                    self.update_metrics(segment, pesq_val, stoi_val, sisdr_val)
+                else:
+                    self.update_metrics(tasks[task_idx].data, pesq_val, stoi_val, sisdr_val)
+        except Exception as e:  # noqa: BLE001
```
mohammadaaftabv: This catches all SQUIM computation failures, logs an error, and then returns the tasks as if the stage succeeded. In the tagging tutorial/benchmark, SQUIM metrics are part of the promised quality output, so a CUDA/model/audio failure can silently produce a successful manifest with missing `pesq_squim`, `stoi_squim`, and `sisdr_squim`.

Can we fail loudly by default, or add an explicit policy such as `on_error: "raise" | "mark"`? If marking is preferred, please write a per-entry/segment error field and metrics counter so downstream filtering does not treat missing SQUIM values as valid quality scores.

nemo_curator/stages/audio/tagging/text/pnc.py
```diff
@@ -0,0 +1,370 @@
...
+        llm_chars = "".join(c for c in llm_no_pnc if c != " ")
+        digit_present = any(c.isdigit() for c in llm_no_pnc)
+
+        if asr_chars != llm_chars:
+            cer, _, _, _, _ = word_error_rate_detail(
+                hypotheses=[asr_no_pnc],
+                references=[llm_no_pnc],
+                use_cer=True,
+            )
+            if cer > self.cer_threshold or digit_present or not self.is_valid_text(llm_cleaned, self._full_vocab_set):
```
mohammadaaftabv: This still writes `use_bert_pnc=True`, but this PR no longer includes a BERT PnC fallback stage. The field name now describes a removed downstream implementation rather than the actual state of the output, and the README also tells users that this is for downstream BERT handling.

Can we rename this to an implementation-neutral field such as `pnc_validation_failed`, `pnc_rejected`, or `use_original_text`, and update tests/fixtures/docs accordingly? If BERT PnC is planned later, that should be introduced with the actual fallback stage rather than encoded into this PR’s schema.

nemo_curator/stages/audio/metrics/wer.py
```diff
@@ -0,0 +1,324 @@
...
+
+    def __post_init__(self) -> None:
+        if self.num_words_look_back >= self.num_words_threshold:
+            msg = (
+                f"num_words_look_back ({self.num_words_look_back}) must be less than "
+                f"num_words_threshold ({self.num_words_threshold})"
+            )
+            raise ValueError(msg)
+
+    def inputs(self) -> tuple[list[str], list[str]]:
```
mohammadaaftabv, line 80: The stage supports a top-level fallback in `process()` when `segments_key` is absent, but `inputs()` requires `segments_key`. With the default base validation path, a top-level-only task can be rejected before this fallback executes.

Can we either remove the top-level fallback and require segments everywhere, or make the input contract match the supported behavior? If both modes are intended, this likely needs explicit scope/config or custom validation that accepts either `segments` or the configured top-level hypothesis/reference keys.

mohammadaaftabv (changes requested):

sarahyurick (approved):
LGTM from a Curator design standpoint. I will let @mohammadaaftabv re-review for any audio-specific concerns.

nemo_curator/stages/audio/metrics/squim.py
```diff
@@ -0,0 +1,223 @@
...
+        """
+        if not tasks:
+            return tasks
+
+        # Collect all valid waveforms with their origin (task_idx, segment_idx)
+        all_waveform_metadata: list[tuple[int, int, torch.Tensor]] = []
+        for task_idx, task in enumerate(tasks):
+            all_waveform_metadata.extend(self._collect_waveforms_for_entry(task_idx, task.data))
+
+        if not all_waveform_metadata:
```
mohammadaaftabv, line 214: SQUIM can silently return unscored tasks when audio loading/collection fails.

nemo_curator/stages/audio/metrics/__init__.py
```diff
@@ -0,0 +1,25 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Audio quality and accuracy metrics stages."""
+
+from nemo_curator.stages.audio.metrics.bandwidth import BandwidthEstimationStage
+from nemo_curator.stages.audio.metrics.squim import TorchSquimQualityMetricsStage
+from nemo_curator.stages.audio.metrics.wer import ComputeWERStage
+
+__all__ = [
```
mohammadaaftabv: Audio metrics package eagerly imports optional-heavy dependencies. Are we sure we need it here?

nemo_curator/stages/audio/metrics/wer.py
```diff
@@ -0,0 +1,328 @@
...
+        return [], [self.segments_key, self.hypothesis_text_key, self.reference_text_key]
```
mohammadaaftabv: ComputeWER input contract rejects segment-level ASR output before WER runs.

Proposed fix:

Override process_batch() because this stage supports an OR-shaped input contract:

segment mode: task.data[segments_key][i][hypothesis_text_key]
entry mode: task.data[hypothesis_text_key]
inputs() cannot express that OR shape today.

Suggested implementation shape:

    # The real validation is shape-dependent and happens in process_batch().
    return [], []

def process_batch(self, tasks: list[AudioTask]) -> list[AudioTask]:
    for task in tasks:
        data = task.data

        if self.segments_key in data:
            segments = data[self.segments_key]
            if not isinstance(segments, list):
                raise ValueError(f"{self.segments_key} must be a list")

            for idx, segment in enumerate(segments):
                if self.hypothesis_text_key not in segment or self.reference_text_key not in segment:
                    raise ValueError(
                        f"Segment {idx} missing WER keys: "
                        f"{self.hypothesis_text_key}, {self.reference_text_key}"
                    )
                self.get_wer(segment)
        else:
            if self.hypothesis_text_key not in data or self.reference_text_key not in data:
                raise ValueError(
                    f"Task missing WER keys: "
                    f"{self.hypothesis_text_key}, {self.reference_text_key}"
                )
            self.get_wer(data)

    return tasks
A slightly softer version could skip individual segments missing one key, matching current get_wer() behavior, but for this tutorial I’d prefer fail-loud because WER is explicitly configured and missing text_2 means the second ASR pass did not produce the expected output.

nemo_curator/stages/audio/metrics/wer.py
```diff
@@ -0,0 +1,328 @@
...
+
+        metrics = audio_segment.get("metrics", {})
+
+        hypothesis_pnc, hypothesis_clean = self.normalize_and_clean_text(audio_segment[self.hypothesis_text_key])
+        reference_pnc, reference_clean = self.normalize_and_clean_text(audio_segment[self.reference_text_key])
+
+        metrics["char_rate"] = self.get_char_rate(audio_segment[self.hypothesis_text_key], duration)
+        metrics["word_rate"] = self.get_word_rate(audio_segment[self.hypothesis_text_key], duration)
+
+        wer_val, tokens, ins_rate, del_rate, sub_rate = word_error_rate_detail(
```
mohammadaaftabv, line 211: Empty normalized references can emit inf WER/CER and serialize as Infinity.
The issue is that reference_clean can become empty after normalization/cleaning, even if the original field exists. Examples:

""
"   "
"..."
"<unk>"
"⁇"
NeMo’s helper returns inf when the reference has zero tokens. I verified locally that [""] vs [""] and ["a"] vs [""] return infinite rates.

That can later be written to JSONL as Infinity, which is not valid strict JSON.

Proposed fix:

if not reference_clean:
    metrics["wer"] = None
    metrics["cer"] = None
    metrics["metric_skip_reason"] = "empty_reference"
    audio_segment["metrics"] = metrics
    return
If they want to preserve one-sided empty cases, use explicit finite semantics instead of raw inf.

nemo_curator/stages/audio/metrics/bandwidth.py
```diff
@@ -0,0 +1,131 @@
...
+    top_db: float = 100.0
+    frequency_threshold: float = -50.0
+    audio_filepath_key: str = "audio_filepath"
+    segments_key: str = "segments"
+
+    # Stage metadata
+    name: str = "BandwidthEstimation"
+
+    def inputs(self) -> tuple[list[str], list[str]]:
+        return [], [self.audio_filepath_key, self.segments_key, "duration"]
```
mohammadaaftabv: Bandwidth top-level fallback is unreachable under real backend process_batch() execution.

This is the same backend-contract issue as WER, but less severe.

BandwidthEstimationStage.process() supports two modes:

if self.segments_key in data_entry:
    score each segment
else:
    score the top-level entry
But inputs() says:

return [], [self.audio_filepath_key, self.segments_key, "duration"]
So under real backend execution, base process_batch() validates segments before calling process(). If segments is missing, the top-level fallback never runs.

The new test calls stage.process() directly, so it bypasses validation and passes.

Proposed fix:

def inputs(self):
    return [], [self.audio_filepath_key]
Then validate inside process():

if self.segments_key in data_entry:
    ...
else:
    if "duration" not in data_entry:
        raise ValueError("duration is required for top-level bandwidth scoring")
    self.get_bandwidth(data_entry, audio, sample_rate)

Or better yet modify https://github.com/NVIDIA-NeMo/Curator/blob/3b3e07f564c2ae9a1b1b9e21a14ece8aef8ace2b/nemo_curator/stages/base.py#L131
to support new type of input validation keeping backward compatibility. This arises because we are trying to support 2 modes of input data types for a single processor.

nemo_curator/stages/audio/metrics/__init__.py
mohammadaaftabv: I know some audio packages do eager re-exports already, but `audio/tagging/__init__.py` uses lazy imports specifically to avoid loading heavy optional deps. Should metrics follow that lazy/empty pattern as well, or is the intent that all audio metric dependencies are always installed together?

mohammadaaftabv (changes requested):

nemo_curator/stages/audio/metrics/bandwidth.py
sushmitha-deva-09: I added support for multi pair of inputs validation in base.py. 
@sarahyurick , could you also please check changes in nemo_curator/stages/base.py ?

tests/stages/audio/tagging/e2e/configs/asr_pipeline.yaml
```diff
@@ -0,0 +1,68 @@
...
+    text_key: ${text_key2}
+    infer_segment_only: True
+    compute_timestamps: False
+    resources:
+      gpus: 0.45
+
+  - _target_: nemo_curator.stages.audio.metrics.wer.ComputeWERStage
+    name: "ComputeWER"
+    language: ${language_short}
+    hypothesis_text_key: "text"
```
mohammadaaftabv: reverses the WER direction in the ASR e2e config.

ASRAlignment2 writes the second-pass transcript into text_2, but the test config sets hypothesis_text_key: "text" and reference_text_key: ${text_key2}. The tutorial pipeline does the opposite at tutorials/audio/tagging/asr_pipeline.yaml (line 166). WER direction matters because the reference side is the denominator. This means the e2e test can bless metrics that do not match the actual intended ASR pipeline behavior. Fix: make test config use hypothesis_text_key: ${text_key2} and reference_text_key: ${text_key1}, then regenerate/update the reference manifest.

nemo_curator/stages/audio/metrics/squim.py
```diff
@@ -0,0 +1,227 @@
...
+        audio_segment["metrics"]["pesq_squim"] = pesq_val
+        audio_segment["metrics"]["stoi_squim"] = stoi_val
+        audio_segment["metrics"]["sisdr_squim"] = sisdr_val
+
+    def process(self, task: AudioTask) -> AudioTask:
+        """Delegate single-task processing to process_batch."""
+        msg = f"[{self.name}] is a GPU/batched inference stage. Use process_batch() instead."
+        raise NotImplementedError(msg)
+
+    def process_batch(self, tasks: list[AudioTask]) -> list[AudioTask]:
```
mohammadaaftabv, line 198: silently returns unscored tasks when audio loading fails.

SquimEstimationStage.process_batch() bypasses the base stage validation path, _collect_audio_waveforms() returns [] on missing/read/load errors at squim.py (line 121), squim.py (line 132), and squim.py (line 139), then the batch is returned unchanged at squim.py (line 201). Since SQUIM is an explicit quality metric stage in the tagging pipelines, a successful pipeline can produce no SQUIM metrics without failing. Fix: raise on unreadable required audio, or write an explicit error/skip metric that downstream checks can detect.

nemo_curator/stages/audio/metrics/bandwidth.py
```diff
@@ -0,0 +1,131 @@
...
+        """Estimate bandwidth for audio entry."""
+        data_entry = task.data
+        audio_path = data_entry.get(self.audio_filepath_key)
+        if not audio_path:
+            logger.error(
+                f"[{self.name}] Missing '{self.audio_filepath_key}' for entry: "
+                f"{data_entry.get('audio_item_id', 'unknown')}"
+            )
+            return task
+        try:
```
mohammadaaftabv, line 130: silent-missing-metric behavior for unreadable audio.

If librosa.load() fails, the stage logs and returns the original task at bandwidth.py (line 123). The manifest key exists, so base validation passes, but output lacks bandwidth metrics. Fix similarly: fail the task/stage or mark a structured bandwidth_error/skip reason.

tutorials/audio/tagging/main.py
mohammadaaftabv: logs manifest dictionary keys as “output files”.

ManifestWriterStage returns tasks whose data is still the manifest entry, so output_files.extend(task.data) extends keys like audio_filepath, not written file paths. Fix: log cfg.final_manifest, or change the writer/result contract if the CLI wants actual output file paths.

mohammadaaftabv (changes requested):

nemo_curator/stages/audio/metrics/bandwidth.py
```diff
@@ -0,0 +1,131 @@
...
+    segments_key: str = "segments"
+
+    # Stage metadata
+    name: str = "BandwidthEstimation"
+
+    def inputs(self) -> tuple[list[str], list[list[str]]]:
+        return [], [[self.audio_filepath_key, self.segments_key], [self.audio_filepath_key, "duration"]]
+
+    def outputs(self) -> tuple[list[str], list[str]]:
+        return [], [[self.audio_filepath_key, self.segments_key], [self.audio_filepath_key, "metrics"]]
```
sarahyurick: Can you explain for these cases, when we can expect it to be `[self.audio_filepath_key, self.segments_key]` and when we can expect it to be `[self.audio_filepath_key, "duration"]`/`[self.audio_filepath_key, "metrics"]`? Like is it dependent on the input `AudioTask` or something else?

nemo_curator/stages/base.py
```diff
@@ -130,32 +130,60 @@ def num_workers(self) -> int | None:
 
     def validate_input(self, task: Task) -> bool:
         """Validate input task meets requirements.
+
+        ``inputs()`` returns ``(required_top_level_attrs, required_data_attrs)``.
+
+        *required_data_attrs* supports two shapes:
+
+        - **Flat list** ``["a", "b"]``: all keys must be present (AND semantics).
+        - **Nested list** ``[["segments"], ["text", "pred_text"]]``: at least one
```
sarahyurick: Maybe rather than editing it here, you can just override with the custom `validate_input` for the relevant stages?

nemo_curator/stages/audio/metrics/bandwidth.py
sushmitha-deva-09: We can reuse the same core logic across different AudioTask input types; only the dictionary iteration logic changes. So, I added support for tagging pipeline inputs that require "segments" within task.data, while still allowing the same stage to work with any generic task.data containing an "audio_filepath" field.

mohammadaaftabv (approved):

ayushdg (approved):

PR #1865, opened 2026-04-23
Add audio checkpointing foundation with stable sample keys and resumable stage runner

ssh-meister:
## Description

This PR adds the first checkpointing foundation for `AudioTask`-based pipelines in NeMo Curator.

The goal is to make audio pipeline execution resumable without reprocessing already completed samples, while keeping the existing `Pipeline` abstraction intact. Instead of changing executor internals, this introduces an audio-specific checkpointing layer that orchestrates stage-by-stage execution on top of the current pipeline model.

### Stable audio identity
`AudioTask` now has a stable top-level `sample_key`, plus shared helpers for audio sample identity:

- `build_audio_sample_key(...)`
- `ensure_sample_key(...)`
- `carry_sample_key(...)`
- `derive_child_sample_key(...)`

This separates runtime task identity (`task_id`) from stable sample identity and provides a consistent basis for checkpointing, resumability, durable artifact naming, and deterministic child identity in fan-out stages.

### Reader-side sample key generation
Stable `sample_key` generation is now supported when `AudioTask`s are first created by:

- `JsonlReader(task_type="audio")` / `JsonlAudioReaderStage`
- `TarredAudioManifestReaderStage`
- `ALMManifestReaderStage`

If an input manifest already provides `sample_key`, it is preserved. Otherwise a deterministic key is derived from stable audio identity fields.

### Propagation through additional audio stages
This PR also extends `sample_key` handling beyond the reader path:

- `InferenceSortformerStage` now preserves identity with `carry_sample_key(...)`
- `VADSegmentationStage` now derives deterministic child keys for segment fan-out
- `SpeakerSeparationStage` now derives deterministic child keys for per-speaker fan-out
- `SegmentConcatenationStage` now preserves identity with `carry_sample_key(...)`

### Durable tar materialization
`MaterializeTarredAudioStage` now supports:

- `materialization_dir: str | None = None`

When unset, the stage keeps the existing temporary-file behavior. When set, materialized audio is written to a deterministic durable path derived from `sample_key`.

This provides the storage primitive needed for restart-safe execution around tarred audio materialization.

### Safe handling of `pipe:` tar streams
For `pipe:`-based tar inputs, this PR adds a `SIGPIPE`-tolerant path in `_PipeStream` for tar materialization reads, so early-close behavior does not surface as a false failure when all required tar members have already been found.

### Audio checkpointing package
This PR introduces a new checkpointing package for audio pipelines:

- `nemo_curator.checkpointing.audio.serialization`
- `nemo_curator.checkpointing.audio.store`
- `nemo_curator.checkpointing.audio.runner`

The new `AudioCheckpointRunner` provides a first resumable execution path for `AudioTask` pipelines. It stores:

- pipeline metadata
- per-stage metadata
- per-sample records
- output manifests for reload/resume

It currently supports:

- `checkpoint_dir` as checkpoint root
- optional `materialization_dir`
- `ignore_failed`
- `link_stages_via_io` via `pipeline.config` for large intermediate outputs

### Stage linking modes
The checkpoint runner supports two linking modes between successful audio stages:

- default: keep `AudioTask`s in memory and also checkpoint them to disk
- `link_stages_via_io=True`: checkpoint to disk and reload via output manifests between stages

This supports both:
- faster in-memory chaining
- lower-memory / manifest-linked execution for large intermediate outputs

### Consistency and retry hardening
The checkpointing and tarred-audio paths were additionally hardened with:

- deterministic stage fingerprinting for set-valued configs
- normalized `pipeline.config` serialization in checkpoint metadata
- no in-place mutation of caller-owned stage objects during runner preparation
- forced `sample_key` materialization before checkpoint record accounting
- explicit assigned-ID range validation in the audio JSONL reader
- retry cleanup for temporary artifacts
- retry-safe tarred materialization behavior when tasks are partially mutated before retry
- shared atomic write helpers for checkpoint metadata and manifests

## Scope

This PR adds the checkpointing foundation and the initial resumable runner/store/serialization pieces.

It is intended for `AudioTask` pipelines, including the current tarred ASR-style flow:

- `TarredAudioManifestReader`
- `MaterializeTarredAudioStage`
- `InferenceAsrNemoStage`
- `GetPairwiseWerStage`
- `CleanupTemporaryAudioStage`

It also brings `sample_key` support to a broader set of audio stages that create or transform `AudioTask`s, including ALM, Sortformer, VAD, speaker separation, and segment concatenation.

`AudioToDocumentStage` and `JsonlWriter` are intentionally not checkpointed; they are treated as finalization stages that can be rerun from the last saved `AudioTask` manifest.

## Why this design

This PR keeps checkpointing outside the existing executor internals and places it in an audio-specific orchestration layer. That keeps:

- `Pipeline` as the declarative stage graph
- executors as execution backends
- checkpointing as resumable execution policy

This provides a practical first version without forcing checkpoint logic directly into `XennaExecutor` or `RayDataExecutor`.

## Validation

This PR adds focused tests for:

- `AudioTask` sample key behavior
- JSONL / tarred / ALM audio reader sample key generation
- sortformer / VAD / speaker separation / concatenation sample key handling
- durable materialization behavior
- `SIGPIPE` handling for `pipe:` materialization
- checkpoint serialization, store, and runner behavior
- deterministic config fingerprinting
- retry cleanup and manifest-linked stage chaining

Lint and syntax checks were also run on the touched files.


nemo_curator/checkpointing/audio/runner.py
```diff
@@ -0,0 +1,175 @@
...
+from nemo_curator.tasks.audio_task import ensure_checkpoint_shard_id, ensure_sample_key
+
+from .io_utils import normalize_for_json, write_json_atomic
+from .store import SampleCheckpointRecord, StageCheckpointStore, fingerprint_stage
+
+if TYPE_CHECKING:
+    from nemo_curator.backends.base import BaseExecutor
+    from nemo_curator.stages.base import ProcessingStage
+    from nemo_curator.tasks import AudioTask
+class AudioCheckpointRunner:
```
sarahyurick, line 36: I am not sure why the Ruff linter isn't catching it but I am noticing a few spacing errors throughout the PR:
```suggestion
if TYPE_CHECKING:
    from nemo_curator.backends.base import BaseExecutor
    from nemo_curator.stages.base import ProcessingStage
    from nemo_curator.tasks import AudioTask


class AudioCheckpointRunner:
```

nemo_curator/checkpointing/audio/store.py
```diff
@@ -0,0 +1,202 @@
...
+
+if TYPE_CHECKING:
+    from nemo_curator.tasks import AudioTask
+
+RecordStatus = Literal["done", "filtered", "failed_retriable", "failed_permanent"]
+
+
+def _utcnow() -> str:
+    return datetime.now(tz=timezone.utc).isoformat()
+def fingerprint_stage(stage: Any) -> str:  # noqa: ANN401
```
sarahyurick, line 47: Here too, I thought the linter would catch something like this:
```suggestion
def _utcnow() -> str:
    return datetime.now(tz=timezone.utc).isoformat()


def fingerprint_stage(stage: Any) -> str:  # noqa: ANN401
```

tests/checkpointing/audio/__init__.py
sarahyurick, line 1: No copyright needed for an empty file.

tests/checkpointing/__init__.py
sarahyurick, line 1: No copyright needed for an empty file.

sarahyurick (commented):
Hi, I just left some minor comments about formatting for now. Can do a more in-depth review later, thanks.

nemo_curator/checkpointing/audio/runner.py
mohammadaaftabv: +1

nemo_curator/stages/audio/io/tarred.py
```diff
@@ -525,7 +529,10 @@ def _materialize_task(
     def _should_segment(self, task: AudioTask, member_name: str) -> bool:
         if not self.segment_if_offset_present:
             return False
-        original_path = task.data.get(self.audio_filepath_key, "")
+        original_path = task.data.get(
```
mohammadaaftabv: The fix to read manifest_audio_filepath_key before audio_filepath_key is correct — this addresses the retry-mutation bug where audio_filepath is overwritten to a temp path during batch processing, causing _should_segment to compare against the wrong value on retry. The new test_materialize_retry_keeps_member_mode_after_audio_filepath_mutation test covers this path well.

However, the fallback chain task.data.get(self.manifest_audio_filepath_key, task.data.get(self.audio_filepath_key, "")) has three distinct cases:

_manifest_audio_filepath exists (tarred pipeline, normal case)
_manifest_audio_filepath missing, audio_filepath exists (non-tarred pipeline)
Neither exists (edge case, returns "")
The test only covers case 1 (the retry-mutation scenario). Consider adding a comment or assertion documenting that case 3 (empty string) results in member_name != "" which is True, meaning _should_segment defaults to True — which may or may not be the intended behavior for unknown provenance tasks.

nemo_curator/stages/audio/segmentation/vad_segmentation.py
```diff
@@ -274,6 +275,16 @@ def process(self, task: AudioTask) -> AudioTask | list[AudioTask]:
                     data=seg_data,
                     task_id=f"{task.task_id}_seg_{i}",
                     dataset_name=task.dataset_name,
+                    sample_key=derive_child_sample_key(
+                        task,
+                        child_kind="vad_segment",
+                        child_identity={
```
mohammadaaftabv, line 281: These are runtime float values from Silero VAD that could vary slightly across runs (e.g., due to non-deterministic GPU kernel scheduling, float rounding, or model version differences). If the same audio is re-processed, the child keys may differ, which defeats durable artifact naming.

The segment_index alone would provide deterministic child identity within a parent. If the intent is to also detect changes in VAD output between runs (i.e., invalidate checkpoint when VAD boundaries shift), this is fine — but that intent should be documented.

For comparison, SpeakerSeparationStage uses {"speaker_id": speaker_id} which is stable.

Suggested approach: Use segment_index as the primary identity, and include start_ms/end_ms only if checkpoint invalidation on VAD output change is intentional:

child_identity={"segment_index": i},

nemo_curator/checkpointing/audio/store.py
```diff
@@ -0,0 +1,202 @@
...
+            ensure_sample_key(task)
+            ensure_checkpoint_shard_id(task)
+        if input_tasks is not None:
+            for task in input_tasks:
+                ensure_sample_key(task)
+                ensure_checkpoint_shard_id(task)
+
+        failed_records = failed_records or []
+        failed_by_key = {record.sample_key: record for record in failed_records}
+        output_by_key = {task.sample_key: task for task in output_tasks}
```
mohammadaaftabv, line 146: This builds a dict keyed by sample_key. In fan-out stages (VAD, SpeakerSeparation), multiple output tasks will have different sample_keys (thanks to derive_child_sample_key), so this is fine for fan-out.

However, for 1:1 stages, if two input tasks somehow produce the same sample_key (e.g., due to identical manifest entries without unique audio_filepath), the dict comprehension silently drops duplicates. The ensure_sample_key call above derives keys from data + dataset_name, so true duplicates would mean identical data — which is likely a data issue, not a code issue. But a warning or assertion here would make debugging easier:

if len(output_by_key) != len(output_tasks):
    logger.warning(f"Duplicate sample_keys detected in {len(output_tasks)} output tasks")
The Greptile P2 about empty sample_key collisions is resolved by the ensure_sample_key(task) call at line 121 which materializes the key before it reaches output_by_key.

tests/stages/text/io/reader/test_jsonl.py
```diff
@@ -270,6 +270,29 @@ def test_audio_stage_generates_and_assigns_stable_ids(self, tmp_path: Path) -> N
...
+
+        generation_stage = JsonlAudioReaderStage(_generate_ids=True)
+        generation_stage.setup()
+        generation_stage.process(task)
+
+        assign_stage = JsonlAudioReaderStage(_assign_ids=True, fields=["audio_filepath"])
+        assign_stage.setup()
+
+        with pytest.raises(RuntimeError, match=r"Assigned-ID range \[0, 1\] \(2 ids\) does not match 1 tasks"):
+            assign_stage.process(task)
```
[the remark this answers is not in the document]
mohammadaaftabv: +1
This test is the sole CI failure in this PR. The Unit_Test_stages-text_CPU job fails with:

FAILED tests/stages/text/io/reader/test_jsonl.py::TestJsonlAudioReader::test_audio_stage_assign_ids_raises_on_task_count_mismatch
- Failed: DID NOT RAISE <class 'RuntimeError'>
The test creates a manifest with 2 entries (both have audio_filepath):

json.dumps({"audio_filepath": "a.wav", "text": "alpha"}),
json.dumps({"audio_filepath": "b.wav", "text": "beta"}),
Step 1: generation_stage.process(task) reads 2 entries and reserves IDs [0, 1] in the ID generator.

Step 2: assign_stage = JsonlAudioReaderStage(_assign_ids=True, fields=["audio_filepath"]) — the fields parameter controls which keys to keep per entry (line 192: entry = {field: entry[field] for field in self.fields if field in entry}), not which entries to include. Since both entries have audio_filepath, both entries produce an AudioTask.

Result: _apply_assigned_ids receives range [0, 1] (2 IDs) and 2 tasks. 2 != 2 is False. No mismatch. RuntimeError is never raised.

Fix: To trigger the mismatch, the test needs the assign stage to produce a different number of tasks than the generation stage. For example, one manifest entry should lack a required field so fields produces an empty dict (though the current code still creates a task for empty dicts), or a separate manifest with fewer entries should be used for the assign step:

# Option A: Use a 1-entry manifest for the assign step
assign_manifest = tmp_path / "audio_ids_assign.jsonl"
assign_manifest.write_text(json.dumps({"audio_filepath": "a.wav", "text": "alpha"}))
assign_task = FileGroupTask(task_id="audio_task", dataset_name="audio_dataset", data=[str(assign_manifest)], _metadata={})
assign_stage.process(assign_task)

nemo_curator/checkpointing/audio/runner.py
```diff
@@ -0,0 +1,175 @@
...
+                raise
+
+            current_tasks = store.load_output_tasks() if self._link_stages_via_io() else output_tasks
+
+        if tail_stages:
+            tail_pipeline = Pipeline(name=f"{self.pipeline.name}_finalize", stages=tail_stages)
+            return tail_pipeline.run(executor=self.executor, initial_tasks=current_tasks)
+        return current_tasks
+
+    def _prepare_stages(self, stages: list[ProcessingStage]) -> list[ProcessingStage]:
```
mohammadaaftabv: def _prepare_stages(self, stages: list[ProcessingStage]) -> list[ProcessingStage]:
    return list(stages)
This method is called once (in run()) and only returns a shallow copy of the input list. No subclass of AudioCheckpointRunner exists in this PR, and the list copy serves no functional purpose since the runner doesn't mutate the stage list. If this is intended as a subclass extension point (e.g., for injecting ensure_sample_key stages or configuring materialization dirs before execution), a docstring should say so. Otherwise consider inlining the list(stages) call at the call site.

nemo_curator/checkpointing/audio/runner.py
```diff
@@ -0,0 +1,175 @@
...
+            if not self.ignore_failed:
+                raise
+
+            self._cleanup_retry_artifacts(input_tasks)
+            logger.warning(
+                f"Stage {stage._name} failed for a batch of {len(input_tasks)} tasks, retrying one-by-one: {error}"
+            )
+            outputs: list[AudioTask] = []
+            failed_records: list[SampleCheckpointRecord] = []
+            for task in retry_tasks or input_tasks:
```
mohammadaaftabv: retry_tasks = deepcopy(input_tasks) if self.ignore_failed else None
...
for task in retry_tasks or input_tasks:
When we reach the except block, self.ignore_failed is necessarily True (the if not self.ignore_failed: raise guard above ensures this). Therefore retry_tasks is always a deepcopy(input_tasks) and never None at this point. The or input_tasks fallback is dead code.

This was also flagged by Greptile's initial review. Simplifying to just for task in retry_tasks: (with appropriate typing) would be cleaner.

mohammadaaftabv (changes requested):

nemo_curator/checkpointing/audio/runner.py
ssh-meister: Thank you, fixed

tests/checkpointing/audio/__init__.py
ssh-meister: Thank you, fixed

tests/checkpointing/__init__.py
ssh-meister: Thank you, fixed

nemo_curator/checkpointing/audio/store.py
ssh-meister: Thank you, fixed

nemo_curator/stages/audio/segmentation/vad_segmentation.py
ssh-meister: Left as is intentionally. In derive_child_sample_key() for VAD, we still include segment_index, start_ms, end_ms, and duration, because this preserves the semantics: if VAD boundaries change, the child identity changes as well. So this is not a bug, but a deliberate checkpointing/invalidation design choice.

nemo_curator/checkpointing/audio/runner.py
ssh-meister: Thank you!
Fixed and even improved:
	•	removed the dead fallback or input_tasks
	•	retries now operate only on retry_tasks
	•	additionally, deepcopy(input_tasks) is now done before the batch attempt, so retries actually start from a clean snapshot rather than already mutated task objects

nemo_curator/checkpointing/audio/runner.py
ssh-meister: Fixed: the method has been removed because it provided no functional value. In run(), we now simply do stages = list(self.pipeline.stages).

nemo_curator/checkpointing/audio/store.py
ssh-meister: Fixed: added logger.warning(...) if len(output_by_key) != len(output_tasks) so that potential duplicates are no longer silently collapsed.

nemo_curator/stages/audio/io/tarred.py
ssh-meister: Fixed logically and documented: the logic remains correct for the retry case, and a comment has been added next to the fallback chain explaining that:
	•	first _manifest_audio_filepath is used
	•	then audio_filepath
	•	and an empty string as the final fallback intentionally makes the behavior conservative for unknown provenance tasks

nemo_curator/checkpointing/audio/runner.py
```diff
@@ -0,0 +1,172 @@
...
+            )
+            retry_tasks = deepcopy(input_tasks)
+            outputs: list[AudioTask] = []
+            failed_records: list[SampleCheckpointRecord] = []
+            for task in retry_tasks:
+                stage_outputs, failed_record = self._run_single_task_with_retry(stage, task)
+                outputs.extend(stage_outputs)
+                if failed_record is not None:
+                    failed_records.append(failed_record)
+            return outputs, failed_records
```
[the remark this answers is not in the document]
ssh-meister: fixed [febcbe4](https://github.com/NVIDIA-NeMo/Curator/pull/1865/commits/febcbe43deb9e6545376180268cb1e19f863b409)

nemo_curator/checkpointing/audio/runner.py
```diff
@@ -0,0 +1,175 @@
...
+        self.ignore_failed = ignore_failed
+
+    def run(self) -> list[Any] | None:
+        self.pipeline.build()
+        stages = list(self.pipeline.stages)
+        self._write_pipeline_metadata(stages)
+        audio_stages, tail_stages = self._split_stages(stages)
+
+        current_tasks: list[AudioTask] | None = None
+        for stage_index, stage in enumerate(audio_stages):
```
sarahyurick, line 59: Maybe I am misinterpreting but it seems this loop is just individually calling `_run_single_stage` per stage right? Isn't that very anti-Curator since the whole point of Curator is to gain performance by running pipelines consisting of multiple stages, which takes care of resource allocation and balancing, etc.?

nemo_curator/stages/audio/alm/alm_manifest_reader.py
```diff
@@ -0,0 +1,122 @@
...
+                        count += 1
+            logger.info(f"ALMManifestReaderStage: loaded {count} entries from {manifest}")
+        return results
+
+    def ray_stage_spec(self) -> dict[str, Any]:
+        return {"is_fanout_stage": True}
+
+
+@dataclass
+class ALMManifestReader(CompositeStage[_EmptyTask, AudioTask]):
```
sarahyurick, line 76: I thought we already had a `ManifestReader` is there a reason we can't use that?

tests/checkpointing/__init__.py
sarahyurick, line 1: No copyright needed for an empty file.

sarahyurick (changes requested):
Hi @ssh-meister I am a little confused about why this PR exists. Can you provide more context about why this was created and why it is audio-specific?

I am not sure if it is supposed to be similar and/or supplemental to Onur's work in https://github.com/NVIDIA-NeMo/Curator/pull/1798 .

PR #1880, opened 2026-04-28
fix(filters): escape `-` in regex_url character class (#1601)

### LINKED ISSUE #1601: Bug: Overly permissive url regex

SAY-5:
Closes #1601.

## Summary
- The `regex_url` pattern in `nemo_curator/stages/text/utils/constants.py` had the unescaped class `[$-_@.&+]`. Python's `re` engine reads `$-_` as a **range** from U+0024 (`$`) through U+005F (`_`), which silently includes `<`, `>`, `;`, `:`, `=`, etc.
- Result: `regex_url.findall("see http://x.com<bad> end")` previously returned `['http://x.com<bad>']` instead of `['http://x.com']`. The pattern is used by `UrlsFilter` and `RepeatingTopNGramsFilter` (called transitively via `repeating_n_gram` paths), so this bleed inflates the `urls_ratio` score on documents that interleave URLs with HTML or punctuation.
- Fix: escape the `-` (`[$\-_@.&+]`) so it is treated as a literal character. The class now contains exactly the set of URL-safe characters the original author intended.

## Behavior change
Characters that the buggy range *accidentally* admitted (`<`, `>`, `;`, `:`, `=`, ASCII 0x25, 0x5F that weren't otherwise listed) are no longer captured. Characters the original code *intentionally* listed (`$`, `_`, `@`, `.`, `&`, `+`, `-`, `!`, `*`, `(`, `)`, `,`, percent-encoded escapes) all still match.

## Test plan
- [x] Added `tests/stages/text/utils/test_regex_url.py` with four regression cases:
  - `<bad>` after a URL no longer extends the match,
  - `;next` after a URL no longer extends the match,
  - all explicitly-allowed characters still match,
  - both `http://` and `https://` continue to match.
- [x] Verified the new tests fail against the unfixed regex and pass against the fixed one (standalone Python `re` reproduction; full pytest run requires the Ray/cosmos_xenna stack which CI will exercise).

## Authorship
Signed off per DCO. The patch is one line; the regression tests document why.


nemo_curator/stages/text/utils/constants.py
```diff
@@ -72,6 +72,11 @@
 regex_alpha = regex.compile("[[:alpha:]]")
 regex_digit = regex.compile("[[:digit:]]")
 regex_alphanum = re.compile("[a-zA-Z0-9\n?!,.]")
-regex_url = re.compile("http[s]?://(?:[a-zA-Z]|[0-9]|[$-_@.&+]|[!*\\(\\),]|(?:%[0-9a-fA-F][0-9a-fA-F]))+")
+# NOTE: the `-` inside the character class is escaped on purpose. With
+# `[$-_…]` the regex engine treats `$-_` as a *range* spanning U+0024
+# through U+005F, which silently includes `<`, `>`, `;`, `:`, etc., so
+# matches bleed past the actual URL into surrounding HTML/punctuation.
+# See https://github.com/NVIDIA-NeMo/Curator/issues/1601.
+regex_url = re.compile(r"http[s]?://(?:[a-zA-Z]|[0-9]|[$\-_@.&+]|[!*\(\),]|(?:%[0-9a-fA-F][0-9a-fA-F]))+")
```
sarahyurick: Should it be this?
```suggestion
regex_url = re.compile("http[s]?://(?:[a-zA-Z]|[0-9]|[$\\-_@.&+]|[!*\\(\\),]|(?:%[0-9a-fA-F][0-9a-fA-F]))+")
```

tests/stages/text/utils/test_regex_url.py
sarahyurick, line 1: Please add the parameters as described here: https://github.com/NVIDIA-NeMo/Curator/issues/1601#issuecomment-4070198012

Then tests can be added to the existing pytest file. I do not think we need to add a separate test file here.

sarahyurick (changes requested):
Hi @SAY-5 thanks for opening! Left some requests.

nemo_curator/stages/text/utils/constants.py
```diff
@@ -72,6 +72,11 @@
 regex_alpha = regex.compile("[[:alpha:]]")
 regex_digit = regex.compile("[[:digit:]]")
 regex_alphanum = re.compile("[a-zA-Z0-9\n?!,.]")
-regex_url = re.compile("http[s]?://(?:[a-zA-Z]|[0-9]|[$-_@.&+]|[!*\\(\\),]|(?:%[0-9a-fA-F][0-9a-fA-F]))+")
+# NOTE: the `-` inside the character class is escaped on purpose. With
+# `[$-_…]` the regex engine treats `$-_` as a *range* spanning U+0024
+# through U+005F, which silently includes `<`, `>`, `;`, `:`, etc., so
+# matches bleed past the actual URL into surrounding HTML/punctuation.
+# See https://github.com/NVIDIA-NeMo/Curator/issues/1601.
```
sarahyurick: Remove link please.

tests/stages/text/modules/test_filters.py
```diff
@@ -555,6 +555,59 @@ def test_urls(self) -> None:
         )
         assert all_equal(expected_data, filtered_data), f"Expected {expected_data} but got {filtered_data}"
 
+    def test_url_regex_does_not_swallow_html_tags(self) -> None:
+        # Regression for #1601: previously the `[$-_…]` range matched `<`,
+        # `>`, `;`, `:`, etc., so a URL match bled past the URL into the
+        # surrounding HTML/punctuation.
+        from nemo_curator.stages.text.utils.constants import regex_url
```
sarahyurick: Can you move this import to the top of the file instead?

tests/stages/text/modules/test_filters.py
```diff
@@ -555,6 +555,59 @@ def test_urls(self) -> None:
...
+
+        assert regex_url.findall("see http://x.com<bad> for details") == ["http://x.com"]
+        assert regex_url.findall("click http://example.com;next") == ["http://example.com"]
+
+    def test_url_regex_still_matches_allowed_characters(self) -> None:
+        # The fix must not regress on characters the original character
+        # class intended to allow: letters, digits, `$`, `_`, `@`, `.`,
+        # `&`, `+`, `-`, `!`, `*`, `(`, `)`, `,`, and percent-encoded escapes.
+        from nemo_curator.stages.text.utils.constants import regex_url
```
sarahyurick: Same comment as above.

tests/stages/text/modules/test_filters.py
```diff
@@ -555,6 +555,59 @@ def test_urls(self) -> None:
...
+            data=pd.DataFrame({"text": ["no urls here!", "https://www.nvidia.com/en-us/"]}),
+            task_id="batch_1_urls_ratio",
+            dataset_name="test_1",
+        )
+        assert all_equal(expected_data, filtered_data), f"Expected {expected_data} but got {filtered_data}"
+
+    def test_urls_filter_accepts_compiled_pattern(self) -> None:
+        # The custom regex argument should also accept a pre-compiled
+        # `re.Pattern` instance, not just a string.
+        import re
```
sarahyurick: Same comment as above.

sarahyurick (commented):
This is great, thank you @SAY-5 ! I left some tiny comments.

sarahyurick (approved):

SAY-5:
Amended 455cbfa with the missing `Signed-off-by` trailer to clear DCO.

nemo_curator/stages/text/utils/constants.py
```diff
@@ -72,6 +72,10 @@
 regex_alpha = regex.compile("[[:alpha:]]")
 regex_digit = regex.compile("[[:digit:]]")
 regex_alphanum = re.compile("[a-zA-Z0-9\n?!,.]")
-regex_url = re.compile("http[s]?://(?:[a-zA-Z]|[0-9]|[$-_@.&+]|[!*\\(\\),]|(?:%[0-9a-fA-F][0-9a-fA-F]))+")
+# NOTE: the `-` inside the character class is escaped on purpose. With
+# `[$-_…]` the regex engine treats `$-_` as a *range* spanning U+0024
+# through U+005F, which silently includes `<`, `>`, `;`, `:`, etc., so
+# matches bleed past the actual URL into surrounding HTML/punctuation.
+regex_url = re.compile(r"http[s]?://(?:[a-zA-Z]|[0-9]|[$\-_@.&+]|[!*\(\),]|(?:%[0-9a-fA-F][0-9a-fA-F]))+")
```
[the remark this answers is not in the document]
sarahyurick: @SAY-5 WDYT?

nemo_curator/stages/text/utils/constants.py
```diff
@@ -72,6 +72,12 @@
...
 regex_digit = regex.compile("[[:digit:]]")
 regex_alphanum = re.compile("[a-zA-Z0-9\n?!,.]")
-regex_url = re.compile("http[s]?://(?:[a-zA-Z]|[0-9]|[$-_@.&+]|[!*\\(\\),]|(?:%[0-9a-fA-F][0-9a-fA-F]))+")
+# NOTE: the `-` inside the character class is escaped on purpose. With
+# `[$-_…]` the regex engine treats `$-_` as a *range* spanning U+0024
+# through U+005F, which silently includes `<`, `>`, `;`, `:`, etc., so
+# matches bleed past the actual URL into surrounding HTML/punctuation.
+# `/` is listed explicitly because it was previously included only as a
+# side effect of that broken range and is required for URL path segments.
+regex_url = re.compile(r"http[s]?://(?:[a-zA-Z]|[0-9]|[$\-_@.&+/]|[!*\(\),]|(?:%[0-9a-fA-F][0-9a-fA-F]))+")
```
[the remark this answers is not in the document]
sarahyurick: How is this: `regex_url = re.compile(r"https?://(?:[A-Za-z0-9$\-_@.&+/:=?#~]|[!*\(\),]|(?:%[0-9A-Fa-f]{2}))+")` ?

PR #1888, opened 2026-04-28
Lazy import for text module

oyilmaz-nvidia:
### Problem

Cold-importing any symbol from `nemo_curator.stages.text.classifiers` took **26.4 s**
on GB200 (aarch64) vs 4.58 s on x86_64 H100. The bottleneck was eager top-level loading
of `torch`, `transformers`, `numpy`, and `huggingface_hub` across all classifier and model
modules, combined with slower shared-library initialization on aarch64.

#### `classifiers/__init__.py` — lazy package loader (biggest impact)

Replaced seven eager `from .xxx import Yyy` statements with a PEP 562 `__getattr__`
dispatcher backed by a `_LAZY` name→module dict. No classifier module is imported until
its name is first accessed; importing the package itself is now effectively free.

#### `classifiers/base.py`, `aegis.py`, `prompt_task_complexity.py` — factory pattern for `nn.Module` subclasses

Classes that inherit from `torch.nn.Module` require `torch` at class-definition time.
Each affected file now wraps its model class(es) in an `@lru_cache(maxsize=1)` factory
function (`_make_deberta`, `_make_aegis_model`, `_make_instruction_data_guard_net`,
`_make_prompt_task_models`). The class is defined only on the first `_setup()` call;
subsequent calls return the cached class object at zero cost.

#### `classifiers/fineweb_edu.py`, `models/model.py`, `models/tokenizer.py` — inline deferred imports

For `ProcessingStage` subclasses, heavy imports are moved into the method that first needs
them. Python caches every import in `sys.modules`, so repeated calls pay only a dict lookup.

| Method | Deferred import |
|---|---|
| `setup_on_node()` | `huggingface_hub.snapshot_download` |
| `_setup()` / `setup()` | `AutoTokenizer`, `AutoModel`, `AutoConfig` |
| `process()` / `yield_next_batch()` / `collect_outputs()` | `torch`, `numpy` |
| `teardown()` | `torch.cuda.empty_cache` |

Type-only annotations (`pd.DataFrame`, `np.ndarray`, `torch.Tensor`, `NodeInfo`,
`WorkerMetadata`, `Generator`) are moved into `if TYPE_CHECKING:` blocks, safe because
all files now carry `from __future__ import annotations`.

#### `tests/stages/text/models/` — update mock patch targets

Patch paths pointed at the old module-level attribute locations (e.g.
`nemo_curator.stages.text.models.model.snapshot_download`). After deferral those
attributes no longer exist at module scope, so patches are updated to their canonical
source locations (`huggingface_hub.snapshot_download`, `transformers.AutoTokenizer`,
`transformers.AutoConfig`, `torch.cuda.empty_cache`, `gc.collect`).

### Verification

```shell
ruff format  # 0 reformats
ruff check   # 0 errors
pytest tests/stages/text/ -m 'not gpu'  # 567 passed, 1 skipped
pytest tests/stages/text/models/        # 21 passed

### CODE REVIEW

File: nemo_curator/stages/text/classifiers/aegis.py [hunk 1]
```diff
@@ -47,102 +51,134 @@
...
-        add_instruction_data_guard: bool = False,
-    ):
-        super().__init__()
+# ---------------------------------------------------------------------------
+# Lazy factories — define nn.Module subclasses on first call so that torch is
+# never imported at module-parse time.
+# ---------------------------------------------------------------------------
+
+
+@lru_cache(maxsize=1)
```
VibhuJawa: Unsure  why are using `lru_cache` here ?

nemo_curator/stages/text/classifiers/__init__.py
```diff
@@ -15,26 +15,31 @@
...
-    "FineWebEduClassifier",
-    "FineWebMixtralEduClassifier",
-    "FineWebNemotronEduClassifier",
-    "InstructionDataGuardClassifier",
-    "MultilingualDomainClassifier",
-    "PromptTaskComplexityClassifier",
-    "QualityClassifier",
-]
+# Lazy-load every classifier so that importing this package does not pull in
+# torch / transformers / numpy at module-parse time.  Each name is resolved
```
VibhuJawa, line 20: Did not we inline  `torch / transformers / numpy` , why did we add both ?

nemo_curator/stages/text/classifiers/aegis.py
```diff
@@ -47,102 +51,134 @@
 HIDDEN_TEXT_FIELD = "_curator_hidden_text"
 MAX_SEQ_LENGTH = 4096
 TOKENIZER_PADDING_SIDE = "left"
-TORCH_DTYPE = torch.bfloat16
```
VibhuJawa, line 50: We added constants in one place for code alignment . We can probably do a  string  here ? Will pass in string here work too ?

nemo_curator/stages/text/classifiers/__init__.py
sarahyurick: +1 shouldn't the changes here be enough? I think it is more straightforward to lazy load here and leave the `aegis.py`, etc. scripts untouched.

nemo_curator/stages/text/classifiers/aegis.py
sarahyurick: +1

sarahyurick (changes requested):

nemo_curator/stages/text/classifiers/__init__.py
oyilmaz-nvidia: Ok, makes sense. WIll only update this file.

sarahyurick (approved):
Thank you!

tests/stages/text/classifiers/test_classifiers_init.py
```diff
@@ -0,0 +1,173 @@
+# Copyright (c) 2025, NVIDIA CORPORATION.  All rights reserved.
```
sarahyurick, line 1:
```suggestion
# Copyright (c) 2026, NVIDIA CORPORATION.  All rights reserved.
```

PR #1890, opened 2026-04-29
Lazy init for openai client

oyilmaz-nvidia:
## Summary

- `OpenAIClient.query_model()` and `AsyncOpenAIClient._query_model_impl()` now lazy-initialize `self.client` via `setup()` if it hasn't been called yet
- Eliminates a confusing `AttributeError: object has no attribute 'client'` that gave no actionable hint to users who skipped the `setup()` lifecycle step
- Code that already calls `setup()` explicitly is unaffected; the Ray executor pattern (which calls `setup()` on each worker) continues to work unchanged


VibhuJawa (approved):
LGTM

PR #1895, opened 2026-04-29
Bump Ray 2.55.1 / CUDA 12.9.1 / torch 2.10 / vllm 0.18; HAProxy ingress

praateekmahajan:
## Summary

- Bump CUDA base image to 12.9.1 and align `torch` / `torchaudio` / `torchvision` to the 2.10 line (cu129 wheels). Move the pytorch wheel index to `cu129`.
- Bump Ray to 2.55.1 (`[default,data]` and `[serve,llm]` extras).
- Cap `vllm<0.19` to keep the stack on the `transformers<5` / `huggingface-hub<1.0` path. Resolves to vLLM 0.18.1.
- Cap `torchcodec<0.11` to match torch 2.10's ABI (torchcodec 0.11 is built for torch 2.11). Resolves to torchcodec 0.10.0+cu129.
- Enable Ray Serve's HAProxy ingress:
  - Install `haproxy` (built from source) + `socat` in the Curator image and set `RAY_SERVE_HAPROXY_BINARY_PATH`.
  - `init_cluster` opportunistically opts in via `RAY_SERVE_ENABLE_HA_PROXY=1` when both binaries resolve on `PATH`. Both are required: Ray Serve uses `socat` to drive HAProxy's admin socket — without it, the controller's healthcheck silently returns `False` and trips a 5s timeout.
  - Also pins a free `RAY_SERVE_HAPROXY_METRICS_PORT` so multiple Curator clusters on a single host don't collide on HAProxy's prometheus bind.
- Extend `BaseModelConfig.merge_runtime_envs` to handle Ray's dict-form `pip` / `uv` runtime_env entries (previously list+list only). Backwards compatible — existing list-form call sites continue to produce list-form output.
- Anchor the Dynamo backend's runtime dir under Ray's session dir (`<ray_temp_dir>/<session_name>/nemo_curator_dynamo_<short_id>/`) so subprocess logs / manifests sit alongside Ray's own session logs. Pin `FLASHINFER_WORKSPACE_BASE` per-run so cubin artifacts compiled by a prior actor venv don't leak into a fresh session.

## Test plan

- [ ] Image builds end-to-end via `bash benchmarking/tools/build_docker.sh`.
- [ ] Existing unit tests pass (Dynamo runtime_env merging, backend start/stop, BaseModelConfig).
- [ ] HAProxy probe path: `init_cluster` emits the env vars when both binaries are on `PATH`, falls back silently otherwise.
- [ ] Full nightly benchmark on this branch (4 GPUs, all entries).

## Note on vLLM 0.19

We also explored bumping vLLM to 0.19.1, which transitively requires `transformers>=5` and `huggingface-hub>=1.0` (and follow-on Curator-side fixes for the boundary changes those introduce). We backed off because **vLLM 0.19.1 isn't compatible with Ray 2.55.1** — `ray.llm._infer_supports_vision` crashes during `LLMServer.start()` for the model configs we hit. Sticking with vLLM 0.18.1 keeps the upgrade scope minimal until the upstream Ray ↔ vLLM 0.19 path stabilizes.

### DIFF
--- a/docker/common/install_haproxy.sh
+++ b/docker/common/install_haproxy.sh
@@ -1,0 +1,65 @@
+#!/bin/bash
+# Copyright (c) 2026, NVIDIA CORPORATION.  All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -xeuo pipefail
+
+# Build HAProxy from source so Ray Serve's HAProxy ingress mode (RAY_SERVE_ENABLE_HA_PROXY=1)
+# has the binary on $PATH. Mirrors ray-project/ray docker/base-slim/Dockerfile (Ray 2.55+).
+#
+# Fetched from ray-project/haproxy-release (a GitHub release mirror) because
+# www.haproxy.org's wildcard TLS cert expired 2026-04-17 and the release tarball
+# disappeared from the upstream download site. Switch back to www.haproxy.org once
+# the cert is renewed and the tarball is republished upstream.
+HAPROXY_VERSION=2.8.20
+HAPROXY_SHA256=c8301de11dabfbf049db07080e43b9570a63f99e41d4b0754760656bf7ea00b7
+
+for i in "$@"; do
+    case $i in
+        --HAPROXY_VERSION=?*) HAPROXY_VERSION="${i#*=}";;
+        --HAPROXY_SHA256=?*) HAPROXY_SHA256="${i#*=}";;
+        *) ;;
+    esac
+done
+
+export DEBIAN_FRONTEND=noninteractive
+
+apt-get update
+apt-get install -y --no-install-recommends \
+    build-essential \
+    ca-certificates \
+    curl \
+    libc6-dev \
+    liblua5.3-dev \
+    libpcre3-dev \
+    libssl-dev \
+    zlib1g-dev \
+    liblua5.3-0 \
+    socat
+rm -rf /var/lib/apt/lists/*
+
+BUILD_DIR=$(mktemp -d)
+curl --retry 5 --retry-all-errors --connect-timeout 20 --max-time 300 \
+     -fsSL -o "${BUILD_DIR}/haproxy.tar.gz" \
+     "https://github.com/ray-project/haproxy-release/releases/download/${HAPROXY_VERSION}/haproxy-${HAPROXY_VERSION}.tar.gz"
+echo "${HAPROXY_SHA256}  ${BUILD_DIR}/haproxy.tar.gz" | sha256sum -c -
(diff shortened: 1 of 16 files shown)


praateekmahajan:
/claude review
@greptileai

praateekmahajan:
/claude review
@greptileai

pyproject.toml
```diff
@@ -296,10 +296,10 @@ override-dependencies = [
     "numpy>=2.0.0,<=2.2.0", # Override nemo-toolkits constraint of <2.0.0, upperbounds for Numba compatibility
     "protobuf>=5.29.5,<7.0",  # Override nemo-toolkits constraint of ~=5.29.5; <7.0 due to ray serve FieldDescriptor API breakage
     "setuptools>=80.10.1", # Override setuptools range in other dependencies to address CVE GHSA-58pv-8j8x-9vj2
-    "torch>=2.8.0", # Override whisperx's torch~=2.8.0 upper bound; cap must match video_cuda12 and build groups
-    "torchaudio>=2.8.0", # Override whisperx's torchaudio~=2.8.0
-    "torchvision>=0.23.0", # Match torch>=2.8.0
-    "torchcodec>=0.9.0; platform_machine == 'x86_64' and platform_system != 'Darwin'", # Must match torch 2.9.x; override pyannote-audio's >=0.7.0 floor; gated since aarch64 lacks wheels
+    "torch==2.10.0", # Override whisperx's <2.9 cap to match cu129 / vllm 0.18.x
```
ayushdg, line 300: I guess going forward the override here always has to strictly match what vllm supports?

pyproject.toml
```diff
@@ -296,10 +296,10 @@ override-dependencies = [
...
+    "torchaudio==2.10.0", # Override whisperx's <2.9 cap to match cu129 / vllm 0.18.x
+    "torchvision==0.25.0", # Match torch==2.10.0
+    "torchcodec>=0.9.0,<0.11; platform_machine == 'x86_64' and platform_system != 'Darwin'", # cap below torchcodec 0.11 (built for torch 2.11) to match our torch 2.10 ABI; override pyannote-audio's >=0.7.0 floor; gated since aarch64 lacks wheels
```
ayushdg: Shouldn't we pin this at 0.10 to ensure codec is compatible with torch?

docker/common/install_haproxy.sh
```diff
@@ -0,0 +1,65 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -xeuo pipefail
+
+# Build HAProxy from source so Ray Serve's HAProxy ingress mode (RAY_SERVE_ENABLE_HA_PROXY=1)
```
ayushdg, line 18: Is there a TODO to remove this in favor of the ray-ha proxy package in the future?

.dockerignore
```diff
@@ -0,0 +1,45 @@
...
+tutorials/**/*.h5
+tutorials/**/*.zip
+tutorials/**/*.gz
+tutorials/**/*.log
+
+# Misc large/temp files
+data/
+ne
+*.bak
+repro*.py
```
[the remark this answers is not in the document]
ayushdg: Valid?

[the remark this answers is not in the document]
praateekmahajan: Removing `.dockerignore` so can add in another PR

pyproject.toml
praateekmahajan: Yup that's my understanding, unless we start doing sub venvs

sarahyurick (approved):

PR #1897, opened 2026-04-30
fix: support batched MinHash processing

### LINKED ISSUE #1465: Support batchsize/process_batch in the minHash stage
GPUs typically do better with larger batches (1-3GB) for processing but CPU stages do better with smaller (256MB or so) batches or run into OOMs. 

To avoid this we'd like to implelement/support `proccess_batch` for the MinHashStage to ensure that we can use a smaller blocksize during read but saturate the GPU better for minhash.

Implement `process_batch` for the minHash stage that reads the files represented by all the incoming task.data one by one and then concat it. There's still some penalty of having to read it one by one (read not saturated), but the minhash and write can be on larger blocks.

Similar approach to #1332 but arguably simpler since it's the typically `process/process_batch` type stage.

**Additional context**

Saibabu7770:
Fixes #1465.
Adds `process_batch` support to `MinHashStage` so multiple `FileGroupTask`s can be combined into one larger cuDF batch before computing MinHash signatures.
This lets upstream readers use smaller CPU-friendly file partitions while allowing the GPU MinHash stage to process larger batches and better utilize GPU throughput.
## Usage
```python
stage = MinHashStage(
    output_path="minhash-output",
    read_format="jsonl",
    batch_size=2,
)

### DIFF
--- a/nemo_curator/stages/deduplication/fuzzy/minhash.py
+++ b/nemo_curator/stages/deduplication/fuzzy/minhash.py
@@ -213,6 +213,8 @@ class MinHashStage(ProcessingStage[FileGroupTask, FileGroupTask], DeduplicationI
         Additional keyword arguments for reading input files
     write_kwargs : dict[str, Any] | None, default=None
         Additional keyword arguments for writing output files
+    batch_size : int, default=1
+        Number of FileGroupTasks to combine before computing minhashes
 
     Examples
     --------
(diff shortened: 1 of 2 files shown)


Saibabu7770:
# Description
Fixes #1465.
Adds `process_batch` support to `MinHashStage` so multiple `FileGroupTask`s can be combined into one larger cuDF batch before computing MinHash signatures.
This lets upstream readers use smaller CPU-friendly file partitions while allowing the GPU MinHash stage to process larger batches and better utilize GPU throughput.
## Usage
```python
stage = MinHashStage(
    output_path="minhash-output",
    read_format="jsonl",
    batch_size=2,
)

PR #1899, opened 2026-04-30
Nemotron OCR SDG Pipeline

suiyoubi:
## Description

Adds the **Nemotron OCR SDG pipeline** — a multimodal synthetic data generation pipeline that converts images into structured OCR + QA conversation data for vision-language model training.

### Pipeline stages

| Stage | Model | Output |
|---|---|---|
| \`NemotronOCRV2Stage\` | NemotronOCR-v2 | Dense word-level OCR with bounding boxes |
| \`OCRScoringQAStage\` | Gemini 3 Pro (NVIDIA Inference API) | Scoring, validation, and missing-region detection |
| \`OCRConversationalizeStage\` | — | 11 output format variants → \`ConversationSample\` |
| \`OCRDenseQAStage\` | — | 6 QA types (bbox↔text, point↔text, dense dump) |

### Key components

- **\`nemo_curator/models/omni/\`** — \`NVInferenceModel\` base class for NVIDIA Inference API-backed VLMs; \`Gemini3Pro\` concrete implementation
- **\`nemo_curator/models/client/nvinference_client.py\`** — thin streaming client helpers (\`get_nvinference_api_key\`, \`create_openai_client\`, \`stream_chat_completion_text\`)
- **\`nemo_curator/stages/synthetic/omni/base.py\`** — \`VLMProcessingStage\` and \`ModelProcessingStage\` base classes with batched inference, per-prompt error isolation, and setup/teardown lifecycle
- **\`nemo_curator/stages/synthetic/omni/io.py\`** — \`HFDatasetImageReader\`, \`TarImageReader\`, \`ParquetReader\`, \`SkipProcessedStage\`, \`ResultWriterStage\`
- **\`nemo_curator/tasks/ocr.py\`** — \`OCRDenseWord\` and \`OCRData\` task data classes
- **\`docker/Dockerfile\`** — installs \`nemotron-ocr-v2\` from source (no-build-isolation, CUDA arch list for A100/A10/RTX Ada/H100)
- **\`tutorials/synthetic/omni/hf_ocr_pipeline.py\`** — end-to-end example using HuggingFace datasets

### Tests
63 new unit tests across \`tests/tasks/\`, \`tests/models/\`, and \`tests/stages/synthetic/omni/\` — all CPU-only, no GPU required.

## Checklist


suiyoubi:
/ok to test [742e3d1](https://github.com/NVIDIA-NeMo/Curator/pull/1899/commits/742e3d12dc06cf7c5b505f66cd9cae963bbd4784)

nemo_curator/stages/synthetic/omni/ocr_conversationalize.py
```diff
@@ -0,0 +1,183 @@
...
+class OCRConversationData(OCRData):
+    """OCRData with a conversation field added by OCRConversationalizeStage."""
+
+    conversation: ConversationSample | None = None
+
+    def to_dict(self) -> dict[str, Any]:
+        d = dataclasses.asdict(self)
+        if self.conversation is not None:
+            d["conversation"] = self.conversation.to_dict()
+        return d
```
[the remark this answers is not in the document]
suiyoubi: @greptile-apps dataclasses.asdict(self) produces d["conversation"] = {"conversation":
  [...messages...]} (because ConversationSample is a dataclass with a conversation field), and self.conversation.to_dict() produces the identical structure. The
  override exists solely to preserve the "t" type discriminator on media fragments, not to change the nesting.

suiyoubi:
/ok to test [f840341](https://github.com/NVIDIA-NeMo/Curator/pull/1899/commits/f840341487d969ff8894026c89832098993ed595)

suiyoubi:
/ok to test [3829629](https://github.com/NVIDIA-NeMo/Curator/pull/1899/commits/38296296337cb81ee8f37943228beb34af9d5916)

nemo_curator/models/omni/base.py
```diff
@@ -0,0 +1,264 @@
...
+    """NVIDIA Inference API-specific configuration."""
+
+    base_url: str = "https://inference-api.nvidia.com"
+    api_key_env_var: str = "NVINFERENCE_API_KEY"
+
+    def __post_init__(self) -> None:
+        pass
+
+
+class NVInferenceModel(VLMModel):
```
voegtlel: For discussion: Well, we can let it be "nvinference", or generic "OpenAI-API-compatible" or something? Basically we're not specifically targeting nvinference, but rather any openai compatible api. Although internally we would of course focus on nvinference.

nemo_curator/models/omni/base.py
```diff
@@ -0,0 +1,264 @@
...
+
+
+@dataclass(kw_only=True)
+class NVInferenceModelConfig(ModelConfig):
+    """NVIDIA Inference API-specific configuration."""
+
+    base_url: str = "https://inference-api.nvidia.com"
+    api_key_env_var: str = "NVINFERENCE_API_KEY"
+
+    def __post_init__(self) -> None:
```
voegtlel: can be deleted

nemo_curator/models/omni/base.py
```diff
@@ -0,0 +1,264 @@
...
+            content.append(
+                {
+                    "type": "image_url",
+                    "image_url": {"url": image_url},
+                }
+            )
+        content.append({"type": "text", "text": prompt})
+        return content
+
+    def generate(
```
voegtlel: Meaning, we only have static batching for now, no dynamic batching?

nemo_curator/tasks/ocr.py
```diff
@@ -0,0 +1,135 @@
...
+from typing import TYPE_CHECKING, Any
+
+if TYPE_CHECKING:
+    from collections.abc import Iterable
+
+from nemo_curator.tasks.image import ImageTaskData
+
+
+@dataclass(kw_only=True)
+class OCRDenseWord:
```
voegtlel: I guess we should rename. It's not necessarily a word, but can also be a line (if using the line prediction mode).

nemo_curator/models/omni/gemini.py
```diff
@@ -0,0 +1,34 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from nemo_curator.models.omni.base import NVInferenceModel, NVInferenceModelConfig
+
+DEFAULT_MODEL_ID = "gcp/google/gemini-3-pro"
+
+
+class Gemini3Pro(NVInferenceModel):
```
voegtlel: Actually, was thinking if we should set our just released omni model as default?

voegtlel (changes requested):

nemo_curator/models/omni/base.py
suiyoubi: I think we should just leave it for nvinference for now. promoting NV inference infra isn't a bad idea

nemo_curator/models/omni/base.py
suiyoubi: thats true. Do we expect perf gain for inference API from using dynamic batching ?

nemo_curator/models/client/nvinference_client.py
```diff
@@ -0,0 +1,75 @@
...
+    """Create an OpenAI-compatible client configured for NVIDIA Inference API.
+
+    Note: Return type is `Any` to avoid importing `openai` at import-time for typing.
+    """
+    from openai import OpenAI  # local import to keep module thin
+
+    return OpenAI(base_url=base_url, api_key=api_key)
+
+
+def stream_chat_completion_text(  # noqa: PLR0913
```
praateekmahajan: Is there a reason to not reuse / extend our OpenAiClietn / AsyncOpenAiClient
https://github.com/NVIDIA-NeMo/Curator/blob/02dfb3fc792cae89eefd79f64a746e712c226499/nemo_curator/models/client/openai_client.py#L86

nemo_curator/models/omni/base.py
```diff
@@ -0,0 +1,261 @@
...
+    def __post_init__(self) -> None:
+        if self.gpu_memory_utilization < 0.0 or self.gpu_memory_utilization > 1.0:
+            msg = "GPU memory utilization must be between 0.0 and 1.0"
+            raise ValueError(msg)
+        if self.tensor_parallel_size < 1:
+            msg = "Tensor parallel size must be greater than 0"
+            raise ValueError(msg)
+
+
+class Model(ABC):
```
praateekmahajan: I'm really wary of new generic base classes... I believe we already have https://github.com/NVIDIA-NeMo/Curator/blob/main/nemo_curator/models/base.py 

Can we extend those? What does this give?
Similarly what does ModelConfig base class give rather than just having kwargs? vLLM accepts multiple kwargs these just being three of them... 

Is there a reason?

nemo_curator/stages/synthetic/omni/base.py
```diff
@@ -0,0 +1,309 @@
...
+        """Set the current CUDA device for this stage, if configured."""
+        if self.cuda_devices is None or len(self.cuda_devices) == 0:
+            return
+        import torch
+
+        if not torch.cuda.is_available():
+            msg = "CUDA is not available"
+            raise RuntimeError(msg)
+        torch.cuda.set_device(self.cuda_devices[0])
+        os.environ["CUDA_VISIBLE_DEVICES"] = ",".join(f"{d}" for d in self.cuda_devices)
```
praateekmahajan: Why are we doing this instead of letting Xenna / Ray Data do it?

nemo_curator/stages/synthetic/omni/ocr_scoring_qa.py
```diff
@@ -0,0 +1,323 @@
...
+
+# Reuse exact same prompt as OCRScoringVerificationStage
+_PROMPT = """\
+Please check if the following OCR bounding boxes are correct and respond ONLY with JSON \
+in this exact format:
+{{
+  "ocr_mode": "word" or "line",
+  "text": [
+    {{
+      "idx": <integer matching input idx>,
```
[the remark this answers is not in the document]
suiyoubi: @greptile-apps not fixed. The regex runs on text that's already had ```json / ``` fences stripped, so the dominant failure shape — a reasoning model emitting <scratchpad with {…}> ```json {answer} ```  — collapses to the answer JSON before the regex ever sees it. We haven't observed parse failures from this in practice on Nemotron-Nano-Omni outputs against this prompt.

nemo_curator/models/omni/base.py
```diff
@@ -0,0 +1,141 @@
...
+from typing import Any
+
+from loguru import logger
+from PIL import Image
+
+from nemo_curator.models.base import ModelInterface
+from nemo_curator.models.client import OpenAIClient
+
+
+class NVInferenceModel(ModelInterface):
```
praateekmahajan: why is this a ModelInterface..

Can this just be a `ProcessingStage`?

See https://github.com/NVIDIA-NeMo/Curator/blob/main/nemo_curator/stages/synthetic/qa_multilingual_synthetic.py#L33

To which we pass in a client which can be OpenAIClient or AsyncOpenAIClient

nemo_curator/models/omni/base.py
```diff
@@ -0,0 +1,141 @@
...
+        if self.is_loaded:
+            return
+
+        api_key = os.environ.get(self.api_key_env_var, "").strip()
+        if not api_key:
+            msg = f"{self.api_key_env_var} is not set"
+            raise RuntimeError(msg)
+
+        logger.info(f"Initializing NVIDIA Inference client for model {self.model_id}")
+        self._client = OpenAIClient(base_url=self.base_url, api_key=api_key)
```
praateekmahajan: TAL at AsyncOpenClient, I'd imagine for perf you need that

nemo_curator/stages/synthetic/omni/base.py
```diff
@@ -0,0 +1,159 @@
...
+from nemo_curator.tasks.image import SingleDataTask
+
+T = TypeVar("T")
+
+
+class SkipSample(Exception):  # noqa: N818
+    """Exception to be raised in build_prompt to skip a sample."""
+
+
+class ModelProcessingStage(ProcessingStage[SingleDataTask[T], SingleDataTask[T]], Generic[T]):
```
praateekmahajan: SingleDataTask is really vague, can we use something else please? Do we need something like that?

nemo_curator/stages/synthetic/omni/base.py
```diff
@@ -0,0 +1,159 @@
...
+    """Base stage for cloud-API VLM inference.
+
+    Concrete subclasses implement ``build_prompt`` and ``handle_response``;
+    image loading and batch dispatch are handled here. Sampling kwargs
+    (temperature, top_p, max_tokens, etc.) live on the model instance —
+    set them at construction.
+
+    GPU placement is delegated to the executor (Xenna/Ray) via the
+    ``resources`` declaration; the stage itself never touches CUDA.
```
praateekmahajan: AI comments? Can you review the code yourself to make sure nothing too obvious / excessive is written

nemo_curator/stages/synthetic/omni/io.py
```diff
@@ -0,0 +1,383 @@
...
+from nemo_curator.backends.utils import RayStageSpecKeys
+from nemo_curator.stages.base import ProcessingStage
+from nemo_curator.stages.resources import Resources
+from nemo_curator.tasks import _EmptyTask
+from nemo_curator.tasks.image import ImageTaskData, SingleDataTask
+
+T_TaskData = TypeVar("T_TaskData", bound=ImageTaskData)
+
+
+class HFDatasetImageReaderStage(ProcessingStage[_EmptyTask, SingleDataTask[T_TaskData]], Generic[T_TaskData]):
```
praateekmahajan: What does this `Generic[T_TaskData]` mean? I thought we only need 

```python
HFDatasetImageReaderStage(ProcessingStage[_EmptyTask, SingleDataTask[T_TaskData]])
```

nemo_curator/stages/synthetic/omni/io.py
```diff
@@ -0,0 +1,383 @@
...
+    # Dataset loading
+    # ------------------------------------------------------------------
+
+    def _load_dataset(self) -> Dataset:
+        """Load a HuggingFace Dataset, handling hub, save_to_disk, and imagefolder paths."""
+        from datasets import load_dataset, load_from_disk
+
+        local_path = Path(self.dataset_name)
+
+        if local_path.exists():
```
praateekmahajan, line 124: All our other I/O stages support cloud, maybe we should by default support cloud in the newer code pats too

nemo_curator/stages/synthetic/omni/io.py
```diff
@@ -0,0 +1,383 @@
...
+        raise ValueError(msg)
+
+    # ------------------------------------------------------------------
+    # Stage entry point
+    # ------------------------------------------------------------------
+
+    def process(self, _: _EmptyTask) -> list[SingleDataTask[T_TaskData]]:
+        self.image_dir.mkdir(parents=True, exist_ok=True)
+        dataset = self._load_dataset()
+        dataset_tag = Path(self.dataset_name).name.replace("/", "_")
```
praateekmahajan, line 166: same here re cloud.. or rather anywhere we you rely on `Path`

nemo_curator/stages/synthetic/omni/io.py
```diff
@@ -0,0 +1,383 @@
...
+            )
+
+        logger.info(
+            f"hf_dataset_image_reader: {len(tasks)} unique images from "
+            f"{self.dataset_name}/{self.split}" + (f" (limit={self.limit})" if self.limit else "")
+        )
+        return tasks
+
+
+class ResultWriterStage(ProcessingStage[SingleDataTask[T_TaskData], SingleDataTask[T_TaskData]], Generic[T_TaskData]):
```
praateekmahajan: The name is too open-ended.. can it be more specific to the work it's doing? Since every writer in Curator is technically "ResultWriterStage"

nemo_curator/stages/synthetic/omni/io.py
```diff
@@ -0,0 +1,383 @@
...
+        # with mode "w", causing truncation; use single_file=False for distributed runs.
+        base_path = Path(self.output_path)
+        if self._worker_id and not use_single_file:
+            suffix = base_path.suffix or ".jsonl"
+            output = base_path.parent / f"{base_path.stem}_worker{self._worker_id}{suffix}"
+        else:
+            output = base_path
+
+        output.parent.mkdir(parents=True, exist_ok=True)
+        self._file = open(output, "w", encoding="utf-8")  # noqa: SIM115
```
praateekmahajan, line 242: If the actor dies mid process won't the file remain open? Why not do it safely with 

```python
with open(...) as _file:
   _file.write(...)
```

nemo_curator/stages/synthetic/omni/io.py
```diff
@@ -0,0 +1,383 @@
...
+        image_parent: str | None = None,
+        single_file: bool = False,
+    ) -> None:
+        """Initialize the result writer stage.
+
+        Args:
+            output_path: Base path for output JSONL file.
+            valid_only: If True, only write valid records.
+            image_parent: If provided, make image paths relative to this directory.
+            single_file: If True, write to exactly output_path without worker_id suffix.
```
praateekmahajan: I'm curious how do you support single_file on a distributed executor i.e. curator? Why is this here? 

If someone had to support single_file then you would set num_workers=1 so that no concurrent writers exist.. or am i missing something here?

nemo_curator/stages/synthetic/omni/ocr_nemotron_v2.py
```diff
@@ -0,0 +1,147 @@
...
+            if not task.data.is_valid:
+                continue
+            try:
+                self._process_one(task)
+            except Exception as e:  # noqa: BLE001
+                logger.error(f"{self.name}: error on task {task.task_id}: {e}")
+                task.data.is_valid = False
+                task.data.error = f"{self.name}: {e}"
+
+        return tasks
```
praateekmahajan: why are we overwriting both process and process_batch?

nemo_curator/stages/synthetic/omni/ocr_nemotron_v2.py
```diff
@@ -0,0 +1,147 @@
...
+
+    def _resolve_model_dir(self) -> str:
+        if self.model_dir is not None:
+            return str(self.model_dir)
+        from huggingface_hub import snapshot_download
+
+        snapshot = snapshot_download(repo_id=_HF_REPO_ID)
+        return str(Path(snapshot) / _DEFAULT_SUBDIR)
+
+    def setup(self, _worker_metadata: dict | None = None) -> None:
```
praateekmahajan: we don't need a setup_on_node?

nemo_curator/tasks/image.py
```diff
@@ -67,3 +72,37 @@ def validate(self) -> bool:
...
+        """Convert to dictionary for JSON serialization.
+
+        Path objects are left as-is; callers should use default=str when
+        passing to json.dumps to handle Path → str conversion.
+        """
+        return dataclasses.asdict(self)
+
+
+@dataclass(kw_only=True)
+class SingleDataTask(Task[T], Generic[T]):
```
praateekmahajan: This name is really vague.. can you rename to something more relevant for Omni? Since something like this might just exist for every modality out there

tutorials/synthetic/omni/ocr_pipeline.py
```diff
@@ -0,0 +1,296 @@
...
+    output_tasks = pipeline.run(executor)
+    elapsed = time.perf_counter() - start
+
+    client.stop()
+
+    logger.info(f"Pipeline completed in {elapsed:.1f}s ({elapsed / 60:.1f} min)")
+    logger.info(f"Tasks processed: {len(output_tasks)}")
+
+    merged = merge_output_shards(Path(args.output_path))
+    logger.info(f"Output: {merged}")
```
praateekmahajan, line 300: We should probably add this to benchmark.. PS for that you'll also need to copy dataset for which this will work (so that we don't re download from the internet each time)..

copy the data to eos and dgx-a100 and then update the benchmarking/script/....yaml with the right path..

Feel free to ask @rlratzel for more questions around that

praateekmahajan (changes requested):

nemo_curator/stages/synthetic/omni/io.py
suiyoubi: This is the HF dataset, and I don't think cloud supported is that necessary given the fact most of the time it will just be downlaoding from HF directly. I would insist sticking with loading local files for this dataset for now. wdyt ?

nemo_curator/stages/synthetic/omni/io.py
suiyoubi: same above

nemo_curator/stages/synthetic/omni/io.py
suiyoubi: so for now we have the `open` in the setup func, are you suggest to repoen the file everytime we do `process` ? the file is flushed on every write, so we won't really lose data on a crash (but yes the file would remain open)

if we use `with open()` in the process, seems to be too many IO ops to me, or do you think this is actually fine

nemo_curator/stages/synthetic/omni/io.py
suiyoubi: yes I should have removed this.. this was comming from the old code where they run this with slurm array so always one core.. Will refactor

nemo_curator/stages/synthetic/omni/ocr_nemotron_v2.py
suiyoubi: we need to implement `process` because it is a abstractmethod.. we are only using `pprocess_batch` here

nemo_curator/tasks/image.py
suiyoubi: this is put to `tasks/image.py` so we should not really make it too "omni-specific", the class should be reusable for others as well, but maybe good to rename as `ImageSampleTask` ?

nemo_curator/stages/synthetic/omni/ocr_nemotron_v2.py
praateekmahajan: In that case just raise an error, see kmeans.py
```python
        msg = "BlaBla does not support single-task processing"
        raise NotImplementedError(msg)
```

nemo_curator/stages/synthetic/omni/base.py
suiyoubi: renamed to `imageSampleTask`

nemo_curator/models/omni/base.py
suiyoubi: This is just the modeling class. Similar to how we have in video for example that we have `QwenLM` class and then we invoke the model in `CaptionGenerationStage`.

Here we just have the `NVInferenceModel` which is supposed to work for any API endpoint from build.nvidia.com . and we use this model in the [ModelProcessingStage](https://github.com/NVIDIA-NeMo/Curator/pull/1899/changes#diff-1c0290c5baf09f9eafeab55ade535d6abea0ddd4ae1df00cbf209aa9b59bb25cR33) (which is a `ProcessingStage)`

nemo_curator/models/omni/base.py
suiyoubi: make sense. I'll add the async option

nemo_curator/stages/synthetic/omni/io.py
suiyoubi: rename to `JsonlSampleWriterStage`

tutorials/synthetic/omni/ocr_pipeline.py
suiyoubi: I have the benchmark setup here in another branch: https://github.com/NVIDIA-NeMo/Curator/tree/aot/omni_sdg_benchmark 
let's merge that later once we finalizing on the benchmark details so we know what to measure (i.e. GPU SKU)

suiyoubi:
/ok to test [7d4d113](https://github.com/NVIDIA-NeMo/Curator/pull/1899/commits/7d4d1130b34b566bd46b0535e0e84605ad89520e)

suiyoubi:
/ok to test [025bc07](https://github.com/NVIDIA-NeMo/Curator/pull/1899/commits/025bc070eb21619034825374b60033844c929363)

praateekmahajan (approved):

suiyoubi:
/ok to test [85c82d5](https://github.com/NVIDIA-NeMo/Curator/pull/1899/commits/85c82d5dfccafad8ee48934c9e5ff8c2fb3fb8a2)

suiyoubi:
/ok to test [fc82dc0](https://github.com/NVIDIA-NeMo/Curator/pull/1899/commits/fc82dc04a9ad782354ed00188d1d75501c258438)

suiyoubi:
/ok to test [146c38b](https://github.com/NVIDIA-NeMo/Curator/pull/1899/commits/146c38bab5d6b036e93b37effb98729efa207132)

thomasdhc (approved):

PR #1909, opened 2026-05-05
[benchmarking] Add `--use-host-curator-benchmarking` and `--skip-curator-image-build*` options

rlratzel:
[benchmarking] Add `--use-host-curator-benchmarking` and `--skip-curator-image-build*` options

## Summary

This PR adds features to the benchmarking tools to make it easier to benchmark specific image builds (such as RC builds).
These features were first used in [this PR](https://github.com/NVIDIA-NeMo/Curator/pull/1431) but were never merged.

- **`gen_runscript_vars.py`**: Add `--use-host-curator-benchmarking` option that mounts only the `benchmarking/` subdirectory into the container, rather than the full curator repo as `--use-host-curator` does. Useful for iterating on benchmarking tools while using the Curator version installed in the image. The two flags are mutually exclusive.
- **`build_docker.sh`**: Replace the simple `--tag-as-latest` string check with a proper argument parser loop. Add two new options:
  - `--skip-curator-image-build`: pull the curator image instead of building it locally
  - `--skip-curator-image-build-and-pull`: skip both build and pull, using an existing local image

## Test plan

- [x] Verify `--use-host-curator-benchmarking` mounts only `benchmarking/` and not the full repo
- [x] Verify `--use-host-curator` and `--use-host-curator-benchmarking` together produce a clear error
- [x] Verify `--skip-curator-image-build` pulls the image rather than building
- [x] Verify `--skip-curator-image-build-and-pull` skips both steps and uses the existing local image
- [x] Verify `--tag-as-latest` still works correctly with the new argument parser

### DIFF
--- a/benchmarking/tools/build_docker.sh
+++ b/benchmarking/tools/build_docker.sh
@@ -17,12 +17,32 @@
 # Exit immediately on error, unset vars are errors, pipeline errors are errors
 set -euo pipefail
 
-# Tag the images with ':latest' tag if --tag-as-latest flag is present.
-# This is not the default to prevent name collisions from multiple users.
+# Parse command-line arguments
 TAG_AS_LATEST=false
-if [[ "${*}" == *"--tag-as-latest"* ]]; then
-  TAG_AS_LATEST=true
-fi
+SKIP_CURATOR_BUILD=false
+SKIP_CURATOR_BUILD_AND_PULL=false
+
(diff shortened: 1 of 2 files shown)


rlratzel:
Re-opened using fork branch [here](https://github.com/NVIDIA-NeMo/Curator/pull/1910).

PR #1931, opened 2026-05-06
resumability fanout only

oyilmaz-nvidia:
## Summary

Adds pipeline resumability: when `pipeline.run(checkpoint_path=...)` is set, completed source partitions are tracked in an LMDB database and skipped on subsequent runs. This allows interrupted pipelines to continue from where they left off rather than reprocessing everything from scratch.

- **`Pipeline.run(checkpoint_path=...)`** — new optional argument. When set, two internal stages are injected around the pipeline stages: `_CheckpointFilterStage` (after the first source stage) and `_CheckpointRecorderStage` (at the end). The pipeline's own stage list is never mutated.
- **`_CheckpointFilterStage`** — validates that source tasks carry a `resumability_key`, stamps a stable `_resumability_uuid`, and drops tasks whose partition was already fully completed in a prior run.
- **`_CheckpointRecorderStage`** — records each leaf task as completed via a singleton Ray actor backed by LMDB.
- **`BaseStageAdapter`** — gains `_propagate_resumability_metadata()` (copies `resumability_key` from inputs to outputs that lack it) and `_record_checkpoint_events()` (handles fan-out by calling `add_expected` synchronously before dispatching leaf tasks, and handles full-drop by immediately marking the input as completed).
- **`FilePartitioningStage`** — marked as `is_source_stage()`, now sets `resumability_key = sha256(sorted_file_paths + partition_index)` and restores `task_id` to `f"file_group_{i}"`.
- **`CheckpointManager` / `_CheckpointActor`** — LMDB-backed singleton Ray actor serializing all checkpoint reads/writes. Tracks per-partition expected completion count and completed UUID set.
- **Image curation example** — updated with `--checkpoint-path` argument.


nemo_curator/backends/base.py
```diff
@@ -113,6 +114,54 @@ def setup(self, worker_metadata: WorkerMetadata | None = None) -> None:
...
+                task._metadata["resumability_key"] = source_key
+
+    def _record_checkpoint_events(self, input_tasks: list[Task], output_tasks: list[Task]) -> None:
+        """Detect fan-out and full-drop events; update checkpoint state accordingly."""
+        if self._checkpoint_actor is None:
+            return
+        if not input_tasks:
+            return
+
+        is_fanout = len(output_tasks) > len(input_tasks) and len(input_tasks) == 1
```
praateekmahajan: Let's imagine we kill the idea of user definied task_id.. Most stages / users anyway don't use it.
So task_id becomes something we control and relies on order of tasks produce by stage..

Imagine a simple pipeline `Stage1 -> Stage2 -> Stage3 -> Stage4 -> Stage5 -> Stage6`

1. Stage 1 : Fanout : EmptyTask -> List[FileGroupTask] (file partitioning; produces 10 tasks)
2. Stage 2 : Normal : FileGroupTask -> DocumentBatch (reader)
3. Stage 3 : Normal : DocumentBatch -> DocumentBatch (transformation)
3. Stage 4 : Fanout: DocumentBatch -> List[DocumentBatch] (row splitter; each task splits 2/3 tasks)
4. Stage 5: Normal : DocumentBatch -> DocumentBatch (transformation)
5. Stage 6: Normal : Document Batch ->  FilGroupTask (Writer)


Here is how the tasks get processed

1. Stage1 starts with `task0` and produces `task_0_0`, `task_0_1`........ `task_0_9` (**10 tasks**)
2. Stage2 / Stage3 are normal so they don't modify task_id at all (**10 tasks**)
3. Stage 4 is another fanout, so it accepts `task_0_0` and produces two outputs `task_0_0_0, task_0_0_1` and `task_0_1` produces three outputs -> `task_0_1_0, task_0_1_1, task_0_1_2`... and remaining tasks produced two outputs only s.t. we have `task_0_2_0, task_0_2_1`..... `task_0_9_0, task_0_9_1` (i.e. **21 tasks**)
4. Stage5 maintains is normal so doesn't modify `task_id` at all (**21 tasks**)
5. Stage6 writes them out. (**21 tasks**)

So now somehow we can use this information to ensure how we write it out, so that if `task_0_1_2` was dropped we don't end up rewriting the whole lot of `task_0_1_0, task_0_1_1, task_0_1_2` again just `task_0_1_2`

oyilmaz-nvidia: I think this can work. Should we create a new ID and put in the  _metadata, or overwrite the existing task id? Maybe in some pipelines, task id has a purpose? 

And since these IDs that we'll create, will be unique. I can mark them as completed, and find a way to not the writer stage rewrite it. Or, find a way to filter it before. Let me work on this this weekend. THanks!

praateekmahajan: I'm fairly confident task_id created by users is redundant (if you see stages where its not, please share with me).. When we wrote Curator, the intent was that task_id sovles this purpose but we never got to it.. ~1 year later we're just carrying it as dead-weight. So I'd say let's reuse it and  

I remember @VibhuJawa also added task_uuid too.. 

So you could perhaps even break it up into two PRs 1) where you make task_id automated (i.e. instead of users setting it) 2) where you use task_id for checkpointing

oyilmaz-nvidia: But we still need a unique and deterministic task id from an initial stage. Or find a way to create deterministic and unique task ids for every stages.

Ok, I'll create a PR for overwriting the task ids first.

oyilmaz-nvidia: Actually, if we can figure out the deterministic and unique tasks ids, then we can filter out tasks at any stage. We don't need to wait till the end. 

With a basic DFS algo, we can visit all the nodes and check if the children nodes are marked completed, a node can mark itself completed.

tests/backends/ray_data/test_pdf_resumability.py
```diff
@@ -0,0 +1,278 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""End-to-end resumability test for the Nemotron-Parse PDF tutorial.
+
+Downloads 2 real PDFs via the tutorial's ``download_data.py``, runs them
+through a CPU-only pipeline (Partitioning -> Preprocess -> FanOut -> PassThrough),
```
abhinavg4, line 18: HI,

(1) Can we include a writer stage at the end, just a simulation of one (no need to write to the disk)
(2), It seems like these stages need to be made in a very specific way: _MarkFirstFanoutLeafStage has some setup fn. It that expected? I thought the interface should be simpler.

oyilmaz-nvidia: @abhinavg4 

1) Sure, will add it.

2) Yeah, you just need to add this to your source stage 

```
def is_source_stage(self) -> bool:
        return True
```

like [here](https://github.com/oyilmaz-nvidia/Curator/blob/onur/resumability-fanout-only/nemo_curator/stages/file_partitioning.py#L101) for file_partitioning.py

oyilmaz-nvidia:
@abhinavg4 Could you please review again? Addressed your feedback.

PR #1934, opened 2026-05-06
ci: Update container registry

thomasdhc:
## Usage

```python
```
## Checklist

### DIFF
--- a/.github/workflows/cicd-main.yml
+++ b/.github/workflows/cicd-main.yml
@@ -138,11 +138,8 @@ jobs:
       - name: main
         uses: ./.github/actions/build-container
         with:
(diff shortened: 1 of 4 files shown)


kajalj22 (approved):

.github/workflows/cicd-main.yml
chtruong814, line 42: You'll want to update this to now use the ephemeral aws runners for non-nvidia path.

On Mbridge, I think Oliver updated these to be the full runner name and then just replaced references downstream where used.

* nvidia runner would be `nemo-ci-aws-gpu-x2`
* non-nvidia runner would be `nemo-ci-aws-gpu-x2-ephemeral`.

PR #1957, opened 2026-05-07
chore: Update dynamo to 1.1.0

praateekmahajan:
## Description
Now dynamo 1.1.0 uses vllm 0.19 and curator base is on vLLM 0.18, this means the dynamo setup time will be fractional of what it was before, since FA2 doesn't need to be rebuilt, as torch version remains the same.

We also see a perf improvement (likely because we went from vLLM 0.16 to vLLM 0.19) for the dynamo workers
image
 
Minor small hack needed, because dynamo[vllm] now pins ray to 2.55, while curator is on 2.55.1
So we create a file on all the nodes containing uv override.

Also pulled the per-node fan-out shared by `execute_setup_on_node` and the new override-file write into a small `run_on_each_node` helper in `nemo_curator/utils/ray_utils.py` (with `get_head_node_id` / `is_head_node` colocated there). Both call sites go through it now, so the override write respects `CURATOR_IGNORE_RAY_HEAD_NODE` the same way executor scheduling does.

## Usage

```python
```
## Checklist

### DIFF
--- a/nemo_curator/backends/utils.py
+++ b/nemo_curator/backends/utils.py
@@ -16,14 +16,14 @@ import os
 import time
 from copy import deepcopy
 from enum import Enum
-from typing import TYPE_CHECKING, Any
+from typing import TYPE_CHECKING
 
 import ray
 from loguru import logger
-from ray.util.scheduling_strategies import NodeAffinitySchedulingStrategy
 
 from nemo_curator.backends.base import NodeInfo, WorkerMetadata
 from nemo_curator.stages.base import ProcessingStage
+from nemo_curator.utils.ray_utils import get_head_node_id, run_on_each_node
 
 if TYPE_CHECKING:
     import loguru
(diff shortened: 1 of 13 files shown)


nemo_curator/backends/utils.py
```diff
@@ -225,29 +196,26 @@ def _setup_stage_on_node(stage: ProcessingStage, node_info: NodeInfo, worker_met
...
 
 
 def execute_setup_on_node(stages: list[ProcessingStage], ignore_head_node: bool = False) -> None:
     """Execute setup on node for a stage."""
-    head_node_id = get_head_node_id()
-    ray_tasks = []
+    head_node_id = get_head_node_id() if ignore_head_node else None
     for node in ray.nodes():
+        if not node.get("Alive"):
+            continue
```
ayushdg, line 214: Curious about scenarios you might've hit to need this check

nemo_curator/backends/utils.py
```diff
@@ -225,29 +196,26 @@ def _setup_stage_on_node(stage: ProcessingStage, node_info: NodeInfo, worker_met
...
-            )
-    ray.get(ray_tasks)
+    for stage in stages:
+        run_on_each_node(
+            _setup_stage_on_node,
+            stage,
+            ignore_head_node=ignore_head_node,
+            num_cpus=stage.resources.cpus if stage.resources is not None else 1,
+            num_gpus=stage.resources.gpus if stage.resources is not None else 0,
+        )
```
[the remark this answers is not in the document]
ayushdg: Good point. Is this intentional to avoid some issues?

nemo_curator/core/serve/dynamo/vllm.py
```diff
@@ -46,30 +50,52 @@
...
+    any mismatch.
+
+    Must run inside an active Ray context, before any worker spawned with
+    :data:`DYNAMO_VLLM_RUNTIME_ENV` lands. The runtime_env_agent on each
+    worker reads the file from the node-local filesystem; a single
+    driver-side write doesn't reach remote nodes.
+
+    Re-call after cluster topology changes (autoscale, node restart) — this
+    is one-shot and not auto-triggered.
+    """
```
ayushdg, line 90: To confirm we override the ray version from dynamo (insert other venv here) to instead use ray version from driver?

ayushdg (commented):
MInor questions. Nothing blocking

nemo_curator/backends/utils.py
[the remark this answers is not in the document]
praateekmahajan: I had forgotten to actually push the changes. Just pushed now, so it's now resolved

nemo_curator/backends/utils.py
praateekmahajan: I don't think we'll run into this, it's just a safe guard. What I learnt was `ray.nodes()` will also return a "dead node". And if a node is dead we can't submit anything on it

nemo_curator/core/serve/dynamo/vllm.py
praateekmahajan: Yes, we will add the driver ray version, and override dynamo's ray version. They've hard pinned it in 1.1.* but in 1.2.x they should be getting rid of the hardpin https://github.com/ai-dynamo/dynamo/pull/9439/changes

ayushdg (approved):

PR #1976, opened 2026-05-13
Chatterbox tts

Ssofja:
## Description

Add a ChatterboxTTS-based speech synthesis stage (`ChatterboxTTSStage`) to the NeMo Curator audio pipeline for generating multi-speaker conversation audio from text.

**New stage:**
- `ChatterboxTTSStage` — Synthesises conversation-turn audio using [Chatterbox TTS](https://github.com/resemble-ai/chatterbox). Supports both the English-only model (`ChatterboxTTS`) and the multilingual model (`ChatterboxMultilingualTTS`, 23 languages). Speaker voices are automatically assigned from a reference audio dataset and stay consistent within each conversation.

**Key features:**
- Two reference audio layouts: `wavs/<dialog>/<speaker>.wav` (with optional RTTM silence stripping) and MLS `<spk>/<book>/<seg>.flac` (auto-concatenated to target duration).
- Per-conversation random exaggeration range for voice style variation.
- RMS-based audio normalisation with clipping protection.
- Deterministic filenames (MD5-based) enabling idempotent re-runs — existing output files are reused without re-generation.
- Graceful failure handling: TTS errors produce a silence placeholder instead of crashing the pipeline.
- Requires 1 GPU (`Resources(gpus=1)`).

**New files:**
- `nemo_curator/stages/audio/tts/__init__.py`
- `nemo_curator/stages/audio/tts/chatterbox_tts.py`
- `tests/stages/audio/tts/__init__.py`
- `tests/stages/audio/tts/test_chatterbox_tts.py` (55 tests)

## Usage

```python
from nemo_curator.stages.audio import ChatterboxTTSStage

# English TTS with wavs/ reference layout
stage = ChatterboxTTSStage(
    output_audio_dir="/data/tts_output",
    reference_voices_dataset="/data/reference_voices",
    cfg_weight=0.5,
    exaggeration=0.5,
    temperature=0.8,
)

# Multilingual TTS (e.g. Russian) with MLS reference layout
stage_ru = ChatterboxTTSStage(
    output_audio_dir="/data/tts_output_ru",
    reference_voices_dataset="/data/mls_russian",
    language="ru",
    exaggeration=[0.3, 0.7],  # random per conversation
    max_reference_duration=30.0,
)

# Process conversation turns
from nemo_curator.tasks import AudioTask

tasks = [
    AudioTask(
        data={"utterance": "Hello, how are you?", "speaker": "Alice", "conversation_id": "conv001"},
        task_id="t1",
        dataset_name="my_dataset",
    ),
    AudioTask(
        data={"utterance": "I'm doing well, thanks!", "speaker": "Bob", "conversation_id": "conv001"},
        task_id="t2",
        dataset_name="my_dataset",
    ),
]

results = stage.process_batch(tasks)
# Each result.data now contains "audio_filepath", "duration", and "reference_voice"
```

**Supported languages (multilingual mode):**
`ar`, `da`, `de`, `el`, `en`, `es`, `fi`, `fr`, `he`, `hi`, `it`, `ja`, `ko`, `ms`, `nl`, `no`, `pl`, `pt`, `ru`, `sv`, `sw`, `tr`, `zh`

## Checklist


nemo_curator/stages/audio/tts/chatterbox_tts.py
```diff
@@ -0,0 +1,493 @@
...
+        sample_rate: int = 24000,
+        cfg_weight: float = 0.5,
+        exaggeration: float | list[float] = 0.5,
+        temperature: float = 0.8,
+        repetition_penalty: float | None = None,
+        min_p: float = 0.05,
+        top_p: float = 1.0,
+        normalize_audio: bool = True,
+        normalize_level: float = -20.0,
+        **kwargs,
```
sarahyurick: Can this be removed?

nemo_curator/stages/audio/tts/chatterbox_tts.py
```diff
@@ -0,0 +1,493 @@
...
+            except (ImportError, AttributeError):
+                pass
+
+            from chatterbox.mtl_tts import ChatterboxMultilingualTTS
+            self.model = ChatterboxMultilingualTTS.from_pretrained(device=self.device)
+            logger.info(f"Loaded ChatterboxMultilingualTTS (language={self.language})")
+        else:
+            from chatterbox.tts import ChatterboxTTS
+            self.model = ChatterboxTTS.from_pretrained(device=self.device)
+            logger.info("Loaded ChatterboxTTS (English)")
```
sarahyurick, line 396: Can this imports be at the top of the script? Same with the environment variable?

nemo_curator/stages/audio/tts/chatterbox_tts.py
```diff
@@ -0,0 +1,493 @@
...
+        return normalised
+
+    @staticmethod
+    def _output_filename(conversation_id: str, speaker: str, text: str) -> str:
+        """Deterministic filename: ``{conv_id_short}_{speaker}_{text_hash}.wav``."""
+        conv_short = conversation_id[:12] if len(conversation_id) > 12 else conversation_id
+        text_hash = hashlib.md5(text.encode("utf-8")).hexdigest()[:10]
+        return f"{conv_short}_{speaker}_{text_hash}.wav"
+
+    def _ensure_ready(self) -> None:
```
sarahyurick: Let's remove this.

nemo_curator/stages/audio/tts/chatterbox_tts.py
```diff
@@ -0,0 +1,493 @@
...
+        """Generate audio for a batch of conversation turns.
+
+        Each turn is synthesised independently (TTS is autoregressive and
+        not easily vectorised), but batching allows the model to stay warm
+        across turns and avoids repeated setup overhead.
+        """
+        if not tasks:
+            return []
+
+        self._ensure_ready()
```
sarahyurick: Remove. We should never call `setup` in `process_batch`/`process`.

nemo_curator/stages/audio/tts/chatterbox_tts.py
```diff
@@ -0,0 +1,493 @@
...
+        os.makedirs(self.output_audio_dir, exist_ok=True)
```
sarahyurick: Should this be in `setup`?

nemo_curator/stages/audio/tts/chatterbox_tts.py
```diff
@@ -0,0 +1,511 @@
...
+    @staticmethod
+    def _output_filename(conversation_id: str, speaker: str, text: str) -> str:
+        """Deterministic filename: ``{conv_id_hash}_{speaker}_{text_hash}.wav``."""
+        conv_hash = hashlib.md5(conversation_id.encode("utf-8"), usedforsecurity=False).hexdigest()[:12]
+        text_hash = hashlib.md5(text.encode("utf-8"), usedforsecurity=False).hexdigest()[:10]
+        return f"{conv_hash}_{speaker}_{text_hash}.wav"
+
+    def process(self, task: AudioTask) -> AudioTask:
+        """Generate audio for a single conversation turn."""
+        return self.process_batch([task])[0]
```
sarahyurick: We can just raise as not implemented instead.

tests/stages/audio/tts/test_chatterbox_tts.py
```diff
@@ -0,0 +1,698 @@
...
+) -> ChatterboxTTSStage:
+    """Inject a fake model and discover reference files."""
+    os.makedirs(stage.output_audio_dir, exist_ok=True)
+    stage._init_temp_dir()
+    stage.model = _fake_model(sample_rate)
+    stage._load_reference_audio_files()
+    return stage
+
+
+class TestConstruction:
```
sarahyurick: None of the tests in this class are needed IMO.

tests/stages/audio/tts/test_chatterbox_tts.py
```diff
@@ -0,0 +1,698 @@
...
+    def test_invalid_language_raises(self, output_dir: str, ref_dataset: str) -> None:
+        with pytest.raises(ValueError, match="Unsupported language"):
+            _build_stage(output_dir, ref_dataset, language="xx")
+
+    def test_gpu_resources(self, output_dir: str, ref_dataset: str) -> None:
+        stage = _build_stage(output_dir, ref_dataset)
+        assert stage.resources.gpus == 1
+
+
+class TestModelLoading:
```
sarahyurick: Same here, I don't think these tests are needed.

tests/stages/audio/tts/test_chatterbox_tts.py
```diff
@@ -0,0 +1,698 @@
...
+        stage.setup()
+
+        stage._assign_reference("Alice", "conv001")
+        stage.teardown()
+        stage.setup()
+
+        ref_path, _ref_id = stage._assign_reference("Alice", "conv001")
+        assert os.path.exists(ref_path)
+
+class TestEdgeCases:
```
sarahyurick: Let's avoid the overly-specific `class TestXYZ` in favor of a single test class for each `ProcessingStage`.

sarahyurick (changes requested):

nemo_curator/stages/audio/tts/chatterbox_tts.py
```diff
@@ -0,0 +1,512 @@
...
+            with contextlib.suppress(OSError):
+                shutil.rmtree(self.temp_dir)
+            self.temp_dir = None
+
+    def _load_model(self) -> None:
+        """Load ChatterboxTTS or ChatterboxMultilingualTTS."""
+        if self.language:
+            os.environ["TRANSFORMERS_ATTN_IMPLEMENTATION"] = "eager"
+            try:
+                import chatterbox.models.t3.llama_configs as _llama_cfgs
```
sarahyurick: Top-level import?

nemo_curator/stages/audio/tts/chatterbox_tts.py
```diff
@@ -0,0 +1,512 @@
...
+        self.conversation_exaggeration: dict[str, float] = {}
+
+        self.temp_dir: str | None = None
+        self._rng = random.Random()  # noqa: S311
+
+    def setup(self, worker_metadata: object = None) -> None:  # noqa: ARG002
+        """Load the TTS model and discover reference audio files."""
+        os.makedirs(self.output_audio_dir, exist_ok=True)
+        self._init_temp_dir()
+        self._load_model()
```
sarahyurick, line 321: Don't we need `setup_on_node` to download the relevant model first?

tests/stages/audio/tts/test_chatterbox_tts.py
```diff
@@ -0,0 +1,319 @@
...
+
+    def test_setup_raises_when_no_reference_audio(
+        self, output_dir: str, tmp_path: Path
+    ) -> None:
+        empty_dir = tmp_path / "empty"
+        empty_dir.mkdir()
+        stage = _build_stage(output_dir, str(empty_dir))
+        with patch.object(ChatterboxTTSStage, "_load_model", _inject_model):
+            with pytest.raises(ValueError, match="No reference audio found"):
+                stage.setup()
```
sarahyurick: These tests can be removed.

sarahyurick (approved):
LGTM from a Curator patterns perspective. Will let @mohammadaaftabv review for audio-specific patterns before merging.

nemo_curator/stages/audio/tts/chatterbox_tts.py
```diff
@@ -0,0 +1,573 @@
...
+import random
+import shutil
+import tempfile
+from typing import TYPE_CHECKING, Any
+
+import numpy as np
+import soundfile as sf
+import torch
+import torchaudio as ta
+from chatterbox.models.t3 import llama_configs as _llama_cfgs
```
mohammadaaftabv: chatterbox not declared in pyproject.toml; top-level import breaks optional-dep hygiene
chatterbox_tts.py:33-35 imports chatterbox at module top level; tts/__init__.py re-exports the stage; audio/__init__.py:50 eagerly imports it. Any from nemo_curator.stages.audio import ChatterboxTTSStage (or loading the audio package on the PR branch) requires chatterbox installed, but this PR adds no optional extra (cf. pyannote-audio under audio_common in pyproject.toml). This violates lens 2.2 (lazy import + declared extra) and is a recurring corpus theme (optional/lazy imports, dependency declaration).

nemo_curator/stages/audio/tts/chatterbox_tts.py
mohammadaaftabv, line 1: Need tutorial for this stage please check out existing tutorials and make one for this stage. Please follow all standardisations and structures of the existing audio tutorials.

nemo_curator/stages/audio/tts/chatterbox_tts.py
```diff
@@ -0,0 +1,573 @@
...
+            )
+        logger.info(f"Pre-downloaded ChatterboxTTS weights from {_CHATTERBOX_REPO_ID}")
+
+    def setup(self, worker_metadata: object = None) -> None:  # noqa: ARG002
+        """Load the TTS model and discover reference audio files."""
+        os.makedirs(self.output_audio_dir, exist_ok=True)
+        self._init_temp_dir()
+        self._load_model()
+        self._load_reference_audio_files()
+
```
mohammadaaftabv, line 323: missing inputs() / outputs() stage contract
ChatterboxTTSStage defines neither method. Runtime validation won't check required manifest keys; downstream stages can't discover the schema. Per nemo_curator/stages/base.py:131-158 and lens 2.1.

Fix:

    return [], ["utterance", "text", "speaker", "conversation_id"]
    return [], ["audio_filepath", "duration", "reference_voice"]
(Or declare only the keys actually required — at minimum conversation_id + one of utterance/text.)

nemo_curator/stages/audio/tts/chatterbox_tts.py
```diff
@@ -0,0 +1,573 @@
...
+    def _cleanup_temp_dir(self) -> None:
+        if self.temp_dir and os.path.exists(self.temp_dir):
+            with contextlib.suppress(OSError):
+                shutil.rmtree(self.temp_dir)
+            self.temp_dir = None
+
+    def _load_model(self) -> None:
+        """Load ChatterboxTTS or ChatterboxMultilingualTTS."""
+        if self.language:
+            os.environ["TRANSFORMERS_ATTN_IMPLEMENTATION"] = "eager"
```
mohammadaaftabv: multilingual _load_model() mutates process-global state; teardown() never restores
chatterbox_tts.py:247-249 sets os.environ["TRANSFORMERS_ATTN_IMPLEMENTATION"] = "eager" and mutates _llama_cfgs.LLAMA_CONFIGS in-place. teardown():224-231 clears the model and maps but does not revert either. Any other transformer-based stage in the same Ray/Xenna worker after a multilingual TTS run silently inherits eager attention (Greptile summary flagged this; not in the 4 OPEN inline threads).

Fix: save previous env/config in _load_model(), restore in teardown():
Add a test asserting env/config is restored after teardown().

nemo_curator/stages/audio/tts/chatterbox_tts.py
```diff
@@ -0,0 +1,573 @@
...
+        self.speaker_to_reference.clear()
+        self._speaker_to_original_wav.clear()
+        self.speaker_to_ref_id.clear()
+        self.conversation_exaggeration.clear()
+        self._cleanup_temp_dir()
+
+    def _init_temp_dir(self) -> None:
+        if self.temp_dir is None or not os.path.exists(self.temp_dir):
+            self.temp_dir = tempfile.mkdtemp(prefix="chatterbox_ref_")
+            atexit.register(self._cleanup_temp_dir)
```
mohammadaaftabv: atexit.register(self._cleanup_temp_dir) is unsafe in distributed workers
chatterbox_tts.py:236 registers an instance-bound cleanup on process exit. In Ray/Xenna multi-worker runs this leaks registrations across stage instances and may delete another worker's temp dir on interpreter shutdown.

Fix: drop atexit.register; rely on teardown() only (already called by backends for actor stages).

nemo_curator/stages/audio/tts/chatterbox_tts.py
mohammadaaftabv: Please also add readmes

nemo_curator/stages/audio/tts/chatterbox_tts.py
```diff
@@ -0,0 +1,573 @@
...
+
+        if conversation_id not in self.conversation_exaggeration:
+            lo, hi = self.exaggeration_range
+            self.conversation_exaggeration[conversation_id] = self._rng.uniform(lo, hi)
+        return self.conversation_exaggeration[conversation_id]
+
+    def _generate_turn_audio(
+        self, text: str, reference_wav: str, conversation_id: str
+    ) -> np.ndarray:
+        """Run ChatterboxTTS inference for a single turn."""
```
mohammadaaftabv: Does this have to be one-by-one inference?

nemo_curator/stages/audio/tts/chatterbox_tts.py
```diff
@@ -0,0 +1,573 @@
...
+
+        Each turn is synthesised independently (TTS is autoregressive and
+        not easily vectorised), but batching allows the model to stay warm
+        across turns and avoids repeated setup overhead.
+        """
+        if not tasks:
+            return []
+
+        output_tasks: list[AudioTask] = []
+        for task in tasks:
```
mohammadaaftabv, line 915: ChatterboxTTSStage sets Resources(gpus=1) but no batch_size. Ray Data warns and defaults GPU batch size to 1 (ray_data/adapter.py:44-46) — harmless here since turns are processed serially in process_batch(), but declaring batch_size = 1 matches convention in asr_nemo.py:51 and makes intent explicit.

Also can batch size more than 1 not work? see comment above

mohammadaaftabv (changes requested):

mohammadaaftabv (changes requested):
Please address comments

Ssofja:
@mohammadaaftabv Can you please check the changes?

pyproject.toml
```diff
@@ -116,6 +116,7 @@ audio_common = [
     "opencc-python-reimplemented",
     "pyannote-audio>=4.0.0; platform_machine == 'x86_64' and platform_system != 'Darwin'",
     "whisperx>=3.8.4; platform_machine == 'x86_64' and platform_system != 'Darwin'",
+    "chatterbox-tts>=0.1.4; platform_machine == 'x86_64' and platform_system != 'Darwin'",
```
mohammadaaftabv: This adds `chatterbox-tts>=0.1.4` to `audio_common`, while Curator requires
`transformers>=4.56.0,<5.0` at line 334. Both `uv lock --check` and a locked
`audio_cuda12` dry-run fail: Chatterbox 0.1.4-0.1.6 pin Transformers 4.46.3,
and 0.1.7 pins 5.2.0. `uv.lock` was not updated and contains no Chatterbox
entry. Because `audio_common` feeds both CPU and CUDA audio extras, this blocks
unrelated supported x86_64 Linux audio installs too. Please use a compatible release/fork
or isolated TTS runtime/dedicated extra, reconcile the Transformers constraint
for all supported Python versions, regenerate the lock, and test both extras
from a clean environment.

nemo_curator/stages/audio/tts/chatterbox_tts.py
```diff
@@ -0,0 +1,628 @@
...
+        else:
+            self.exaggeration_range = None
+            self.exaggeration = float(exaggeration)
+
+        self.model = None
+        self.reference_wavs_list: list[str] | None = None
+        self._reference_layout: str = "wavs"
+        self._speaker_audio_map: dict[str, list[str]] = {}
+
+        self.speaker_to_reference: dict[str, str] = {}
```
mohammadaaftabv, line 268: Worker-local random maps do not preserve a conversation's voice assignment
on a multi-GPU run.**

Ray Data and Xenna create multiple actor copies for this one-GPU stage and
work-steal its batch-size-one turns. Each copy owns independent
`speaker_to_reference`, `conversation_exaggeration`, and unseeded RNG state, so
two turns for the same `(conversation_id, speaker)` can reach different actors
and receive different voices/styles. A two-instance reproduction assigned two
different reference IDs to the same key. Please make assignment stateless and
deterministic, plan it before GPU fanout, or partition by conversation, and add
a multi-actor/out-of-order test.

mohammadaaftabv: Please test thoroughly on multi-gpu and multi-node settings, i am suspicious if code in current state works correctly and preserves exactly info of each row without jumbling up turns across rows

nemo_curator/stages/audio/tts/chatterbox_tts.py
```diff
@@ -0,0 +1,628 @@
...
+
+    @staticmethod
+    def _output_filename(
+        conversation_id: str, speaker: str, text: str, ref_id: str
+    ) -> str:
+        """Deterministic filename including reference voice for cache correctness."""
+        conv_hash = hashlib.md5(conversation_id.encode("utf-8"), usedforsecurity=False).hexdigest()[:12]
+        text_hash = hashlib.md5(text.encode("utf-8"), usedforsecurity=False).hexdigest()[:10]
+        ref_hash = hashlib.md5(ref_id.encode("utf-8"), usedforsecurity=False).hexdigest()[:8]
+        return f"{conv_hash}_{speaker}_{text_hash}_{ref_hash}.wav"
```
mohammadaaftabv: The filename now includes the reference ID, but it omits model/revision,
language, actual exaggeration, all sampling settings, normalization, sample
rate, and reference content. Different English and French configurations in a
local reproduction produced the identical filename, so a valid config change
silently returns stale audio. This is broader than the existing ref-ID thread.
Please hash a canonical manifest of every effective generation input plus
model and reference-content identity, validate a sidecar on hits, and publish
new cache entries atomically.

nemo_curator/stages/audio/tts/chatterbox_tts.py
```diff
@@ -0,0 +1,628 @@
...
+            audio_path = os.path.join(self.output_audio_dir, filename)
+
+            try:
+                if os.path.exists(audio_path):
+                    audio_data, _ = sf.read(audio_path)
+                else:
+                    audio_data = self._generate_turn_audio(
+                        text, reference_wav, conversation_id
+                    )
+                    sf.write(audio_path, audio_data, self.sample_rate)
```
mohammadaaftabv: **`sample_rate` changes the WAV header without resampling model output.**

Generated samples are returned unchanged and written using the user-configured
rate; duration is divided by the same value. In a model-free reproduction,
24,000 samples from a 24 kHz model written with `sample_rate=12000` became a
12 kHz file reported as 2.0 seconds, changing playback speed and pitch. Please
write at the model's native rate or explicitly resample, use the actual file
rate on cache reads, and test a non-default rate.

nemo_curator/stages/audio/tts/chatterbox_tts.py
```diff
@@ -0,0 +1,628 @@
...
+            except (OSError, RuntimeError) as e:
+                logger.warning(f"Failed to load {fpath}: {e}")
+
+        if not chunks:
+            return files[0], chosen
+
+        try:
+            concatenated = torch.cat(chunks, dim=1)
+            out_path = os.path.join(self.temp_dir, f"ref_{chosen}.wav")
+            ta.save(out_path, concatenated, last_sr)
```
mohammadaaftabv: **MLS clips with different sample rates are concatenated without resampling.**

Each clip's raw sample tensor is appended, then the whole tensor is saved using
only `last_sr`. An 8 kHz clip followed by a 16 kHz clip is accepted but the
first segment plays at twice its original speed/pitch. Please resample every
clip to one explicit reference rate before concatenation and cover mixed-rate
inputs.

tests/stages/audio/tts/test_chatterbox_tts.py
```diff
@@ -0,0 +1,404 @@
...
+    }
+    kwargs.update(overrides)
+    return ChatterboxTTSStage(**kwargs)
+
+
+def _inject_model(stage: ChatterboxTTSStage) -> None:
+    stage.model = _fake_model()
+
+
+class TestChatterboxTTSStage:
```
mohammadaaftabv, line 185: **The new implementation is below the repository's coverage target.**

There are 17 tests, not the 55 claimed in the PR description. They pass with
the Chatterbox imports mocked, but branch coverage for `chatterbox_tts.py` is
73% (333 statements, 77 missed, 86 branches, 17 partial), below the 80%
changed-code target. RTTM parsing/truncation, MLS error and mixed-rate paths,
cache corruption, normalization edges, contracts, I/O errors, and provenance
are among the missing cases. Please add focused regressions; real GPU coverage
is already called out in the current Greptile summary.

nemo_curator/stages/audio/tts/chatterbox_tts.py
```diff
@@ -0,0 +1,628 @@
...
+
+            duration = len(audio_data) / self.sample_rate
+
+            out_data = dict(data)
+            out_data["audio_filepath"] = audio_path
+            out_data["duration"] = duration
+            out_data["reference_voice"] = ref_id
+
+            output_tasks.append(
+                AudioTask(
```
mohammadaaftabv: **Successful outputs discard upstream task provenance.**

Constructing a new `AudioTask` with only data/task ID/dataset resets
`_metadata` and `_stage_perf`. The current "preserves metadata" test checks
ordinary data keys only; a reproduction changed non-empty framework metadata
and perf history to `{}` and `[]`. Please mutate and return the input task, or
copy every framework field, and assert the actual provenance fields.

Never create new audiotasks as outputs, very important just return mutated original input audiotask for any processor in audio modality in nemo curator.

tutorials/audio/tts/pipeline.py
```diff
@@ -0,0 +1,201 @@
...
+    )
+
+    pipeline.add_stage(AudioToDocumentStage())
+
+    result_dir = os.path.join(args.output_dir, "result")
+    if args.clean and os.path.isdir(result_dir):
+        shutil.rmtree(result_dir)
+
+    pipeline.add_stage(
+        JsonlWriter(
```
mohammadaaftabv: **Identical tutorial reruns create duplicate UUID-named result shards.**

`AudioToDocumentStage` drops `source_files`, so the default `JsonlWriter`
chooses a fresh UUID for every run. `--clean` defaults off, and the Hydra path
has no cleanup/output-mode control. Writing one identical converted batch twice
locally created two JSONLs. Please preserve deterministic source metadata and
choose/document one rerun policy consistently in both entry points.

tutorials/audio/tts/README.md
```diff
@@ -0,0 +1,208 @@
...
+
+## Composability
+
+The TTS stage pairs naturally with conversation generation upstream and forced alignment downstream:
+
+```python
+from nemo_curator.pipeline import Pipeline
+from nemo_curator.stages.audio.common import ManifestReader
+from nemo_curator.stages.audio.tts import ChatterboxTTSStage
+from nemo_curator.stages.audio.alignment import MFAAlignmentStage
```
mohammadaaftabv: The composability example imports code that is absent from this PR.**

`nemo_curator.stages.audio.alignment.MFAAlignmentStage` does not exist at this
head, so the example raises `ModuleNotFoundError`; the line-195
`../data-generation/` link is also missing. Please remove/defer these references,
stack on their source PR, or replace them with merged public APIs and validate
the example/link.

mohammadaaftabv (changes requested):

mohammadaaftabv:
**Blocking: please reuse the centralized YAML tutorial runner before merge.**

Current `main` already provides the shared runner in [`nemo_curator/config/run.py`](https://github.com/NVIDIA-NeMo/Curator/blob/main/nemo_curator/config/run.py). The canonical audio reference is [`tutorials/audio/qwen_omni_inprocess/`](https://github.com/NVIDIA-NeMo/Curator/tree/main/tutorials/audio/qwen_omni_inprocess): its [README](https://github.com/NVIDIA-NeMo/Curator/blob/main/tutorials/audio/qwen_omni_inprocess/README.md) explicitly states that no tutorial-specific Python runner is needed, and the tutorial contributes only `README.md` plus [`pipeline.yaml`](https://github.com/NVIDIA-NeMo/Curator/blob/main/tutorials/audio/qwen_omni_inprocess/pipeline.yaml).

Please make both #1976 and #1977 follow that structure:

1. Remove the tutorial-local `pipeline.py` and `run.py`. Do not duplicate Hydra setup, pipeline construction, executor selection, logging, cleanup, or execution inside each tutorial.
2. Keep tutorial behavior declarative in `pipeline.yaml`.
3. Rename the private `processors:` list to the shared runner’s `stages:` schema.
4. Put `backend`, Xenna `execution_mode`, stage parameters, resources, and batch sizes in the YAML.
5. Use the audio-native `ManifestWriterStage` with one explicit `output_path`, as the Qwen tutorial does, instead of `AudioToDocumentStage` + `JsonlWriter` and tutorial-specific result-directory cleanup.
6. Update the README to invoke the centralized runner:

   ```bash
     --config-path ../../tutorials/audio/<tts-or-alignment> \
     --config-name pipeline \
     ...
   ```

7. Add a configuration test following [`test_qwen_tutorial_yaml_matches_reference_runner_config`](https://github.com/NVIDIA-NeMo/Curator/blob/main/tests/config/test_run.py) that composes the real tutorial YAML through `create_pipeline_from_yaml()` / `create_executor_from_yaml()` and verifies stage order, important parameters, resources, batch size, output path, and backend.

PR-specific requirements:

- **#1976:** `ChatterboxTTSStage` already declares `batch_size = 1` and its GPU resources, so the tutorial’s `isinstance(...).with_(batch_size=1)` hook is redundant. Keep all supported synthesis controls in the YAML.
- **#1977:** the configurable MFA batch size does not justify a custom runner. Declare `batch_size: int = 256` as a configurable field on `MFAAlignmentStage` and set `batch_size: ${batch_size}` in its YAML stage entry.

These PRs currently add 611 lines of duplicated tutorial orchestration—298 in #1976 and 313 in #1977—and the copies are already drifting in backend defaults and overwrite behavior. If cleanup, timing, or another execution feature is genuinely missing, implement and test it once in the shared runner rather than adding another tutorial-specific implementation.

tests/stages/audio/tts/test_chatterbox_tts.py
```diff
@@ -0,0 +1,830 @@
...
+    """Test suite for ChatterboxTTSStage."""
+
+    @patch("chatterbox.tts.ChatterboxTTS")
```
mohammadaaftabv, line 188: tests/stages/audio/tts/test_chatterbox_tts.py:135-170` patches targets under
the real `chatterbox` package and directly imports
`chatterbox.models.t3.llama_configs`. However,
`tests/L0_Unit_Test_CPU.sh:24-36` installs `audio_cpu` and runs every non-GPU
test under `tests/stages/audio`, while this PR intentionally keeps Chatterbox
out of `pyproject.toml` and `uv.lock`. `unittest.mock.patch()` must import its
target before replacing it, so a clean-equivalent invocation reproduced
`ModuleNotFoundError: No module named 'chatterbox'` at the first decorated
test. The current GitHub check list does not include this CPU unit job.

Keep Chatterbox out of the shared extras (adding it would restore the dependency
conflict), but inject a fake Chatterbox module hierarchy through `sys.modules`
or introduce and mock a Curator-owned import/loader boundary. Add a clean
`audio_cpu` test invocation to the validation evidence

nemo_curator/stages/audio/tts/chatterbox_tts.py
```diff
@@ -0,0 +1,923 @@
...
+
+_CHATTERBOX_REPO_ID = "ResembleAI/chatterbox"
+_ENGLISH_MODEL_FILES = (
+    "ve.safetensors",
+    "t3_cfg.safetensors",
+    "s3gen.safetensors",
+    "tokenizer.json",
+    "conds.pt",
+)
+_MULTILINGUAL_MODEL_FILES = (
```
mohammadaaftabv, line 113: nemo_curator/stages/audio/tts/chatterbox_tts.py:85-92` prefetches
`t3_23lang.safetensors` and `mtl_tokenizer.json`, while the Chatterbox 0.1.7
selected by the unbounded `chatterbox-tts>=0.1.4` spec loads
`t3_mtl23ls_v2.safetensors` and
`grapheme_mtl_merged_expanded_v1.json`. The immutable upstream loader also
shows that `from_pretrained(device=...)` has no `cache_dir` argument:
https://github.com/resemble-ai/chatterbox/blob/59bc590b3cad826e5d5987745bf6844627a21ad5/src/chatterbox/mtl_tts.py#L179-L220.
Consequently, `:281-300` can report a successful node prefetch into the
user-supplied cache, but `:368-373` then searches the default Hugging Face
cache and downloads the actual assets again. Workers without egress can fail
after a nominally successful prefetch.

Pin an exact Chatterbox release and immutable model revision, prefetch the
files that release actually consumes, and load the returned snapshot with
`from_local` (or configure one identical HF cache for both phases). Add an
offline-worker test proving that actor setup performs no second download.

nemo_curator/stages/audio/tts/chatterbox_tts.py
```diff
@@ -0,0 +1,923 @@
...
+            )
+            return
+
+        msg = (
+            f"No reference audio found in {self.reference_voices_dataset}. "
+            f"Expected wavs/*/*.wav or */*/*.flac"
+        )
+        raise ValueError(msg)
+
+    def _process_audio_with_rttm(self, audio_filepath: str, rttm_filepath: str) -> str:  # noqa: C901
```
mohammadaaftabv: For the `wavs/` layout, `nemo_curator/stages/audio/tts/chatterbox_tts.py:416-467`
uses RTTM bytes to choose the conditioning waveform, but `:603-607` hashes only
the original WAV. The manifest at `:737-757` therefore stays identical when an
RTTM file changes. A reproduction changed only the RTTM segment from
0.0-0.5 s to 1.0-1.5 s: the processed reference audio changed while the cache
manifest/hash did not. The next run silently serves audio generated from the
old speech selection.

Include RTTM presence/content plus the preprocessing schema in the reference
identity, or hash a deterministic representation of the effective processed
reference. Add a regression that edits only RTTM and requires a new cache key.

nemo_curator/stages/audio/tts/chatterbox_tts.py
```diff
@@ -0,0 +1,923 @@
...
+        """Strip silences using RTTM speech segments, up to max_reference_duration."""
+        if not os.path.exists(rttm_filepath):
+            return audio_filepath
+
+        try:
+            audio, sr = ta.load(audio_filepath)
+
+            speech_segments: list[tuple[float, float]] = []
+            with open(rttm_filepath) as f:
```
mohammadaaftabv: nemo_curator/stages/audio/tts/chatterbox_tts.py:425-432` converts RTTM start
and duration fields with `float()`, but `:463-465` catches only `OSError` and
`RuntimeError`. A valid-shaped line containing `SPEAKER d 1 nope 1.0 ...`
reproduced an uncaught `ValueError`. `_assign_reference()` also runs before the
per-task guard begins at `:896`, **_so all remaining turns in that batch are lost._**

Parse each RTTM line defensively: catch `ValueError`, require finite,
non-negative start/duration values and sensible bounds, and skip/reject the
bad record with its line number. Move reference assignment/preprocessing under
the per-task error boundary and add malformed-RTTM coverage.

nemo_curator/stages/audio/io/convert.py
```diff
@@ -76,12 +76,20 @@ def process_batch(self, tasks: list[AudioTask]) -> list[DocumentBatch]:
             return []
         df = pd.DataFrame([self._sanitize(t.data) for t in tasks])
         perf = []
+        source_files: list[str] = []
         for t in tasks:
             perf.extend(t._stage_perf)
+            source_files.extend(t._metadata.get("source_files", []))
+        # Without this, writers downstream (e.g. JsonlWriter) never see
```
mohammadaaftabv: The comment at `nemo_curator/stages/audio/io/convert.py:83-87` claims that
propagating `source_files` makes identical reruns overwrite a stable shard.
But this stage aggregates N `AudioTask`s into one `DocumentBatch`; the adapter
therefore takes the ambiguous N-to-1 branch at
`nemo_curator/backends/base.py:198-204` and assigns a fresh `r<uuid>` task ID.
`BaseWriter` hashes both `source_files` and that random ID at
`nemo_curator/stages/text/io/writer/base.py:75-80`. Two identical local adapter
runs produced different IDs and therefore different downstream hashes. The new
unit tests call `process_batch()` directly and never exercise adapter ID
assignment.

Remove this unrelated converter change now that the TTS tutorial uses
`ManifestWriterStage`, or design a framework-supported deterministic identity
for N-to-1 aggregation and test converter-to-writer reruns through an executor.

tutorials/audio/tts/README.md
```diff
@@ -0,0 +1,239 @@
...
+Curator's pinned stack (`transformers>=4.56,<5.0`, `torch==2.10.0`) and would
+make the audio extras unresolvable. Instead, `ChatterboxTTSStage` declares a Ray
+`runtime_env` that pip-installs `chatterbox-tts` (plus `setuptools<81`, which
+still provides the `pkg_resources` its watermarker needs) into an **isolated
+virtualenv** at runtime, so its conflicting pins never touch the main
+environment. The first run therefore provisions that environment (one-time
+download of chatterbox and its deps); subsequent runs reuse Ray's cached
+virtualenv.
+
+> **Requires `backend=ray_data`.** Only the Ray Data (and Ray actor pool)
```
mohammadaaftabv: tutorials/audio/tts/README.md:49-53` and `:151-158` tell Xenna users to install
Chatterbox's conflicting dependencies into Curator's main environment. Current
Curator does the opposite: `nemo_curator/backends/xenna/adapter.py:29-48` wraps
the complete Ray-format environment, including `pip`, and `:83-93` returns it
for the stage. Constructing the current adapter with this stage yields a
`CuratorRuntimeEnv` whose Ray environment contains the Chatterbox package list.
The stale assertion is also repeated in `tests/config/test_run.py:626-627`.

Document that both Ray Data and Xenna honor the per-stage isolated environment,
remove the main-environment installation advice, and assert the Chatterbox
`pip` environment through both backend adapters.

mohammadaaftabv (changes requested):

Ssofja:
@mohammadaaftabv please review

PR #1980, opened 2026-05-13
caption quality evaluation

weijiac0619:
## Description

 Adds a video caption quality evaluation tool using Summarize-then-Align with [CosmosEmbed1](https://huggingface.co/collections/nvidia/cosmos-embed1).

**What it does:**
  - Summarizes VLM captions to ≤80 words using an LLM, then encodes with CosmosEmbed1 and computes cosine similarity against pre-computed video embeddings
  - Includes a one-time dataset construction script (`build_benchmark_dataset.py`) that samples videos, runs K-means clustering on embeddings, and selects 200 diverse representative clips
  - Provides baseline scores on [OpenVid-1M](https://huggingface.co/datasets/nkp37/OpenVid-1M): Qwen2.5-VL ~0.39, Qwen3-VL ~0.39, Nemotron-VL ~0.38
 
  **Code structure:**
  ```
  eval/
    video/
      README.md                      # Method, dataset, baselines, quick start
      build_benchmark_dataset.py     # One-time: sample → embed → K-means → select 200 clips
      caption_clipscore.py           # Score captions against video embeddings
  ```
  ## Usage

  ```bash
  # Score captions from multiple models
  python eval/video/caption_clipscore.py \
      --embedding-dir /path/to/ce1_embd \
      --cosmos-model-dir /path/to/models \
      --summarizer-model /path/to/Qwen2.5-14B-Instruct \
      --caption-dirs qwen25=/path/to/captions_qwen25 nemotron=/path/to/captions_nemotron \
      --save-summaries summaries.json \
      --output-csv results.csv
  ```
  See eval/video/README.md for full documentation.

## Checklist


suiyoubi:
is it possible to add [caption_clipscore.py](https://github.com/NVIDIA-NeMo/Curator/pull/1980/changes#diff-b1c156b9e6c1335c4b3d5a499d4d5a38087c89fe06f66022b4cea6654a12ba62) to the benchmark suites ? We can pre-cache the dataset, just need to make sure we have consistent clipscore for the same model

tutorials/video/eval/build_benchmark_dataset.py
```diff
@@ -0,0 +1,380 @@
...
+    """Configuration for the embedding pipeline."""
+
+    input_dir: str
+    output_dir: str
+    model_dir: str
+    gpus: str | None
+    split_duration: float
+    aesthetic_threshold: float
+    pipeline_script: str
+    ffmpeg_dir: str | None = None
```
suiyoubi: why do we need this ?

tutorials/video/eval/build_benchmark_dataset.py
```diff
@@ -0,0 +1,380 @@
...
+    ]
+    env = os.environ.copy()
+    if cfg.ffmpeg_dir is not None:
+        env["PATH"] = cfg.ffmpeg_dir + ":" + env.get("PATH", "")
+    if cfg.gpus is not None:
+        env["CUDA_VISIBLE_DEVICES"] = cfg.gpus
+    # Prevent Ray from resetting CUDA_VISIBLE_DEVICES on zero-GPU actors
+    env["RAY_ACCEL_ENV_VAR_OVERRIDE_ON_ZERO"] = "0"
+    if cfg.gpus is not None:
+        print(f"  Running pipeline on GPUs {cfg.gpus} ...")
```
suiyoubi: could we always use logger instead of print ?

tutorials/video/eval/build_benchmark_dataset.py
```diff
@@ -0,0 +1,380 @@
...
+    else:
+        print("  Running pipeline on all available GPUs ...")
+    env["LOGURU_LEVEL"] = "ERROR"
+    subprocess.run(cmd, env=env, check=True)  # noqa: S603
```
suiyoubi: I am not a big fan of using subprocess with setting bunch of env variables for running the pipeline. In stead, could you do something similar like: https://github.com/NVIDIA-NeMo/Curator/blob/main/benchmarking/scripts/video_pipeline_benchmark.py#L65C5-L65C13 to run the embedding pipeline in this script ?

eval/video/build_benchmark_dataset.py
```diff
@@ -0,0 +1,380 @@
...
+def _build_output(
+    selected: list[tuple[str, dict]],
+    pipeline_output_dir: str,
+    output_dir: str,
+) -> None:
+    """Create the benchmark output directory with symlinks."""
+    emb_dir = f"{pipeline_output_dir}/ce1_embd"
+    clip_dir = f"{pipeline_output_dir}/clips"
+
+    os.makedirs(f"{output_dir}/ce1_embd", exist_ok=True)
```
suiyoubi, line 188: do we need to create dirs for these ? these should be created by the pipeline already ? I would lean towards asserting those exists (so we make sure the embedding pipeline works )

tutorials/video/eval/build_benchmark_dataset.py
```diff
@@ -0,0 +1,380 @@
...
+        required=True,
+        help="Output directory for the benchmark dataset.",
+    )
+    parser.add_argument(
+        "--model-dir",
+        required=True,
+        help="Root model directory for the pipeline (contains CosmosEmbed1 weights etc.).",
+    )
+    parser.add_argument(
+        "--pipeline-script",
```
suiyoubi: I think this arg is useless: I don't expect this script to work with any other scripts. better keep this constant ( you can load it with rel import )

tutorials/video/eval/build_benchmark_dataset.py
```diff
@@ -0,0 +1,380 @@
...
+        default=3.5,
+        help="Minimum aesthetic score for clip filtering (default: 3.5).",
+    )
+    parser.add_argument(
+        "--gpus",
+        default=None,
+        help="Comma-separated GPU IDs for the pipeline. Uses all available GPUs if not set.",
+    )
+    parser.add_argument(
+        "--ffmpeg-dir",
```
suiyoubi: why we need to pass this ?

tutorials/video/eval/caption_clipscore.py
```diff
@@ -0,0 +1,429 @@
...
+
+def _cosine_sim(a: torch.Tensor, b: torch.Tensor) -> float:
+    a = a / a.norm()
+    b = b / b.norm()
+    return float((a * b).sum())
+
+
+def _clean_summary(text: str) -> str:
+    """Post-process LLM output: strip meta-commentary, truncate at revision."""
+    for pattern in [
```
suiyoubi: just curious if these pattern too aggressive ?

suiyoubi (changes requested):

eval/video/build_benchmark_dataset.py
```diff
@@ -0,0 +1,330 @@
+# Copyright (c) 2026, NVIDIA CORPORATION.  All rights reserved.
```
VibhuJawa, line 1: I think we should move this to a new head folder folder  called eval and then have `eval/video` and move it there. 
Like eval should become first class citizen of curator (and code that we have to maintain going forward) . I am sure there will be utils we will use across benchmarks so will be easier for code structuring (similar to how we do for benchmarking) and automate it too (not today but later on)

VibhuJawa (changes requested):
Thanks again @weijiac0619  for working on this. I had suggestions around code organization. 

I think we should move this outside tutorials to a new eval subfolder ! In that eval folder we should have a eval.video.caption and place this here. 

The other main ask is if we can use a huggingface dataset here as input so that our customers and we have a opensource dataset we can use . 

The other unknown is if we can have some expected baseline results or some guidance on what a good vs bad resulted and an ability to check that in an automated way.

weijiac0619:
> is it possible to add [caption_clipscore.py](https://github.com/NVIDIA-NeMo/Curator/pull/1980/changes#diff-b1c156b9e6c1335c4b3d5a499d4d5a38087c89fe06f66022b4cea6654a12ba62) to the benchmark suites ? We can pre-cache the dataset, just need to make sure we have consistent clipscore for the same model

Good idea. Will add caption quality scoring as a standalone benchmark entry in a follow-up. plan to wire up the 200-clip dataset + captioning + scoring as a single entry, and validate the pass/fail thresholds on nightly.

weijiac0619:
@VibhuJawa thanks for the comments. Code moved to `eval/video/`. HF dataset upload and automated baseline validation will be follow-up.

eval/video/README.md
```diff
@@ -0,0 +1,135 @@
+# Video Caption Quality Evaluation
+
+Use this alignment score as a fast regression signal to capture large semantic
+drift. Run scoring after model updates or pipeline refactors.
+
+## Method
+
+CosmosEmbed1 is a video-text embedding model (ViT + QFormer) with understanding
```
VibhuJawa: Can we add a HF link here  ,  also this is a collection of models right ? 

[Cosmos-Embed1-224p](https://huggingface.co/nvidia/Cosmos-Embed1-224p) (optimized with 8 frames and 224x224 input resolution, 256-dim output text and video embeddings)?

[Cosmos-Embed1-336p](https://huggingface.co/nvidia/Cosmos-Embed1-336p) (optimized with 8 frames and 336x336 input resolution, 768-dim output text and video embeddings)

[Cosmos-Embed1-448p](https://huggingface.co/nvidia/Cosmos-Embed1-448p) (optimized with 8 frames and 448x448 input resolution, 768-dim output text and video embeddings)

eval/video/README.md
```diff
@@ -0,0 +1,135 @@
...
+## Quick Start
+
+### 1. Build benchmark dataset (one-time)
+
+Samples videos from a large dataset, generates CosmosEmbed1 embeddings, and
+selects diverse representatives via K-means clustering.
+
+```bash
+python build_benchmark_dataset.py \
+    --video-dir /path/to/video_dataset \
```
VibhuJawa, line 65: Is there a public dataset , our customers can use, can we link/test with that here please .

eval/video/build_benchmark_dataset.py
weijiac0619: This script is run once to build the cached benchmark dataset, not part of nightly.
Pipeline output dirs are already asserted (lines 181-186). The os.makedirs create the benchmark output directory structure. these don't exist yet, so they need to be created.

eval/video/README.md
```diff
@@ -0,0 +1,135 @@
...
+    --embedding-dir /path/to/captions_new_model/ce1_embd \
+    --cosmos-model-dir /path/to/models \
+    --summarizer-model /path/to/Llama-3.1-8B-Instruct \
+    --caption-dirs new_model=/path/to/captions_new_model \
+    --uid-list /path/to/benchmark_200/selected_uids.txt \
+    --save-summaries /path/to/captions_new_model/summaries.json \
+    --output-csv new_model_results.csv
+```
+
+Compare the new model score against the cached baseline scores directly.
```
VibhuJawa, line 157: What is this comparison, what is expected number here, how much of deviation is allowed ? Could we add some details and possibly a script here.

weijiac0619: added baseline scores section.

eval/video/README.md
weijiac0619: Added CosmosEmbed1 HF links with variants, OpenVid-1M public dataset link

eval/video/build_benchmark_dataset.py
```diff
@@ -0,0 +1,330 @@
...
+
+    uid_to_meta = {}
+    for uid in uids:
+        meta_path = f"{meta_dir}/{uid}.json"
+        if os.path.exists(meta_path):
+            with open(meta_path) as fh:
+                uid_to_meta[uid] = json.load(fh)
+
+    logger.info(f"K-means K={num_clusters} ...")
+    kmeans = KMeans(n_clusters=num_clusters, random_state=seed, n_init=10, max_iter=300)
```
sarahyurick: Should these be configurable?
```suggestion
    kmeans = KMeans(n_clusters=num_clusters, random_state=seed, n_init=n_init, max_iter=max_iter)
```

eval/video/caption_clipscore.py
```diff
@@ -0,0 +1,428 @@
...
+    tasks: list[tuple[str, str, str]],
+    summaries: list[str],
+    vid_emb_cache: dict[str, torch.Tensor],
+    cosmos_model_dir: str,
+    variant: str,
+) -> dict[str, dict[str, float]]:
+    """Encode summaries with CosmosEmbed1 and compute per-clip scores."""
+    logger.info(f"Loading CosmosEmbed1-{variant} from {cosmos_model_dir} ...")
+    model = CosmosEmbed1(variant=variant, utils_only=False, model_dir=cosmos_model_dir)
+    model.setup()
```
sarahyurick, line 229: Could it be worth doing any of this in a Curator pipeline or no?

eval/video/README.md
```diff
@@ -0,0 +1,157 @@
...
+200-clip benchmark subset via the K-means procedure described above.
+
+## Baseline Scores
+
+Expected scores: Qwen2.5-VL and Qwen3-VL ~0.39, Nemotron-VL ~0.38. If
+scores drop >5% from these baselines, spot-check the 10 worst-scoring clips to
+determine whether it is a real quality regression.
+
+Scores are deterministic across runs on the same machine when using
+`enforce_eager=True` and `CUBLAS_WORKSPACE_CONFIG=:4096:8`.
```
VibhuJawa, line 54: QQ: Is the understanding that these result in same exact scores, but without it we get some variation but that variation is within 5% ? 

@claude , Can you suggest better phrasing for Baseline Scores ?

sarahyurick (commented):
Hi, just familiarizing myself with the PR and left minor comments if they are useful. This is awesome work, thank you!

eval/video/build_benchmark_dataset.py
```diff
@@ -0,0 +1,330 @@
...
+
+# Import the pipeline creation function from the video tutorial
+_REPO_ROOT = Path(__file__).parent.parent.parent
+sys.path.insert(0, str(_REPO_ROOT / "tutorials" / "video" / "getting-started"))
+
+from video_split_clip_example import (  # noqa: E402
+    create_video_splitting_argparser,
+    create_video_splitting_pipeline,
+)
+
```
VibhuJawa, line 66: @suiyoubi / @weijiac0619,  Should we move/copy this code to eval or some other  place ?

Like it feels generally useful for other evals will use this and benchmarks use this too. 

Feels like an awkward place to pull getting started code for these functions  ?

eval/video/build_benchmark_dataset.py
weijiac0619: Hi @sarahyurick , These are scikit-learn's defaults (n_init=10, max_iter=300), which means this is equivalent to KMeans(n_clusters=num_clusters, random_state=seed) without specifying them. They're explicit in the code just so readers know the values at a glance.

eval/video/caption_clipscore.py
weijiac0619: This is a standalone evaluation tool instead of a data processing step. i'm worried that wrapping it in a Curator pipeline would add overhead

eval/video/build_benchmark_dataset.py
weijiac0619: agree but video_split_clip_example.py is the existing pipeline entry point. what do you think about folder structure? @suiyoubi

suiyoubi: I think moving it to `eval/` and use `sys.path.insert` is fine

eval/video/build_benchmark_dataset.py
suiyoubi: I think @sarahyurick 's point is to also make these two args configurable. If you think we should not/no need to let the user configure this, I would suggest just keep it `KMeans(n_clusters=num_clusters, random_state=seed)` for simplicity .

eval/video/build_benchmark_dataset.py
suiyoubi: when you run `_run_embedding_pipeline` (in step2), isn't these dirs should be created already ?

weijiac0619:
HI @suiyoubi , @VibhuJawa and @sarahyurick , I believe the comments are resolved. could you take another look?

eval/video/build_benchmark_dataset.py
VibhuJawa: >  think moving it to eval/ and use sys.path.insert is fine

We can do this later on too , But are we saying we will use this this for tutorials or from eval ?

weijiac0619: Hi @VibhuJawa , do you mean moving the two scrips in my PR to eval, or video_split_clip_example.py to eval? I have done the former

VibhuJawa: Asking if we should move video_split_clip_example to eval or utils. **Code that is supposed to  be used for actual features (like eval), should not need to import tutorials. Also will need test cases !!!! **

VibhuJawa (changes requested):
Lets add some test cases please. Ensure at least some parts of the scripts run under CI. 

Overall looks great ! Thanks again @weijiac0619  for doing this work . Lets just move things around, and ensure code coverage.

eval/video/build_benchmark_dataset.py
suiyoubi: @VibhuJawa `video_split_clip_example.py` was originally created as a reference for how to use the video pipeline. I think we should still have it in the tutorial folder to demonstrate the user for how to use the video pipeline. It has also been imported by the benchmark subdir for running our nightly benchmark as well.

I would suggest to keep it in the tutorial folder for now, we can decide whether to refactor it somewhere else in a separate PR ( i don't think it should lands in the `eval/`, if we want to refactor, it should be landed in somewhere more centeral)

VibhuJawa: Sounds good. Agreed. 

@claude , can you create an issue for us to track this please.

tests/eval/video/conftest.py
```diff
@@ -0,0 +1,7 @@
+"""Configure sys.path for eval test imports."""
+
+import os
+import sys
+
+_REPO_ROOT = os.path.abspath(os.path.join(os.path.dirname(__file__), "..", "..", ".."))
+sys.path.insert(0, os.path.join(_REPO_ROOT, "eval", "video"))
```
VibhuJawa: @thomasdhc , Could you guide here, wondering what is a good way to get this under CI , without packaging the `.eval` things for now ?

suiyoubi (approved):
LGTM, thanks for the hard work

tests/eval/video/conftest.py
weijiac0619: @thomasdhc should I  add eval/ to pytest's pythonpath in pyproject.toml, or add the test to the existing CPU test script (tests/L0_Unit_Test_CPU.sh)

thomasdhc: @VibhuJawa @weijiac0619 Please add the new folder path here: https://github.com/NVIDIA-NeMo/Curator/blob/main/.github/workflows/cicd-main.yml#L65

VibhuJawa: @weijiac0619 , i think  we should not do this but something like this : 
```
 from eval.video.caption_clipscore import
```

VibhuJawa (commented):
Minor ask to ensure we do import cleanly in the lib

weijiac0619:
> Member

@VibhuJawa just resolved in the new commit

VibhuJawa (approved):
LGTM

PR #1981, opened 2026-05-14
fix: fail fast for empty Ray Data fanout

### LINKED ISSUE #1387: Text readers hang instead of failing fast when using `RayDataExecutor`
I noticed that if I am using the `RayDataExecutor` and I accidentally set the wrong input data path or the wrong file reader (e.g., try to use `JsonlReader` on a directory of Parquet files), it doesn't fail. It just runs forever with 0 tasks, 0 actors, 0 resources, 0B object store...

Minimal reproducer:
```
from nemo_curator.backends.experimental.ray_data import RayDataExecutor
from nemo_curator.stages.text.io.reader import JsonlReader

pipeline = Pipeline(name="test")
# Incorrectly use JsonlReader on a directory of Parquet files
pipeline.add_stage(JsonlReader("/path/to/parquet/dir"))
executor = RayDataExecutor()
result = pipeline.run(executor=executor)
```

Note that the above example fails fast when using the `XennaExecutor`, which is the desired behavior.

nightcityblade:
## Description
Closes #1387.

This makes `RayDataExecutor` eagerly run an initial fanout stage that starts from the default `EmptyTask`. If that stage produces no tasks (for example, a text reader points at a missing path or a directory with non-matching file extensions), the executor raises a `ValueError` before constructing the Ray Data pipeline instead of hanging with no work scheduled.

## Usage
```python
from nemo_curator.backends.experimental.ray_data import RayDataExecutor
from nemo_curator.stages.text.io.reader import JsonlReader

pipeline = Pipeline(name="test")
pipeline.add_stage(JsonlReader("/path/to/parquet/dir"))
pipeline.run(executor=RayDataExecutor())  # raises ValueError when no .jsonl files match
```

## Checklist

## Testing
- `uv run ruff check nemo_curator/backends/experimental/ray_data/executor.py tests/backends/experimental/ray_data/test_max_calls_pid.py`
- `uv run python -m py_compile nemo_curator/backends/experimental/ray_data/executor.py tests/backends/experimental/ray_data/test_max_calls_pid.py`
- `git diff --check`

Note: I attempted the targeted pytest locally, but this machine is macOS and NeMo-Curator raises its Linux-only platform guard during test collection.

### DIFF
--- a/nemo_curator/backends/experimental/ray_data/executor.py
+++ b/nemo_curator/backends/experimental/ray_data/executor.py
@@ -19,7 +19,7 @@ from loguru import logger
 from ray.data import DataContext, Dataset
 
 from nemo_curator.backends.base import BaseExecutor
-from nemo_curator.backends.experimental.utils import execute_setup_on_node
+from nemo_curator.backends.experimental.utils import RayStageSpecKeys, execute_setup_on_node
 from nemo_curator.backends.utils import register_loguru_serializer
 from nemo_curator.tasks import EmptyTask, Task
 
@@ -60,8 +60,9 @@ class RayDataExecutor(BaseExecutor):
         # This prevents verbose logging from Ray Data about serialization of the dataclass
         DataContext.get_current().enable_fallback_to_arrow_object_ext_type = True
         # Initialize with initial tasks if provided, otherwise start with EmptyTask
-        tasks: list[Task] = initial_tasks if initial_tasks else [EmptyTask]
+        tasks: list[Task] = initial_tasks or [EmptyTask]
         output_tasks: list[Task] = []
+        stages_to_run = stages
         try:
             # Initialize ray and explicitly set NOSET to empty
             # This ensures if Xenna was used before which was setting NOSET, we end up overriding it.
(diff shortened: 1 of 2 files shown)


nemo_curator/backends/experimental/ray_data/executor.py
```diff
@@ -101,6 +107,26 @@ def execute(self, stages: list["ProcessingStage"], initial_tasks: list[Task] | N
             ray.shutdown()
         return output_tasks
 
+    def _get_initial_fanout_tasks(self, stage: "ProcessingStage") -> list[Task]:
+        """Run the first fanout stage eagerly so empty input fails before Ray Data execution."""
+        execute_setup_on_node([stage], ignore_head_node=self.ignore_head_node)
+        tasks = stage.process(EmptyTask)
```
sarahyurick, line 113: This is similar to how another user originally proposed to do it in https://github.com/NVIDIA-NeMo/Curator/pull/1516, but I don't think we should be running the first stage on the driver node like this.

Instead, maybe consider adding the fix in the file partitioning stage and/or reader logic?

sarahyurick (changes requested):

nightcityblade:
Thanks for the review — I agree the driver-side execution approach is the wrong layer for this fix. Rather than keep the PR open with that design, I'm going to close it and revisit the behavior from the file partitioning / reader side in a smaller follow-up.

nightcityblade:
Closing this version because the requested fix belongs lower in the file partitioning / reader path, not in driver-side executor logic.

PR #1993, opened 2026-05-16
Add unique and deterministic id generation

oyilmaz-nvidia:
### Why

`Task._uuid` is generated by `uuid.uuid4()` on construction, so running the same pipeline twice on the same inputs produces a completely different set of IDs. That makes any kind of caching, checkpointing, or "resume from intermediate output" impossible — and it already affects real code: dedup stages name Parquet files after `_uuid` ([minhash.py:309](nemo_curator/stages/deduplication/fuzzy/minhash.py#L309), [buckets_to_edges.py:83](nemo_curator/stages/deduplication/fuzzy/buckets_to_edges.py#L83), [kmeans.py:235](nemo_curator/stages/deduplication/semantic/kmeans.py#L235)), and `AddId` prefixes document IDs with it ([add_id.py:71](nemo_curator/stages/text/modules/add_id.py#L71)).

### What changed

Two new fields on `Task`, derived from the task's path through the pipeline DAG:

| Field | Meaning |
|---|---|
| `_uuid` | **Unchanged** — still `uuid.uuid4()`. There are existing calls and existing call sites keep working. |
| `_lineage_path` | Underscore-separated index path through the DAG. Propagated to children. E.g. `"3_0_7"` = 4th root, 1st child, 8th grandchild. |
| `_udid` | `sha256(_lineage_path)[:32]` — 32-char hex, deterministic across runs. Use this for cache keys / output filenames. |

Lineage assignment happens at one chokepoint — the default `ProcessingStage.process_batch` calls a new module-level helper, `nemo_curator.stages.base.assign_child_lineage(parent_paths, result)`, on the output of each `process()` call. Stages that override `process_batch` are responsible for calling the helper themselves (documented in the docstring); if they forget, outputs have empty `_lineage_path`/`_udid` — a loud, easy-to-catch failure rather than a silent mis-id.

### Examples

| `_lineage_path` | `_udid` |
|---|---|
| `"3"` (root task, 4th output of first stage) | `4e07408562bedb8b60ce05c1decfe3ad` |
| `"3_0"` (its single child) | `sha256("3_0")[:32]` |
| `"3_0_7"` (further descendant) | `sha256("3_0_7")[:32]` |
| `"0_0_1_0_2_0_0"` (fan-in of 3 parents + idx 0) | `sha256(...)[:32]` |

### Files

| File | Change |
|---|---|
| [`nemo_curator/tasks/tasks.py`](nemo_curator/tasks/tasks.py) | Add `_lineage_path` and `_udid` fields + `_set_lineage(parents, idx)` helper on `Task`. `_uuid` untouched. |
| [`nemo_curator/stages/base.py`](nemo_curator/stages/base.py) | Add module-level `assign_child_lineage(parent_paths, result)` helper; default `process_batch` delegates to it; docstring updates the override contract. |
| [`tests/tasks/test_tasks.py`](tests/tasks/test_tasks.py) | Added `test_lineage_path_and_udid_format`, `test_fanout_udid_from_empty_root`, `test_udid_deterministic_across_runs`. Original `_uuid` uniqueness test kept as-is. |
| [`tests/pipelines/test_pipelines.py`](tests/pipelines/test_pipelines.py) | Added `test_pipeline_udid_deterministic_across_runs` and `test_pipeline_udid_fanout_passthrough_fanin_passthrough` (4-stage topology: fan-out → passthrough → fan-in → passthrough; exercises the multi-parent code path). |

### Trade-offs documented in code

- **Hyperparameter changes do not invalidate `_udid`.** Same DAG shape ⇒ same `_udid`. Clear the cache or version output dirs when changing stage config.
- **Stages overriding `process_batch` must call `assign_child_lineage` themselves.** No magic safety net.
- **`_uuid` stays random.** Anything reading `_uuid` today keeps the same per-run behaviour; new code that needs stability reads `_udid`.

### Test plan

- [x] New unit tests cover root tasks, single-parent fanout, multi-parent join, and end-to-end determinism across two independent runs of the same pipeline.
- [x] Original `test_fanout_tasks_have_unique_uuid` still passes (back-compat for `_uuid` confirmed).
- [x] Writer tests in `tests/stages/text/io/writer/` untouched and still pass — their `mock_uuid4.call_count` assertions remain valid because `_uuid` is unchanged.

### DIFF
--- a/nemo_curator/pipeline/pipeline.py
+++ b/nemo_curator/pipeline/pipeline.py
@@ -17,7 +17,7 @@ from typing import Any
 from loguru import logger
 
 from nemo_curator.backends.base import BaseExecutor
-from nemo_curator.stages.base import CompositeStage, ProcessingStage, assign_root_lineage
+from nemo_curator.stages.base import CompositeStage, ProcessingStage
 from nemo_curator.tasks import Task
 
 
@@ -213,5 +213,7 @@ class Pipeline:
                 )
 
         if initial_tasks:
-            assign_root_lineage(initial_tasks)
+            # Assign deterministic root-level lineage to initial pipeline tasks
+            for i, task in enumerate(initial_tasks):
+                task._set_lineage([], i)
         return executor.execute(self.stages, initial_tasks)
--- a/nemo_curator/pipeline/pipeline.py
+++ b/nemo_curator/pipeline/pipeline.py
@@ -17,7 +17,7 @@ from typing import Any
 from loguru import logger
 
 from nemo_curator.backends.base import BaseExecutor
-from nemo_curator.stages.base import CompositeStage, ProcessingStage
+from nemo_curator.stages.base import CompositeStage, ProcessingStage, assign_root_lineage
 from nemo_curator.tasks import Task
 
 
@@ -212,4 +212,6 @@ class Pipeline:
                     "The executor will schedule GPU stages on GPUs not held by Serve."
                 )
 
+        if initial_tasks:
+            assign_root_lineage(initial_tasks)
         return executor.execute(self.stages, initial_tasks)
--- a/nemo_curator/stages/base.py
+++ b/nemo_curator/stages/base.py
@@ -35,6 +35,37 @@ Y = TypeVar("Y", bound=Task)  # Output task type
 _STAGE_REGISTRY: dict[str, type[ProcessingStage]] = {}
 
 
+def assign_child_lineage(
+    parent_paths: list[str],
+    result: Task | list[Task] | None,
+) -> list[Task]:
+    """Normalize a stage's ``process()`` result and assign deterministic lineage.
+
+    Each surviving ``children[i]`` gets ``_lineage_path`` and ``_uuid`` derived
+    from ``(parent_paths, i)`` so that the same pipeline run twice on the same
+    inputs produces byte-identical task IDs. Call this from any custom
+    ``process_batch`` override to keep outputs consistent with the rest of the
+    pipeline.
+
+    Args:
+        parent_paths: One element per logical parent (typically
+            ``[task._lineage_path]`` for 1:N stages, or multiple paths for
+            join/aggregate stages).
+        result: Whatever ``process()`` (or your custom batch logic) returned for
+            this parent set — a single task, a list, or ``None``.
+
+    Returns:
+        The normalized list of children with lineage assigned. May be empty.
+    """
+    if result is None:
+        return []
+    children = result if isinstance(result, list) else [result]
+    children = [c for c in children if c is not None]
+    for i, child in enumerate(children):
+        child._set_lineage(parent_paths, i)
+    return children
+
+
 class StageMeta(ABCMeta):
     """Metaclass that automatically registers concrete Stage subclasses.
     A class is considered *concrete* if it directly inherits from
@@ -179,9 +210,21 @@ class ProcessingStage(ABC, Generic[X, Y], metaclass=StageMeta):
             - Single task: For 1-to-1 transformations
             - List of tasks: For 1-to-many transformations
             - None: If the task should be filtered out
-        Note: The returned list should have the same length as the input list,
-        with each element corresponding to the result of processing the task
-        at the same index.
+
+        Lineage contract: every emitted child must have its ``_lineage_path``
+        and ``_uuid`` set so the pipeline produces deterministic IDs. The
+        default implementation below delegates to
+        :func:`assign_child_lineage` per input task. If you override this
+        method, you are responsible for calling ``assign_child_lineage`` on
+        each chunk of outputs that share parentage, e.g.::
+
+            outputs = []
+            for task in tasks:
+                raw = self.my_batched_process(task)
+                outputs.extend(assign_child_lineage([task._lineage_path], raw))
+            return outputs
+
+        Outputs that skip this step will carry empty ``_uuid``/``_lineage_path``.
         """
         # Default implementation: process tasks one by one
         # This is only used as a fallback if a stage doesn't override this method
@@ -192,10 +235,7 @@ class ProcessingStage(ABC, Generic[X, Y], metaclass=StageMeta):
                 raise ValueError(msg)
 
             result = self.process(task)
-            if isinstance(result, list):
-                results.extend(result)
-            else:
-                results.append(result)
+            results.extend(assign_child_lineage([task._lineage_path], result))
         return results
 
     def setup_on_node(self, node_info: NodeInfo | None = None, worker_metadata: WorkerMetadata | None = None) -> None:
--- a/nemo_curator/stages/base.py
+++ b/nemo_curator/stages/base.py
@@ -41,11 +41,16 @@ def assign_child_lineage(
 ) -> list[Task]:
     """Normalize a stage's ``process()`` result and assign deterministic lineage.
 
-    Each surviving ``children[i]`` gets ``_lineage_path`` and ``_uuid`` derived
+    Each surviving ``children[i]`` gets ``_lineage_path`` and ``_udid`` derived
     from ``(parent_paths, i)`` so that the same pipeline run twice on the same
     inputs produces byte-identical task IDs. Call this from any custom
     ``process_batch`` override to keep outputs consistent with the rest of the
     pipeline.
+
+    Children whose ``_udid`` is already set are passed through unchanged. This
+    happens when a stage mutates and returns the same task instance it received
+    (e.g. an embedder that writes results onto the input task): the framework
+    must not treat such a task as a new child of itself.
 
     Args:
         parent_paths: One element per logical parent (typically
@@ -212,7 +217,7 @@ class ProcessingStage(ABC, Generic[X, Y], metaclass=StageMeta):
             - None: If the task should be filtered out
 
         Lineage contract: every emitted child must have its ``_lineage_path``
-        and ``_uuid`` set so the pipeline produces deterministic IDs. The
+        and ``_udid`` set so the pipeline produces deterministic IDs. The
         default implementation below delegates to
         :func:`assign_child_lineage` per input task. If you override this
         method, you are responsible for calling ``assign_child_lineage`` on
@@ -224,7 +229,12 @@ class ProcessingStage(ABC, Generic[X, Y], metaclass=StageMeta):
                 outputs.extend(assign_child_lineage([task._lineage_path], raw))
             return outputs
 
-        Outputs that skip this step will carry empty ``_uuid``/``_lineage_path``.
+        In-place returns are supported: if ``process()`` mutates and returns the
+        same task it received, ``assign_child_lineage`` will preserve that
+        task's existing ``_lineage_path`` / ``_udid`` rather than treating it as
+        a new child of itself.
+
+        Outputs that skip this step will carry empty ``_udid``/``_lineage_path``.
         """
         # Default implementation: process tasks one by one
         # This is only used as a fallback if a stage doesn't override this method
--- a/nemo_curator/stages/base.py
+++ b/nemo_curator/stages/base.py
@@ -71,22 +71,6 @@ def assign_child_lineage(
     return children
 
 
-def assign_root_lineage(tasks: list[Task]) -> list[Task]:
-    """Assign deterministic root-level lineage to initial pipeline tasks.
-
-    Each ``tasks[i]`` gets ``_lineage_path = str(i)`` and the corresponding
-    ``_udid``. Tasks whose ``_udid`` is already set are left untouched
-    (``_set_lineage`` early-returns), so calling this twice is a no-op.
-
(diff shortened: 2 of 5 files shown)


oyilmaz-nvidia:
@VibhuJawa @praateekmahajan @sarahyurick Could you please review this PR when you a chance?

VibhuJawa (commented):
So the logic looks sound to me but will really lean on @praateekmahajan  because i am not familiar with this side of the codebase.

nemo_curator/stages/base.py
```diff
@@ -35,6 +35,58 @@
...
+    if result is None:
+        return []
+    children = result if isinstance(result, list) else [result]
+    children = [c for c in children if c is not None]
+    for i, child in enumerate(children):
+        child._set_lineage(parent_paths, i)
+    return children
+
+
+def assign_root_lineage(tasks: list[Task]) -> list[Task]:
```
sarahyurick: Nit but should this function live in `pipeline.py` instead? Since that is the only place it is being used right?

sarahyurick (commented):
Logic LGTM. Left a minor comment.

nemo_curator/stages/base.py
```diff
@@ -192,10 +245,7 @@ def process_batch(self, tasks: list[X]) -> list[Y]:
                 raise ValueError(msg)
 
             result = self.process(task)
-            if isinstance(result, list):
-                results.extend(result)
-            else:
-                results.append(result)
+            results.extend(assign_child_lineage([task._lineage_path], result))
```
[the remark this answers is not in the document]
sarahyurick: I am also a bit tripped up about the counting part, but I think the concern about the collision is wrong, I don't see a collision happening because either:

1. we start with a `[EmptyTask]` and so it produces e.g., a list of `FileGroupTask`s which will be handled correctly
2. we start with something else, in which case it is assigned a root lineage in `pipeline.py`

But perhaps I am thinking too narrowly about the possibilities of scenario 1.

nemo_curator/stages/base.py
oyilmaz-nvidia: In the next PRs, we'll have related functions in the base stage, and they are all for resumability.

oyilmaz-nvidia:
@sarahyurick could you please approve? @praateekmahajan could you please review?

nemo_curator/tasks/tasks.py
```diff
@@ -40,11 +41,31 @@ class Task(ABC, Generic[T]):
     _stage_perf: list[StagePerfStats] = field(default_factory=list)
     _metadata: dict[str, Any] = field(default_factory=dict)
     _uuid: str = field(init=False, default_factory=lambda: str(uuid.uuid4()))
+    # `_lineage_path` is the index-based path of this task through the pipeline
+    # DAG (e.g. "3_0_7" = 4th root task, then 1st child, then 8th grandchild).
+    # It is propagated to children and hashed into `_udid`, the deterministic
+    # task id.
+    _lineage_path: str = field(init=False, default="")
+    _udid: str = field(init=False, default="")
```
praateekmahajan, line 49: Instead of adding `udid` on top of `uuid`, `task_id` I think we can just keep one of them and remove others. I believe uuid is only used by add id and I'm pretty sure it'll be unaffected too

nemo_curator/tasks/tasks.py
```diff
@@ -40,11 +41,31 @@ class Task(ABC, Generic[T]):
...
 
     def __post_init__(self) -> None:
         """Post-initialization hook."""
         self.validate()
 
+    def _set_lineage(self, parent_lineage_paths: list[str], child_index: int) -> bool:
+        """Assign deterministic lineage to this task.
+
+        Returns ``True`` if lineage was newly assigned, ``False`` if ``_udid``
+        was already set — which signals the task was returned in place by an
```
praateekmahajan:
```suggestion
        was already set - which signals the task was returned in place by an
```

Claude em-dashes lol

nemo_curator/tasks/tasks.py
oyilmaz-nvidia: Some of the stages use the uuid so I couldn't touch it. Task_id is an integer number and not sure if we should touch it.

nemo_curator/stages/base.py
sarahyurick: So it mainly is because you want to keep all lineage related things together? After taking a quick look at 1994, 1995, 1997, it doesn't look like it is being used anywhere else right? Or maybe it should wind up in `lineage_store.py`?

Again is not a blocker IMO, just asking because I was initially a bit confused while scrolling this PR.

sarahyurick: I mean like, it seems we technically don't need the function at all and could just put the for loop as-is in the `pipeline.py` script?

nemo_curator/tasks/tasks.py
praateekmahajan: Can you share which ones, I can help decide if we can yank them. Would prefer to not add new _id fiields. Similarly for task_id..  Would want to avoid as much as possible to create new fields without depreciating similar old fields..

tests/pipelines/test_pipelines.py
```diff
@@ -69,3 +73,304 @@ def test_raises_when_ray_serve_active_with_xenna_and_gpu_stages() -> None:
...
+            _SimpleTask(
+                task_id=f"{task.task_id}_{i}",
+                dataset_name=task.dataset_name,
+                data=task.data,
+            )
+            for i in range(self.times)
+        ]
+
+
+def _drive(pipeline: Pipeline, initial_tasks: list[Task]) -> list[Task]:
```
praateekmahajan, line 115: I think this test might better live inside `tests/backends/test_integration.py` within backends or you can create a new one. 

Instead of having `drive` as the functionality, let's just run the pipeline how users run without interfacing with these lower-level functionality, primarily because of tomorrow the contract changes then we don't want to be testing things with a lower level utility

tests/pipelines/test_pipelines.py
```diff
@@ -69,3 +73,304 @@ def test_raises_when_ray_serve_active_with_xenna_and_gpu_stages() -> None:
...
+    current = initial_tasks
+    for stage in pipeline.stages:
+        current = BaseStageAdapter(stage).process_batch(current)
+    return current
+
+
+def test_pipeline_udid_deterministic_across_runs():
+    def run_once() -> tuple[list[str], list[str]]:
+        pipeline = Pipeline(name="det", stages=[_Repeat(times=2), _Repeat(times=3)])
+        pipeline.build()
```
praateekmahajan, line 129: `build()` is a no-op so far, AI continues to consider them important.. Same as above, let's test how users run our pipelines. i..e. `Pipeiline(...).run()` and then check the output tasks.

tests/pipelines/test_pipelines.py
```diff
@@ -69,3 +73,304 @@ def test_raises_when_ray_serve_active_with_xenna_and_gpu_stages() -> None:
...
+        return assign_child_lineage([t._lineage_path for t in tasks], merged)
+
+
+def test_pipeline_udid_fanout_passthrough_fanin_passthrough():
+    """End-to-end: a 4-stage pipeline that exercises 1:N, 1:1, N:1, 1:1 and
+    verifies the exact `_lineage_path` / `_udid` values at every step.
+
+    Pipeline topology:
+
+        Input ─▶ FanOut(3) ─▶ Passthrough ─▶ FanIn ─▶ Passthrough ─▶ Output
```
praateekmahajan, line 226: super nit : How is ruff passing on these characters?
Can we keep it to `Input -> FanOut(3) -> Passthrough -> FanIn`...

nemo_curator/stages/base.py
oyilmaz-nvidia: Yeah, it's a simple function. I can just move that code into pipeline. I am Ok with that. Will update it.

tests/pipelines/test_pipelines.py
oyilmaz-nvidia: Not sure but it's passing :D

tests/pipelines/test_pipelines.py
```diff
@@ -69,3 +73,304 @@ def test_raises_when_ray_serve_active_with_xenna_and_gpu_stages() -> None:
...
+    pipeline.build()
+
+    root = _SimpleTask(task_id="r", dataset_name="d", data=[1])
+    assign_root_lineage([root])
+
+    # Drive stage-by-stage so we can inspect each intermediate set of tasks.
+    after_fanout = BaseStageAdapter(pipeline.stages[0]).process_batch([root])
+    after_passthrough_1 = BaseStageAdapter(pipeline.stages[1]).process_batch(after_fanout)
+    after_fanin = BaseStageAdapter(pipeline.stages[2]).process_batch(after_passthrough_1)
+    after_passthrough_2 = BaseStageAdapter(pipeline.stages[3]).process_batch(after_fanin)
```
praateekmahajan, line 254: I think there is a bug here.. your `FanIn` stage if you do `pipeline.run()` isn't technically a fan-in..

Once you modify your tests to how I shared above you'll see, that the output tasks woudn't be what you expect. 

For a fanin to be you need to define `batch_size` property on the stage. So a good reason to move the tests to follow similar pattern to `tests/backends/test_integration.py` 
I could be wrong but some of these fan outs and fan ins might already be there with some other properties being tested. So if we can reuse existing tests while covering same functionality I'd prefer that 100%

nemo_curator/tasks/tasks.py
oyilmaz-nvidia: Ok, will use task_id and won't add the _udid. Will update the code.

nemo_curator/tasks/tasks.py
```diff
@@ -40,11 +41,31 @@ class Task(ABC, Generic[T]):
...
+
+        Returns ``True`` if lineage was newly assigned, ``False`` if ``_udid``
+        was already set - which signals the task was returned in place by an
+        earlier stage and its existing lineage must be preserved.
+        """
+        if self._udid:
+            return False
+        parts = [*[p for p in parent_lineage_paths if p], str(child_index)]
+        self._lineage_path = "_".join(parts)
+        self._udid = hashlib.sha256(self._lineage_path.encode()).hexdigest()[:32]
```
praateekmahajan, line 66: Maybe move this to tasks/utils.py (a separate staticmethod here in this calass) and then also have a test in `test_tasks.py` that we indeed to that the same hasing we expect and that we pass lineage_path to it

tests/tasks/test_tasks.py
```diff
@@ -69,3 +71,75 @@ def test_fanout_tasks_have_unique_uuid():
     assert len(output) == 3
     uuids = [t._uuid for t in output]
     assert len(set(uuids)) == 3, f"Expected unique _uuid per task, got {uuids}"
+
+
+def _sha256_32(s: str) -> str:
+    return hashlib.sha256(s.encode()).hexdigest()[:32]
```
praateekmahajan, line 77: related to my comment on staticmethod for the shaing

nemo_curator/pipeline/pipeline.py
```diff
@@ -212,4 +212,6 @@ def run(self, executor: BaseExecutor | None = None, initial_tasks: list[Task] |
                     "The executor will schedule GPU stages on GPUs not held by Serve."
                 )
 
+        if initial_tasks:
+            assign_root_lineage(initial_tasks)
```
praateekmahajan: Nice!

praateekmahajan: Though I agree with @sarahyurick's comment that the `assign_root_lineage` can just be implemented here

nemo_curator/stages/base.py
```diff
@@ -192,10 +261,7 @@ def process_batch(self, tasks: list[X]) -> list[Y]:
...
```
praateekmahajan, line 248: Can we not have this functionality inside `stages/base.py:process_batch`.. 
Users for batch style stages often override it therefore users will be at risk of not calling `assign_child_lineage`..

I believe another `nemo_curator/backends/base.py:process_batch` would work

oyilmaz-nvidia: @praateekmahajan Not sure how another process_batch would work here. That function is called by the executors and there is a contract between the user stages and executors. 

This is the reason I have this assign_child_lineage as a function and if the user doesn't call it, the _udid won't be generated hence I can warn the user if the resumability is enabled later on.

oyilmaz-nvidia: Oh, I misunderstood it. The reason I couldn't do it is that base stage class returns the flattened list of tasks and I lose the parent child relationship.

So, I need an update in the ProcessingStage.process_batch and I don't think we'll be able to do it since many user have already been using it.

nemo_curator/pipeline/pipeline.py
oyilmaz-nvidia: Yeah, moved it.

PR #1769, opened 2026-04-08
docs: add AGENTS.md for AI agent and contributor guidance

arhamm1:
## Summary

Adds `AGENTS.md` — a structured reference for AI coding agents (Claude Code, Copilot, Cursor, etc.) contributing to NeMo Curator.

### What it covers
- Environment setup with `uv` (all extras)
- Test commands, GPU vs CPU split, 80% coverage requirement
- Linting with `ruff` via pre-commit
- Architecture overview: `Task`, `ProcessingStage`, `CompositeStage`, `Pipeline`, `Executor`
- Step-by-step guide to adding a new stage with code examples
- Stage authoring rules (`name`/`resources` must be class attributes, not `@property`; `with_()` pattern; idempotency requirement)
- Commit requirements (DCO `-s` flag)
- PR requirements including AI assistance disclosure and `Co-authored-by` trailer
- Common mistakes table (bare `pip`, `@property` on stage attrs, skipping signoff, etc.)

Also removes `AGENTS.md` from `.gitignore` where it was mistakenly added by a template generator.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2005, opened 2026-05-20
docs(governance): AGENTS.md steward network

lbliii:
## Summary

Migrates NeMo Curator's `AGENTS.md` steward network from eleven independently maintained instruction files to a manifest-driven system:

- `.stewards/manifest.toml` is the control plane for the network, ordered source layers, checks, coverage, and context budget.
- `.stewards/layers/*.toml` gives repository and domain owners focused authoring surfaces for scopes, invariants, guardrails, judgment, and graph edges.
- `.stewards/project.py` resolves base, domain, and optional overlay layers, then deterministically generates the root map and ten scoped `AGENTS.md` maps.
- `.stewards/verify.py` rejects stale maps, broken graph edges, duplicate IDs or paths, missing proof anchors, invalid severities, unsafe paths, orphan generated maps, uncovered code domains, and over-budget active map chains.
- `.stewards/PROTOCOL.md` is opened only for explicit review, audit, or steward-network maintenance; ordinary work reads the root map plus the nearest scoped map.
- A lightweight `steward-network` CI job runs projection, verification, and tooling regression tests on Python 3.11.

The resulting network has **11 stewards and 33 explicit invariants**: 9 machine-backed, 17 evidence-anchored manual, and 7 deliberately marked as unverified rather than overstating enforcement. The largest ordinary active context is under 9 KiB, against a 24 KiB budget.

This remains an advisory system. Agents implement and synthesize; `.github/CODEOWNERS` and human reviewers approve.

## Architecture

```mermaid
flowchart LR
    M[".stewards/manifest.toml<br/>source of truth"] --> P["project.py"]
    P --> R["Root AGENTS.md"]
    P --> S["10 scoped AGENTS.md maps"]
    R --> A["Implementing agent"]
    S --> A

    M --> V["verify.py"]
    R --> V
    S --> V
    V --> CI["steward-network CI gate"]

    T["Explicit review trigger"] -.-> PR["PROTOCOL.md"]
    PR -.-> A
    A --> H["Human / CODEOWNERS review"]
```

### Ordinary work

1. Read the root map and only the nearest scoped map governing the target path.
2. Apply its exact invariants and run only affected checks.
3. Do not load the manifest or protocol merely to translate instructions.

### Explicit review or audit

Triggers such as `ask stewards`, `bugbash`, `review swarm`, or `steward synthesis` open the protocol. Independent affected viewpoints return evidence-backed findings; matching findings increase confidence but do not automatically change severity. Severity follows demonstrated contract or user impact, and accepted P0s require a bounded sibling-pattern sweep.

### Steward maintenance

Each generated map names its guidance owners and exact source layer. Domain owners edit that focused layer, project the maps, and verify the network:

```bash
python .stewards/project.py --explain backends
python .stewards/project.py
python .stewards/verify.py --coverage
```

The manifest remains the control plane for layer order, checks, coverage, and context budgets. `CODEOWNERS` maps each layer to the engineers who own its guidance. Generated maps carry source provenance and are checked for freshness in CI.

## Steward scopes

| Steward | Generated map | Primary contract |
| --- | --- | --- |
| root | `AGENTS.md` | Repository-wide routing, public boundaries, optional dependencies, evidence discipline |
| pipeline | `nemo_curator/AGENTS.md` | Stages, tasks, workflows, resources, lifecycle, and lineage |
| backends | `nemo_curator/backends/AGENTS.md` | Adapter/executor behavior, resumability, and Ray stage specifications |
| deduplication | `nemo_curator/stages/deduplication/AGENTS.md` | Exact/fuzzy/semantic workflows, stable IDs, CUDA extra, Ray Actor Pool scope |
| synthetic | `nemo_curator/stages/synthetic/AGENTS.md` | Model-processing, Data Designer, prompts, and synthetic sample I/O |
| text | `nemo_curator/stages/text/AGENTS.md` | `DocumentBatch` and composable text stages |
| video | `nemo_curator/stages/video/AGENTS.md` | `VideoTask` and actual clip/video/preview/metadata/embedding output layout |
| tests | `tests/AGENTS.md` | GPU registration, Ray fixtures, dependency and network isolation |
| docs | `fern/AGENTS.md` | Canonical Fern documentation and source-backed claims |
| performance | `benchmarking/AGENTS.md` | Reproducible benchmark context and session behavior |
| tutorials | `tutorials/AGENTS.md` | Current imports, extras, prerequisites, and Ruff discipline |

`audio`, `image`, `interleaved`, and `math` currently inherit the pipeline/root contracts through explicit, reasoned coverage exemptions. Adding a new direct stage domain without either a steward map or an exemption fails verification.

## Review-driven corrections

Sarah Yurick's comments are reflected in the current manifest and generated maps:

- Xenna/Ray Data integration evidence is scoped to the exact tested fixtures; the network no longer claims universal executor parity.
- Ray Actor Pool workflow guidance lives with deduplication and is not grouped with general streaming backends.
- Ray adapter guidance uses the actual `RayStageSpecKeys` API. Stateful actor/fan-out behavior is reviewed explicitly rather than described as universal auto-detection.
- Text guidance recommends separate composed tokenizer/model stages when persistent state and CPU/GPU scaling differ, without asserting an exceptionless architecture rule.
- Lightweight Hugging Face downloads are allowed; heavier remote/API tests mock or skip, and token-gated CI requires automation-team coordination.
- Temporary `text/experimental/` structure is not encoded as a permanent boundary.
- Tutorial changes run Ruff without claiming tutorial execution is part of the unit-test matrix.
- Stale benchmark-file and internal Slack references are gone.

The migration also removes or narrows claims that source inspection did not support: automatic convergence to P0, WebDataset video output, an all-GPU end-to-end guarantee, a blanket `print()` ban, universal retry/parity/fallback guarantees, and overbroad batching rules.

## Additional fixes retained from the original PR

- Routes `fern/` review through `@NVIDIA-NeMo/docs_team` in `.github/CODEOWNERS`.
- Corrects stale quickstart references in `api-design.md` and Copilot instructions.
- Aligns the text modifier's `output_fields` documentation/error message with its actual parameter.

## Validation

- [x] `python .stewards/project.py --check`
- [x] `python .stewards/verify.py --coverage`
- [x] `python -m unittest discover -s tests/stewards -p 'test_*.py' -v` — 15 tests
- [x] Ruff format and lint across `.stewards/` and `tests/stewards/`
- [x] `markdownlint-cli2` across the manifest documentation and all 11 generated maps
- [x] `python .github/scripts/check_gpu_test_coverage.py` — all 43 GPU test files registered
- [x] `git diff --check`
- [ ] Focused NeMo functional suites were not runnable on the local macOS host because `nemo_curator/__init__.py` intentionally rejects non-Linux systems. Their exact commands remain registered in the manifest and the existing Linux CPU matrices cover their directories.

## Steward notes

- **Consulted:** root, pipeline, backends, deduplication, synthetic, text, video, tests, docs, performance, and tutorials; Sarah's review was re-audited against the final generated maps.
- **Accepted:** one source of truth, generated scoped views, explicit verification classes, proof anchors, coverage classification, context budgets, path safety, orphan detection, and a dependency-free CI gate.
- **Verification debt kept visible:** broad parity, external-test isolation, benchmark context, tutorial runnability, acceleration claims, and other judgment-heavy contracts remain `verification = "none"` until dedicated checks exist.
- **Human authority:** CODEOWNERS remains authoritative; stewards advise.

🤖 Updated with Codex


nemo_curator/AGENTS.md
```diff
@@ -0,0 +1,109 @@
...
+- All three backend adapters wrap the new contract:
+  `backends/{xenna,ray_data,ray_actor_pool}/adapter.py`
+- `api-design.md` — keep aligned with reality
+- Cursor rules listed above; the "API Design Architecture" section of
+  `.github/copilot-instructions.md`
+- `fern/` concept pages (pipeline, stages, tasks, executors,
+  resources)
+- `tests/{tasks,pipelines,backends/test_integration.py}/`
+- `tutorials/quickstart.py`
+- `CHANGELOG.md`
```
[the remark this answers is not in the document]
lbliii: Disagreeing — the `.cursor/rules/` directory is not empty. All 8 `.mdc` files cited across these AGENTS.md surfaces exist and are tracked on this branch (and on upstream `main`):

```
$ git ls-tree origin/lbliii/refine-pasted-prompt .cursor/rules/
100644 blob 018ffa7d…  .cursor/rules/coding-standards.mdc
100644 blob 1a5b53d7…  .cursor/rules/composite-stage-patterns.mdc
100644 blob 5ca3c995…  .cursor/rules/executors.mdc
100644 blob 7a130dab…  .cursor/rules/modality-structure.mdc
100644 blob 1742bbd4…  .cursor/rules/pipeline-structure.mdc
100644 blob 11440ff7…  .cursor/rules/processing-stage-patterns.mdc
100644 blob 3ba8a10c…  .cursor/rules/resources-configuration.mdc
100644 blob 2e8eea90…  .cursor/rules/task-patterns.mdc
```

Each file has real content (e.g. `.cursor/rules/pipeline-structure.mdc` is a 1.5 KB rule with `alwaysApply: true` front-matter describing `ProcessingStage` composition). Because this PR doesn't modify `.cursor/rules/`, those paths don't appear in the diff — but they're present in the working tree.

Worth noting: this finding is exactly the **Unverified finding regression** Known Regression Pattern this PR introduces in root `AGENTS.md`:

> A steward reports a divergence that a source grep would have disproved. *Verify:* every factual P0/P1 carries a verification status before triage.

The verification recipe for this kind of factual claim is `git ls-tree <ref> <path>` or `ls <path>`. Either would have shown the eight files. The pattern is in the same PR being reviewed — a useful first test of whether it lands.

No code change needed; closing this thread.

benchmarking/AGENTS.md
```diff
@@ -0,0 +1,72 @@
+# Steward: Benchmarking & Performance
+
+You own perf gates. Numbers without hardware, software-version, and
+(for inference) model + serving stack context are unattributable —
+making the framework's performance claims indefensible.
+
+Related: [benchmarking/README.md](README.md),
+[ALM_BENCHMARK.md](ALM_BENCHMARK.md),
+[AUDIO_PROFILING.md](AUDIO_PROFILING.md). Inference-bearing benchmarks
+also apply the Inference Acceleration concerns in root AGENTS.md.
```
sarahyurick: I think these files should be removed tbh. Either way let's not include them in the agents file.

benchmarking/AGENTS.md
```diff
@@ -0,0 +1,72 @@
...
+  schema changes are user-visible.
+- **Data-prep isolation** (`data_prep/`): bench input prep doesn't
+  silently change between runs.
+
+## Contract Checklist
+
+When this domain changes:
+
+- `benchmarking/{run.py,runner/,scripts/,tools/,data_prep/,Dockerfile,test-paths.yaml,nightly-benchmark.yaml}`
+- `benchmarking/{ALM_BENCHMARK,AUDIO_PROFILING,README}.md`
```
sarahyurick: Same comment as above.

benchmarking/AGENTS.md
```diff
@@ -0,0 +1,72 @@
...
+- `CHANGELOG.md` for user-visible perf regressions or improvements
+
+## Advocate
+
+- **Regression detection** — compare current results against a
+  baseline and flag > N% slowdowns.
+- **A "minimum viable benchmark" recipe** for new modality work so
+  perf gates exist from day one.
+- **Per-executor cost/throughput reporting** (Xenna vs Ray Data vs
+  Ray Actor Pool).
```
sarahyurick: We are usually only comparing Xenna vs Ray Data. We use Ray Actor Pool for dedup.

benchmarking/AGENTS.md
```diff
@@ -0,0 +1,72 @@
...
+- **Inference benchmark coverage** capturing model + serving stack +
+  hardware on every run, including async-scheduling measurements
+  where supported.
+
+## Own
+
+**Code:** `benchmarking/` (entire tree).
+
+**Docs:** `benchmarking/README.md`, `ALM_BENCHMARK.md`,
+`AUDIO_PROFILING.md`; `fern/` performance pages if present.
```
sarahyurick: Same comment as above.

nemo_curator/backends/AGENTS.md
```diff
@@ -0,0 +1,100 @@
+# Steward: Executor Parity & Backend Adapters
+
+You own executor parity. Same `Pipeline` + same input must produce
+equivalent output across Xenna, Ray Actor Pool, and Ray Data — and the
```
sarahyurick: Similar comment as above, I don't think it is appropriate to mention Ray Actor Pool here. We generally want to compare Xenna vs Ray Data and leave Ray Actor Pool for dedup.

nemo_curator/stages/video/AGENTS.md
```diff
@@ -0,0 +1,84 @@
...
+When this domain changes:
+
+- `nemo_curator/stages/video/`, `nemo_curator/tasks/video.py`
+- `tests/stages/video/`
+- `pyproject.toml` `video_cpu` and `video_cuda12` extras groups
+- `fern/` video pages — codec/container coverage, GPU prereqs,
+  configuration
+- `tutorials/video/`
+- `benchmarking/` — video benchmark configs and runners
+  (`ALM_BENCHMARK.md` is audio, not video)
```
sarahyurick: Similar comment as above.

tutorials/AGENTS.md
```diff
@@ -0,0 +1,86 @@
...
+- `pyproject.toml` extras (mentions match `<modality>_cpu` /
+  `<modality>_cuda12` naming)
+- `fern/` tutorial / how-to pages that reference these tutorials
+- `README.md` — its quickstart aligns with `tutorials/quickstart.py`
+- `CHANGELOG.md` if tutorial changes reflect user-visible API moves
+
+## Advocate
+
+- **CI smoke job** that imports/parses every tutorial Python file.
+  Export notebooks to `.py` for static checks.
```
sarahyurick: Not sure this needs to be included. Folks should run Ruff on tutorials for sure. Besides that we are not trying to have CI/CD test it right?

AGENTS.md
```diff
@@ -0,0 +1,375 @@
...
+  `pyproject.toml` entry, argparse declaration, Pydantic field, or
+  dataclass field. Grep the source.
+- **Stage-contract drift.** Doc claims `ProcessingStage` input/output
+  types or resource shapes that no longer match
+  `stages/base.py` or the stage's definition. *Verify:* read
+  `inputs`, `outputs`, `process`, `resources` and cross-check.
+- **Executor parity drift.** Doc claims a behavior across executors
+  but the implementation differs across
+  `backends/{xenna,ray_data,ray_actor_pool}/`. *Verify:* grep all
+  three adapters for the named feature.
```
sarahyurick: Should only need to verify for Xenna and Ray Data.

sarahyurick (commented):
Hi, did a quick pass, thanks.

lbliii:
Pushed three commits responding to @sarahyurick's review and integrating Praateek's developer guide:

1. **\`docs(governance): apply Sarah's PR review — Ray Actor Pool is dedup-only\`** — reframed Ray Actor Pool as the dedup-batch executor across root, backends, benchmarking, video, tutorials. Dropped \`ALM_BENCHMARK.md\` / \`AUDIO_PROFILING.md\` references. Tutorials get ruff, not CI execution.

2. **\`docs(governance): seed stewards with Praateek's architectural discipline\`** — encoded the patterns from the NeMo Curator Developer Guide into the relevant stewards. The Pipeline & Stage Contract steward now carries the \`__init__\` / \`setup_on_node\` / \`setup\` lifecycle discipline, \`process\` vs \`process_batch\` rules, \`_metadata\` / \`_stage_perf\` propagation (especially fan-out), task-size sweet spot, \`with_()\` for variable resources, \`CompositeStage\`, and \`Workflow\`. Text / Video / SDG stewards cross-reference the setup discipline for model loading. Dedup steward gains the \`Workflow\` callout and explicit Ray Actor Pool scope. Tests steward gains the "no \`TestThisClassFunction\` over-classification" rule.

3. **\`docs(governance): integrate Praateek's What-NOT-to-Do + new KRPs\`** — root \`AGENTS.md\` Anti-Patterns is now a Don't / Do table (shape from #1769), and six new Known Regression Patterns the swarm hunts by default: stage lifecycle drift, \`process_batch\` mis-use, metadata/_stage_perf drop on fan-out, Ray Data spec missing (\`IS_ACTOR_STAGE\`, \`IS_FANOUT_STAGE\`), Ray Actor Pool mis-attribution, test over-classification, and \`EmptyTask\` first-stage worker waste. Each pattern names a verification recipe so it's machine-checkable in autopilot mode.

The \`What NOT to Do\` shape and several concrete rules are borrowed from #1769 — the two PRs are complementary (#1769 = root-level developer onboarding; #2005 = scoped domain network). When both land they reinforce each other.

Verifying lint clean across all 11 \`AGENTS.md\` files.

benchmarking/AGENTS.md
```diff
@@ -0,0 +1,104 @@
...
+
+1. Add a `.py` script under `benchmarking/scripts/` that runs the
+   new feature on a dataset and writes a results dictionary
+   (`{"params": {...}, "metrics": {...}, "tasks": [...]}`).
+2. Add an entry to a configuration `.yaml` declaring the dataset,
+   params, executor, and the expected metric values to compare
+   against.
+3. The nightly cron runs all entries in `nightly-benchmark.yaml` on
+   4×A100; results post to the rapids-workflows-nightly-tests
+   channel.
```
sarahyurick: This seems to be referencing an internal Slack channel.

benchmarking/AGENTS.md
```diff
@@ -0,0 +1,104 @@
...
+schema, search `benchmarking/`, `fern/`, `README.md`, and
+`.github/copilot-instructions.md` for:
+
+- `test-paths.yaml`, `nightly-benchmark.yaml` entries
+- Benchmark script names you renamed under `benchmarking/scripts/`
+- Result schema field names (params, metrics, tasks)
+- Hardware references (H100, L40S, A100, GB200) tied to specific
+  workloads
+- Cost-per-token / cost-per-hour-of-video claims
+- Headline speedup numbers (e.g., RedPajama-v2 fuzzy-dedup figures
```
sarahyurick: I thought it was CC not red pajama?

nemo_curator/AGENTS.md
```diff
@@ -0,0 +1,177 @@
...
+  - `setup_on_node()` runs **once per node**. Use for
+    pre-downloading model weights or verifying they exist on disk.
+    Never load a model into memory here — the next worker on the
+    same node would reload it.
+  - `setup()` runs **once per worker (replica)**. Use for loading
+    models into memory and moving them to device
+    (`self.model = AutoModel.from_pretrained(...).to("cuda")`).
+    GPU-resident state lives here.
+  - A stage that overrides `setup()` is auto-routed to a Ray Data
+    **Actor**; stateless stages become Ray Data **Tasks**. Overriding
```
sarahyurick: I know this is true for text filters and modifiers but I don't think this auto detection is true across the board right? I think you normally have to specify if you want it to be an actor (appropriate for stages that persist some type of object in memory per actor, like a tokenizer or model).

AGENTS.md
```diff
@@ -0,0 +1,476 @@
...
+- **Deploy anywhere** — Python APIs, runnable on CSP and on-prem
+
+The architecture rests on three design pillars:
+
+- **Task-centric** — Tasks (`DocumentBatch`, `ImageBatch`, `VideoTask`, `AudioTask`) are the unit of data flowing through pipelines
+- **Map-style** — every stage transforms tasks to tasks; this constraint enables auto-balancing and streaming
+- **Fault tolerant** — stages survive preemption and reschedule; partial state is recoverable
+
+The same pipeline definition must run unchanged across Xenna, Ray Actor
+Pool, and Ray Data executors.
```
sarahyurick: Lingering comment should not group Ray Actor Pool with Xenna and Ray Data.

nemo_curator/stages/text/AGENTS.md
```diff
@@ -0,0 +1,112 @@
...
+  Domain Classifier (26 classes — Finance, Health, Business and
+  Industrial, Science, Law and Government, Internet and Telecom,
+  Jobs and Education, News, Computers and Electronics, Shopping,
+  and 16 others) and the Quality Classifier (High / Medium / Low
+  against a rubric of content accuracy, clarity, coherence,
+  grammar, depth of information, and overall usefulness) are
+  published on HuggingFace. Renames, removals, default changes, or
+  rubric changes are major-version events.
+- **Filter and modifier semantics** (`text/filters/`,
+  `text/modifiers/`, `text/modules/`). Document edge cases
```
sarahyurick: Nit but the modules directory is just for add ID now right?

nemo_curator/stages/text/AGENTS.md
```diff
@@ -0,0 +1,112 @@
...
+  limits / robots / licensing; surface failures explicitly;
+  retry-safe under preemption.
+- **GPU/CPU paths.** Lazy-import GPU libs; declare resources
+  honestly; CPU fallback loads.
+- **Model loading in `setup()`, not `__init__`.** Classifiers and
+  embedders must load model weights and call `.to("cuda")` inside
+  `setup()`. Loading in `__init__` serializes the model to every
+  replica. Downloading weights belongs in `setup_on_node()`. See the
+  setup-discipline rule in [parent](../../AGENTS.md).
+- **Tokenizer and model artifact handling.** HuggingFace-pinned
```
sarahyurick: This isn't necessarily text-specific but since it is common for text, I would like us to specify that tokenization and model inference should always be split into 2 stages and defined together in a composite stage. Like in general, if we are loading some type of model object in memory, it is usually better for each model to be its own stage. This is especially true for the model forward pass which has a huge advantage when it has more GPU resources available in the pipeline, as opposed to having to share the time and resources with the tokenization step.

tests/AGENTS.md
sarahyurick, line 1: Maybe worth mentioning, if you are adding a model test that requires a HF_TOKEN, then you will need to coordinate with the automation team to get the model configured with the token in CI/CD.

fern/AGENTS.md
```diff
@@ -0,0 +1,122 @@
...
+  argparse, pipeline examples type-check.
+- **Cross-page consistency.** Same fact reads identically across
+  `fern/`, `README.md`, `CONTRIBUTING.md`, `api-design.md`, cursor
+  rules, copilot instructions, and tutorials.
+- **Broken-link tooling.** `fern/_fix_broken_links.py` is
+  authoritative for Fern. `docs/broken_links_*.json` is the
+  deprecated Sphinx site — ignore for Fern.
+- **Variable substitution.** `fern/substitute_variables.py` rewrites
+  `{{ current_release }}` / `<release/>`. Don't hand-pin versions
+  where substitution would apply.
```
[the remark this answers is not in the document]
lbliii: Disagreeing — `fern/versions/latest.yml` does exist on this branch:

```
$ ls fern/versions/
latest.yml
main
main.yml
v25.09
v25.09.yml
v26.02
v26.02.yml
v26.04
v26.04.yml
```

The file is referenced as the redirect-only manifest (no matching `latest/` directory, which the steward already calls out two lines later: *"`latest.yml` is redirect-only — no `latest/` directory"*). This is the second "Unverified finding regression" pattern on this PR from this tool — the Known Regression Pattern is in root `AGENTS.md` for a reason. No code change needed.

benchmarking/AGENTS.md
lbliii: Good catch — dropped the internal Slack channel name. Now reads "team's results sink" in 02f6c75f6.

benchmarking/AGENTS.md
lbliii: The current public README headline is RedPajama v2: *"16× faster fuzzy deduplication on 8 TB RedPajama v2 (1.78 trillion tokens)"*. Nemotron-CC is cited separately as the end-to-end recipe. To avoid pinning a specific corpus that may drift, I softened the steward to point at README first as the source of truth (02f6c75f6). If we're swapping the headline benchmark to Nemotron-CC, the README + the public site update first, and the steward then auto-discovers via grep.

nemo_curator/AGENTS.md
lbliii: Good correction. Praateek's deck described auto-routing as a general behavior; you're saying it's opportunistic in practice. Reframed in 02f6c75f6: auto-detection works for simpler text filter/modifier stages but isn't reliable across the board, and any stage holding a tokenizer / model / lookup table in memory should explicitly set `RayStageSpecKeys.IS_ACTOR_STAGE: True` in `ray_stage_spec`.

AGENTS.md
lbliii: Fixed in 02f6c75f6 — root now reads: *"The same pipeline definition must run unchanged across the streaming executors (Xenna and Ray Data). Ray Actor Pool is the dedup-batch executor — appropriate for shuffle-based dedup and full-data-state passes, not for general streaming workloads."*

nemo_curator/stages/text/AGENTS.md
lbliii: Verified — `text/modules/` contains `add_id.py`, `joiner.py`, `splitter.py` (utility code, not filter/modifier). Removed `text/modules/` from the filter/modifier grouping and added a note that modules is utility code (02f6c75f6).

nemo_curator/stages/text/AGENTS.md
lbliii: Strong guidance — added in 02f6c75f6 as a new Protect bullet on the text steward:\n\n> **Tokenization and model inference are always separate stages.** When a pipeline needs both (classifier inference, embedder inference), split them into two stages and compose them with `CompositeStage`. Tokenization is CPU-bound; inference is GPU-bound. Keeping them in one stage forces the GPU stage to share its replicas with tokenization, hurting throughput. Each model loaded in memory deserves its own stage so the GPU stage can scale independently.\n\nI kept it on the text steward since that's where it activates most often, but the convention is general — happy to move it to Pipeline steward if you'd prefer it broader.

tests/AGENTS.md
lbliii: Added in 02f6c75f6:\n\n> **HF_TOKEN-gated tests require automation-team coordination.** If a test loads a HuggingFace model that needs an access token, the test will fail in CI without the token configured. Coordinate with `@NVIDIA-NeMo/automation` to add the token (and any model-specific access grants) to the CI secrets before merging. Use `pytest.skip` if the token is absent locally — never `pytest.fail`, which hard-blocks contributors without the secret.\n\nThe steward swarm flagged three test files using `pytest.fail` on missing HF_TOKEN — those are tracked as follow-up fixes.

tests/AGENTS.md
```diff
@@ -0,0 +1,115 @@
...
+  and `tests/data/` only cover audio.
+- **Per-module coverage reporting** so low-coverage hotspots are
+  visible.
+- **Faster CPU CI** via path-based test selection on changed files.
+- **Snapshot or determinism tests** for dedup and synthetic-data
+  outputs.
+
+## Do Not
+
+- Use real network calls (HuggingFace download, remote APIs) in CPU
```
sarahyurick: I think small Hugging Face downloads are okay. We use them for classifier and embedder tests.

nemo_curator/stages/text/AGENTS.md
```diff
@@ -0,0 +1,121 @@
...
+- `tutorials/text/`, `tutorials/quickstart.py`
+- `.cursor/rules/modality-structure.mdc`
+- `CHANGELOG.md`
+
+## Advocate
+
+- **Programmatic embedder registry** mirroring
+  `classifiers/__init__.py:_LAZY` so docs don't hand-count.
+  Auto-generate `fern/` reference pages from `__all__`.
+- **Clearer separation** between `text/experimental/` and stable
```
sarahyurick: Maybe let's remove this point. I think `text/experimental/` will not always be there.

sarahyurick: Bumping, thanks.

tests/AGENTS.md
```diff
@@ -0,0 +1,115 @@
...
+  tests; mock or skip.
```
sarahyurick:
```suggestion
- Use real network calls (HuggingFace download, remote APIs) in lightweight
  tests; mock or skip otherwise
```

.stewards/manifest.toml
```diff
@@ -0,0 +1,538 @@
...
+
+[judgment.pipeline]
+advocate = ["Prefer a small composable stage over backend-specific branching in shared pipeline code."]
+do_not = ["Do not change shared descriptors to accommodate one executor without checking the others."]
+serves = ["Modality stages, backend adapters, workflow authors, and downstream library users."]
+
+[judgment.backends]
+advocate = ["Make retry, resume, failure, and task-lineage behavior observable in focused tests."]
+do_not = ["Do not claim parity beyond the backends and cases exercised by evidence."]
+serves = ["Pipeline authors running locally, on Ray Data, and on Ray Actor Pool."]
```
sarahyurick: Not sure this is accurate. Pipelines are currently run on either Xenna (default) or Ray Data. And Ray Actor Pool is for dedup.

nemo_curator/backends/AGENTS.md
```diff
@@ -0,0 +1,42 @@
...
+
+- Make retry, resume, failure, and task-lineage behavior observable in focused tests.
+
+## Do Not
+
+- Do not claim parity beyond the backends and cases exercised by evidence.
+
+## Serves
+
+- Pipeline authors running locally, on Ray Data, and on Ray Actor Pool.
```
sarahyurick: Same comment as before, not sure what this is supposed to mean.

nemo_curator/stages/text/AGENTS.md
```diff
@@ -0,0 +1,31 @@
...
+
+| Invariant | Sev | Backing | Proof / anchor |
+| --- | --- | --- | --- |
+| The text classifier namespace exports its lazy registry without eagerly initializing optional classifier implementations. | P1 | machine-backed | `uv run pytest tests/stages/text/classifiers/test_classifiers_init.py -q -m 'not gpu'` (`text-registry`) |
+| Text pipeline stages preserve the DocumentBatch table/data-frame task boundary unless a public migration is approved. | P1 | manual | nemo_curator/tasks/document.py · `class DocumentBatch` |
+| Classifier, embedder, download, I/O, filter, and modifier concerns remain independently composable where their public stages permit it. | P2 | none | — |
+
+## Guardrails
+
+- Keep classifier registration lazy so importing the namespace does not initialize optional frameworks.
```
sarahyurick: I see a lot of emphasis on this throughout the PR but I am not so sure it is that important to verify repeatedly, since we already do this. Presumably the agent would pick up on the pattern already if it was contributing a new classifier (not even sure we will be contributing new classifiers anyway). Plus the pattern isn't specific to classifiers, lazy imports should be used for any classes/files that import heavy dependencies and/or dependencies not needed for the entire modality.

tests/stewards/__init__.py
sarahyurick, line 1: No copyright needed for an empty file.

AGENTS.md
```diff
@@ -0,0 +1,81 @@
...
+- Public stage, task, workflow, backend, configuration, or output-layout changes include focused regression proof and user-facing collateral or an explicit no-impact rationale.
+- Keep CPU-only tests isolated from optional GPU, service, and cloud dependencies; use registered markers and shared fixtures where applicable.
+- Do not broaden an invariant beyond what its proof command or evidence anchor demonstrates.
+- Review triggers such as `ask stewards`, `bugbash`, `review swarm`, and `steward synthesis` open the protocol and solicit independent affected viewpoints.
+- Cross-surface work records Steward Notes naming consulted maps, accepted or deferred findings, proof run, collateral, and unresolved dissent.
+- Avoid unrelated refactors, speculative knobs, silent exception handling, and new suppressions unless they are part of the demonstrated fix.
+
+## Network
+
+| Steward | Map | Invariants | Automated backing |
```
sarahyurick: Not sure I understand what does this table mean?

sarahyurick (changes requested):
It seems like a lot of the details from previous commits have been removed? Not sure how useful the PR is in this current state, since it is not really providing as much guidance as it was before.

lbliii:
Thanks for taking another pass. One important piece of context is that the approach changed substantially since your first review.

```mermaid
flowchart LR
    subgraph Old["Earlier approach"]
        O1["Hand-maintained root AGENTS.md"] --> OA["Agent"]
        O2["Hand-maintained scoped files"] --> OA
    end

    subgraph New["Current approach"]
        M[".stewards/manifest.toml"] --> G["Generate and verify"]
        G --> R["Root map"]
        G --> S["Nearest scoped map"]
        R --> NA["Agent"]
        S --> NA
    end
```

The `AGENTS.md` files are now compact, generated views rather than independently maintained guides. Agents normally read only the root file and the nearest scoped file.

We tested this on the same tasks and then transferred it to two other repositories. It maintained or improved correctness while reducing unnecessary searching by 83–98% and instruction size by 20–60%.

Your comments still identify places where the generated output should be clearer. We’ll correct the backend wording, generalize the lazy-import guidance, and simplify or remove the confusing network table.

nemo_curator/backends/AGENTS.md
```diff
@@ -0,0 +1,42 @@
+<!-- generated from .stewards/manifest.toml — edit the manifest, not this file -->
```
sarahyurick: So folks should not edit this file? I think there would need to be a larger conversation with Curator engineers about what an agent-friendly Curator would look like. It was my understanding that a good approach would be to have core engineers update the AGENTS.md as guidance for comment Claude/Codex bugs/assumptions that we have run into ourselves.

.github/CODEOWNERS
```diff
@@ -4,12 +4,27 @@
...
+.stewards/layers/repository.toml @NVIDIA-NeMo/curator_reviewers
+.stewards/layers/pipeline.toml @NVIDIA-NeMo/curator_reviewers
+.stewards/layers/backends.toml @oyilmaz-nvidia @praateekmahajan @abhinavg4 @ayushdg
+.stewards/layers/deduplication.toml @ayushdg @praateekmahajan
+.stewards/layers/synthetic.toml @huvunvidia
+.stewards/layers/text.toml @sarahyurick @praateekmahajan @VibhuJawa
+.stewards/layers/video.toml @suiyoubi @abhinavg4
+.stewards/layers/tests.toml @NVIDIA-NeMo/curator_reviewers
+.stewards/layers/docs.toml @NVIDIA-NeMo/docs_team
+.stewards/layers/performance.toml @rlratzel @praateekmahajan @sarahyurick @ayushdg
```
sarahyurick, line 18: Could we call this benchmarking instead of performance?

nemo_curator/stages/synthetic/AGENTS.md
```diff
@@ -0,0 +1,34 @@
...
+Ordinary work: use this map directly with the root map and run only affected checks.
+Open `.stewards/` only for explicit review, audit, or steward maintenance.
+
+## Protects
+
+| Invariant | Sev | Backing | Proof / anchor |
+| --- | --- | --- | --- |
+| ModelProcessingStage preserves the setup and sample-processing behavior covered by its focused unit tests. | P1 | machine-backed | `uv run pytest tests/stages/synthetic/omni/test_base.py -q -m 'not gpu'` (`synthetic-base`) |
+| Nemotron-CC system prompts are behavior-bearing inputs and changes receive output-quality review. | P2 | manual | nemo_curator/stages/synthetic/nemotron_cc/prompts.py · `NEMOTRON_CC_SYSTEM_PROMPT` |
+| DataDesignerStage remains a DocumentBatch-to-DocumentBatch processing stage. | P2 | manual | nemo_curator/stages/synthetic/nemo_data_designer/data_designer.py · `class DataDesignerStage` |
```
sarahyurick, line 20: Sorry just wanted to clarify, would the expectation be for folks to update the toml files or the md files? Since right now I see they cover the same information.

sarahyurick:
Closing as out of date. Let's open a fresh PR if we decide to add this feature.

PR #2024, opened 2026-05-22
fix(utils): include port 65535 in get_free_port() scan range

### LINKED ISSUE #2023: fix: get_free_port() off-by-one — port 65535 never checked despite appearing in error message
## Description

\`\`\`python
# line 103
    ...
# line 118
\`\`\`

## Steps to reproduce

## Suggested fix

\`\`\`diff
\`\`\`

## Impact

miclaldogan:
## Summary

- `range(start_port, 65535)` in `get_free_port()` excluded port 65535, the last valid TCP port
- Changed upper bound to `65536` so the full valid range 0–65535 is covered
- The existing error message already referenced 65535, making the off-by-one visible

## Change

```diff
- for port in range(start_port, 65535):
+ for port in range(start_port, 65536):
```

## Test plan

- [ ] Verify `get_free_port(65535)` now succeeds when port 65535 is free
- [ ] Verify existing port allocation behavior is unchanged for normal port ranges

Closes #2023


VibhuJawa (approved):
LGTM

sarahyurick (approved):

PR #2025, opened 2026-05-24
fix: auto-detect Ray fanout stages

### LINKED ISSUE #1613: Automatically detect when `IS_FANOUT_STAGE` should be set to True
Whenever a stage does `def process(self, task: Task) -> list[Task]`, it is considered a fanout stage. We should add logic to the base `ProcessingStage` to automatically do this, instead of the child stages having to do it every time.

For example, in https://github.com/NVIDIA-NeMo/Curator/blob/main/nemo_curator/stages/text/download/base/url_generation.py we explicitly do:
```
def ray_stage_spec(self) -> dict[str, Any]:
    return {
        "is_fanout_stage": True,
    }
```
but since `def process(self, task: _EmptyTask) -> list[FileGroupTask]` is returning a list, the base `ProcessingStage` should just have some extra logic to automatically do it instead. That way, we can remove it from our existing stages like `URLGenerationStage`, and contributors do not need to remember to explicitly set it when creating new stages.

This could be done with `get_type_hints` or something similar.

We should make sure we are handling cases like `def process(self, task: Task) -> Task | list[Task]`, etc. appropriately too. When we do `IS_FANOUT_STAGE`, there is a perf cost to it. Under the hood it runs `StreamingRepartition` which is an expensive operation at scale.

nightcityblade:
## Description
Closes #1613.

Automatically marks Ray Data stages as fanout stages when their `process` return annotation is `list[...]` or a union that includes `list[...]`. This lets stages like `URLGenerationStage` rely on the base `ProcessingStage` default instead of manually setting `is_fanout_stage`.

## Usage
```python
class MyFanoutStage(ProcessingStage[InputTask, OutputTask]):
    def process(self, task: InputTask) -> list[OutputTask]:
        ...

assert MyFanoutStage().ray_stage_spec() == {"is_fanout_stage": True}
```

## Checklist

Tests:
- `uv run --python 3.12 ruff check nemo_curator/stages/base.py nemo_curator/stages/text/download/base/url_generation.py tests/stages/common/test_base.py`
- `uv run --python 3.12 --with pytest --with ray --with loguru --with pandas --with pyarrow --with fsspec python -m pytest tests/stages/common/test_base.py tests/stages/text/download/base/test_url_generation.py` (blocked on macOS by NeMo-Curator's Linux-only runtime check)


nemo_curator/stages/text/download/base/url_generation.py
```diff
@@ -77,11 +77,6 @@ def process(self, task: _EmptyTask) -> list[FileGroupTask]:
             for i, url in enumerate(urls)
         ]
 
-    def ray_stage_spec(self) -> dict[str, Any]:
-        return {
-            "is_fanout_stage": True,
```
sarahyurick, line 81: Can you do this for all stages to make sure that this PR works for all of them? And ensure that each existing stage has a pytest to check that it is being set?

tests/stages/common/test_base.py
```diff
@@ -51,6 +51,24 @@ def outputs(self) -> tuple[list[str], list[str]]:
...
+    def process(self, task: MockTask) -> list[MockTask]:
+        return [task]
+
+
+class MaybeFanoutProcessingStage(ProcessingStage[MockTask, MockTask]):
+    """ProcessingStage that may return multiple tasks."""
+
+    name = "MaybeFanoutProcessingStage"
+
+    def process(self, task: MockTask) -> MockTask | list[MockTask]:
```
sarahyurick, line 68: I am undecided what should happen in this case. Maybe it should be up to the user?

sarahyurick (changes requested):
Thanks @nightcityblade !

nemo_curator/stages/text/download/base/url_generation.py
nightcityblade: Thanks — I broadened the coverage beyond URLGenerationStage in the follow-up commit. Existing list-returning stages now have pytest coverage for the Ray fanout spec in common/text/audio/image tests (URLGenerationStage and FilePartitioning already had assertions; I added coverage for ALMManifestReaderStage, CreateInitialManifestFleursStage, AudioToDocumentStage, and ImageReaderStage). I also kept existing explicit ray_stage_spec overrides intact.

tests/stages/common/test_base.py
nightcityblade: Good point. I changed the inference to be conservative: `Task | list[Task]` is no longer auto-marked as a Ray fanout stage. If a maybe-fanout stage wants Ray fanout behavior, it can opt in by overriding `ray_stage_spec()`. I added tests for both the conservative default and explicit opt-in.

nemo_curator/stages/base.py
```diff
@@ -302,8 +303,30 @@ def ray_stage_spec(self) -> dict[str, Any]:
...
+
+        return cls._annotation_includes_list(return_annotation)
+
+    @staticmethod
+    def _annotation_includes_list(annotation: object) -> bool:
+        """Return whether an annotation is or includes a list type."""
+        origin = get_origin(annotation)
+        if origin is list:
+            return True
+        if origin in (UnionType, Union) or isinstance(annotation, UnionType):
```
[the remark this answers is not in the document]
nightcityblade: Addressed in 947e029: removed the unreachable `isinstance(annotation, UnionType)` branch while refining the detection logic.

nemo_curator/stages/base.py
```diff
@@ -302,8 +303,30 @@ def ray_stage_spec(self) -> dict[str, Any]:
...
+            return {"is_fanout_stage": True}
         return {}
 
+    @classmethod
+    def _process_returns_list(cls) -> bool:
+        """Return whether the stage's process annotation can return a list."""
+        try:
+            return_annotation = get_type_hints(cls.process).get("return")
+        except (NameError, TypeError):
+            return_annotation = cls.process.__annotations__.get("return")
```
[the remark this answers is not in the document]
nightcityblade: Addressed in 947e029: the fallback now handles raw string annotations like `"list[OutputTask]"` so list-returning stages are still detected when `get_type_hints()` cannot resolve the annotation. I added a regression test that forces that fallback path.

nemo_curator/stages/text/download/base/url_generation.py
sarahyurick: Thanks, I see some test edits but let's remove the `ray_stage_spec` function from the actual stages and make sure the tests still pass. The ones I found:

- https://github.com/NVIDIA-NeMo/Curator/blob/main/nemo_curator/stages/text/download/base/url_generation.py (this file)
- https://github.com/NVIDIA-NeMo/Curator/blob/main/nemo_curator/stages/image/io/image_reader.py
- https://github.com/NVIDIA-NeMo/Curator/blob/main/nemo_curator/stages/audio/datasets/fleurs/create_initial_manifest.py
- https://github.com/NVIDIA-NeMo/Curator/blob/main/nemo_curator/stages/audio/common.py
- https://github.com/NVIDIA-NeMo/Curator/blob/main/nemo_curator/stages/video/clipping/clip_extraction_stages.py
- https://github.com/NVIDIA-NeMo/Curator/blob/main/nemo_curator/stages/audio/alm/pretrain/extraction.py
- https://github.com/NVIDIA-NeMo/Curator/blob/main/nemo_curator/stages/audio/segmentation/speaker_separation.py
- https://github.com/NVIDIA-NeMo/Curator/blob/main/nemo_curator/stages/file_partitioning.py
- https://github.com/NVIDIA-NeMo/Curator/blob/main/nemo_curator/stages/audio/datasets/readspeech/create_initial_manifest.py
- https://github.com/NVIDIA-NeMo/Curator/blob/main/nemo_curator/stages/audio/alm/pretrain/io.py
- https://github.com/NVIDIA-NeMo/Curator/blob/main/nemo_curator/stages/deduplication/semantic/pairwise_io.py
- https://github.com/NVIDIA-NeMo/Curator/blob/c0a02ade13a3a4c1d434f2258c9e5fba67b0e5bb/nemo_curator/stages/audio/segmentation/vad_segmentation.py#L112

Do all of these work with this PR? Are there any that should be considered exceptions/explicitly keep `ray_stage_spec` for some reason?

nightcityblade:
Thanks, I pushed a focused follow-up in ec6a75f.

What changed:
- Removed redundant `ray_stage_spec()` fanout overrides from the list-returning stages that still had them on this branch: `FilePartitioningStage`, `ALMManifestReaderStage`, and `ClusterWiseFilePartitioningStage`.
- Added a regression assertion for `ClusterWiseFilePartitioningStage` so it is covered by the base return-annotation inference too.
- `URLGenerationStage` was already relying on the base inference from the previous commit, and the existing image/audio/Fleurs tests continue to cover those list-returning stages.

I intentionally did not remove the clip extraction override because that stage's `process()` returns a single `VideoTask`, not a list, so the new base inference would not mark it as fanout and removing it would change behavior. The other stage-specific `ray_stage_spec()` methods I left are for non-fanout Ray behavior (actor/RAFT/LSH/shuffle/max-calls), not redundant fanout inference.

Validation:
- `uv run ruff check nemo_curator/stages/file_partitioning.py nemo_curator/stages/audio/alm/alm_manifest_reader.py nemo_curator/stages/deduplication/semantic/pairwise_io.py tests/stages/deduplication/semantic/test_pairwise_io.py` passed.
- I attempted the relevant pytest set locally, but this machine is macOS and the package exits during test collection with the existing Linux-only guard; CI should be the source of truth for pytest here.

nightcityblade:
Followed up on the fanout-inference review and pushed one more cleanup commit.

What changed:
- removed the remaining explicit `ray_stage_spec()` override from `ClipTranscodingStage`
- updated `ClipTranscodingStage.process()` to declare its real return type (`list[VideoTask]`), so it now participates in the same inference path as the other fanout stages in this PR

On the remaining exceptions question: after this change, the stage-specific overrides I found are the ones that still carry non-fanout Ray behavior/config (for example actor-stage / backend-specific settings), rather than plain `is_fanout_stage=True` duplication.

Validation:
- `python3 -m compileall nemo_curator/stages/base.py nemo_curator/stages/video/clipping/clip_extraction_stages.py`

I also tried to run the targeted pytest files, but the local environment here is missing `ray`, so test collection stops in `tests/conftest.py` before those stage tests can run.

nightcityblade:
Thanks, I pushed a small follow-up in effa59a to address the Greptile finding on the clip fanout contract.\n\nWhat changed:\n- Updated the no-clips early-return path in `ClipTranscodingStage.process()` to return `[task]`, matching the `list[VideoTask]` annotation and Ray fanout behavior.\n- Updated the existing clip transcoding pytest to assert that the no-clips path also returns a one-item task list.\n\nValidation:\n- `uv run ruff check nemo_curator/stages/video/clipping/clip_extraction_stages.py tests/stages/video/clipping/test_clip_transcoding_stage.py` passed.\n- `python3 -m compileall nemo_curator/stages/video/clipping/clip_extraction_stages.py tests/stages/video/clipping/test_clip_transcoding_stage.py` passed.\n- I also attempted the targeted pytest set for the fanout-related stage tests, but local collection still stops on macOS with the existing Linux-only package guard in `nemo_curator/__init__.py`, so CI/Linux should validate pytest execution.

nightcityblade:
I refreshed the branch sign-offs and opened a replacement PR here: #2056. This branch now has the same focused 13-file diff with all local commits signed off.

PR #2032, opened 2026-05-27
test: add image reader sample data

### LINKED ISSUE #1555: Fail tests if the test data is not found

nightcityblade:
## Description
Add the missing image reader WebDataset fixture and enable the GPU image reader test so missing test data fails instead of being silently skipped.

Fixes #1555

## Usage
```python
# N/A - test fixture only
```
## Checklist


nightcityblade:
Addressed the fixture-path issue Greptile flagged: the GPU image reader test now resolves the repo root with `parents[4]`, so it points at `tests/image_data/00000.tar` instead of one directory above the repo.

Updated in commit 8a5b3ce.

Verification:
- `python3 - <<'PY' ...` confirmed `parents[4] / "tests/image_data/00000.tar"` exists while `parents[5]` does not
- `python3 -m pytest tests/stages/image/io/test_image_reader.py -k 'resources_without_cuda'` is currently blocked locally because `ray` is not installed in this checkout

sarahyurick (approved):
LGTM, thanks! Let's check with @huvunvidia too.

PR #2033, opened 2026-05-27
Pipeline resumability via source-level counter checkpointing

abhinavg4:
## What

Adds **opt-in pipeline resumability** via source-level counter checkpointing,
built on top of the sentinel-task refactor in **#2062** (this PR is stacked on
that branch — review/merge #2062 first).

`Pipeline.run(checkpoint_path=...)` tracks which source partitions have fully
drained through the pipeline and skips already-completed ones on a rerun, so an
interrupted run resumes without reprocessing finished work.

## How

- **Sentinels** (on #2062's `SentinelTask` base): `NoneTask` / `FailedTask` are
  bare subclasses — no identity of their own (`dataset_name` `"none"`/`"failed"`,
  `task_id` assigned by the adapter like any task).
- **`backends/base.py` `process_batch`** (always-on): a returned `None`
  ("filter this slot") becomes a `NoneTask` via a single inline comprehension so
  every output is a real `Task` and gets a `task_id`; sentinels are stripped
  before the next stage.
- **`_apply_resumability_counters`** (gated on `_is_active`, counter-only): a
  source stamps `_source_id`, skips completed sources, fires `+1`; a non-source
  fires `-1`/`0` per output (`NoneTask` → `-1`, `FailedTask` → no delta, so its
  source stays pending and reruns). Counters key on the **parent's** identity —
  which is why the sentinels need none of their own. Ambiguous `M→K` batches
  warn + skip rather than misattribute.
- **LMDB actor + client.** `lmdb` is a (locked) dependency but stays opt-in at
  import time: `ACTOR_NAME` lives in `resumability_client` so the always-imported
  worker path never imports `lmdb`; it loads only when resumability is used.
- **SLURM-array safe.** State lives in `<checkpoint_path>/.nemo_curator_metadata`
  with **one LMDB file per writer** (`<host>-<pid>`), not a single shared file
  (LMDB can't be safely shared across hosts on a networked FS like Lustre). Each
  actor writes only its own file and reads the **union** of completed sources
  across all files on startup, so the tasks of a SLURM array can checkpoint to
  the same directory without contention.

## Testing

- `tests/tasks/test_sentinels.py` — `SentinelTask` hierarchy (bare construction,
  payload rejection, `EmptyTask` rooted at `"0"`, `task_id` not user-settable).
- `tests/backends/test_resumability_adapter.py` — counter math + an end-to-end
  `None`→`NoneTask`→strip case (actor RPCs mocked).
- `tests/utils/test_resumability_actor.py` — counter dedup, anomaly recovery,
  lifecycle, and **multi-writer union** (SLURM-array safety).
- Verified end-to-end in the `nemo-curator` container: a
  `Source → Flaky(random FailedTask) → Sink` pipeline re-run against one on-disk
  LMDB checkpoint converges, skips already-completed sources on resume, and
  processes each source exactly once.

### DIFF
--- a/nemo_curator/backends/base.py
+++ b/nemo_curator/backends/base.py
@@ -17,9 +17,13 @@ from abc import ABC, abstractmethod
 from dataclasses import dataclass
 from typing import TYPE_CHECKING, Any
 
+from loguru import logger
+
 from nemo_curator.core.utils import ignore_ray_head_node
 from nemo_curator.tasks import Task
+from nemo_curator.tasks.sentinels import FailedTask, NoneTask
 from nemo_curator.utils.performance_utils import StageTimer
+from nemo_curator.utils.resumability_client import _flush_deltas, _is_active, _skip_completed_sources
 
 if TYPE_CHECKING:
     from nemo_curator.stages.base import ProcessingStage
@@ -85,8 +89,22 @@ class BaseStageAdapter:
             # Use the batch processing logic
             results = self.stage.process_batch(tasks)
 
+        # A returned ``None`` ("filter this slot") becomes a NoneTask so every
+        # output is a real Task that gets a task_id. Sentinels (NoneTask /
+        # FailedTask) carry no identity and are stripped again before this
+        # method returns.
+        results = [NoneTask() if r is None else r for r in results]
+
         # Guarantee every emitted task has a task_id (derived id, or uuid fallback).
         results = self._post_process_task_ids(tasks, results)
+
+        # Opt-in resumability: fire per-source counter deltas. A no-op (the
+        # client helpers self-disable) when no resumability actor is registered.
+        if _is_active():
+            results = self._apply_resumability_counters(tasks, results)
+
+        # Sentinels never propagate to the next stage.
+        results = [r for r in results if not isinstance(r, (NoneTask, FailedTask))]
 
         # Log performance stats and add to result tasks
         _, stage_perf_stats = self._timer.log_stats()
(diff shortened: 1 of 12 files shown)


nemo_curator/tasks/tasks.py
```diff
@@ -41,6 +42,20 @@ class Task(ABC, Generic[T]):
...
     _uuid: str = field(init=False, default_factory=lambda: str(uuid.uuid4()))
 
+    # --- resumability fields ---
+    # `_lineage_path` is the index-based path of this task through the
+    # pipeline DAG (e.g. "3_0_7" = 4th root task, then 1st child, then 8th
+    # grandchild). It is propagated to children via `_set_lineage` and
+    # hashed into `_deterministic_lineage_path_hash`, the deterministic
+    # task id used for retry-deduplication on the resumability actor.
+    _lineage_path: str = field(init=False, default="")
+    _deterministic_lineage_path_hash: str = field(init=False, default="")
```
abhinavg4: Do we really need this? We have task_id, _uuid and _deterministic_lineage_path_hash. They can be combined into a single variable. But before that we need to ensure the usage of task_id and uuid is not affected. Eg:- Writer stage might use task_id right now, we should change to this.

nemo_curator/pipeline/pipeline.py
```diff
@@ -80,6 +88,42 @@ def build(self) -> None:
...
+        explicit_terminals = [s for s in self.stages if s._is_terminal_stage]
+        if len(explicit_terminals) > 1:
+            names = [type(s).__name__ for s in explicit_terminals]
+            msg = (
+                f"Pipeline has multiple terminal stages marked: {names}. "
+                f"Resumability supports at most one."
+            )
+            raise ValueError(msg)
+        if not explicit_terminals:
+            self.stages[-1]._is_terminal_stage = True
```
abhinavg4: Rename terminal stage to sink. Source and sink is better than source and terminal

nemo_curator/backends/base.py
```diff
@@ -90,11 +96,146 @@ def process_batch(self, tasks: list[Task]) -> list[Task]:
...
+                out.append(r)
+
+        _flush_deltas(per_task)
+        return out
+
+    def _handle_source_stage(
+        self,
+        tasks: list[Task],
+        results: list[Task | None],
+    ) -> list[Task]:
```
abhinavg4: For this, it currently does not support a dynamic dataset. Can we have a content-based hash instead of 0,1,2,.... I think our reader might already have something like this, so we can use that. Else fallback to 0,1,2. In case the reader does not have this content-based hash.

nemo_curator/utils/resumability_actor.py
```diff
@@ -0,0 +1,182 @@
...
+
+    def _record_error(self, e: Exception) -> None:
+        self._errors.append(e)
+
+    def _apply_impl(self, per_task: list[tuple[str, str, int]]) -> None:
+        newly_done: list[str] = []
+        for task_hash, sid, d in per_task:
+            existing = self._applied.get(task_hash)
+            if existing is not None:
+                if existing != d:
```
abhinavg4: Here, instead of erroring out, can we just rewrite the new value? So we make delta as -delta_old + delta_new or something. It will also allow us to remove the whole error fire and forget logic so that resumability will never error out. After this error is removed we can get rid of NonDeterministicTaskError. 

For NegativePendingCountError, I think we do this wrong. delta might always be -1, so we go ot 0 and then sid gets popped and gets added _completed, in next run we just continue with it. I think we should do something like this: check if task_id is not in applied but sid is in completed, no matter what delta is, this means theres something terribly wrong with the code. After this, print something like "sid got an update with delta {delta} after being completed, please create an issue at https://github.com/NVIDIA-NeMo/Curator", 

Do not raise so that we can remove the whole erroring out logic here and from base.

abhinavg4 (commented):
Improvements we should probably apply to v0. 

Priority:
1. Rename terminal stage to sink. Source and sink is better than source and terminal
2. Here, instead of erroring out, can we just rewrite the new value? and removing the whole demon logic
3. Unify task_id, _uuid and _deterministic_lineage_path_hash.
4.  support a dynamic dataset

nemo_curator/pipeline/pipeline.py
```diff
@@ -212,4 +264,51 @@ def run(self, executor: BaseExecutor | None = None, initial_tasks: list[Task] |
                     "The executor will schedule GPU stages on GPUs not held by Serve."
                 )
 
-        return executor.execute(self.stages, initial_tasks)
+        if initial_tasks:
+            assign_root_lineage(initial_tasks)
+
+        if checkpoint_path is None:
+            return executor.execute(self.stages, initial_tasks)
+        return self._run_with_resumability(executor, initial_tasks, checkpoint_path)
```
abhinavg4, line 286: I do not like this, the reason this is being done is probably coz we call ray init inside the executor and not here, so we need to differentiate that. Not ideal for sure. Maybe we can make a function for _start_resumability_actor and call it inside each executor. Design choice.

nemo_curator/backends/base.py
```diff
@@ -99,6 +116,15 @@ def process_batch(self, tasks: list[Task]) -> list[Task]:
...
         return results
 
+    @staticmethod
+    def _normalize_none(tasks: list[Task], results: list[Task | None]) -> list[Task | None]:
+        """Replace ``None`` with ``NoneTask`` in the positional case so a
+        filtered slot keeps its position (and gets a task_id). In the fan-out
+        case (lengths differ) ``None`` children are simply dropped."""
+        if len(results) == len(tasks):
+            return [NoneTask.from_input(tasks[i]) if r is None else r for i, r in enumerate(results)]
+        return [r for r in results if r is not None]
```
sarahyurick: Not really clear to me why we would just drop `None` children during a fanout stage.

nemo_curator/backends/base.py
```diff
@@ -99,6 +116,15 @@ def process_batch(self, tasks: list[Task]) -> list[Task]:
 
         return results
 
+    @staticmethod
+    def _normalize_none(tasks: list[Task], results: list[Task | None]) -> list[Task | None]:
```
sarahyurick: Nit but everywhere else is calling them input and output tasks:
```suggestion
    def _normalize_none(input_tasks: list[Task], output_tasks: list[Task | None]) -> list[Task | None]:
```

nemo_curator/backends/base.py
sarahyurick: Furthermore it looks like `_post_process_task_ids` will already drop any `None` objects right?

nemo_curator/backends/base.py
```diff
@@ -168,6 +194,82 @@ def _post_process_task_ids(self, input_tasks: list[Task], output_tasks: list[Tas
...
 
+    # ------------------------------------------------------------------ #
+    # Resumability (opt-in). Runs only when a resumability actor is
+    # registered. task_ids are already assigned by _post_process_task_ids;
+    # this layer only stamps _source_id, fires per-source counter deltas, and
+    # drops already-completed sources. Sentinels are stripped by the caller.
+    # ------------------------------------------------------------------ #
+    def _apply_resumability_counters(self, input_tasks: list[Task], output_tasks: list[Task]) -> list[Task]:  # noqa: C901
+        stage = self.stage
+        if getattr(stage, "is_source_stage", False):
```
sarahyurick, line 197: General question: why would the user ever want something other than source stage being the first stage and sink stage being the last stage of the pipeline? Like if the last stage failed but the second to last stage was the sink stage, they just don't want to rerun the last stage?

nemo_curator/backends/base.py
```diff
@@ -168,6 +194,82 @@ def _post_process_task_ids(self, input_tasks: list[Task], output_tasks: list[Tas
...
+            return self._source_counters(output_tasks)
+
+        # Pre-source stages: inputs carry no _source_id, so there's nothing to
+        # track yet. Leave outputs untouched.
+        if all(not t._source_id for t in input_tasks):
+            return output_tasks
+
+        is_sink = stage.is_sink_stage
+        per_task: list[tuple[str, str, int]] = []
+        real = [t for t in output_tasks if not isinstance(t, (NoneTask, FailedTask))]
```
sarahyurick, line 207: It seems like since `real` is only used in the `if len(input_tasks) == 1 and len(output_tasks) != 1:` block, let's move it there for easier reading?

nemo_curator/backends/base.py
```diff
@@ -168,6 +194,82 @@ def _post_process_task_ids(self, input_tasks: list[Task], output_tasks: list[Tas
...
+        return output_tasks
+
+    def _source_counters(self, output_tasks: list[Task]) -> list[Task]:
+        """Source stage: each output is a source partition. Its ``_source_id``
+        is its own (last) id segment — the content id or index assigned by
+        ``_post_process_task_ids``. Already-completed sources are dropped; each
+        surviving source fires a ``+1``."""
+        sources = [t for t in output_tasks if not isinstance(t, (NoneTask, FailedTask))]
+        for t in sources:
+            t._source_id = t.task_id.rsplit("_", 1)[-1]
```
sarahyurick, line 253: I am a little confused why the source ID here is the last part of the `X_Y_Z` chain? I guess maybe I am confused about how `task_id` versus `_source_id` are constructed/formatted.

abhinavg4:
Heads-up: this PR was **inadvertently squash-merged into the `abhinavg/sentinel-task` feature branch** (not `main`) by a stale auto-merge rule that fired when I retargeted its base. No code reached `main`. I reset the feature branch to its approved state, so the sentinel refactor (#2062) is unaffected. Resumability now continues in #2063 (same diff, stacked on #2062).

PR #2035, opened 2026-05-27
[benchmarking] Multiple fixes to stabilize the nightly benchmark suite

rlratzel:
## Summary

Bundles **eight independent fixes** uncovered while triaging the Curator
nightly benchmark suite against `main` (reference nemo-ci pipelines
[52840568](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/pipelines/52840568),
[53098506](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/pipelines/53098506),
[53323567](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/pipelines/53323567),
[53333039](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/pipelines/53333039),
and [53423863](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/pipelines/53423863)),
plus **three environmental/observability changes** added to enable fair A100
vs EOS comparison runs.

Pass-rate progression: 23/37 → 28/37 → 6/8 (scoped) → 1/2 (scoped) → 40/41
(full suite, 53423863). Sole remaining failure is `video_embedding_xenna`,
a new entry merged in from upstream main that needs a missing NVENC
library — out of scope for this PR.

### 1. Make undefined env vars in config non-fatal by default (`b737202d`)

- `resolve_env_vars` previously raised `ValueError` on the first undefined
  `${VAR}` reference, halting the entire benchmark session even when the
  missing var was used by only a few entries.
- Default behavior is now to substitute an empty string and log a warning
  so unrelated entries can still run.
- Adds `--strict-config-check` CLI flag to `run.py` to restore the old
  fail-fast behavior.

### 2. New `--entries-exact` flag, exact-match entry filtering (`ab1d4385`, `e8c6a52f`, `7b351693`)

Fixes a substring-aliasing bug in CI per-job invocations:

- `--entries` uses pytest's `-k` substring expression evaluator, so
  `--entries audio_tagging_tts_xenna` also matches
  `audio_tagging_tts_xenna_repeat`. In CI, where each per-job script runs
  `--entries <entry-name>`, this caused the non-`_repeat` job to also
  execute the `_repeat` entry, polluting that entry's per-entry results
  dir and crashing the subsequent legitimate `_repeat` SLURM job with
  `Capture file ... already exists` at Ray cluster setup.
- Adds `--entries-exact` accepting a comma-separated list of exact entry
  names. Every supplied name must match a configured (enabled) entry,
  otherwise the run aborts with a `ValueError` listing the unknown names
  alongside the available entry names.
- Mutually exclusive with `--entries` (CLI and `Session.from_dict` both
  enforce this).
- `benchmarking/tools/ci_benchmark_launcher.sh` switched from `--entries`
  to `--entries-exact`.
- Interactive `--entries` substring expression semantics are unchanged.

### 3. Bump too-tight `timeout_s` values for several entries

Multiple `nightly-benchmark.yaml` entries had `timeout_s` values that
didn't cover the actual wall time on the EOS H100 runner. These split
into two categories:

* **Pre-existing tight limits** that overshot during Ray teardown (#3.1).
* **"No-op" timeouts** — set back when the entry effectively did no work
  (e.g. rpv2 data was unreadable, or fuzzy_id_generator artifacts were
  stale and the test failed fast); once the underlying issue was fixed
  and the test ran for real, the historical ceiling was too tight (#3.2,
  #3.3).

#### 3.1 `ndd_ray_serve_dp4`: 700 → 1200 (`ab1d4385`)

- SLURM-killed at the 700s `--time` ceiling ~13s after its benchmark
  subprocess succeeded — Ray teardown overhead pushed total wall over the
  wire.
- 1200s gives ~70% headroom over the observed warm-run wall.
- Reference failed job:
  https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/jobs/327777542

#### 3.2 `exact_dedup_identification`: 500 → 1500 (`21647b4f`)

- The 500s value was set when the test was effectively a no-op (the rpv2
  dataset was unreadable due to filesystem permissions, so the test failed
  fast on `stat()`). Once rpv2 access was restored, the test SLURM-killed
  at 68% (516/755) of the "Inserting into shuffler" phase.
- Shuffler runs at ~1.83 it/s × 755 items = ~410s on EOS H100, plus Ray
  cluster setup + dataset stat + post-shuffle dedup compute + cleanup ≈
  realistic wall 800-1000s.
- 1500s gives ~50% headroom.
- Reference failed job:
  https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/jobs/329434352

#### 3.3 `dedup_removal_*`: 1100/1500 → 1800/1800 (`c2f3a6b0`)

Once the upstream `fuzzy_id_generator` artifacts were refreshed (see
section 4 — the `delete_scratch: false` fix made this possible), the
dedup_removal tests actually ran the workload for the first time:

- `dedup_removal_raydata`: observed wall ~1419s (92% complete when killed
  at the prior 1100s ceiling) → 1800s (~27% headroom over the estimated
  1500s full wall).
- `dedup_removal_xenna`: passed at 1523s wall with the prior 1500s ceiling
  (zero margin) → 1800s as a preemptive bump to match raydata and absorb
  run-to-run variance.

Reference: nemo-ci pipeline 53333039 / leaf 53333328 — raydata killed by
SLURM TIME LIMIT during normal pipeline execution while xenna finished
with zero headroom.

### 4. Install `lynx` in the CI benchmark launcher (`d0eac901`)

The math benchmarks (`math_preprocess`, `math_preprocess_classifier`,
`math_preprocess_llm_cleanup`) shell out to the `lynx` text browser via
`nemo_curator/stages/math/download/html_extractors/lynx.py` for HTML
extraction. `lynx` is not in the Curator container, so those benchmarks
fail with `RuntimeError: lynx executable not found in PATH`.

`lynx` is GPL-licensed, so we deliberately do **not** bake it into the
redistributable Curator image. Instead it is installed transiently in the
existing benchmark container at CI run time, used during the run, and
discarded with the container. The published image stays GPL-free; the
apt-installed `lynx` only lives for the lifetime of each CI container.

### 5. (Reverted) Preserve scratch dir for `fuzzy_dedup_identification` (`63691509` then `a3c38ac6`)

Initial commit `63691509` added `delete_scratch: false` to the
`fuzzy_dedup_identification` entry to keep its scratch artifacts around for
the downstream `dedup_removal_*` benchmarks. In practice the consumption
path is the **canonical dataset path**:

```
{datasets_path}/cleaned_exact_dedup_all_cc_fuzzy_output_nightly_container_paths/
  fuzzy_id_generator.json
  FuzzyDuplicateIds/
```

Operators promote a known-good fuzzy_dedup output to this path once, and
the dedup_removal entries consume it from there on every subsequent
pipeline — there is no same-pipeline dependency. Given that workflow, the
per-entry `delete_scratch: false` override is unnecessary and just leaves
unused data on lustre across runs. Reverted by `a3c38ac6`; the entry now
uses the session-level default (`delete_scratch: true`).

### 6. Range `domain_label_games_count` metric +/-1% (`0c375b7e`, tightened in `ab31cf11`)

Both `domain_classification_raydata` and `domain_classification_xenna`
had a requirement on `domain_label_games_count` with `exact_value: 149816`.
Run-to-run output of the classifier drifts by several hundred to a few
thousand classifications, causing repeated false failures while the actual
benchmark (throughput, total docs, number of domains predicted) is
healthy.

Loosens the metric to a +/-1% range (`min_value: 148318`,
`max_value: 151314`) which still catches genuine regressions of the
classifier output while tolerating normal run-to-run variability.
`domain_label_news_count` is left at `exact_value: 2817` pending further
investigation.

### 7. Raise A100 container memory cap to host capacity (`fd6e9cfd`)

The benchmarking container on the A100 host was capped at 1 TiB RAM (and
512 GiB shm, derived as 50% of container memory) by
`_max_container_memory_bytes = 1 * _TB` in
`benchmarking/tools/gen_runscript_vars.py`. The A100 host actually has
~2 TiB RAM and ~1008 GiB `/dev/shm` — comparable to the EOS H100 host.
The artificial cap made A100's `env.json` report ~50% of host RAM and
~25% of host shm, surfacing a fake environmental delta in A100 vs EOS
comparison runs even though the workload (Ray plasma capped at 500 GB
on both sides) was bounded identically.

Raises the cap to `2 * _TB` so A100's container memory and shm reflect
host capacity. After this change, A100 `env.json` reports
`total_system_memory_bytes` ≈ 2.16 × 10¹² and `shm_size_bytes` ≈
1.08 × 10¹² — matching EOS.

Also removes an experimental `CURATOR_SHM_SIZE_BYTES` env-var block in
`ci_benchmark_launcher.sh` that attempted to cap EOS-side `/dev/shm` via
`mount -o remount,size=…`. Pyxis containers on EOS don't grant
`CAP_SYS_ADMIN`, so the remount silently fell back to the WARNING branch
and the cap never applied. Raising A100 up to match EOS is the
maintainable direction (the EOS shm ceiling cannot be capped down without
nemo-ci infrastructure changes), so the toggle is no longer needed.

### 8. Background per-GPU stats recorder (`c2fd47e6`)

Adds `benchmarking/runner/gpu_stats_recorder.py` — `GPUStatsRecorder`,
a context-managed daemon thread that polls all GPUs via `gpustat`
(already a curator dep) and writes one row per (timestamp, GPU) pair to
`{session_entry_path}/gpustats.csv` while each entry's benchmark
subprocess runs.

CSV columns:

| column | content |
|---|---|
| `timestamp_utc` | ISO-8601 timestamp |
| `gpu_id` | NVML index |
| `utilization_gpu_pct` | compute utilization 0-100 |
| `utilization_memory_pct` | `memory_used / memory_total * 100` |
| `temperature_c` | GPU temperature in °C |
| `processes` | JSON-encoded list of `{pid, username, command, gpu_memory_usage}` |

Configurable via a new top-level YAML key in `nightly-benchmark.yaml`:

```yaml
gpu_stats_recorder:
  interval_s: 1.0   # 1 Hz; set to 0 to disable
```

The recorder polls all NVML-visible GPUs, **independent of
`CUDA_VISIBLE_DEVICES`**. This lets us verify post-run whether Ray/Xenna
actually honored the visible-device mask in cross-host comparison runs
(any nonzero util on a masked GPU index = leakage).

Replaces the prior shell-based `nvidia-smi -l` poller (block removed from
`ci_benchmark_launcher.sh`) which lived outside the runner lifecycle,
wrote ad-hoc file names, and required a separate env-var toggle to enable.

### 9. Wire `NEMO_CI_*` env vars + Slack viewer-URL block (`da4c6760`)

Curator-side plumbing for an upcoming nemo-ci-hosted developer-launch
wrapper (`curator_benchmark_launch.py`, still under design). All three
edits are gated behind new env vars / flags so today's nightly schedule
is byte-identical to before.

* **`ci_benchmark_launcher.sh`**:
  * Honors `NEMO_CI_SESSION_NAME` when set; otherwise picks `nightly-<TS>`
    for scheduled (cron) pipelines and the legacy `benchmark_run_<id>`
    name for direct `launch_pipeline.py` launches.
  * Composes a run-viewer URL only when `NEMO_CI_VIEWER_HOST` is set,
    using the host-side lustre path and the resolved session name.
  * Passes `--viewer-url` and `--run-reason` through to `run.py` when set.

* **`run.py`**: new `--viewer-url` and `--run-reason` CLI flags. When
  `--run-reason` is set it lands in `env_dict` (and thus `env.json` +
  the Slack environment block). When `--viewer-url` is set the Slack
  sink's `sink_config["viewer_url"]` is patched in-process before
  `sink.initialize(...)` — no YAML schema change required.

* **`slack_sink.py`**: `SlackSink` reads optional `viewer_url` from
  `sink_config`; `SlackParentMessage` accepts `viewer_url` and renders
  a `*Results viewer:* <{url}|open run>` section between the
  overall-status block and the environment table, only when set.

The new `NEMO_CI_*` namespace is shared with the upcoming nemo-ci
launcher work and is intentionally generic (not curator-specific) so
other modules can opt in later.

## Verification

Across multiple nemo-ci pipelines:

* [53098506](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/pipelines/53098506)
  — first verification of commits 1, 2, 3.1, 3.2: **28/37** pass (up from
  23/37 baseline). Confirmed `audio_tagging_tts_xenna_repeat` works
  (`--entries-exact`), `ndd_ray_serve_dp4` works (timeout bump),
  `exact_dedup_identification` works (rpv2 + timeout),
  `audio_tagging_tts_xenna` works without `HF_SECRET_KEY` set.
* [53323567](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/pipelines/53323567)
  — scoped 8-entry run of commits 4, 5, 6: **6/8** pass (`math_preprocess*`
  all pass thanks to lynx install; `domain_classification_*` pass with the
  metric range; `fuzzy_dedup_identification` passes). The two failures
  (`dedup_removal_*`) were caused by a same-pipeline race with
  fuzzy_dedup_identification — fixed by operators promoting fresh
  `fuzzy_id_generator.json` artifacts to the canonical dataset path
  between runs.
* [53333039](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/pipelines/53333039)
  — scoped 2-entry run with fresh artifacts in place: `dedup_removal_xenna`
  passes (1523s, no margin); `dedup_removal_raydata` SLURM-killed by TIME
  LIMIT at 92% completion → addressed by commit `c2f3a6b0` (#3.3).
* [53423863](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/pipelines/53423863)
  — full-suite verification after #3.3: **40/41 pass**. Sole failure is
  `video_embedding_xenna` (new entry merged from upstream main, missing
  NVENC library — separate follow-up).
* [53584394](https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/pipelines/53584394)
  — in flight. Verifies #7 (cap bump doesn't break EOS) and #8 (recorder
  always-on writes `gpustats.csv`); also captures empirical evidence of
  whether `CUDA_VISIBLE_DEVICES=0,1,2,3` actually restricts Ray/Xenna to
  4 GPUs on EOS.

### Undefined env vars
- [x] `audio_tagging_tts_xenna` ran successfully in 53098506 with
  `HF_SECRET_KEY` unset.
- [ ] Run with `--strict-config-check`; confirm it still exits with the
  original `ValueError`.
- [ ] Run an unmodified config (all env vars defined); confirm no
  behavior change.

### `--entries-exact`
- [x] CI per-job invocation ran exactly one entry per job in 53098506
  and 53323567; no cross-entry pollution observed.
- [ ] `--entries-exact <typo>` exits with an error listing the unknown
  name and the available entry names.
- [ ] `--entries-exact a,b,c` runs only those three entries in YAML order.
- [ ] Passing both `--entries` and `--entries-exact` exits with a
  "mutually exclusive" error.

### Timeout bumps
- [x] `ndd_ray_serve_dp4` finished without TIME LIMIT cancellation in
  53098506.
- [x] `exact_dedup_identification` finished without TIME LIMIT in 53098506.
- [x] `dedup_removal_raydata` and `dedup_removal_xenna` both finish
  without TIME LIMIT cancellation (verified in 53398187 / 53423863).

### lynx install
- [x] `math_preprocess*` entries all passed in 53323567 — `RuntimeError:
  lynx executable not found` no longer fires.

### fuzzy_dedup scratch retention (now reverted)
- [x] One-time artifact promotion completed: fresh
  `fuzzy_id_generator.json` + `FuzzyDuplicateIds/` written to the
  canonical dataset path.
- [x] `dedup_removal_*` consume the promoted artifacts in a clean
  pipeline (verified in 53398187).
- [x] After artifact promotion, the per-entry `delete_scratch: false`
  override on `fuzzy_dedup_identification` is no longer needed and has
  been reverted (commit `a3c38ac6`).

### domain_label_games_count range
- [x] `domain_classification_*` both passed in 53323567 with the new
  range satisfied.

### A100 container memory cap (#7)
- [ ] On the A100 nightly host, after pulling commit `fd6e9cfd`, a fresh
  run produces `env.json` with `total_system_memory_bytes` ≈ host total
  (~2.16 × 10¹²) and `shm_size_bytes` ≈ ~1.08 × 10¹².
- [x] EOS pipelines continue to succeed (no regression observed from
  the unrelated `ci_benchmark_launcher.sh` shm-block removal —
  verified in 53584394 run-up).

### GPUStatsRecorder (#8)
- [ ] After 53584394 completes, `gpustats.csv` exists under each
  per-entry session dir on EOS lustre.
- [ ] CSV header matches `timestamp_utc,gpu_id,utilization_gpu_pct,utilization_memory_pct,temperature_c,processes`.
- [ ] Setting `gpu_stats_recorder.interval_s: 0` in the YAML disables
  the recorder (no file written); default (omitted key) gives 1 Hz.
- [ ] Recorder polls 8 rows per timestamp on an EOS node regardless of
  `CUDA_VISIBLE_DEVICES` (NVML-visible, not CUDA-runtime-visible).


benchmarking/nightly-benchmark.yaml
```diff
@@ -939,7 +939,7 @@ entries:
       --executor=ray_data
       --engine-kwargs='{"tensor_parallel_size": 1}'
       --autoscaling-config='{"min_replicas": 4, "max_replicas": 4}'
-    timeout_s: 700
+    timeout_s: 1200  # warm-run wall ~700s observed; headroom added for cold vLLM model load (cf. ndd_dynamo_dp4: 2700)
```
praateekmahajan: I'm wondering why this happened now?

praateekmahajan: Because even dynamo should not take 2700s rn since we upgraded versions to 1.1.0 (i.e.we can reduce that)

praateekmahajan (approved):

benchmarking/nightly-benchmark.yaml
rlratzel: I need to investigate that. I know we had success on the DGX-A100 machine with 700s, but I don't know yet if a larger timeout is needed for the other machine running nightlies because it's slower in general, or if something else is causing a longer runtime on the other machine.

benchmarking/runner/gpu_stats_recorder.py
```diff
@@ -0,0 +1,132 @@
...
+            mem_total = gpu.memory_total or 0
+            mem_pct = (gpu.memory_used / mem_total * 100.0) if mem_total else 0.0
+            procs = [{k: p.get(k) for k in ("pid", "username", "command", "gpu_memory_usage")} for p in gpu.processes]
+            self._csv_writer.writerow(
+                [
+                    ts,
+                    gpu.index,
+                    gpu.utilization,
+                    round(mem_pct, 2),
+                    gpu.temperature,
```
praateekmahajan, line 134: Can we add power_draw / power_limit  / fan_speed as well?

benchmarking/nightly-benchmark.yaml
```diff
@@ -204,9 +214,13 @@ entries:
...
         exact_value: 26
-      # We choose two domains randomly and ensure that the counts are as expected
+      # We choose two domains randomly and ensure that the counts are as expected.
+      # domain_label_games_count is given a +/-5% tolerance window (observed
+      # 149816 +/- ~7500) because the classifier output drifts run-to-run;
+      # domain_label_news_count remains exact pending further investigation.
       - metric: domain_label_games_count
-        exact_value: 149816
+        min_value: 142325
+        max_value: 157307
```
praateekmahajan: 1% is more than enough, instead of 5%

benchmarking/nightly-benchmark.yaml
```diff
@@ -273,7 +287,7 @@ entries:
       --input-filetype=parquet
       --text-field=raw_content
       --input-blocksize=3GiB
-    timeout_s: 500
+    timeout_s: 1500  # observed shuffler ~410s on EOS H100 on real data; +headroom for setup + post-shuffle stages. 500s was sufficient on a faster system — revisit if EOS becomes a bottleneck
```
praateekmahajan: can remove this comment, wdyt?

benchmarking/nightly-benchmark.yaml
```diff
@@ -369,7 +383,7 @@ entries:
       --id-field=_curator_dedup_id
       --duplicate-id-field=_curator_dedup_id
       --blocksize=1.5GiB
-    timeout_s: 1100
+    timeout_s: 1800  # observed wall ~1419s on EOS with real fuzzy_id_generator data (92% complete at the prior 1100s ceiling); +headroom for variance
```
praateekmahajan: can remove this comment wdyt?

praateekmahajan (approved):
LGTM! Thanks for this 🙏

PR #2036, opened 2026-05-27
Unify task identity: collapse _uuid into deterministic task_id

abhinavg4:
## Summary

Collapses the random per-task \`_uuid\` field into a deterministic \`task_id\` set by the framework. \`task_id\` is now derived from the task's lineage path through the pipeline DAG (a sha256-32 hash); two runs of the same pipeline on the same inputs produce byte-identical \`task_id\`s across all tasks.

This is a small, low-risk prerequisite for the resumability work in #2033, which will be rebased on top to use \`task.task_id\` as the per-task dedup key instead of a separate \`_deterministic_lineage_path_hash\` field.

## What changes

- **\`Task\` class** (\`nemo_curator/tasks/tasks.py\`)
  - DROP \`_uuid\` field.
  - ADD \`_lineage_path\` field (\`init=False, default=\"\"\`).
  - ADD \`_set_lineage(parent_lineage_paths, child_segment)\` method. Always overwrites \`_lineage_path\` and \`task_id\` (no idempotency check). \`child_segment\` accepts either a positional \`int\` index or a content-based \`str\` (for source-stage emissions).
  - ADD \`get_deterministic_id() -> str | None\` method (default \`None\`). Subclasses with stable content override.
- **\`FileGroupTask\`** overrides \`get_deterministic_id()\` to return \`get_deterministic_hash(sorted(self.data))\` — stable across runs even if files are added/removed between launches.
- **\`ProcessingStage.process_batch\`** default impl now calls \`_set_lineage\` per emitted child. Stages that override \`process_batch\` are responsible for calling it themselves (docstring updated).
- **Migrate 6 \`_uuid\` use sites** to \`task.task_id\` (all dedup output-filename generators + AddId).

## Why no breaking change at the constructor

\`Task.task_id\` is still a regular \`init=True\` constructor argument. User code that passes \`task_id=\"file_group_3\"\` continues to work — the value is just immediately overwritten by \`_set_lineage\` when the task flows through any stage. The framework retains the ability to enforce \`init=False\` strictly in a future cleanup PR, but doing it here would force migrating ~55 internal constructor calls and isn't needed for the resumability work.

## Test plan

- [x] \`tests/tasks/test_tasks.py\`: updated fanout test to call \`process_batch\` (the path that exercises the new lineage assignment).
- [x] \`tests/tasks/test_lineage.py\` (NEW):
  - \`_set_lineage\` filters empty parent paths, always overwrites, accepts \`str\` and \`int\` segments.
  - \`get_deterministic_id\` returns \`None\` by default; \`FileGroupTask\` returns a stable hash regardless of input list order.
  - Default \`process_batch\` assigns unique lineage paths to fan-out outputs.
  - Same pipeline shape on same inputs produces byte-identical \`task_id\`s.

## Future work

- \"Strict\" mode: make \`task_id\` \`init=False\` so users can't pass it at all. Requires migrating ~55 internal \`task_id=...\` constructor calls — mechanical but big diff. Separate PR.
- Resumability (#2033) rebases on top of this and uses \`task.task_id\` as the per-task dedup key.


nemo_curator/stages/audio/preprocessing/concatenation.py
```diff
@@ -111,7 +111,7 @@ def process(self, task: AudioTask) -> AudioTask | list[AudioTask]:
         segments_sorted = sorted(segments, key=self._seg_sort_key)
         original_file = segments_sorted[0].get("original_file", "unknown")
 
-        combined = self._concatenate(original_file, segments_sorted, task.task_id, task.dataset_name)
+        combined = self._concatenate(original_file, segments_sorted, task.dataset_name)
```
praateekmahajan, line 114: @mohammadaaftabv can you / someone from audio review if this is fine? We're removing task_id

nemo_curator/stages/deduplication/fuzzy/lsh/stage.py
```diff
@@ -155,11 +155,10 @@ def insert_finished(self) -> None:
 
     def extract_and_write(self) -> list[FileGroupTask]:
         self._check_actor_obj()
-        current_band_min, current_band_max = self._current_band_range
+        _current_band_min, _current_band_max = self._current_band_range
```
praateekmahajan, line 158: @ayushdg

nemo_curator/tasks/file_group.py
```diff
@@ -46,3 +46,12 @@ def validate(self) -> bool:
...
         return True
+
+    def get_deterministic_id(self) -> str:
+        """Content-based id derived from the sorted file paths. Stable
+        across runs even if the source stage emits the file group at a
+        different position (e.g. because new files were added or removed
+        between runs)."""
+        from nemo_curator.stages.text.io.writer.utils import get_deterministic_hash
+
+        return get_deterministic_hash(sorted(self.data))
```
praateekmahajan, line 57: Nice!
nit: move this `get_deterministic_hash` out to somewhere in utils outside text? That way the whole text dependencies need not be resolved for a non text modality

nemo_curator/stages/file_partitioning.py
```diff
@@ -64,6 +64,7 @@ class FilePartitioningStage(ProcessingStage[_EmptyTask, FileGroupTask]):
     storage_options: dict[str, Any] | None = None
     limit: int | None = None
     name: str = "file_partitioning"
+    is_source_stage: bool = True
```
praateekmahajan: this should be `field(init=False)` right? i.e. we don't want this to be modifiable by the user

tests/pipelines/test_pipeline_roles.py
```diff
@@ -0,0 +1,97 @@
...
+        assert s2.is_source_stage is False
+
+    def test_defaults_last_stage_to_sink(self) -> None:
+        s0 = _NoopStage(name="s0")
+        s1 = _NoopStage(name="s1")
+        s2 = _NoopStage(name="s2")
+        Pipeline(name="t", stages=[s0, s1, s2]).build()
+        assert s2.is_sink_stage is True
+        assert s0.is_sink_stage is False
+        assert s1.is_sink_stage is False
```
praateekmahajan: nit: AI overdoes on the unittests and writes too many lines.. can we possibly consolidate into a single test maybe within `test_pipelines.py` where we have class `TestPipelineBuild` and that has a test that says `tests_default_first_source_last_sing_stage(self)` which tests both these things?

That way any future changes to pipeline.build(..) also live there

nemo_curator/stages/text/download/base/url_generation.py
```diff
@@ -69,7 +69,6 @@ def process(self, task: _EmptyTask) -> list[FileGroupTask]:
 
         return [
```
praateekmahajan, line 70: This stage should also be a source stage btw

nemo_curator/stages/base.py
```diff
@@ -192,10 +212,21 @@ def process_batch(self, tasks: list[X]) -> list[Y]:
...
+            for i, child in enumerate(children):
+                if child is None:
+                    continue
+                # Source stages: prefer a content-based segment from
+                # `get_deterministic_id()` (e.g. `FileGroupTask` hashes its
+                # file paths). Falls back to the positional index when the
+                # child's Task subclass doesn't implement one. This makes
+                # task_ids stable across reorderings of the source input.
+                if self.is_source_stage:
+                    segment: str | int = child.get_deterministic_id() or i
```
praateekmahajan: `segment` means the same thing as `current_task_id` right? or `current_task_id_suffix`? 
nit: Can we rename it?

nemo_curator/tasks/tasks.py
```diff
@@ -12,34 +12,46 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import uuid
+from __future__ import annotations
```
praateekmahajan: why do we need future?

nemo_curator/tasks/tasks.py
```diff
@@ -54,6 +66,43 @@ def add_stage_perf(self, perf_stats: StagePerfStats) -> None:
...
+                contribute a leading ``"_"`` to the path).
+            child_segment: Either a positional index (``int`` → coerced
+                to ``str``) for plain emissions, or a string id (e.g. a
+                content-based hash from :py:meth:`get_deterministic_id`)
+                for source-stage emissions where stability across input
+                reordering matters.
+        """
+        parts = [*[p for p in parent_lineage_paths if p], str(child_segment)]
+        self._lineage_path = "_".join(parts)
+        self.task_id = hashlib.sha256(self._lineage_path.encode()).hexdigest()[:32]
```
praateekmahajan: Maybe claude knows but I've usually seen
```
import uuid
stable_id = uuid.uuid5(uuid.NAMESPACE_DNS, YOUR_SEED_HERE)
```

tests/tasks/test_lineage.py
```diff
@@ -0,0 +1,285 @@
...
+        t._set_lineage([], 3)
+        assert t._lineage_path == "3"
+        # task_id is the sha256-32 of the lineage path.
+        assert len(t.task_id) == 32
+
+    def test_set_lineage_filters_empty_parent_paths(self) -> None:
+        t = _new()
+        # Empty strings in parent paths are stripped (EmptyTask's default
+        # _lineage_path is "").
+        t._set_lineage(["", "5", ""], 3)
```
praateekmahajan: When can this happen?

tests/tasks/test_task_id.py
```diff
@@ -0,0 +1,285 @@
...
+        assert t._lineage_path == "abc123"
+
+    def test_child_segment_under_parent_lineage(self) -> None:
+        """Source under a non-empty parent lineage encodes both."""
+        t = _new()
+        t._set_lineage(["root"], "abc123")
+        assert t._lineage_path == "root_abc123"
+
+
+class TestGetDeterministicId:
```
praateekmahajan: Can this test be inside tasks/test_file_group_tasks.py where we have a test for file group task where we do this...  under TestFileGroupTask def test_deterministic_ids(..): in which we gtest few different asserts at once..

tests/tasks/test_lineage.py
```diff
@@ -0,0 +1,285 @@
...
+        return [], []
+
+    def outputs(self) -> tuple[list[str], list[str]]:
+        return [], []
+
+    def process(self, task: _SimpleTask) -> _SimpleTask | None:
+        return None  # always filter out
+
+
+class TestDefaultProcessBatchAssignsLineage:
```
praateekmahajan: These tests are good, but can we add a new test method here https://github.com/NVIDIA-NeMo/Curator/blob/main/tests/backends/test_integration.py#L53

Where we do `test_lineage()` / `test_task_ids()`

praateekmahajan (changes requested):
Mostly LGTM, left some feedback re tests and clarifying questions

nemo_curator/stages/file_partitioning.py
abhinavg4: No, user can modify it. If the user provides a source stage, we use that; we assume that the first stage is a source stage. This line should probably be removed.

nemo_curator/tasks/file_group.py
abhinavg4: Yup I agree.

tests/pipelines/test_pipeline_roles.py
abhinavg4: Yup probably we can do that. @oyilmaz-nvidia what do you think?

nemo_curator/stages/text/download/base/url_generation.py
abhinavg4: Hmm other way I think, we should remove is_source_stage from the FilePartitioning as well. The user can provide it else we assume first stage is source

nemo_curator/stages/base.py
abhinavg4: Actually, neither. It means the current task's ID in the lineage tree. For non-source, it's just i, and for source it's deterministic_hash. Re naming, whatever feels good to you as a reader of the PR or to @oyilmaz-nvidia

nemo_curator/tasks/tasks.py
abhinavg4: I think it was since we had int | str or something in the typedef. Probably AI dump to support older Python version too. Can we removed?

nemo_curator/tasks/tasks.py
abhinavg4: Hmmm no idea about this. it was copied over from @oyilmaz-nvidia 's code. He can add/modify

tests/tasks/test_lineage.py
abhinavg4: It cannot. First empty is due to an Empty Task, and the second cannot occur.

abhinavg4: Wait until get_deteministic_hash gives "" as the output. I think unrelated, but this is a bug. We need to ensure that get_deteministic_hash cannot return an empty string. @oyilmaz-nvidia I might be wrong, but that sound correct to you/

nemo_curator/tasks/tasks.py
suiyoubi: why can't we just use lineage_path directly as the task_id ? (without uuid5 or sha256), if lineage_path is supposed to be unique and keeping that as the task_id would make it much easier to debug I suppose ?

abhinavg4: Hmmm, actually we can, it's just that it might look ugly and a lot of long filenames like:

0_ed23fgr_2_4.parquet

0_rd21fgq_0_1.parquet
.......

But I agree it will make the debugging easy. Let me check with the team about this.

abhinavg4:
/ok to test [75b9298](https://github.com/NVIDIA-NeMo/Curator/pull/2036/commits/75b9298be45a58cb5257a82f54b3ecc8ebba945b)

abhinavg4:
/ok to test [976f16d](https://github.com/NVIDIA-NeMo/Curator/pull/2036/commits/976f16d9e5ac65b385beddacaec710b61bb89ec8)

nemo_curator/tasks/tasks.py
```diff
@@ -65,7 +117,15 @@ def validate(self) -> bool:
 
 @dataclass
 class _EmptyTask(Task[None]):
```
VibhuJawa, line 126: Curious: Why did we decide against adding a NoneTask ?  I remember it being discussed at one point

nemo_curator/backends/base.py
```diff
@@ -95,6 +99,58 @@ def process_batch(self, tasks: list[Task]) -> list[Task]:
...
+          empty even when lineage can't be derived. The ``"r"`` prefix flags
+          the id as non-deterministic / lineage-not-tracked (see
+          ``Task.task_id`` docstring).
+
+        ``seg`` is the output's content id (``Task.get_deterministic_id()``)
+        for a source stage when available, else the positional index — so a
+        source partition keeps a stable id across reorderings regardless of
+        whether the source is 1→N or N→N.
+        """
+        out = [r for r in results if r is not None]
```
VibhuJawa: We are decreasing/dropping N the length here, is that fine ?  Why are we doing that here ? Does Xenna support that now ?

nemo_curator/backends/base.py
```diff
@@ -95,6 +99,58 @@ def process_batch(self, tasks: list[Task]) -> list[Task]:
 
         return results
 
+    def _post_process_task_ids(self, tasks: list[Task], results: list[Task | None]) -> list[Task]:
```
VibhuJawa: Nit: 
```suggestion
    def _post_process_task_ids(self, input_tasks: list[Task], output_tasks: list[Task | None]) -> list[Task]:
```

nemo_curator/backends/base.py
```diff
@@ -95,6 +99,58 @@ def process_batch(self, tasks: list[Task]) -> list[Task]:
...
+
+        if len(tasks) == 1:
+            parent_id = tasks[0].task_id
+            for i, r in enumerate(out):
+                suffix = (r.get_deterministic_id() or i) if is_source else i
+                r._set_lineage([parent_id], suffix)
+        elif len(out) == len(tasks):
+            for parent, r in zip(tasks, out, strict=True):
+                suffix = (r.get_deterministic_id() or 0) if is_source else 0
+                r._set_lineage([parent.task_id], suffix)
```
VibhuJawa: Question: 
How is the positional 1:1 mapping guaranteed here?
*With codex*

I searched the current process_batch overrides and there are stages whose batch cardinality is not always 1:1, e.g. PreserveByValueStage.process_batch() returns only surviving tasks after filtering, while some dedup stages aggregate many inputs into one output. 

Like is this possible ? 
```
# Input batch:
tasks = [
    Task(task_id="0_A"),
    Task(task_id="0_B"),
]

# Stage behavior:
# - A fans out into two children
# - B is filtered out (because we seem to be dropping nones now)
results = [
    child_from_A_0,
    child_from_A_1,
]

# len(results) == len(tasks), so current code takes the 1:1 branch:
for parent, r in zip(tasks, results, strict=True):
    r._set_lineage([parent.task_id], 0)

# Assigned lineage:
child_from_A_0.task_id == "0_A_0"  # correct
child_from_A_1.task_id == "0_B_0"  # wrong: this child came from A, not B
```

VibhuJawa (changes requested):
Trying to understand these changes. Mostly questions.

nemo_curator/tasks/tasks.py
abhinavg4: Hmmm I think that is in another PR. The next one. The way I understand it (We can always reduce the number of sential tasks later)

Empty Task: Just the first task that a Pipeline starts with
None Task: Filtered out task (Added in next PR)
Failed Task: A task that is failed and should be retried (Added in next PR.)

nemo_curator/backends/base.py
abhinavg4: Hmm not sure if I get this. But Xenna always supported N->M where N and M can be anything including None. (M can be empty list)

nemo_curator/backends/base.py
abhinavg4: Makes sense to me. Will commit in next PR or later if we have more changes since don't wanna rerun the whole test suite for this.

nemo_curator/backends/base.py
abhinavg4: Yeah, so for a stage, if it goes from N -> M and N != M, there is no simple way to get lineage. 

When N == M, we **assume** that there is a 1:1 positional mapping. We do NOT guarantee it, we just assume it.

For your above example, this assumption will fail. But I think having a pipeline with a stage that fans out and filters at the same time might be rare.

I am not sure if we have any good way of detecting whether 1:1 position mapping is correct. Basically, if N == M, we cannot ensure if it was 1:1 or it was some task that was filtered and other fanned out.

nemo_curator/backends/base.py
abhinavg4: Hmm, ok, I need to rebase with main anyways, so might as well do this.

nemo_curator/backends/base.py
abhinavg4: Btw, we cover PreserveByValueStage and dedup stages (Basically filtering and fan out/in stages). What we do not support is a stage that filters and fans out.

VibhuJawa: I guess, my main gap comes from this ` out = [r for r in results if r is not None]` , like now output can't be None but previously it could be so the length guarantees change more ! 

I am trying to understand why we decide to drop this if you are using length guarantees (or hints of that) .

abhinavg4: Hmmm, fair enough, so for batch process, now it breaks for filter stages. If we have a stage like this:

Input: [Task1, Task2, Task3]

Output: [Task1, None, Task3]

This gets messed up, although the lineage is clear. I think for this PR, I can either (1) not drop none tasks (2) introduce NoneTask and FailedTask in this PR itself.

(1) Might be cleared for this PR. Wdyt ?

VibhuJawa: Yeah, 1 is what i was thinking too. Lets add  NoneTask and FailedTask  later on

VibhuJawa (commented):
Looks good to me.  But will wait on @praateekmahajan   who understands this code path better before landing in.

nemo_curator/tasks/tasks.py
```diff
@@ -54,6 +67,45 @@ def add_stage_perf(self, perf_stats: StagePerfStats) -> None:
         """Add performance stats for a stage."""
         self._stage_perf.append(perf_stats)
 
+    def _set_lineage(self, parent_task_ids: list[str], current_task_id_suffix: str | int) -> None:
```
abhinavg4: set_task_id and not set_lineage. WE do not have any concept of lineage right now?

abhinavg4:
/ok to test [681744d](https://github.com/NVIDIA-NeMo/Curator/pull/2036/commits/681744d192b40f2a7bc652c9adff39de074eaa83)

abhinavg4:
/ok to test [5757fba](https://github.com/NVIDIA-NeMo/Curator/pull/2036/commits/5757fba984e70e17ec9ccfe17c6f1f36ee336545)

nemo_curator/pipeline/pipeline.py
```diff
@@ -80,6 +105,30 @@ def build(self) -> None:
...
+        # marked; if none, the first stage is the source and the last is
+        # the sink. The source flag activates content-based ids in the
+        # default ``process_batch``; the sink flag is used by the
+        # resumability layer in a follow-up PR.
+        self._assign_source_sink_roles()
+
+    def _assign_source_sink_roles(self) -> None:
+        explicit_sources = [s for s in self.stages if s.is_source_stage]
+        if len(explicit_sources) > 1:
+            names = [type(s).__name__ for s in explicit_sources]
```
praateekmahajan: Is `type(s).__name__` different than `stage.name` property?

nemo_curator/pipeline/pipeline.py
```diff
@@ -80,6 +105,30 @@ def build(self) -> None:
...
+            msg = f"Pipeline has multiple source stages marked: {names}. At most one is supported."
+            raise ValueError(msg)
+        if not explicit_sources:
+            self.stages[0].is_source_stage = True
+
+        explicit_sinks = [s for s in self.stages if s.is_sink_stage]
+        if len(explicit_sinks) > 1:
+            names = [type(s).__name__ for s in explicit_sinks]
```
praateekmahajan: Is type(s).__name__ different than stage.name property?

tests/backends/test_integration.py
```diff
@@ -136,6 +136,27 @@ def test_output_tasks(self):
...
+
+        # Task ids are unique per task.
+        assert len(set(task_ids)) == len(task_ids), "task_ids should be unique"
+
+        # Each id is a clean "_"-joined id path: no empty segments, so no
+        # leading/trailing/double underscores (the source's EmptyTask parent,
+        # whose id is "", is filtered out by _set_task_id).
+        for tid in task_ids:
+            segments = tid.split("_")
+            assert all(segments), f"task_id {tid!r} has an empty id segment"
```
praateekmahajan, line 160: If in these pipeline they're deterministic can we just test those ids? Or do we have a fan-in stage here making them non-deterministic?

tests/backends/test_task_id_postprocess.py
```diff
@@ -0,0 +1,150 @@
+# Copyright (c) 2026, NVIDIA CORPORATION.  All rights reserved.
```
praateekmahajan, line 1: So far our tests have one is to one mapping of source file to test file (for the most part), which makes it super easy to reason in a large code base where the logic of that particular thing is implemented? Would there be any hesitation to move this to a corresponding source test file?

praateekmahajan: For me bunch of these tests seem like integration tests of running a series of stages for a given task and getting an output, so can we try to replicate this in our tests/backends/test_integration.py

nemo_curator/stages/base.py
```diff
@@ -182,6 +192,14 @@ def process_batch(self, tasks: list[X]) -> list[Y]:
...
         with each element corresponding to the result of processing the task
         at the same index.
+
+        ``task_id`` is framework-owned: stages must NOT set it. The executor
+        adapter (``BaseStageAdapter._post_process_task_ids``) assigns a
+        deterministic id to every emitted task — regardless of whether
+        a stage uses this default or overrides ``process_batch``. Where the
+        input→output mapping is ambiguous (e.g. a batch aggregation), the
+        adapter falls back to a random ``"r"``-prefixed id (see
+        ``Task.task_id``); there is no way for a stage to supply its own.
```
praateekmahajan, line 202: Are you sure ruff is running on this? Aren't the characters `-` / `→` problematic? Can you ensure pre-commit did run on these because our CI still doesn't catch them

nemo_curator/backends/base.py
```diff
@@ -95,6 +99,75 @@ def process_batch(self, tasks: list[Task]) -> list[Task]:
...
+        ``seg`` is the output's content id (``Task.get_deterministic_id()``)
+        for a source stage when available, else the positional index — so a
+        source partition keeps a stable id across reorderings regardless of
+        whether the source is 1→N or N→N.
+
+        Note: a stage that BOTH filters and fans out within a single batch
+        (returning a flat list rather than a per-input slot) cannot be mapped
+        positionally; if its length happens to equal the input length the 1:1
+        assumption may misattribute parents. That combination is unsupported
+        until per-slot sentinels (NoneTask/FailedTask) land in a later PR.
```
praateekmahajan, line 137: Are `NoneTask`'s   or can we have `None`?

nemo_curator/backends/base.py
```diff
@@ -95,6 +99,75 @@ def process_batch(self, tasks: list[Task]) -> list[Task]:
...
+        """
+        is_source = getattr(self.stage, "is_source_stage", False)
+
+        if len(input_tasks) == 1:
+            # Fan-out (incl. a source reading from EmptyTask): every non-None
+            # output is a child of the single input.
+            parent_id = input_tasks[0].task_id
+            out: list[Task] = [t for t in output_tasks if t is not None]
+            for i, task in enumerate(out):
+                suffix = (task.get_deterministic_id() or i) if is_source else i
```
praateekmahajan, line 147: If something was not a source and wanted to still override get_deterministic_id do we not want to support that?

nemo_curator/pipeline/pipeline.py
```diff
@@ -19,6 +19,31 @@
...
+    NOTE: we deliberately use the positional index here, NOT
+    ``get_deterministic_id()``, even for content-bearing tasks like
+    ``FileGroupTask``. The source stage is the single place content-based
+    ids are assigned (to its outputs); hashing here too would put the
+    content hash at two levels of the id path (``"0_<hashA>_<hashB>"``).
+    Passing initial tasks directly is rare; if you need reorder-stable
+    source ids, let a source stage emit them.
+    """
+    for i, task in enumerate(initial_tasks):
+        if isinstance(task, _EmptyTask):
```
praateekmahajan, line 43: why doesn't _EmptyTask need `i`?

praateekmahajan: nvm I see it, it's set to 0

nemo_curator/tasks/tasks.py
```diff
@@ -54,6 +67,45 @@ def add_stage_perf(self, perf_stats: StagePerfStats) -> None:
         """Add performance stats for a stage."""
         self._stage_perf.append(perf_stats)
 
+    def _set_task_id(self, parent_task_ids: list[str], current_task_id_suffix: str | int) -> None:
```
praateekmahajan: Why is parent_task_ids a list? Since only 1-> 1 is supported or N->1 will parent ever be a list? It should be just that parent's single task id right?

abhinavg4: Yeah it should just be a string

nemo_curator/backends/base.py
abhinavg4: Nope. I think the current code assumption is that only the source can have a deterministic ID. This makes a few assumptions easy; all other IDs will be numbers except one. Very easy to say which one is the source ID (We do store the source ID as well). Finally, I think there is no specific purpose that deterministic_ids are serving. Later down the road, we might implement a uniqueness check for the source ID, but doing it for 2 ids would be much more difficult.

nemo_curator/backends/base.py
abhinavg4: We can have None, we repalce None with NoneTask anyways in the next PR> not sure what is question is though

nemo_curator/stages/base.py
abhinavg4: The precommits are def running. Not sure why this was not caught. Can look at this later.

tests/backends/test_task_id_postprocess.py
abhinavg4: Sure. We can move these tests around.

nemo_curator/pipeline/pipeline.py
abhinavg4: They should be the same

tests/backends/test_integration.py
abhinavg4: Hmm, wondering if we know the exact IDs...

praateekmahajan (approved):

VibhuJawa (approved):
Lgtm

PR #2040, opened 2026-05-30
fix: include port 65535 in free port search

### LINKED ISSUE #2023: fix: get_free_port() off-by-one — port 65535 never checked despite appearing in error message

nightcityblade:
## Description

Closes #2023.

Updates `get_free_port()` to include port 65535 in the search range, matching the valid TCP port range and the existing error message. Adds a focused regression test for starting the search at 65535.

## Usage

```python
from nemo_curator.core.utils import get_free_port

port = get_free_port(65535)
```

## Checklist

## Testing

- `uv run --python 3.12 --group linting ruff check nemo_curator/core/utils.py tests/core/test_utils.py` — passed
- `uv run --python 3.12 --group test python -m pytest tests/core/test_utils.py` — blocked on macOS because `nemo_curator` raises `ValueError` on non-Linux systems during import
- Direct boundary check with `get_free_port(65535)` and a fake socket — passed

### DIFF
--- a/nemo_curator/core/utils.py
+++ b/nemo_curator/core/utils.py
@@ -100,7 +100,7 @@ def get_free_port(start_port: int, get_next_free_port: bool = True) -> int:
     If not, it will get the next free port starting from start_port if get_next_free_port is True.
     Else, it will raise an error if the free port is not equal to start_port.
     """
-    for port in range(start_port, 65535):
+    for port in range(start_port, 65536):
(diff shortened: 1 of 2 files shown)


sarahyurick:
Closing in favor of https://github.com/NVIDIA-NeMo/Curator/pull/2024 which was opened first.

PR #2044, opened 2026-06-02
Fix get_free_port checking port 65535

### LINKED ISSUE #2023: fix: get_free_port() off-by-one — port 65535 never checked despite appearing in error message

nightcityblade:
## Description
Fixes the inclusive upper bound in `get_free_port()` so port 65535 is actually checked before raising.

Closes #2023.

## Usage
```python
from nemo_curator.core.utils import get_free_port

port = get_free_port(65534)
```

## Checklist

Tests:
- `uv run ruff check nemo_curator/core/utils.py tests/core/test_utils.py`
- `uv run python - <<'PY' ...` custom focused `get_free_port(65534) == 65535` harness

Note: `uv run pytest tests/core/test_utils.py -q` could not run on this macOS host because NeMo-Curator raises on non-Linux systems during import.

### DIFF
--- a/nemo_curator/core/utils.py
+++ b/nemo_curator/core/utils.py
@@ -100,7 +100,7 @@ def get_free_port(start_port: int, get_next_free_port: bool = True) -> int:
     If not, it will get the next free port starting from start_port if get_next_free_port is True.
     Else, it will raise an error if the free port is not equal to start_port.
     """
-    for port in range(start_port, 65535):
+    for port in range(start_port, 65536):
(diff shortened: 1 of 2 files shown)


sarahyurick:
Closing in favor of https://github.com/NVIDIA-NeMo/Curator/pull/2024 which was opened first.

PR #2053, opened 2026-06-05
fix(dynamo): multimodal chat processor, hybrid KV cache compat, subprocess env passthrough

praateekmahajan:
## Summary

- **Multimodal frontend processor**: Dynamo's native-Rust chat processor serializes OpenAI multimodal \`content\` arrays instead of flattening them, corrupting prompts for models with pass-through chat templates (e.g. Nemotron-Parse's \`{{ message['content'] }}\`). \`DynamoBackend._frontend_router_kwargs\` is now a pure passthrough — callers must set \`dyn_chat_processor="vllm"\` explicitly in \`router_kwargs\` when needed (see Nemotron-Parse example below).

- **Hybrid KV cache manager compat** (prefill/decode + disagg paths): vLLM rejects \`--kv-events-config\` when \`--no-disable-hybrid-kv-cache-manager\` is also passed, even if \`enable_kv_cache_events=False\`. Both \`_launch_vllm_worker\` and \`_launch_disagg_role\` now skip \`--kv-events-config\` when the caller has explicitly opted in to the hybrid KV cache manager via \`disable_hybrid_kv_cache_manager=False\` in \`engine_kwargs\`.

- **Subprocess env passthrough**: \`backend_cfg.subprocess_env\` is now merged into the base env dict passed to frontend and worker processes, so user-supplied vars like \`DYN_TCP_REQUEST_TIMEOUT\` actually reach the subprocesses.

## Nemotron-Parse setup

All three fixes above are required to serve [Nemotron-Parse](https://huggingface.co/nvidia/NVIDIA-Nemotron-Parse-v1.2) via \`InferenceServer\`:

```python
from nemo_curator.core.serve import InferenceServer
from nemo_curator.core.serve.dynamo import DynamoServerConfig, DynamoRouterConfig
from nemo_curator.core.serve.dynamo.vllm import DynamoVLLMModelConfig

server = InferenceServer(
    models=[DynamoVLLMModelConfig(
        model_identifier="/path/to/NVIDIA-Nemotron-Parse-v1.2",
        engine_kwargs={
            "trust_remote_code": True,
            "dtype": "bfloat16",
            "limit_mm_per_prompt": {"image": 1},
            "enable_prefix_caching": False,   # encoder-decoder models crash with APC
            "disable_hybrid_kv_cache_manager": False,  # fix: skip --kv-events-config
        },
        dynamo_kwargs={"enable_multimodal": True},
    )],
    backend=DynamoServerConfig(
        request_plane="tcp",
        router=DynamoRouterConfig(router_kwargs={
            "dyn_chat_processor": "vllm",     # fix: native-Rust corrupts multimodal arrays
            "trust_remote_code": True,
            "chat_template_content_format": "string",
        }),
        subprocess_env={"DYN_TCP_REQUEST_TIMEOUT": "180"},  # fix: env now reaches subprocess
    ),
)
```

## Test plan

- [ ] `tests/core/serve/dynamo/test_backend.py` — passthrough (no auto-inject), explicit processor forwarded, text-only unchanged, TCP env passthrough
- [ ] `tests/core/serve/dynamo/test_vllm.py` — hybrid KV cache manager skip path for both worker and disagg roles
- [ ] `tests/core/serve/dynamo/test_infra.py`


praateekmahajan:
@claude review

nemo_curator/core/serve/dynamo/vllm.py
```diff
@@ -321,13 +320,17 @@ def _launch_vllm_worker(  # noqa: PLR0913
...
-        model_config,
-        pg=pg,
-        bundle_index=node_rank,
-        port_seed=20080 + replica_index + node_rank,
-        enabled=kv_events_enabled,
-    )
+    kv_events_config = None
+    # vLLM treats any non-None kv_events_config as incompatible with explicitly
+    # enabled hybrid KV cache manager, even when enable_kv_cache_events=False.
+    if kv_events_enabled or not explicit_hybrid_kv_cache_manager_enabled(model_config.engine_kwargs):
```
oyilmaz-nvidia: Do we need this same check in the _launch_disagg_role() as well?

oyilmaz-nvidia (commented):
Do we have a documentation or example that users can follow? If yes, those need updates. If no, we need that documentation and/or examples.

ayushdg (approved):

PR #2056, opened 2026-06-08
fix: auto-detect Ray fanout stages

### LINKED ISSUE #1613: Automatically detect when `IS_FANOUT_STAGE` should be set to True

nightcityblade:
## Description
Closes #1613.

Automatically marks Ray Data stages as fanout stages when their `process` return annotation is a concrete `list[...]`. This lets stages like `URLGenerationStage` rely on the base `ProcessingStage` default instead of manually setting `is_fanout_stage`.

Supersedes #2025 with the same branch after refreshing the DCO sign-offs.

## Usage
```python
class MyFanoutStage(ProcessingStage[InputTask, OutputTask]):
    def process(self, task: InputTask) -> list[OutputTask]:
        ...

assert MyFanoutStage().ray_stage_spec() == {"is_fanout_stage": True}
```

## Checklist

Tests:
- `python3 -m compileall nemo_curator/stages/base.py nemo_curator/stages/video/clipping/clip_extraction_stages.py tests/stages/video/clipping/test_clip_transcoding_stage.py`
- Earlier targeted ruff checks passed on the changed stage/test files; targeted pytest collection is blocked locally on macOS by NeMo-Curator's Linux-only runtime check.


sarahyurick:
Hi @nightcityblade can you resolve the merge conflicts here?

nightcityblade:
Resolved the merge conflicts by merging the latest `main` into `fix/issue-1613-auto-fanout` and keeping the fanout auto-detection changes aligned with the current base branch.

Also addressed the Greptile review feedback by catching `AttributeError` around `get_type_hints(cls.process)` and added coverage for that dotted missing-symbol annotation fallback.

Validation:
- `uv run python -m py_compile nemo_curator/stages/base.py tests/stages/common/test_base.py nemo_curator/stages/deduplication/semantic/pairwise_io.py nemo_curator/stages/file_partitioning.py nemo_curator/stages/video/clipping/clip_extraction_stages.py` ✅
- `uv run pytest tests/stages/common/test_base.py -q` could not run on this macOS host because NeMo-Curator now raises its Linux-only platform guard during import.

sarahyurick:
Hi @nightcityblade can you check if any of the changes in https://github.com/NVIDIA-NeMo/Curator/pull/2086 are relevant for this PR? Thanks in advance!

sarahyurick:
Closing in favor of https://github.com/NVIDIA-NeMo/Curator/pull/2157.

PR #2057, opened 2026-06-08
fix(audio): download FLEURS via hf_hub_download to avoid HTTP 429

mohammadaaftabv:
## Summary
Fixes the `audio_fleurs_raydata` / `audio_fleurs_xenna` nightly benchmark failures caused by Hugging Face returning `HTTP Error 429: Too Many Requests` while downloading the FLEURS dataset.

**Root cause:** The download is *not* run by many concurrent threads within a single pipeline run — the create-initial-manifest stage runs once on a single seed task and fetches its files sequentially. The 429 came from **repeated unauthenticated pulls of the same artifacts**: `audio_fleurs_xenna` and `audio_fleurs_raydata` run back-to-back (plus nightly reruns and a shared CI egress IP), each previously re-downloading FLEURS via raw `wget` into a separate per-run scratch dir, with no retry and no cache.

## Approach
FLEURS is itself a Hugging Face dataset, and the repo already standardizes on `huggingface_hub` for HF assets (e.g. the audio `band`/`sigmos` stages and `nemo_curator/utils/hf_download_utils.py`). So rather than hand-roll retry + caching, FLEURS now downloads through `hf_hub_download`, which retries transient HTTP errors (incl. 429) with backoff **and** caches by content hash natively.

## Changes
1. **FLEURS → `hf_hub_download`** (`nemo_curator/stages/audio/datasets/fleurs/create_initial_manifest.py`):
   - Replace `get_fleurs_url_list` + raw `wget`/`download_file` with `get_fleurs_filenames` + `hf_hub_download(repo_id="google/fleurs", repo_type="dataset", ...)`.
   - Add a `cache_dir` field; `process_transcript()` now takes an explicit `audio_root` (decoupled from the transcript file location, which now lives in the HF cache).
2. **Benchmark** (`benchmarking/`): `--cache-dir` configures the HF cache (default `$CURATOR_FLEURS_CACHE_DIR` or `/tmp/curator/fleurs_cache`) so sibling entries / reruns reuse the cached download; the archive extracts into a per-run scratch dir. Both FLEURS nightly entries point at the shared cache.

_Note: `file_utils.py` is intentionally left unchanged — `hf_hub_download` provides retry/caching for FLEURS. The ReadSpeech stage (Azure blob, not HF) still uses `download_file`; hardening that path can be a separate follow-up._

## Root cause (from CI log)
```
download_file:37 - Not found file .../audio_fleurs_raydata/scratch/armenian/fleurs/train.tsv
...
urllib.error.HTTPError: HTTP Error 429: Too Many Requests
```

## Test plan
- [x] `pytest tests/stages/audio/datasets/test_fleurs_create_initial_manifest.py` — FLEURS stage with `hf_hub_download` mocked.
- [x] `pytest tests/stages/audio/datasets/test_readspeech_create_initial_manifest.py` — no regression on the existing download path.
- [x] 18 passed; `ruff check` clean; benchmark script compiles and the nightly YAML parses.

### Local end-to-end nightly verification
Ran the **full nightly framework** (`benchmarking/run.py`, no Docker) for both FLEURS entries back-to-back on a single GPU, reproducing the exact CI sequence that triggered the 429:

| Entry | Result | `is_success` | Tasks | Throughput |
|---|---|---|---|---|
| `audio_fleurs_xenna` | exit 0 | `true` | 403 | 2.26 tasks/s |
| `audio_fleurs_raydata` | exit 0 | `true` | 403 | 2.22 tasks/s |

→ `Session ... completed with overall success: True`.

- **Zero** `HTTP Error 429` / `Too Many Requests`.
- **Cache reuse confirmed:** `audio_fleurs_xenna` downloaded `train.tar.gz` once into the shared `--cache-dir`; `audio_fleurs_raydata` reused the *same* HF snapshot/blob with no re-download — the exact behavior that eliminates the cross-run rate-limiting.

<sub>Local single-consumer-GPU caveat (not part of this PR): RNNT CUDA-graph decoding fails to capture on an RTX 3080 Ti, so the run was done with `greedy.use_cuda_graph_decoder=false`. This is environment-specific and is **not** changed here, since CUDA-graph decoding works on the A100-based CI runners.</sub>


mohammadaaftabv:
/ok to test https://github.com/NVIDIA-NeMo/Curator/commit/fe8d4d750cfb03c3d097ebee69bbe4acbfb301c2

benchmarking/scripts/audio_fleurs_benchmark.py
```diff
@@ -36,6 +37,13 @@
 from nemo_curator.stages.resources import Resources
 from nemo_curator.stages.text.io.writer import JsonlWriter
 
+# Stable Hugging Face cache directory for the FLEURS download. hf_hub_download
+# caches by content hash here, so repeated benchmark entries (e.g.
+# audio_fleurs_xenna then audio_fleurs_raydata) and nightly reruns reuse the
+# same files instead of re-downloading them from Hugging Face -- which is what
+# triggered unauthenticated rate-limiting (HTTP 429) on the dataset host.
+DEFAULT_FLEURS_CACHE_DIR = "/tmp/curator/fleurs_cache"  # noqa: S108
```
sarahyurick, line 45: Couldn't we just download the dataset once (by adding the download script here: https://github.com/NVIDIA-NeMo/Curator/tree/main/benchmarking/data_prep) and then just keep the dataset on the benchmarking machine and not re-download it every time we run the benchmarks? Or do you want to include dataset download as part of the overall benchmark here?

mohammadaaftabv: Good call — done in d1d40c77.

I added `benchmarking/data_prep/prepare_fleurs_data.py` to stage FLEURS once into `{datasets_path}/fleurs/<lang>` (transcript + extracted audio), and both nightly entries now read it via `--raw-data-dir={dataset:fleurs_hy_am,files}`, mirroring how ReadSpeech consumes its pre-staged data.

The stage now downloads from Hugging Face **only when the dataset isn't already on disk** (the first run, or ahead of time via the prep script) and reuses it on every subsequent run — so it keeps the dataset on the benchmarking machine and never re-fetches, which removes the 429 path entirely. I kept an auto-download-to-scratch fallback for standalone/local runs.

mohammadaaftabv:
/ok to test https://github.com/NVIDIA-NeMo/Curator/pull/2057/commits/d1d40c77e598416aa5885356730c59fbca1d0c03

benchmarking/nightly-benchmark.yaml
```diff
@@ -113,6 +113,15 @@ datasets:
...
       path: "{datasets_path}/read_speech"
+  # Persistent on-disk location for the FLEURS dataset (transcript <split>.tsv +
+  # extracted <split>/ audio). The benchmark downloads it here exactly once (on
+  # the first run, or ahead of time via
+  # benchmarking/data_prep/prepare_fleurs_data.py) and reuses it on every
+  # subsequent run, so it never re-fetches from Hugging Face (avoids HTTP 429).
+  - name: "fleurs_hy_am"
+    formats:
+    - type: "files"
+      path: "{datasets_path}/fleurs/hy_am"
```
rlratzel, line 127: We need to make sure this dataset is present in `{datasets_path}` on our benchmark machine(s). @mohammadaaftabv I'll sync with you offline to get those in place.

benchmarking/nightly-benchmark.yaml
```diff
@@ -667,6 +676,7 @@ entries:
     args: >-
       --benchmark-results-path={session_entry_dir}
       --scratch-output-path={session_entry_dir}/scratch
+      --raw-data-dir={dataset:fleurs_hy_am,files}
```
rlratzel, line 682: Should we also add the new `--no-auto-download` flag so we fail if the datasets are not found? I'm worried we may be silently downloading datasets if users specify the `--raw-data-dir` option incorrectly.

mohammadaaftabv: Hi Rick,

Re: `--no-auto-download` — agreed, done in 75198b9f. Both FLEURS nightly entries now pass `--no-auto-download`, so a missing/mis-specified `--raw-data-dir` fails loudly instead of silently downloading — matching what the ReadSpeech nightly entry already does.

Re: the dataset on the benchmark machine — yes, with `--no-auto-download` in place the data does need to be present beforehand; it won't auto-populate on first run. It's a one-time setup, though: I added `benchmarking/data_prep/prepare_fleurs_data.py` (mirrors `prepare_readspeech_data.py`) which creates `{datasets_path}/fleurs/hy_am` and downloads the transcript + extracted audio once. After that, every nightly run just reuses it.

The exact one-time command (run on the benchmark machine, with `{datasets_path}` expanded to its real value):

```
python benchmarking/data_prep/prepare_fleurs_data.py --output-path {datasets_path}/fleurs
```

It defaults to `--lang hy_am --split train` (matching the nightly entries); pass `--verify-only` to re-check an existing stage without downloading. Happy to sync offline on placement.

benchmarking/data_prep/prepare_fleurs_data.py
```diff
@@ -0,0 +1,187 @@
...
+        cache_dir=cache_dir,
+        auto_download=True,
+    )
+    # Extracts audio into <lang_dir>/<split>/ and returns the transcript path
+    # (which lives in the Hugging Face cache).
+    tsv_src, audio_root = stage.download_extract_files(str(lang_dir))
+
+    tsv_dst = lang_dir / f"{split}.tsv"
+    shutil.copyfile(tsv_src, tsv_dst)
+    logger.info(f"Copied transcript to {tsv_dst}")
```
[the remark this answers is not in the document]
mohammadaaftabv: Good catch — fixed in 90971c79. `download_extract_files` already stages the transcript at `<lang_dir>/<split>.tsv` and returns that path, so the extra copy was both redundant and the source of the `SameFileError`. Removed it (and the now-unused `import shutil`) and updated the stale comment. Added a regression test (`test_prepare_fleurs_stage_dataset_does_not_recopy`) that runs `stage_dataset` with mocked Hugging Face downloads so this can't regress.

rlratzel (approved):

benchmarking/nightly-benchmark.yaml
rlratzel: I chatted with @mohammadaaftabv offline and was able to get the dataset installed on the benchmark machine.

sarahyurick (approved):

PR #2058, opened 2026-06-08
Strip user-set task_id from tutorials & getting-started script

abhinavg4:
## What

PR #2036 made `Task.task_id` `init=False` (framework-owned — assigned by the executor adapter at each stage boundary). The `tutorials/` examples and the getting-started verify script were not swept and still pass `task_id=` to `Task` constructors, so they now crash:

```
TypeError: FileGroupTask.__init__() got an unexpected keyword argument 'task_id'
```

This was reported for the math `cc_index_lookup` pipeline; the same break exists across audio / text / synthetic / slurm / quickstart tutorials and the getting-started CPU verify script.

## Change

Remove the `task_id=` kwarg at every construction site (`FileGroupTask`, `DocumentBatch`, `AudioTask`, `SampleTask`) — the framework assigns the id. Where a loop index existed only to build the removed `task_id`, drop the index (`for _ ...` / `for batch in ...`). Read-only uses of `task_id` (logging, a hash seed, the audio checkpoint payload dict) are left unchanged.

13 files, 8 insertions / 29 deletions — no behavior change beyond ids now being framework-assigned.

Labeled `docs-only` (tutorials/examples) so it skips CI, per @praateekmahajan.

### DIFF
--- a/.claude/skills/getting-started/scripts/verify_pipeline_cpu.py
+++ b/.claude/skills/getting-started/scripts/verify_pipeline_cpu.py
@@ -38,7 +38,6 @@ class TaskCreationStage(ProcessingStage[_EmptyTask, SampleTask]):
         return [
             SampleTask(
                 data=pd.DataFrame({"text": ["Hello world", "Test sentence"]}),
-                task_id="1",
                 dataset_name="test",
             )
         ]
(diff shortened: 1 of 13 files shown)


praateekmahajan (approved):

abhinavg4:
/ok to test [e83798c](https://github.com/NVIDIA-NeMo/Curator/pull/2058/commits/e83798c4d3960a2966302135b452a34e7f6a0bff)

PR #2059, opened 2026-06-09
Add support for Slurm arrays

sarahyurick:
TODO:

- [x] Add retry support
- [x] Add `FailedTask` support
- [x] Add a tutorial
- [ ] Add `nemo-curator-slurm-cli` (not planned for this PR)
- [x] Address case when `SLURM_ARRAY_TASK_COUNT` > cluster limit
- [x] Add tests


nemo_curator/stages/text/deduplication/semantic.py
```diff
@@ -76,6 +76,10 @@ class TextSemanticDeduplicationWorkflow:
     embedding_vllm_init_kwargs: dict[str, Any] | None = None
     hf_token: str | None = None
     model_cache_dir: str | None = None
+    enable_array_partitioning: bool = False
```
sarahyurick: Not sure if it is useful here.

sarahyurick: cc @praateekmahajan I blindly added this here, but thinking more about it this probably does not make sense since the next workflows need to be run globally (in a single Slurm job)...

nemo_curator/backends/base.py
```diff
@@ -85,9 +145,20 @@ def process_batch(self, tasks: list[Task]) -> list[Task]:
...
+        # A returned ``None`` ("filter this slot") becomes a NoneTask so every
+        # output is a real Task that gets a task_id. Sentinels (NoneTask /
+        # FailedTask) carry no identity and are stripped again before this
+        # method returns.
+        results = [NoneTask() if r is None else r for r in results]
+
         # Guarantee every emitted task has a task_id (derived id, or uuid fallback).
         results = self._post_process_task_ids(tasks, results)
 
+        self._record_failed_tasks([r for r in results if isinstance(r, FailedTask)])
```
sarahyurick: Discussed with @abhinavg4 . For now the PR keeps track of `FailedTask` instances by looking for a user-set `FAILED_TASKS_DIR_ENV_VAR = "NEMO_CURATOR_FAILED_TASKS_DIR"` and writing a JSON file per failed task in the specified directory.

I did the environment variable and write approach because it seems more reliable than trying to handle a global Python variable, etc. And the reason it is an environment variable is so that `BaseStageAdapter` does not have to propagate an additional parameter for every single stage (which I think would involve having to update the executors as well?). Open to other suggestions.

nemo_curator/stages/text/deduplication/semantic.py
sarahyurick: Removing changes to this file since `TextSemanticDeduplicationWorkflow` needs to be run globally and cannot be split across multiple Slurm jobs.

praateekmahajan (commented):
Took a super quick look, here are some general thoughts
1. Instead of adding the same 3/4 fields to every "source" stage, can we have a base class and inherit that?
2. Alternatively (or maybe in addition), `pipeline.build` iirc now dynamically sets the first stage as is_source_stage=True, so can we just rely on those? If we do then inside backends/base.py we can say "if this is a source stage AND slurm is enabled then just use `task_id` as my key and decide which shard it belongs to"... this is something @abhinavg4 and I had discussed, this reduces the number of changes needed across curator code base, and also generalizes, since source_stage have task_id which is likely assigned using `get_determenistic_task_id` which is a hash(metadat['source_files'])

sarahyurick:
> Took a super quick look, here are some general thoughts
> 
> 1. Instead of adding the same 3/4 fields to every "source" stage, can we have a base class and inherit that?
> 2. Alternatively (or maybe in addition), `pipeline.build` iirc now dynamically sets the first stage as is_source_stage=True, so can we just rely on those? If we do then inside backends/base.py we can say "if this is a source stage AND slurm is enabled then just use `task_id` as my key and decide which shard it belongs to"... this is something @abhinavg4 and I had discussed, this reduces the number of changes needed across curator code base, and also generalizes, since source_stage have task_id which is likely assigned using `get_determenistic_task_id` which is a hash(metadat['source_files'])

For 1, sure.

For 2, we could but it makes this PR dependent on the resumability PR, which is what we were trying to avoid I thought... also, I guess it is not immediately obvious to me how it can work for source stages that are not a `FilePartitioningStage`. I get the general idea I guess but I am not convinced that it could always work, it sounds to me like how it would probably have to work is convert all unselected tasks to `NoneTask` maybe?

nemo_curator/backends/base.py
```diff
@@ -12,19 +12,144 @@
...
 
 if TYPE_CHECKING:
     from nemo_curator.stages.base import ProcessingStage
 
 
+FAILED_TASKS_DIR_ENV_VAR = "NEMO_CURATOR_FAILED_TASKS_DIR"
+SLURM_ARRAY_ENABLED_ENV_VAR = "NEMO_CURATOR_SLURM_ARRAY_ENABLED"
+SLURM_ARRAY_SHARD_INDEX_ENV_VAR = "NEMO_CURATOR_SLURM_ARRAY_SHARD_INDEX"
+SLURM_ARRAY_TOTAL_SHARDS_ENV_VAR = "NEMO_CURATOR_SLURM_ARRAY_TOTAL_SHARDS"
+SLURM_ARRAY_MINIMUM_SHARD_INDEX_ENV_VAR = "NEMO_CURATOR_SLURM_ARRAY_MINIMUM_SHARD_INDEX"
```
abhinavg4: Do we need all these variables, or can they be self-inferred? I think the initial design that @praateekmahajan had in mind was we just add --array=1-100 to the slurm submit command, and everything else works OOTB. Currently, it seems like the effort from the user side is a bit more than that?

nemo_curator/backends/base.py
```diff
@@ -12,19 +12,144 @@
...
+
+_TRUE_ENV_VALUES = {"1", "true", "yes", "on"}
+
+
+def _get_int_env_var(env_var: str, fallback_name: str | None = None, default: int | None = None) -> int:
```
abhinavg4: Did not review these since should be trivial but 2 comments:

1. Can you add a simple docstring
2. Move them to utils or something?

nemo_curator/backends/base.py
```diff
@@ -12,19 +12,144 @@
...
+
+    try:
+        return int(env_value)
+    except ValueError as e:
+        msg = f"Environment variable {env_var} must contain an integer, got {env_value!r}"
+        raise ValueError(msg) from e
+
+
+@dataclass
+class SlurmArrayConfig:
```
abhinavg4: Same as above. 
Can you add a simple docstring
Move them to utils or something?

nemo_curator/backends/base.py
```diff
@@ -12,19 +12,144 @@
...
+        flags |= os.O_DIRECTORY
+
+    dir_fd = os.open(path, flags)
+    try:
+        os.fsync(dir_fd)
+    finally:
+        os.close(dir_fd)
+
+
+def _write_failed_task_marker(marker_dir: Path, stage_name: str, task: FailedTask) -> None:
```
abhinavg4: Again same as above

nemo_curator/backends/base.py
abhinavg4: Ok I think a lot of the util functions are coming because of this feature, and there might be an easier way for this. Continuing in DMs.

abhinavg4 (changes requested):
I think I still have some major comments about the algorithm. Especially at scale, how will the writing and reading work. Will continue in the DMs.

nemo_curator/backends/base.py
sarahyurick: So the idea is we want to give the user full control if they need to override total shards (needed for reruns) or minimum shard index (needed to get around any Slurm array size limits). Really to enable Slurm array partitioning, the only thing explicitly needed is:
```
NEMO_CURATOR_SLURM_ARRAY_ENABLED=1
```
and it can automatically grab the environment variables without any issues. And then the user can override with `NEMO_CURATOR_SLURM_ARRAY_SHARD_INDEX`, etc. as desired (but not required).

nemo_curator/backends/base.py
sarahyurick: Sure.

nemo_curator/backends/base.py
sarahyurick: Sure, but the only place in the codebase that `SlurmArrayConfig` is used is in this file, so that is why it is here...

nemo_curator/backends/base.py
sarahyurick: Same reply as above, this is the only place in the whole codebase that the function is used. But I can move it if desired.

nemo_curator/backends/base.py
sarahyurick: The functions are really only:

1. Write info about failed tasks
2. Use task IDs to filter by Slurm array index

but I can move to util scripts if that makes it easier to read.

nemo_curator/backends/base.py
```diff
@@ -85,9 +210,22 @@ def process_batch(self, tasks: list[Task]) -> list[Task]:
...
+
         # Guarantee every emitted task has a task_id (derived id, or uuid fallback).
         results = self._post_process_task_ids(tasks, results)
 
+        self._record_failed_tasks([r for r in results if isinstance(r, FailedTask)])
+
+        # Sentinels never propagate to the next stage.
+        results = [r for r in results if not isinstance(r, (NoneTask, FailedTask))]
+
+        results = self._filter_slurm_array_source_tasks(results)
```
sarahyurick: I could probably be convinced that the order should be:
```suggestion
        results = self._filter_slurm_array_source_tasks(results)

        # Sentinels never propagate to the next stage.
        results = [r for r in results if not isinstance(r, (NoneTask, FailedTask))]
```
instead... but I don't have any strong opinion either way.

tutorials/slurm/array_pipeline.py
```diff
@@ -0,0 +1,188 @@
...
+    )
+    args = parser.parse_args()
+
+    slurm_array = SlurmArrayConfig.from_env()
+    if slurm_array is None:
+        parser.error("Slurm array filtering is not enabled. Set NEMO_CURATOR_SLURM_ARRAY_ENABLED=1.")
+
+    ray_client = SlurmRayClient() if args.slurm else RayClient()
+    is_driver_process = is_slurm_array_driver_process(args.slurm)
+    retry_manifest = build_slurm_array_retry_manifest(
```
sarahyurick: fyi @abhinavg4 this retry manifest logic is a bit different than we discussed (creating and deleting a manifest file at the beginning and end of the `pipeline.run()` function), but it is more robust and can capture more generic errors and log more generic statistics. Happy to discuss.

nemo_curator/backends/base.py
```diff
@@ -85,9 +92,29 @@ def process_batch(self, tasks: list[Task]) -> list[Task]:
             # Use the batch processing logic
             results = self.stage.process_batch(tasks)
 
+        # A returned ``None`` ("filter this slot") becomes a NoneTask so every
+        # output is a real Task that gets a task_id. Sentinels (NoneTask /
+        # FailedTask) carry no identity and are stripped again before this
+        # method returns.
+        results = [NoneTask() if r is None else r for r in results]
```
abhinavg4, line 110: We might have a merge conflict on lin 99, 113. But I think the intention is the same. So one of can resolve it whoever merges later.

nemo_curator/backends/slurm_array.py
```diff
@@ -0,0 +1,197 @@
...
+    """Source-task sharding settings for one Slurm array task."""
+
+    shard_index: int
+    total_shards: int
+    minimum_shard_index: int = 0
+
+    @classmethod
+    def from_env(cls) -> "SlurmArrayConfig | None":
+        """Build config from Curator env vars, falling back to Slurm env vars."""
+        enabled = os.environ.get(SLURM_ARRAY_ENABLED_ENV_VAR, "")
```
abhinavg4: Why are we gating on this. This would mean usre needs to enable this variable ?

nemo_curator/backends/slurm_array.py
```diff
@@ -0,0 +1,197 @@
...
+    return assigned_tasks
+
+
+def raise_for_failed_source_tasks_with_slurm_array(
+    stage_name: str,
+    failed_tasks: list[FailedTask],
+    slurm_array: SlurmArrayConfig | None,
+) -> None:
+    """Reject source-stage FailedTasks, which cannot be retried by shard."""
+    if failed_tasks and slurm_array is not None:
```
abhinavg4: Why do we need this? Can't this just mean we ignore this task? Like, if we have

[task_1, task_2, failed, task_4] and sluirm array of 2. Then can't we do task_1 and task_2 first and task_4 for another?

Also, why not treat None the same way then? I think for resumability, also we need think about this. For now, if a task is failed at the source stage, I assume it needs to rerun. 

I'm ok with having a contract that source stages cannot emit failed or none tasks. but it should be global then and not just for Slurm array.

nemo_curator/backends/failed_task_markers.py
```diff
@@ -0,0 +1,80 @@
...
+
+
+def record_failed_tasks(stage_name: str, failed_tasks: list[FailedTask]) -> None:
+    """Record FailedTask markers when ``NEMO_CURATOR_FAILED_TASKS_DIR`` is set."""
+    marker_dir = os.environ.get(FAILED_TASKS_DIR_ENV_VAR)
+    if not marker_dir or not failed_tasks:
+        return
+
+    marker_path = Path(marker_dir)
+    for task in failed_tasks:
```
abhinavg4: Still not sure why we need it for every task in failed task. This makes writing difficult and reading diffiult? 

For each Slurm array idx, would it not help to just have a single marker saying this idx failed and that's all? Also, can we move "if failed task" to the base.py code for readability? 

Just if condition and call this function

nemo_curator/backends/slurm_array.py
```diff
@@ -0,0 +1,197 @@
...
+    return int(digest[:16], 16) % slurm_array.total_shards + slurm_array.minimum_shard_index
+
+
+def filter_slurm_array_source_tasks(
+    tasks: list[Task],
+    slurm_array: SlurmArrayConfig | None,
+    stage_name: str,
+) -> list[Task]:
+    """Keep only source tasks assigned to the active Slurm array shard."""
+    if slurm_array is None:
```
abhinavg4, line 173: Can we move this into a single if in base.py? Currently, I feel the code is very less readable since if I'm not using Slurm arrays. I'm coming to a ton of functions and just returning.

tutorials/slurm/array_pipeline.py
abhinavg4: @sarahyurick Will this require the user to call is_slurm_array_driver_process everytime ?

nemo_curator/backends/slurm_array.py
sarahyurick: Yes, see here too: https://github.com/NVIDIA-NeMo/Curator/pull/2059#discussion_r3462120961

IMO we need a way to explicitly opt into Slurm array partitioning, and this variable is how users can do it.

nemo_curator/backends/slurm_array.py
sarahyurick: This if statement is just acting as an additional barrier to ensure we are only using this function if Slurm array partitioning has been enabled. I can add more comments to base.py if you think it can help?

I mainly want to avoid having base.py reason about Slurm array partitioning at all in favor of delegating the work to these functions here, so that base.py does not get clobbered up with a bunch of Slurm-specific things.

nemo_curator/backends/slurm_array.py
sarahyurick: I added this because I am worried about there being edge cases when a source stage produces failed tasks, specifically with Slurm array partitioning enabled.

Like with the FilePartitioningStage, the list of file names is used to generate the deterministic IDs, but if there is a failed task then won't that affect how a deterministic ID is generated and thus how it is bucketed into a Slurm array index? So would this mess up which Slurm index it is assigned to upon retry?

I haven't thought very thoroughly about the above example so maybe it is not a big concern. But since Slurm array partitioning is supposed to work with any source stage, I wanted to add this restriction until we could be more confident here.

Wdyt?

nemo_curator/backends/failed_task_markers.py
sarahyurick: Yeah I can see your point.

For me it seems easy enough to check if there is anything in the failed tasks directory to know whether or not to rerun.

I thought it could be useful to understand how many failed tasks there were (1 vs 1 million) to maybe inform whether or not you even want to rerun it. (Since the files are tiny I am not worried about disk space.)

And if each manifest saves which stage produced the failed task, then you know which stages you need to potentially update or edit before rerunning.

But I am open to other approaches too! Let me know what you think.

tutorials/slurm/array_pipeline.py
sarahyurick: Yes, I was thinking only the Slurm driver process should be dealing with the retry manifest. Let me know if this sounds reasonable.

nemo_curator/backends/slurm_array.py
sarahyurick: If you want me to make it more general and crash if the source stage emits any failed tasks (regardless of if we are doing Slurm array partitioning), then I can do that.

nemo_curator/backends/slurm_array.py
sarahyurick: Oops misread the comment I think. Sure I can move the if statement only to base.py, to make it more readable.

tutorials/slurm/array_pipeline.py
```diff
@@ -0,0 +1,185 @@
...
+        pipeline = build_pipeline(
+            args.input_dir,
+            args.input_file_type,
+            args.output_dir,
+            args.output_file_type,
+            args.files_per_partition,
+        )
+        logger.info(f"\n{pipeline.describe()}")
+
+        pipeline.run()
```
sarahyurick: Update to:
```suggestion
        pipeline.run(checkpoint_path=args.checkpoint_path)
```
after https://github.com/NVIDIA-NeMo/Curator/pull/2063 is merged.

nemo_curator/backends/base.py
```diff
@@ -98,13 +103,32 @@ def process_batch(self, tasks: list[Task]) -> list[Task]:
...
+        is_source_stage = getattr(self.stage, "is_source_stage", False)
+        failed_tasks = [r for r in results if isinstance(r, FailedTask)]
+        if failed_tasks and is_source_stage:
+            msg = (
+                f"Source stage {self.stage.name} emitted FailedTask, which is not supported."
+            )
+            raise ValueError(msg)
+
+        # Record failed tasks for later inspection or retry bookkeeping.
+        record_failed_tasks(failed_tasks)
```
abhinavg4: If failed_task is not None:  record_failed_tasks(), I do not think we need to pass the whoel array in the fucntion ?

nemo_curator/backends/base.py
```diff
@@ -98,13 +103,32 @@ def process_batch(self, tasks: list[Task]) -> list[Task]:
...
+
+        # Filter tasks based on the Slurm array configuration.
+        slurm_array = resolve_slurm_array_config(is_source_stage=is_source_stage)
+        if slurm_array is not None:
```
abhinavg4: if slurm_array is not None and is_source_stage: 
results = filter_slurm_array_source_tasks(results, slurm_array, self.stage.name)

?

nemo_curator/backends/base.py
```diff
@@ -144,22 +168,20 @@ def _post_process_task_ids(self, input_tasks: list[Task], output_tasks: list[Tas
...
             return out
 
         if len(output_tasks) == len(input_tasks):
-            # Positional 1:1. None is kept above so a filtered slot still lines
-            # up with its own parent; drop the None slots from the result.
+            # Positional 1:1. A NoneTask sentinel remains aligned with the
+            # parent whose output was filtered.
             out = []
             for parent, task in zip(input_tasks, output_tasks, strict=True):
-                if task is None:
```
abhinavg4, line 161: I think this needs to be rebased with main,

sarahyurick: It is already up to date with main I think. We already do:
```
results = [NoneTask() if r is None else r for r in results]

# Assign every emitted task a task_id (derived, or uuid fallback).
results = self._post_process_task_ids(tasks, results)
```
so we should never run into `if task is None` here?

abhinavg4 (approved):

Issue #1544, opened 2026-02-24
Consider using a minimum of 0 resources in our Ray data actor pool strategy.


ayushdg:
**Describe the bug**

When running in resource constrained environments (like where number of stages is greater than number of resources) ray data should technically be able to scale down some stages to 0 resources to process others. However our [actor pool strategy set's a minimum of 1 actor](https://github.com/NVIDIA-NeMo/Curator/blob/main/nemo_curator/backends/experimental/ray_data/utils.py#L53) which prevents it from scaling down to 0 leading to hangs.

**Steps/Code to reproduce bug**

2 Stage pipeline in a 1 CPU cluster. 1 actor based stage that requires 1 Cpu and another stage that also requires 1 CPU. 
Our calculation will assign 1 CPU to the actor and the other stage will never get any resources to run and will be permanently stuck in a hang.

**Expected behavior**

**Additional context**

nightcityblade:
Hi, I'd like to work on this. I'll submit a PR shortly.

PR #2077, opened 2026-06-16
Update Ray Data Backend to use `compute` strategy + Allow min/max/initial workers to be specified

praateekmahajan:
## Description
This PR updates the Ray Data backend to use Ray Data's `compute` API for actor stages instead of the deprecated `concurrency` argument.

Key changes:
- Use `ray.data.ActorPoolStrategy` for Ray Data actor stages.
- Treat `num_workers` as an exact fixed-size actor pool with `ActorPoolStrategy(size=n)`.
- Add `ray_stage_spec` overrides for autoscaling actor pools: `min_workers`, `max_workers`, and `initial_workers`.
- Leave Ray Data task stages on Ray Data's default task scheduling path, and warn when actor-pool worker sizing options are configured on task stages.
- Warn when `num_workers` takes precedence over actor-pool min/max/initial sizing.
- Simplify the Ray Data actor concurrency helper now that Ray Data owns actor-pool scaling and resource scheduling. (i.e. `max_size=None` is the same as the math we used to do before)
- Update the custom Ray actor pool backend so `num_workers` still requests an exact worker count when resources fit, but warns and caps to available resources when they do not.
- Tighten tests around Ray Data compute kwargs, actor-pool sizing, and `max_calls` behavior without relying on exact Ray worker PID reuse.

## Usage
```python
from nemo_curator.backends.utils import RayStageSpecKeys


class MyActorStage(...):
    def ray_stage_spec(self) -> dict:
        return {
            RayStageSpecKeys.IS_ACTOR_STAGE: True,
            RayStageSpecKeys.MIN_WORKERS: 2,
            RayStageSpecKeys.MAX_WORKERS: 8,
            RayStageSpecKeys.INITIAL_WORKERS: 4,
        }
```

If `num_workers()` returns a positive value for an actor stage, it takes precedence and creates a fixed-size actor pool:

```python
def num_workers(self) -> int:
    return 4
```
## Checklist

### DIFF
--- a/nemo_curator/backends/ray_data/adapter.py
+++ b/nemo_curator/backends/ray_data/adapter.py
@@ -25,6 +25,8 @@ from nemo_curator.stages.base import ProcessingStage
 
 from .utils import get_actor_compute_strategy_for_stage, get_configured_actor_pool_sizing_keys, is_actor_stage
 
+CURATOR_MANAGED_MAP_BATCHES_KWARGS = {"compute", "max_calls", "num_cpus", "num_gpus"}
+
 
 class RayDataStageAdapter(BaseStageAdapter):
     """Adapts ProcessingStage to Ray Data operations.
@@ -71,11 +73,12 @@ class RayDataStageAdapter(BaseStageAdapter):
         # For Task objects, we return them in the 'item' column
         return {"item": results}
 
-    def process_dataset(self, dataset: Dataset) -> Dataset:
+    def process_dataset(self, dataset: Dataset, ignore_head_node: bool = False) -> Dataset:
         """Process a Ray Data dataset through this stage.
 
         Args:
             dataset (Dataset): Ray Data dataset containing Task objects
+            ignore_head_node (bool): Whether to exclude head-node resources from the actor-pool max size.
 
         Returns:
             Dataset: Processed Ray Data dataset
@@ -85,7 +88,9 @@ class RayDataStageAdapter(BaseStageAdapter):
 
         if stage_is_actor:
             map_batches_fn = create_actor_from_stage(self.stage)
-            map_batches_kwargs = {"compute": get_actor_compute_strategy_for_stage(self.stage)}
+            map_batches_kwargs = {
+                "compute": get_actor_compute_strategy_for_stage(self.stage, ignore_head_node=ignore_head_node)
+            }
         else:
             map_batches_fn = create_task_from_stage(self.stage)
             map_batches_kwargs = {}
@@ -119,6 +124,14 @@ class RayDataStageAdapter(BaseStageAdapter):
         # caches an isolated virtualenv for this stage's workers.
         if self.stage.runtime_env:
             ray_remote_args["runtime_env"] = self.stage.runtime_env
+
+        colliding_ray_remote_args = sorted(CURATOR_MANAGED_MAP_BATCHES_KWARGS & ray_remote_args.keys())
+        if colliding_ray_remote_args:
+            msg = (
+                f"ray_remote_args for Ray Data stage {self.stage.name} must not override "
+                f"Curator-managed map_batches arguments {colliding_ray_remote_args}."
+            )
+            raise ValueError(msg)
 
         map_batches_kwargs.update(ray_remote_args)
 
--- a/nemo_curator/backends/ray_data/adapter.py
+++ b/nemo_curator/backends/ray_data/adapter.py
@@ -23,7 +23,7 @@ from nemo_curator.backends.base import BaseStageAdapter
 from nemo_curator.backends.utils import RayStageSpecKeys, get_worker_metadata_and_node_id
 from nemo_curator.stages.base import ProcessingStage
 
-from .utils import calculate_concurrency_for_actors_for_stage, is_actor_stage
+from .utils import get_actor_compute_strategy_for_stage, get_configured_actor_pool_sizing_keys, is_actor_stage
 
 
 class RayDataStageAdapter(BaseStageAdapter):
@@ -71,7 +71,7 @@ class RayDataStageAdapter(BaseStageAdapter):
         # For Task objects, we return them in the 'item' column
         return {"item": results}
 
-    def process_dataset(self, dataset: Dataset, ignore_head_node: bool = False) -> Dataset:
+    def process_dataset(self, dataset: Dataset) -> Dataset:
         """Process a Ray Data dataset through this stage.
 
         Args:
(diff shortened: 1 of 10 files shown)


praateekmahajan:
/claude review
@greptileai

nemo_curator/backends/ray_data/adapter.py
```diff
@@ -80,43 +80,54 @@ def process_dataset(self, dataset: Dataset, ignore_head_node: bool = False) -> D
...
+        if stage_is_actor:
             map_batches_fn = create_actor_from_stage(self.stage)
-            concurrency_kwargs = {
-                "concurrency": calculate_concurrency_for_actors_for_stage(
-                    self.stage, ignore_head_node=ignore_head_node
-                ),
-            }
+            map_batches_kwargs = {"compute": get_actor_compute_strategy_for_stage(self.stage)}
         else:
             map_batches_fn = create_task_from_stage(self.stage)
```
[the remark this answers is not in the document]
praateekmahajan: The original intent of `ignore_head_node` was for running Curator on Anyscale. Anyscale turns off head-node scheduling by default for multi-node clusters, and only schedules on the head if the cluster is single-node or the head explicitly defines Ray logical resources: https://docs.anyscale.com/configuration/compute/advanced#control-head-node-scheduling

Given that, I don’t think Curator should compute a static `max_size` when `max_workers` is unset. `ActorPoolStrategy(max_size=None)` lets Ray Data/Anyscale autoscale based on Ray’s resource scheduler. If someone needs strict no-head placement outside Anyscale defaults, then we can come back to this problem to solve that. But since we don't have a use case for it I'd rather keep it simple.

nemo_curator/backends/ray_data/utils.py
```diff
@@ -12,45 +12,54 @@
...
-    # Take the minimum of CPU and GPU constraints
-    max_actors = min(max_cpu_actors, max_gpu_actors)
-    return (1, int(max_actors))
+        actor_pool_sizing_keys = get_configured_actor_pool_sizing_keys(stage.ray_stage_spec())
+        if actor_pool_sizing_keys:
+            logger.warning(
+                f"Stage {stage.name} uses num_workers={num_workers}; ignoring ray_stage_spec "
+                f"actor-pool sizing keys {actor_pool_sizing_keys}."
+            )
+        return ActorPoolStrategy(size=num_workers)
```
ayushdg, line 51: Note for the future: We might want to combine this with either runtime_env or some method to set placement strategy for the num_workers (1 per node etc).

nemo_curator/backends/ray_data/utils.py
```diff
@@ -12,45 +12,54 @@
...
+
+    ray_stage_spec = stage.ray_stage_spec()
+    min_size = ray_stage_spec.get(RayStageSpecKeys.MIN_WORKERS, 1)
```
ayushdg, line 54: Should this be 0?

nemo_curator/backends/utils.py
```diff
@@ -130,6 +130,9 @@ class RayStageSpecKeys(str, Enum):
     IS_LSH_STAGE = "is_lsh_stage"
     IS_SHUFFLE_STAGE = "is_shuffle_stage"
     MAX_CALLS_PER_WORKER = "max_calls_per_worker"
+    MIN_WORKERS = "min_workers"
+    MAX_WORKERS = "max_workers"
+    INITIAL_WORKERS = "initial_workers"
```
ayushdg, line 135: Do you envision INITIAL_WORKERS as something being manually set by users/Agents that based on how they see scaling perform on workflows? Maybe we also have better defaults here than what Ray data might select or perhaps select it as a fraction of total works rather than a fixed value?

tests/backends/ray_data/test_max_calls_pid.py
```diff
@@ -169,15 +163,10 @@ def test_max_calls_not_fused_with_actor_stage(max_calls_per_worker: int | None):
         results = executor.execute(stages=[pid_stage, actor_stage], initial_tasks=tasks)
         all_logs = log_buffer.getvalue()
 
-    # Verify PID recycling
+    # Verify max_calls without assuming exact Ray worker reuse.
     pids = [r.data["worker_pid"].iloc[0] for r in results]
-    expected_unique_pids = math.ceil(num_tasks / max_calls_per_worker) if max_calls_per_worker else 1
```
ayushdg, line 174: what's the failure/issue here?

tests/backends/ray_data/test_max_calls_pid.py
```diff
@@ -266,3 +250,14 @@ def test_max_calls_not_fused_with_task_stage(max_calls_per_worker: int | None):
...
         )
+
+
+def _assert_max_calls_respected(pids: list[int], max_calls_per_worker: int | None) -> None:
+    assert pids
+    if max_calls_per_worker is None:
+        return
+
+    pid_counts = Counter(pids)
+    assert max(pid_counts.values()) <= max_calls_per_worker, (
```
ayushdg, line 261: does `==` fail here?

ayushdg (commented):
Some open non blocking questions about behavior.

Also open question: I read through the greptile thread but are we removing `ignore_head_node` support here?

nemo_curator/backends/ray_data/utils.py
praateekmahajan: btw https://docs.ray.io/en/latest/data/api/doc/ray.data.ActorPoolStrategy.html doesn't seem to have anything about placement strategy

nemo_curator/backends/ray_data/utils.py
praateekmahajan: As per the code https://github.com/ray-project/ray/blob/822c8f57f7bd131745d4aa25f6344ec1af55ca87/python/ray/data/_internal/compute.py#L147-L148 minimum is 1.

However if all tasks for this stage are done then autoscale will force it to zero

nemo_curator/backends/utils.py
praateekmahajan: Yes the reason I wanted to add INITIAL_WORKERS was for that we could choose polciies such as Xenna like policy where we divide N resources over S stages evenly and start each stage with those many workers.

tests/backends/ray_data/test_max_calls_pid.py
praateekmahajan: Yes, `==` might be too strong here. IIUC `max_calls` means a worker should process *at most* N batches before being recycled; it does not require Ray to pack exactly N batches onto every worker PID.

nemo_curator/backends/ray_data/utils.py
ayushdg: Thanks for confirming. We'll leave this for later to see if something like 1 actor per node is doable or not: https://github.com/ray-project/ray/issues/19856

nemo_curator/backends/ray_data/utils.py
ayushdg: I see. Thanks for confirming. xref: #1544

nemo_curator/backends/utils.py
ayushdg: Okay so to confirm, set manually for now, in the future might be set by Curator based on some strategy like N/S?

ayushdg (approved):

tests/backends/ray_data/test_max_calls_pid.py
praateekmahajan: There isn’t a new issue here. The only reason this assertion changed is that exact unique PID count is not a Ray guarantee. Similar to other thread.

nemo_curator/backends/ray_data/utils.py
```diff
@@ -12,45 +12,54 @@
...
+    max_size = ray_stage_spec.get(RayStageSpecKeys.MAX_WORKERS)
```
weijiac0619, line 55: compared to the deleted code : 
```
if stage.resources.cpus > 0:
        max_cpu_actors = available_cpus // stage.resources.cpus 
```

does that mean we allow ray data to oversubscribe if needed?

nemo_curator/backends/ray_data/utils.py
```diff
@@ -12,45 +12,54 @@
...
+from loguru import logger
+from ray.data import ActorPoolStrategy
+
+from nemo_curator.backends.utils import RayStageSpecKeys
 from nemo_curator.stages.base import ProcessingStage
 
+ACTOR_POOL_SIZING_KEYS = (
+    RayStageSpecKeys.MIN_WORKERS,
+    RayStageSpecKeys.MAX_WORKERS,
+    RayStageSpecKeys.INITIAL_WORKERS,
```
weijiac0619, line 26: to confirm my understanding is correct,  will ray data start at  MIN_WORKERS worker and increase incrementally if initial workers here is not set?

nemo_curator/backends/ray_data/adapter.py
```diff
@@ -71,7 +73,7 @@ def _process_batch_internal(self, batch: dict[str, Any]) -> dict[str, Any]:
         # For Task objects, we return them in the 'item' column
         return {"item": results}
 
-    def process_dataset(self, dataset: Dataset, ignore_head_node: bool = False) -> Dataset:
+    def process_dataset(self, dataset: Dataset) -> Dataset:
```
weijiac0619, line 76: question: the ignore_head_node is removed here. does that mean ray data may allocate head node resources to the actors?

nemo_curator/backends/ray_data/utils.py
praateekmahajan: Ray won't over subscribe since the autoscaling algorithm will only scale UPTO the number of cluster resources available. In our old code we used to do `max_size=N` but Ray has since introduced `ActorPoolStrategy(max_size=None)` which should solve the problem

nemo_curator/backends/ray_data/adapter.py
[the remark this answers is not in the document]
praateekmahajan: If we want Curator to be run on Anyscale we want folks to use `ignore_head_node=True`   inside `pieline.run(...)`. Outside of the ability to run on Anyscale platform we don't have a use case of `ignore_head_node`

1. In Ray Data we used to manually compute max_num_workers for a stage. Therefore `ignore_head_node` was necessary. It is also used to decide on which workers do we call `setup_on_node(...)`. It is still used there
2. For Xenna, it is ignored since Xenna has no ability to ignore a node.
3. For Ray Actor Pool we also do `calculate_optimal_actors_for_stage` which does the math to decide number of actors, so we still need `ignore_head_node`.

nemo_curator/backends/ray_data/utils.py
praateekmahajan: If MIN_WORKERS is specified and INITIAL_WORKERS=None then, then yes, that's always been the case.

nemo_curator/backends/ray_data/adapter.py
praateekmahajan: Context here https://github.com/NVIDIA-NeMo/Curator/pull/2077#discussion_r3423119697

weijiac0619 (approved):
nice work! approved

PR #2078, opened 2026-06-16
Add option to drop deduplication id field

### LINKED ISSUE #1580: Allow the textRemovalWorkflow to skip writing the id columns/_curator_dedup_id column on removal
When running a series of deduplication steps the removal workflow from each stage also writes out the `_curator_dedup_id` column that ends up messing up removal the next time it's called.

It would be good to ensure that we don't write out this column during removal in the general case or have an option to exclude it. 

The current workaround is to manually specify the fields to write out in write_kwargs to ensure that `_curator_dedup_id` is explicitly excluded. 

**Additional context**

nightcityblade:
## Description
Closes #1580.

Adds an optional `drop_id_field` flag to the text duplicate removal stage/workflow so generated deduplication IDs can be removed from final outputs. The semantic deduplication workflow enables it by default when using generated IDs and no explicit `output_fields` are provided.

## Usage
```python
workflow = TextDuplicatesRemovalWorkflow(
    input_path="input",
    ids_to_remove_path="duplicates",
    output_path="output",
    drop_id_field=True,
)
```

## Checklist

## Testing
- `uv run ruff check nemo_curator/stages/text/deduplication/removal.py nemo_curator/stages/text/deduplication/removal_workflow.py nemo_curator/stages/text/deduplication/semantic.py tests/stages/text/deduplication/test_removal_workflow.py`
- `uv run pytest tests/stages/text/deduplication/test_removal_workflow.py -q` (blocked on macOS: `ValueError: NeMo-Curator currently only supports Linux systems, while the current machine has a darwin system.`)


nemo_curator/stages/text/deduplication/removal_workflow.py
```diff
@@ -61,6 +61,7 @@ class TextDuplicatesRemovalWorkflow(WorkflowBase):
     output_kwargs: dict[str, Any] | None = None
     output_fields: list[str] | None = None
     output_mode: Literal["ignore", "overwrite", "append", "error"] | None = None
+    drop_id_field: bool = False
```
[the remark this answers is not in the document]
sarahyurick: +1 I think this is decent feedback. WDYT @nightcityblade ?

tests/stages/text/deduplication/test_removal_workflow.py
```diff
@@ -281,6 +281,27 @@ def test_initial_tasks_partitioning(self, test_config: "TestTextDuplicateRemoval
...
+
+    stage = TextDuplicatesRemovalStage(
+        ids_to_remove_path=str(ids_to_remove_path),
+        id_field=CURATOR_DEDUP_ID_STR,
+        drop_id_field=True,
+    )
+
+    result = stage.process(task).to_pandas()
+
+    assert result.to_dict(orient="list") == {"text": ["keep"]}
```
sarahyurick, line 301: I misread this test at first and was confused why we were checking that this text was kept. Can you add another assertion that explicitly checks that `CURATOR_DEDUP_ID_STR` is not in the result?

nemo_curator/stages/text/deduplication/removal_workflow.py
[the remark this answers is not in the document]
nightcityblade: Good call — I added an early `__post_init__` validation that raises a clear `ValueError` when `drop_id_field=True` conflicts with `output_fields` containing the id field, plus a focused unit test for that configuration.

Pushed: nightcityblade/Curator@2dbe71d
Validation: `uv run ruff check nemo_curator/stages/text/deduplication/removal_workflow.py tests/stages/text/deduplication/test_removal_workflow.py` and `python3 -m py_compile ...` passed locally. Targeted pytest collection is blocked on this macOS host by Curator’s import-time Linux-only guard.

tests/stages/text/deduplication/test_removal_workflow.py
nightcityblade: Added — the test now explicitly asserts that `CURATOR_DEDUP_ID_STR` is absent from the result columns, in addition to checking the remaining row contents.

Pushed in nightcityblade/Curator@2dbe71d.

nemo_curator/stages/text/deduplication/semantic.py
```diff
@@ -374,6 +369,7 @@ def _run_duplicate_removal(self, executor: BaseExecutor) -> WorkflowRunResult |
             output_kwargs=self.write_kwargs,
             output_fields=self.output_fields,
             output_mode="ignore",
+            drop_id_field=self.use_id_generator and self.output_fields is None,
```
sarahyurick, line 377: I guess I am on the fence about whether to silently drop the IDs generated by `use_id_generator` in this case (since `output_fields` default is none). I think it is fine but the information in the tutorial https://github.com/NVIDIA-NeMo/Curator/blob/main/tutorials/text/deduplication/semantic/semantic_e2e.ipynb might need to be updated. Can you check?

Monkiia: Checked and updated the tutorial in `3ff725a`.

The semantic dedup notebook previously said `_curator_dedup_id` would appear in the final deduplicated output when `use_id_generator=True`. That is now outdated with this PR's default `drop_id_field` behavior, so I revised the note to explain that the generated ID is dropped by default and can be preserved only by including it explicitly in `output_fields`.

Validation: the notebook still parses as JSON after the edit.

sarahyurick (approved):
Thanks @nightcityblade !

tests/stages/text/deduplication/test_removal_workflow.py
```diff
@@ -281,7 +281,39 @@ def test_initial_tasks_partitioning(self, test_config: "TestTextDuplicateRemoval
         assert workflow_output.get_metadata("num_duplicates_removed") == expected_removed
 
 
+
+def test_removal_stage_can_drop_id_field(tmp_path: Path):
+    ids_to_remove_path = tmp_path / "ids_to_remove.parquet"
+    pd.DataFrame({"id": [1]}).to_parquet(ids_to_remove_path, index=False)
+    task = DocumentBatch(
+        task_id="task",
```
sarahyurick:
```suggestion
```
see https://github.com/NVIDIA-NeMo/Curator/pull/2036.

ayushdg:
/ok to test [dff6e7c](https://github.com/NVIDIA-NeMo/Curator/pull/2078/commits/dff6e7c5f947c1de84c499d27d5857614fef34a2)

PR #2086, opened 2026-06-16
feat: support stage config overrides via with_

praateekmahajan:
1. Adds `with_` overrides for `ray_stage_spec`, `xenna_stage_spec`, and `num_workers`. Stage spec overrides are shallow-merged, with user-provided keys winning; a typed unset sentinel preserves the difference between omitted and explicit `None`.
2. Applies `num_workers` to Ray Data task stages via `TaskPoolStrategy`; actor stages continue to use `ActorPoolStrategy`.
3. Lets Xenna use `stage.num_workers()` for cluster-wide worker counts, rejects `xenna_stage_spec["num_workers"]`, and raises if a stage sets both `num_workers()` and Xenna `num_workers_per_node`.
4. Reserves `num_workers` as the `ProcessingStage` backend worker hook and renames the ASR dataloader setting to `dataloader_num_workers` so stage-local worker counts do not shadow `num_workers()`.
5. Moves single-input/fanout defaults to `num_workers() == 1` for file partitioning, URL generation, PDF partitioning, cluster-wise pairwise partitioning, audio manifest reader/writer, audio segment extraction, and ALM long-form manifest reading. PDF partitioning now also advertises itself as a Ray fanout stage.
6. No `deduplication/removal` stage behavior was removed; the deduplication change is limited to semantic pairwise partitioning using generic `num_workers()` instead of Xenna-only `num_workers_per_node`.
7. Keeps PyAnnote existing `xenna_num_workers` constructor option as a compatibility alias, but routes it through `num_workers()` instead of `xenna_stage_spec`.
8. Adds coverage for `with_` backend overrides, reserved `num_workers` enforcement, Xenna worker fallback/rejection, Ray Data task worker sizing, and updated singleton-stage defaults.


praateekmahajan:
@claude review

VibhuJawa (approved):
LGTM

weijiac0619:
just out of curiosity, what happened before when num_workers() == 1 didnt apply to ray data for single-input/fanout stages?

nemo_curator/stages/interleaved/pdf/nemotron_parse/partitioning.py
```diff
@@ -80,8 +80,11 @@ def inputs(self) -> tuple[list[str], list[str]]:
     def outputs(self) -> tuple[list[str], list[str]]:
         return [], []
 
-    def xenna_stage_spec(self) -> dict[str, Any]:
-        return {"num_workers_per_node": 1}
+    def ray_stage_spec(self) -> dict[str, object]:
+        return {RayStageSpecKeys.IS_FANOUT_STAGE: True}
```
weijiac0619, line 84: qq: do we set IS_FANOUT_STAGE manually by ourselves now in the pipeline? is there any logic we should follow to set IS_FANOUT_STAGE? did we missed setting it to true for pdf partitioning before this PR?

praateekmahajan: Yes we missed it before this PR. I know @sarahyurick is working with an external contributor to make it dynamic but i haven't taken a close look at it

praateekmahajan:
@weijiac0619  ray-data anyway starts leniently from 1 "concurrent" worker so we never had the problem in case of single-input stages for ray data, only for xenna we had the issue

nemo_curator/stages/interleaved/pdf/nemotron_parse/partitioning.py
weijiac0619: does setting IS_FANOUT_STAGE make that stage much faster? is there any benchmark result to share if possible?

weijiac0619 (approved):

PR #2102, opened 2026-06-22
docs(fern): add local library autodocs without Fern auth

lbliii:
## Description
Adds a no-auth local flow for generating the Python API reference. `npm run generate:library:local` (or `make docs-generate-library-local`) runs `fern/scripts/generate-library-local.sh`, which temporarily injects a path-based `nemo-curator-local` library entry into `docs.yml`, runs Fern's `docs md generate --local` (Docker, no Fern login), and restores `docs.yml` on exit.

The entry is injected rather than committed because the Fern CLI has two conflicting rules (verified in `fern-api/fern` source, present in 5.50.4 and latest 5.65.5):
- `fern docs md generate --local` **requires** a `path` input (git inputs are remote-only and need auth), but
- `fern docs dev` / `fern generate --docs` **reject** any docs.yml containing a `path` input library ("'path' input which is not yet supported. Please use 'git' input.") and render a blank page. `fern check` does not catch this.

CI and publish/preview workflows are pinned to `--library nemo-curator` so the git-based remote generation path is unchanged. Also bumps the Fern CLI pin to 5.50.4 and documents both flows in `fern/README.md`, `fern/AUTODOCS_GUIDE.md`, and the docs agent skill.

## Usage
```bash
cd fern
npm run generate:library:local   # no Fern auth; requires Docker
npm run generate:library         # full reference via git; needs fern login or FERN_TOKEN
```

## Checklist

Made with [Cursor](https://cursor.com)


lbliii:
Pushed 9e68e26: the committed `nemo-curator-local` (`input.path`) entry in `docs.yml` turned out to break `fern docs dev` — the docs config parser throws `Library '...' uses 'path' input which is not yet supported. Please use 'git' input.` and renders a blank page (reproduced by @chtruong814 in Voice-Agent; the throw exists in every fern-api release through latest 5.65.5, and `fern check` doesn't catch it).

Since `docs md generate --local` *requires* a path input while `docs dev` *rejects* one, the entry now lives only in `fern/scripts/generate-library-local.sh`, which injects it into `docs.yml`, generates, and restores the file on exit. `npm run generate:library:local` / `make docs-generate-library-local` are unchanged from the caller's perspective, and `fern docs dev` keeps working.

chtruong814 (approved):

sarahyurick (approved):

PR #2109, opened 2026-06-24
[video] Remove model_does_preprocess

suiyoubi:
## Overview

  This PR simplifies video caption preprocessing by removing the `model_does_preprocess` switch and always passing raw video frames through the vLLM/HF multimodal processor. This restores Qwen2.5 behavior on
  latest `main`, keeps Qwen3 working, and removes a fragile caller-controlled preprocessing mode.

  ## Key Changes

  - Removed `model_does_preprocess` / `--captioning-model-does-preprocess` from captioning stages, tutorial CLI, docs, and tests.
  - Updated Qwen VL setup so Qwen2.5 and Qwen3 always use vLLM/HF preprocessing with explicit resize/rescale/normalize kwargs.
  - Renamed `_convert_to_numpy()` to `_format_raw_video_frames()` and made it format raw decoded frames as contiguous `uint8` arrays for vLLM.
  - Added guards against accidentally passing normalized/invalid frame tensors into captioning.
  - Fixed a frame-windowing bug where windows dropped frames and advanced the frame cursor incorrectly across multiple windows.
  - Added unit coverage for inclusive window end handling and per-window frame preservation.

  ## Why

  In `r1.2.0`, Qwen2.5 effectively received correctly preprocessed inputs. On latest `main`,  the Qwen3 support PR introduced a default path where Qwen2.5 could receive raw/unprocessed frames unless users set
  `--captioning-model-does-preprocess true`.

  Instead of keeping that behavioral toggle, this PR makes the code path explicit and consistent: Curator prepares raw frames, and the model processor owns model-specific resize/rescale/normalize logic.

  ## Validation

  - `ruff check`: passed
  - Focused pytest: `95 passed, 2 deselected`
  - Docker pipeline with `qwen2.5`: passed and generated caption output
  - Docker pipeline with `qwen3`: passed after the frame-windowing fix and generated caption output


weijiac0619 (approved):
LGTM! Thank you for fixing that

suiyoubi:
/ok to test [25e39a0](https://github.com/NVIDIA-NeMo/Curator/pull/2109/commits/25e39a0022684e100ecac1fe3d57ad7bc77bf830)

suiyoubi:
/ok to test [75425e5](https://github.com/NVIDIA-NeMo/Curator/pull/2109/commits/75425e51f2bc5297c621bcf6db0daee7a18d3939)

PR #2106, opened 2026-06-24
Add Lance reader and writer stages

VibhuJawa:
## Description
Adds a first-pass Lance integration for Curator text IO:

- `LanceReader` partitions Lance datasets by fragment and returns Arrow-backed `DocumentBatch`es.
- `LanceWriter` writes Curator batches to Lance and records checkpointed commit payloads.
- `LanceAnnotationWriter` appends/validates annotation columns with Lance `add_columns(...)`, writes fragment-local row updates, and records checkpointed commit payloads.
- Writer-side driver commit helpers live in `writer/lance.py` and publish checkpointed writes with `commit_lance_checkpoint(...)` and annotation-column updates with `commit_lance_annotation_checkpoint(...)`.
- Shared Lance metadata/checkpoint utilities live under `nemo_curator/utils/lance.py` and reuse Curator filesystem/hash helpers.
- Includes focused Lance functionality coverage for read filters/projection, blob v2 round trips, write checkpoint commits, annotation prepare/update, duplicate row-address rejection, and split-fragment update rejection.

## Usage
```python
from nemo_curator.stages.text.io.reader import LanceReader
from nemo_curator.stages.text.io.writer import LanceWriter, commit_lance_checkpoint

reader = LanceReader(
    "s3://bucket/dataset.lance",
    fields=["url", "text", "content_zlib"],
    read_kwargs={"filter": "snapshot_id = 'CC-MAIN-2025-26'"},
)

writer = LanceWriter(
    path="s3://bucket/output.lance",
    commit_path="s3://bucket/checkpoints/output",
    mode="overwrite",
)

# Run the Curator pipeline with the writer, then commit on the driver.
version = commit_lance_checkpoint(
    "s3://bucket/output.lance",
    "s3://bucket/checkpoints/output",
)
```

Append annotation columns back to an existing Lance dataset:

```python
import pyarrow as pa

from nemo_curator.stages.text.io.writer import LanceAnnotationWriter, commit_lance_annotation_checkpoint

annotation_writer = LanceAnnotationWriter(
    path="s3://bucket/dataset.lance",
    commit_path="s3://bucket/checkpoints/annotations",
    schema=pa.schema([pa.field("word_count", pa.int64())]),
    create_columns=True,
)
annotation_writer.prepare()

# Run the Curator pipeline with `annotation_writer`, then commit on the driver.
version = commit_lance_annotation_checkpoint(
    "s3://bucket/dataset.lance",
    "s3://bucket/checkpoints/annotations",
)
```

[...]

PR #2111, opened 2026-06-24
Add Lance reader stage

VibhuJawa:
Split from #2106. This is PR 1 of 3 in the Lance IO stack.

What changed:
- Generalizes text IO readers with a task-oriented BaseReader plus BaseFileReader for existing file-path readers.
- Adds Lance fragment partitioning and LanceReaderStage on top of BaseReader.
- Emits Lance row/fragment metadata for downstream annotation writers and preserves Lance schema metadata for later writes.
- Adds a reader-only lance extra using pylance.

Stack:
1. LanceReader (this PR)
2. LanceWriter on top of this branch
3. LanceAnnotationWriter on top of LanceWriter

Validation:
- uv run --extra lance --group test pytest -q tests/stages/text/io/reader/test_lance.py tests/stages/text/io/reader/test_jsonl.py tests/stages/text/io/reader/test_parquet.py -k "not IdGenerator and not id_generation"
- uv run ruff check nemo_curator/stages/text/io/reader/base.py nemo_curator/stages/text/io/reader/jsonl.py nemo_curator/stages/text/io/reader/parquet.py nemo_curator/stages/text/io/reader/lance.py nemo_curator/utils/lance.py tests/stages/text/io/reader/test_lance.py

Note: the full JSONL ID-generator reader test path was not used for this focused validation because the local untracked outputs/ benchmark artifacts make Ray runtime-env packaging exceed 512 MiB.


nemo_curator/stages/text/io/reader/base.py
```diff
@@ -72,23 +86,23 @@ def setup(self, _: WorkerMetadata | None = None) -> None:
                 )
                 raise RuntimeError(msg) from None
 
-    def process(self, task: FileGroupTask) -> DocumentBatch:
+    def process(self, task: ReaderTask) -> DocumentBatch | None:
```
VibhuJawa: Can this be None ?

nemo_curator/stages/text/io/reader/base.py
```diff
@@ -72,23 +86,23 @@ def setup(self, _: WorkerMetadata | None = None) -> None:
...
             effective_read_kwargs.update(self.read_kwargs)
 
-        # Read the files
-        result = self.read_data(task.data, effective_read_kwargs, self.fields)
+        # Read the task payload
+        read_result = self.read_task(task, effective_read_kwargs, self.fields)
+        if read_result is None:
+            return self.handle_empty_result(task)
+
+        output = read_result if isinstance(read_result, ReaderOutput) else ReaderOutput(read_result)
```
VibhuJawa: When will it not be an instance of ReaderOutput ?

nemo_curator/stages/text/io/reader/base.py
```diff
@@ -72,23 +86,23 @@ def setup(self, _: WorkerMetadata | None = None) -> None:
...
         effective_read_kwargs: dict[str, Any] = {}
         if self.read_kwargs:
             effective_read_kwargs.update(self.read_kwargs)
 
-        # Read the files
-        result = self.read_data(task.data, effective_read_kwargs, self.fields)
+        # Read the task payload
+        read_result = self.read_task(task, effective_read_kwargs, self.fields)
+        if read_result is None:
+            return self.handle_empty_result(task)
```
VibhuJawa: This should just raise Error, than validating .

nemo_curator/stages/text/io/reader/lance.py
```diff
@@ -0,0 +1,256 @@
...
+        table = table.set_column(table.schema.get_field_index(column), column, lance.blob_array(payloads))
+    return table
+
+
+def _add_lance_metadata(table: pa.Table) -> pa.Table:
+    if "_rowaddr" not in table.column_names:
+        msg = "Lance scanner did not return _rowaddr; include_lance_metadata requires row addresses"
+        raise ValueError(msg)
+
+    rowaddrs = table["_rowaddr"].combine_chunks().cast(pa.uint64())
```
VibhuJawa: What does this do and why is this required ?

nemo_curator/stages/text/io/reader/lance.py
```diff
@@ -0,0 +1,256 @@
...
+    fragids = pa.array([int(value) >> 32 for value in rowaddrs.to_pylist()], type=pa.uint64())
```
VibhuJawa: Should we gather frag ids from upstream directly ?

nemo_curator/stages/text/io/reader/lance.py
```diff
@@ -0,0 +1,256 @@
...
+
+    @property
+    def num_items(self) -> int:
+        return len(self.data)
+
+    def validate(self) -> bool:
+        if not isinstance(self.data, list):
+            msg = f"Invalid data type in task {self.task_id}"
+            raise TypeError(msg)
+        return bool(self.data)
```
VibhuJawa: This feels very hacky validation, why do we have this

nemo_curator/stages/text/io/reader/lance.py
```diff
@@ -0,0 +1,256 @@
...
+
+    def get_deterministic_id(self) -> str:
+        lance_metadata = self._metadata.get("lance") or {}
+        parts = [
+            str(lance_metadata.get("path", self.dataset_name)),
+            str(lance_metadata.get("version", "")),
+            *(str(fragment_id) for fragment_id in self.data),
+        ]
+        return get_deterministic_hash(parts)
```
VibhuJawa: Explain logic on why is it needed

VibhuJawa (commented):
Please fix these things

nemo_curator/stages/text/io/reader/base.py
VibhuJawa: No, this no longer returns `None`. `BaseReader.process()` now returns a `DocumentBatch` or raises during validation. Empty Lance scanner results are handled by `allow_empty=True`, but they still produce an empty `DocumentBatch` rather than `None`.

nemo_curator/stages/text/io/reader/base.py
VibhuJawa: It is always a `ReaderOutput` now. I removed the raw-result fallback path and kept `read_task()` typed to return `ReaderOutput`, so the base class no longer needs runtime wrapping/validation here.

nemo_curator/stages/text/io/reader/base.py
VibhuJawa: Done. The default behavior now raises again via `_validate_result()`, matching the previous BaseReader behavior. Lance explicitly opts into `allow_empty=True` because scanner filters can legitimately return a zero-row table for a fragment partition.

nemo_curator/stages/text/io/reader/lance.py
VibhuJawa: This is needed when `include_lance_metadata=True` or when reading `lance.blob.v2` columns. Lance exposes row identity through `_rowaddr`; we preserve it as `__lance_rowaddr` and derive `__lance_fragid` from it so annotation writes can route each row back to the exact source fragment. Blob v2 restoration also needs row addresses to read payloads in row order.

nemo_curator/stages/text/io/reader/lance.py
VibhuJawa: We should not use the upstream partition fragment ids for the output rows. A single read task can own multiple fragments, and scanner filters can remove arbitrary rows, so the task-level fragment list is not precise enough per row. The row address is the reliable source of truth for the row's fragment. Also, with the Lance version in this env, `LanceDataset.get_fragments(filter=...)` documents the filter as unsupported, so we cannot pre-partition filtered fragments directly.

nemo_curator/stages/text/io/reader/lance.py
VibhuJawa: Agreed; I removed that custom type validation. `LanceReadTask.validate()` now only checks that the fragment list is non-empty, and requested fragment IDs are validated in `LancePartitioningStage` against the dataset's actual fragments.

nemo_curator/stages/text/io/reader/lance.py
VibhuJawa: This deterministic id is for stable source task IDs across runs. It includes the dataset path, pinned Lance version, and fragment ids so the same fragment partition gets the same task id for resumability/checkpoint naming, while a new Lance dataset version gets a different id.

nemo_curator/stages/text/io/reader/lance.py
```diff
@@ -0,0 +1,311 @@
...
+        return dataset_kwargs
+
+    def _scanner_kwargs(self, read_kwargs: dict[str, Any], fields: list[str] | None) -> dict[str, Any]:
+        scanner_kwargs = dict(read_kwargs.pop("scanner_options", {}) or {})
+        scanner_kwargs.update(read_kwargs)
+        if fields is not None:
+            scanner_kwargs["columns"] = fields
+        return scanner_kwargs
+
+    def read_task(
```
VibhuJawa, line 155: Main class to review

nemo_curator/stages/text/io/reader/base.py
```diff
@@ -72,24 +84,16 @@ def setup(self, _: WorkerMetadata | None = None) -> None:
...
-        # Validate the result
-        if (
-            (result is None)
-            or (hasattr(result, "empty") and result.empty)
-            or (hasattr(result, "num_rows") and result.num_rows == 0)
-        ):
-            msg = f"No data read from files in task {task.task_id}"
-            raise ValueError(msg)
+    def _effective_read_kwargs(self) -> dict[str, Any]:
+        return dict(self.read_kwargs or {})
```
sarahyurick: Nit but I don't see a reason for having a 1 line helper function.

nemo_curator/stages/text/io/reader/base.py
```diff
@@ -100,16 +104,29 @@ def process(self, task: FileGroupTask) -> DocumentBatch:
         return DocumentBatch(
             dataset_name=task.dataset_name,
             data=result,
-            _metadata=task._metadata,
+            _metadata=self._output_metadata(task, output),
+            _stage_perf=task._stage_perf,
         )
 
+    def _output_metadata(self, task: ReaderTask, _output: ReaderOutput) -> dict[str, Any]:
+        return task._metadata
```
sarahyurick: Same comment as above.

nemo_curator/stages/text/io/reader/lance.py
```diff
@@ -0,0 +1,311 @@
...
+        version = read_kwargs.pop("version", version)
+        if version is not None:
+            dataset_kwargs["version"] = version
+        storage_options = read_kwargs.pop("storage_options", None)
+        if storage_options is not None:
+            dataset_kwargs["storage_options"] = storage_options
+        return dataset_kwargs
+
+    def process(self, _: EmptyTask) -> list[LanceReadTask]:
+        import lance
```
sarahyurick: Top-level import?

nemo_curator/stages/text/io/reader/lance.py
```diff
@@ -0,0 +1,311 @@
...
+
+    Args:
+        path: Path or URI of the Lance dataset.
+        fields: Optional columns to read. Overrides ``columns`` in ``read_kwargs``.
+        read_kwargs: Keyword arguments for Lance dataset and scanner construction.
+        include_lance_metadata: Whether to include row-address and fragment-id metadata columns.
+        allow_empty: Whether filtered reads may return empty tables without raising.
+    """
+
+    path: str = ""
```
sarahyurick: Make it a hard requirement instead of having a check in the post init:
```suggestion
    path: str
```

nemo_curator/stages/text/io/reader/lance.py
```diff
@@ -0,0 +1,311 @@
...
+        output_fields = list(columns or [])
+        if self.include_lance_metadata:
+            output_fields.extend([LANCE_ROWADDR_COLUMN, LANCE_FRAGID_COLUMN])
+        return ["data"], output_fields
+
+    def _output_metadata(self, task: LanceReadTask, output: ReaderOutput) -> dict[str, Any]:
+        return output.metadata if output.metadata is not None else task._metadata
+
+    def _restore_blob_v2_columns(self, dataset: object, table: pa.Table, blob_columns: list[str]) -> pa.Table:
+        import lance
```
sarahyurick: Same here, should it be a top-level import? We can lazy-load the `LanceReader` so that lance is not a hard dependency.

nemo_curator/stages/text/io/reader/lance.py
```diff
@@ -0,0 +1,311 @@
...
+
+        rowaddrs = [int(value) for value in table["_rowaddr"].combine_chunks().to_pylist()]
+        for column in blob_columns:
+            payloads = [
+                payload
+                for _, payload in dataset.read_blobs(column, addresses=rowaddrs, preserve_order=True)  # type: ignore[attr-defined]
```
sarahyurick: General question does `# type: ignore[attr-defined]` matter for the codebase? Like will it break without it?

nemo_curator/stages/text/io/reader/lance.py
```diff
@@ -0,0 +1,311 @@
...
+        return scanner_kwargs
+
+    def read_task(
+        self,
+        task: LanceReadTask,
+        read_kwargs: dict[str, Any] | None,
+        fields: list[str] | None,
+    ) -> ReaderOutput:
+        import lance
+        from lance.schema import schema_to_json
```
sarahyurick: Similar comment as above.

tests/stages/text/io/reader/test_lance.py
```diff
@@ -0,0 +1,154 @@
...
+from nemo_curator.stages.text.io.reader.lance import (
+    LANCE_FRAGID_COLUMN,
+    LANCE_ROWADDR_COLUMN,
+    LancePartitioningStage,
+    LanceReader,
+    LanceReaderStage,
+)
+from nemo_curator.tasks import EmptyTask
+
+pytest.importorskip("lance")
```
sarahyurick: Should we create a `@pytest.mark.lance` or something instead? I kinda worry about using `importorskip` because unless someone is explicitly checking the relevant CI job then it could just silently skip or something.

sarahyurick (commented):
Hi, just walking through the PR and leaving some minor comments for now. Will try to do more of a deep dive soon.

VibhuJawa:
Stable-row-ID follow-up: the Lance reader now exposes `__lance_rowid` alongside row address/fragment metadata and records `has_stable_row_ids` in task metadata. The reader suite passes against PyLance 7.0.

nemo_curator/stages/text/io/reader/base.py
```diff
@@ -28,20 +29,31 @@
...
-    Subclasses must implement the read_data method.
+    Subclasses must implement read_task for their input task type.
     """
 
     fields: list[str] | None = None
     read_kwargs: dict[str, Any] = field(default_factory=dict)
     name: str = ""
     _generate_ids: bool = False
     _assign_ids: bool = False
+    allow_empty: bool = False
```
praateekmahajan, line 56: What's this for?

nemo_curator/stages/text/io/reader/base.py
```diff
@@ -28,20 +29,31 @@
 from nemo_curator.backends.utils import RayStageSpecKeys
 from nemo_curator.stages.base import ProcessingStage
 from nemo_curator.tasks import DocumentBatch, FileGroupTask
+from nemo_curator.tasks.tasks import Task
+
+ReaderTask = TypeVar("ReaderTask", bound=Task)
```
praateekmahajan: Confused with what this means / does?

Shouldn't this be `FileGroupTask | LanceReadTask`?

nemo_curator/stages/text/io/reader/base.py
praateekmahajan: Agreed

nemo_curator/stages/text/io/reader/lance.py
```diff
@@ -0,0 +1,315 @@
...
+    """Task containing Lance fragment ids assigned to one read partition.
+
+    This is created by ``LancePartitioningStage`` and consumed by
+    ``LanceReaderStage``.
+
+    Args:
+        data: Lance fragment ids to read.
+    """
+
+    data: list[int] = field(default_factory=list)
```
praateekmahajan: In my mind, a stage's output should be enough to be the next stage's input. In my mind it's unclear how just from `list[int]` we can know what needs to be read?

IOW
```
files = FilePartitiongStage(...).process(EmptyTask())
document_batch = JsonlReaderStage(...).process(files)
```
where `files` contain everything that needs to be known

But here how does int contain that information, I'm guessing it relies on the init kwargs of the JsonlReader and uses that along with what's inside the task..

does it make sense?

What if we do something like
```python
class LanceReadTask(Task[list[tuple[str, int]]]):
	data : list[tuple[str, int]]
```
where the tuple contains the path AND the fragment id?

nemo_curator/stages/text/io/reader/lance.py
praateekmahajan: Since nemo_curator/stages/text/io/reader/__init__.py imports LanceReader I don't think we should move it to a top-level import since that means even a jsonl reader will need to resolve lance

nemo_curator/stages/text/io/reader/lance.py
```diff
@@ -0,0 +1,315 @@
...
+            if missing:
+                msg = f"Lance dataset does not contain requested fragment ids: {missing[:10]}"
+                raise ValueError(msg)
+
+        tasks = []
+        for start in range(0, len(fragment_ids), self.fragments_per_partition):
+            owned_fragments = fragment_ids[start : start + self.fragments_per_partition]
+            tasks.append(
+                LanceReadTask(
+                    dataset_name=self.path,
```
praateekmahajan: dataset_name is sadly some dead weight we're carrying, but iirc it's usually not path and some strigified version.. you might be able to find some utility that does path to dataset name somewhere in the code..

nemo_curator/stages/text/io/reader/lance.py
```diff
@@ -0,0 +1,315 @@
...
+        else:
+            fragment_ids = sorted(set(self.fragment_ids))
+            missing = sorted(set(fragment_ids) - set(available_fragments))
+            if missing:
+                msg = f"Lance dataset does not contain requested fragment ids: {missing[:10]}"
+                raise ValueError(msg)
+
+        tasks = []
+        for start in range(0, len(fragment_ids), self.fragments_per_partition):
+            owned_fragments = fragment_ids[start : start + self.fragments_per_partition]
```
praateekmahajan:
```suggestion
            fragment_ids_for_task = fragment_ids[start : start + self.fragments_per_partition]
```

nemo_curator/stages/text/io/reader/lance.py
praateekmahajan: agreed

nemo_curator/stages/text/io/reader/lance.py
praateekmahajan: How are we doing the lazy-loading? As long as we lance at each import (of JsonlReader), I'm okay to move this to top level

nemo_curator/stages/text/io/reader/lance.py
```diff
@@ -0,0 +1,315 @@
...
+        read_kwargs = dict(read_kwargs or {})
+        dataset_kwargs = self._dataset_kwargs(read_kwargs, self._task_version(task))
+        scanner_kwargs = self._scanner_kwargs(read_kwargs, fields)
+        dataset = lance.dataset(self.path, **dataset_kwargs)
+        fragments = [dataset.get_fragment(fragment_id) for fragment_id in task.data]
+        requested_columns = scanner_kwargs.get("columns")
+        blob_columns = [
+            field.name
+            for field in dataset.schema
+            if getattr(field.type, "extension_name", None) == "lance.blob.v2"
```
praateekmahajan: can you add a comment explaining what this is? and the need for other methods like `_restore_blob_v2_columns`?

nemo_curator/stages/text/io/reader/lance.py
sarahyurick: I think this is where lazy loading can be used, or just don't include it in the init at all.

nemo_curator/stages/text/io/reader/lance.py
sarahyurick: I was thinking of examples like: https://github.com/NVIDIA-NeMo/Curator/pull/1888 should be sufficient I think?

nemo_curator/utils/lance.py
```diff
@@ -0,0 +1,32 @@
...
+
+import pyarrow as pa
+
+LANCE_ROWADDR_COLUMN = "__lance_rowaddr"
+LANCE_FRAGID_COLUMN = "__lance_fragid"
+
+
+def lance_fragment_ids_from_row_addresses(rowaddr_column: pa.ChunkedArray) -> pa.Array:
+    rowaddrs = rowaddr_column.combine_chunks().cast(pa.uint64())
+    return pa.array([int(value) >> 32 for value in rowaddrs.to_pylist()], type=pa.uint64())
```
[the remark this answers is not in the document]
VibhuJawa: Agreed. This now uses `pyarrow.compute.shift_right`; the shift scalar is explicitly `uint64` so row addresses above `2^63 - 1` remain valid.

nemo_curator/stages/text/io/reader/lance.py
```diff
@@ -0,0 +1,311 @@
...
+                fragments_per_partition=self.fragments_per_partition,
+                fragment_ids=self.fragment_ids,
+                read_kwargs=self.read_kwargs,
+            ),
+            LanceReaderStage(
+                path=self.path,
+                fields=self.fields,
+                read_kwargs=self.read_kwargs,
+                include_lance_metadata=self.include_lance_metadata,
+            ),
```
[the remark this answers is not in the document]
VibhuJawa: `_pop_dataset_kwargs` now centralizes dataset options: nested `dataset_options` is supported, top-level `version` and `storage_options` override it, and remaining options flow to the scanner.

nemo_curator/stages/text/io/reader/base.py
VibhuJawa: Agreed. The defensive copy is inlined in `BaseReader.process()`, and `_effective_read_kwargs()` is removed.

nemo_curator/stages/text/io/reader/base.py
VibhuJawa: Agreed. `_document_batch()` now uses `ReaderOutput.metadata` when provided and otherwise falls back to task metadata; the helper and Lance override are removed.

nemo_curator/stages/text/io/reader/lance.py
VibhuJawa: Removed the duplicate stage path configuration. `LanceReadTask` now carries the required path and version, and the reader opens the dataset from the task.

nemo_curator/stages/text/io/reader/lance.py
VibhuJawa: Each `LanceReadTask` carries the dataset path, version, and fragment IDs, so `LanceReaderStage` can read directly from the partitioning output.

nemo_curator/stages/text/io/reader/lance.py
VibhuJawa: It would not affect runtime; the ignore only suppressed static checking because `dataset` was typed as `object`. The parameter is now `lance.LanceDataset`, so the ignore is unnecessary.

tests/stages/text/io/reader/test_lance.py
VibhuJawa: The CPU test environment installs the `lance` extra, so a missing Lance dependency should fail rather than skip. The tests now use a normal import; no marker is needed.

nemo_curator/stages/text/io/reader/base.py
VibhuJawa: `BaseReader` treats empty reads as errors. A Lance filter can legitimately eliminate every row from one fragment partition, so Lance sets `allow_empty=True`; JSONL and Parquet keep the default `False`.

nemo_curator/stages/text/io/reader/base.py
VibhuJawa: Agreed. `LanceReadTask` now lives in `nemo_curator.tasks`, and `ReaderTask` is the explicit `FileGroupTask | LanceReadTask` alias.

nemo_curator/utils/lance.py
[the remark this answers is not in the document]
VibhuJawa: [Vectorized fragment-ID extraction](https://github.com/VibhuJawa/NeMo-Curator/blob/a9667622361859395c63375f80b2be4462bef293/nemo_curator/utils/lance.py#L23-L25).

nemo_curator/stages/text/io/reader/lance.py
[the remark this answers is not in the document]
VibhuJawa: [Dataset-option parsing](https://github.com/VibhuJawa/NeMo-Curator/blob/a9667622361859395c63375f80b2be4462bef293/nemo_curator/stages/text/io/reader/lance.py#L38-L52) and [scanner-option merging](https://github.com/VibhuJawa/NeMo-Curator/blob/a9667622361859395c63375f80b2be4462bef293/nemo_curator/stages/text/io/reader/lance.py#L169-L175).

nemo_curator/stages/text/io/reader/base.py
VibhuJawa: [The defensive copy is inlined in `BaseReader.process()`](https://github.com/VibhuJawa/NeMo-Curator/blob/a9667622361859395c63375f80b2be4462bef293/nemo_curator/stages/text/io/reader/base.py#L87-L90).

nemo_curator/stages/text/io/reader/base.py
VibhuJawa: [Metadata precedence now lives directly in `_document_batch()`](https://github.com/VibhuJawa/NeMo-Curator/blob/a9667622361859395c63375f80b2be4462bef293/nemo_curator/stages/text/io/reader/base.py#L92-L106).

nemo_curator/stages/text/io/reader/lance.py
VibhuJawa: [`LanceReadTask` carries path and version](https://github.com/VibhuJawa/NeMo-Curator/blob/a9667622361859395c63375f80b2be4462bef293/nemo_curator/tasks/lance.py#L25-L37), and [the reader consumes them](https://github.com/VibhuJawa/NeMo-Curator/blob/a9667622361859395c63375f80b2be4462bef293/nemo_curator/stages/text/io/reader/lance.py#L177-L188).

nemo_curator/stages/text/io/reader/lance.py
VibhuJawa: [Partitioning emits the complete read task](https://github.com/VibhuJawa/NeMo-Curator/blob/a9667622361859395c63375f80b2be4462bef293/nemo_curator/stages/text/io/reader/lance.py#L97-L117), and [the reader consumes it directly](https://github.com/VibhuJawa/NeMo-Curator/blob/a9667622361859395c63375f80b2be4462bef293/nemo_curator/stages/text/io/reader/lance.py#L177-L188).

nemo_curator/stages/text/io/reader/lance.py
VibhuJawa: [The dataset parameter is typed as `lance.LanceDataset`](https://github.com/VibhuJawa/NeMo-Curator/blob/a9667622361859395c63375f80b2be4462bef293/nemo_curator/stages/text/io/reader/lance.py#L155-L157).

tests/stages/text/io/reader/test_lance.py
VibhuJawa: [The Lance tests use a normal module import](https://github.com/VibhuJawa/NeMo-Curator/blob/a9667622361859395c63375f80b2be4462bef293/tests/stages/text/io/reader/test_lance.py#L15-L20).

nemo_curator/stages/text/io/reader/base.py
VibhuJawa: [`ReaderTask` is the explicit union](https://github.com/VibhuJawa/NeMo-Curator/blob/a9667622361859395c63375f80b2be4462bef293/nemo_curator/stages/text/io/reader/base.py#L31-L44), with [`LanceReadTask` in `nemo_curator.tasks`](https://github.com/VibhuJawa/NeMo-Curator/blob/a9667622361859395c63375f80b2be4462bef293/nemo_curator/tasks/lance.py#L25-L37).

nemo_curator/stages/text/io/reader/lance.py
VibhuJawa: `infer_dataset_name_from_path` now accepts [`path_kind="directory"`](https://github.com/VibhuJawa/NeMo-Curator/blob/a9667622361859395c63375f80b2be4462bef293/nemo_curator/utils/file_utils.py#L358-L380), and [Lance uses it](https://github.com/VibhuJawa/NeMo-Curator/blob/a9667622361859395c63375f80b2be4462bef293/nemo_curator/stages/text/io/reader/lance.py#L97-L105) to derive the name from the directory's final component.

nemo_curator/stages/text/io/reader/lance.py
VibhuJawa: The reader package now [lazily exports all public readers](https://github.com/VibhuJawa/NeMo-Curator/blob/a9667622361859395c63375f80b2be4462bef293/nemo_curator/stages/text/io/reader/__init__.py#L15-L31), so [Lance imports remain at module scope](https://github.com/VibhuJawa/NeMo-Curator/blob/a9667622361859395c63375f80b2be4462bef293/nemo_curator/stages/text/io/reader/lance.py#L20-L22) without loading Lance when importing JSONL or Parquet.

nemo_curator/stages/text/io/reader/lance.py
VibhuJawa: The reader package [lazily exports all readers](https://github.com/VibhuJawa/NeMo-Curator/blob/a9667622361859395c63375f80b2be4462bef293/nemo_curator/stages/text/io/reader/__init__.py#L15-L31), allowing [`lance` and `schema_to_json` to remain module-level imports](https://github.com/VibhuJawa/NeMo-Curator/blob/a9667622361859395c63375f80b2be4462bef293/nemo_curator/stages/text/io/reader/lance.py#L20-L22).

nemo_curator/stages/text/io/reader/lance.py
VibhuJawa: `lance` and `schema_to_json` are now [module-level imports](https://github.com/VibhuJawa/NeMo-Curator/blob/a9667622361859395c63375f80b2be4462bef293/nemo_curator/stages/text/io/reader/lance.py#L20-L22), while the [package-level lazy export](https://github.com/VibhuJawa/NeMo-Curator/blob/a9667622361859395c63375f80b2be4462bef293/nemo_curator/stages/text/io/reader/__init__.py#L15-L31) keeps them isolated until `LanceReader` is accessed.

nemo_curator/stages/text/io/reader/lance.py
VibhuJawa: Renamed to [`fragment_ids_for_task`](https://github.com/VibhuJawa/NeMo-Curator/blob/a9667622361859395c63375f80b2be4462bef293/nemo_curator/stages/text/io/reader/lance.py#L97-L115).

nemo_curator/stages/text/io/reader/lance.py
VibhuJawa: Lance scans Blob v2 columns as descriptors rather than payload bytes, so the reader [materializes requested blobs with `read_blobs()` and aligns them by row address](https://github.com/VibhuJawa/NeMo-Curator/blob/a9667622361859395c63375f80b2be4462bef293/nemo_curator/stages/text/io/reader/lance.py#L155-L167). The [scan path returns `large_binary` while preserving the original Blob schema in metadata](https://github.com/VibhuJawa/NeMo-Curator/blob/a9667622361859395c63375f80b2be4462bef293/nemo_curator/stages/text/io/reader/lance.py#L190-L216).

VibhuJawa:
@praateekmahajan / @sarahyurick , Please take another look, should be good for a followup review now

nemo_curator/stages/text/io/reader/lance.py
```diff
@@ -0,0 +1,270 @@
...
+                    _metadata={
+                        "source_files": [self.path],
+                        "lance": {
+                            "path": self.path,
+                            "version": dataset.version,
+                            "fragment_ids": fragment_ids_for_task,
+                        },
+                    },
+                )
+            )
```
sarahyurick, line 108: I am wondering why is it saving as metadata information that is already being saved via the `path`, `version`, and `data` fields?

VibhuJawa: Fair enough, we can skip it now . Let me fix that.

VibhuJawa: codex: Agreed. `LanceReadTask` fields are now the canonical read inputs ([code](https://github.com/NVIDIA-NeMo/Curator/blob/d46cb782f73e26de446edbb5960dbd87b8f75673/nemo_curator/stages/text/io/reader/lance.py#L102-L107)); lineage is added only to the resulting `DocumentBatch` ([code](https://github.com/NVIDIA-NeMo/Curator/blob/d46cb782f73e26de446edbb5960dbd87b8f75673/nemo_curator/stages/text/io/reader/lance.py#L188-L196)). The test confirms path, version, and fragment IDs determine partition identity ([code](https://github.com/NVIDIA-NeMo/Curator/blob/d46cb782f73e26de446edbb5960dbd87b8f75673/tests/stages/text/io/reader/test_lance.py#L74-L80)).

VibhuJawa: Removed it https://github.com/NVIDIA-NeMo/Curator/pull/2111/changes/d46cb78d6b22128bfda59fa2cd555bdb6c92d851 .

Added it in metadata of LanceReaderStage in  DocumentBatch , which i think makes sense because there we should  carry metadata that is specific to lance fields !

sarahyurick (approved):
LGTM!

nemo_curator/stages/text/io/reader/base.py
```diff
@@ -72,24 +84,13 @@ def setup(self, _: WorkerMetadata | None = None) -> None:
...
-        if (
-            (result is None)
-            or (hasattr(result, "empty") and result.empty)
-            or (hasattr(result, "num_rows") and result.num_rows == 0)
-        ):
-            msg = f"No data read from files in task {task.task_id}"
-            raise ValueError(msg)
+    def process(self, task: ReaderTask) -> DocumentBatch:
+        output = self.read_task(task, dict(self.read_kwargs or {}), self.fields)
+        self._validate_result(task, output.data)
```
praateekmahajan, line 89: We might be breaking contract by adding validate here? Technically isn't validate a part of  process_batch?

https://github.com/NVIDIA-NeMo/Curator/blob/2c9786cc7c5ba2c44863d03b8f63937c75bc42fd/nemo_curator/stages/base.py#L243

nemo_curator/stages/text/io/reader/base.py
```diff
@@ -27,21 +28,32 @@
...
+@dataclass(frozen=True)
+class ReaderOutput:
+    data: ReaderData
+    metadata: dict[str, Any] | None = None
 
 
 @dataclass
-class BaseReader(ProcessingStage[FileGroupTask, DocumentBatch]):
-    """Common base for tabular file readers.
+class BaseReader(ProcessingStage[ReaderTask, DocumentBatch]):
```
praateekmahajan, line 44: Sort of confused shouldn't output of BaseReader also be `DocumentBatch | TypesOfOutputLanceReaderSupports`

tests/stages/text/io/reader/test_lance.py
```diff
@@ -0,0 +1,195 @@
...
+        max_rows_per_group=2,
+        data_storage_version="2.2",
+        enable_stable_row_ids=enable_stable_row_ids,
+    )
+
+
+def test_lance_reader_partitions_filters_blobs_and_metadata(tmp_path: Path):
+    dataset_path = tmp_path / "docs.lance"
+    _write_lance_dataset(dataset_path)
+    read_tasks = LancePartitioningStage(path=str(dataset_path), fragments_per_partition=1).process(EmptyTask)
```
praateekmahajan: i think ...?
```suggestion
    read_tasks = LancePartitioningStage(path=str(dataset_path), fragments_per_partition=1).process(EmptyTask())
```

tests/stages/text/io/reader/test_lance.py
```diff
@@ -0,0 +1,195 @@
...
+    assert read_tasks[0].path == str(dataset_path)
+    assert read_tasks[0].version == lance.dataset(str(dataset_path)).version
+    assert {fragment_id for task in read_tasks for fragment_id in task.data} == {0, 1}
+    assert all(task._metadata == {} for task in read_tasks)
+    first_task = read_tasks[0]
+    first_task_id = first_task.get_deterministic_id()
+    assert first_task_id != read_tasks[1].get_deterministic_id()
+    assert first_task_id != replace(first_task, path=f"{first_task.path}.other").get_deterministic_id()
+    assert first_task_id != replace(first_task, version=first_task.version + 1).get_deterministic_id()
+    assert first_task_id != replace(first_task, data=[*first_task.data, 999]).get_deterministic_id()
```
praateekmahajan: nit: probably a good idea to add `tests/tasks/test_lance.py` where you test out a `LanceTask` so that we don't need to do those tests here


```suggestion
```

praateekmahajan (approved):
LGTM, left a few comments re tests / base class inheirtance / validate

nemo_curator/stages/text/io/reader/base.py
VibhuJawa: For now the output is TypesOfOutputLanceReaderSupports is  DocumentBatch only. It is just another data source

nemo_curator/stages/text/io/reader/base.py
VibhuJawa: So this is  validate output (and the code was always structured like this, i just put in a function), i think its fine as is. 

I dont think we have a place where we validate output apart from process.

tests/stages/text/io/reader/test_lance.py
VibhuJawa: Yes.

VibhuJawa: codex: Fixed the source-stage test plumbing to pass `EmptyTask()` instances consistently in [17aae8b](https://github.com/NVIDIA-NeMo/Curator/commit/17aae8b49c9fad8a81c87b8e61da2e6ea3657b9b).

tests/stages/text/io/reader/test_lance.py
VibhuJawa: codex: Moved `LanceReadTask` deterministic-ID coverage into a focused [task test](https://github.com/NVIDIA-NeMo/Curator/blob/c3d792b048e82129b80dea50d339ce3592c65d43/tests/tasks/test_lance.py#L20-L27) and removed it from the reader integration test.

VibhuJawa:
/ok to test [a8285af](https://github.com/NVIDIA-NeMo/Curator/commit/a8285afd5f9ec217a9aa0dc2101fb4e891930f53)

VibhuJawa:
/ok to test [2dcc06a](https://github.com/NVIDIA-NeMo/Curator/commit/2dcc06a5b46add0314c62e098c074edb323dbd9a)

PR #2115, opened 2026-06-25
[benchmarking] Generate stable nightly session names

rlratzel:
## Summary
- Generate a single scheduled-nightly benchmark session name in `generate_ci_tests.py`.
- Emit `NEMO_CI_SESSION_NAME` as a top-level generated pipeline variable so every benchmark job in the pipeline uses the same session directory.
- Simplify `ci_benchmark_launcher.sh` to consume `NEMO_CI_SESSION_NAME` or fall back to the legacy `benchmark_run_<pipeline-id>` name for manual launches.

## Notes
The previous per-job timestamp generation in `ci_benchmark_launcher.sh` could produce a different session name for each benchmark job. This keeps seconds in the nightly name, but derives it once from `CI_PIPELINE_CREATED_AT` while generating the benchmark pipeline.

## Validation
- `bash -n benchmarking/tools/ci_benchmark_launcher.sh`
- `python3 -m py_compile benchmarking/tools/generate_ci_tests.py`
- Scheduled generator run emits `NEMO_CI_SESSION_NAME: nightly-2026_06_25__17_18_19_UTC`.
- Explicit `NEMO_CI_SESSION_NAME=custom-session` is preserved.
- Manual/non-scheduled generator run omits `NEMO_CI_SESSION_NAME`.
- Pre-commit hooks passed during commit.

### DIFF
--- a/benchmarking/tools/generate_ci_tests.py
+++ b/benchmarking/tools/generate_ci_tests.py
@@ -13,6 +13,9 @@
 # limitations under the License.
 
 import argparse
+import datetime as dt
+import os
+from collections.abc import Mapping
 from pathlib import Path
 
 from ruamel.yaml import YAML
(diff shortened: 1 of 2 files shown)


benchmarking/tools/generate_ci_tests.py
```diff
@@ -140,14 +164,17 @@ def main() -> None:
...
-    pipeline = generate_pipeline(args.curator_dir, args.scope)
+    session_name = session_name_from_env()
+    pipeline = generate_pipeline(args.curator_dir, args.scope, session_name=session_name)
 
     output_file = "generated_curator_benchmark_tests.yml"
     with open(output_file, "w") as f:
         yaml.dump(pipeline, f)
 
-    job_count = len([k for k in pipeline if k != "include"])
+    job_count = len([k for k in pipeline if k not in {"include", "variables"}])
```
sarahyurick, line 174: What is this line for?

sarahyurick (approved):
LGTM, added a small question.

benchmarking/tools/generate_ci_tests.py
rlratzel: It's simply to compute the number of individual jobs in the pipeline for the print statement below it. It can be removed along with the print statement, since the print statement is just a sanity check for anyone looking at the stdout.

PR #2116, opened 2026-06-29
Add benchmark max timeout guard

rlratzel:
## Summary
- add a benchmark-level max_timeout_s setting with a 3h59m default
- validate effective entry timeouts after YAML merging/default timeout application
- enforce the same cap in generated CI benchmark jobs
- lower ReadSpeech benchmark hang guards to 3h30m and document the new setting

## Validation
- pre-commit hooks passed during signed-off commit
- CI benchmark generator produced 41 jobs
- verified no enabled nightly entry exceeds max_timeout_s

### DIFF
--- a/tests/benchmarking/runner/test_session.py
+++ b/tests/benchmarking/runner/test_session.py
@@ -1,6 +1,21 @@
+# Copyright (c) 2026, NVIDIA CORPORATION.  All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
 from __future__ import annotations
 
 import sys
+import types
 from pathlib import Path
 from typing import TYPE_CHECKING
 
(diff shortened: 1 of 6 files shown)


tests/benchmarking/runner/test_session.py
```diff
@@ -0,0 +1,87 @@
+from __future__ import annotations
```
sarahyurick, line 15: Can you add a copyright header here?

tests/benchmarking/runner/test_session.py
```diff
@@ -0,0 +1,87 @@
...
+import sys
+from pathlib import Path
+from typing import TYPE_CHECKING
+
+import pytest
+
+if TYPE_CHECKING:
+    from collections.abc import Callable
+
+sys.path.insert(0, str(Path(__file__).resolve().parents[3] / "benchmarking"))
```
sarahyurick, line 27: Can you check that these tests are getting picked up by the CI somewhere? Does something under `.github/` need to be added to run this new test suite?

sarahyurick (changes requested):

sarahyurick:
Fyi @rlratzel if you intend to keep the pytests then we should no longer use the `docs-only` tag on benchmark PRs.

tests/benchmarking/runner/test_session.py
rlratzel: I don't believe they're getting picked up - they were added strictly for local testing (hence docs-only), but I agree that they don't make sense if CI isn't running them.

tests/benchmarking/runner/test_session.py
rlratzel: Done, thanks!

tests/benchmarking/runner/test_session.py
rlratzel: They should be running now in CI.

sarahyurick (approved):

PR #1835, opened 2026-04-20
fix(audio): lazy-load audio stages and fix tutorial notebook paths

mohammadaaftabv:
## Summary

- **Empty `nemo_curator/stages/audio/__init__.py`** — removes eager imports of all 13 audio stages at package load time. Previously, importing *any* audio stage transitively pulled in every optional dependency (e.g. `onnxruntime` for `SIGMOSFilterStage`), causing `ModuleNotFoundError` in tutorials that only need a subset. This now follows the same pattern used by the `image` modality.

- **Fix `tutorials/audio/readspeech/pipeline.py`** — the only internal consumer of the old top-level import. Updated to import `AudioDataFilterStage` from its specific subpackage (`nemo_curator.stages.audio.advanced_pipelines`).

- **Fix ALM tutorial notebook** — use relative paths (`../../../tests/fixtures/...`) instead of relying on `subprocess`/`git rev-parse` for repo root discovery. Also corrects the API to use `ALMManifestReader` (composite stage) instead of `ALMManifestReaderStage`.

- **Fix FLEURS tutorial notebook** — use simple relative path (`./example_audio/fleurs`) for data directory.

- **Standardize `RayClient` usage** across all ALM and FLEURS tutorial scripts (`main.py`, `pipeline.py`, `run.py`) and notebooks (`alm_tutorial.ipynb`, `fleurs_tutorial.ipynb`). `RayClient` manages the Ray cluster lifecycle (start/stop, port allocation, dashboard) — both `XennaExecutor` and `RayDataExecutor` run on top of Ray, so the client is needed regardless of backend. Scripts use `try`/`finally` for cleanup; notebooks use separate cells.

- **Backend comparison in notebooks** — both notebooks now run the pipeline with both Xenna and Ray Data backends, compare results (entries, windows, WER, audio duration) for correctness, and report wall-clock timing with speedup.

- **Add `RayClient` to benchmark scripts** — `alm_pipeline_benchmark.py` and `audio_fleurs_benchmark.py` now start/stop `RayClient` for standalone invocation. When run via the nightly runner, the runner's existing `RayClient` is reused (detected via `RAY_ADDRESS`).

- **Update benchmark docs** — `ALM_BENCHMARK.md` and `AUDIO_PROFILING.md` updated with both-backend results. No performance regression from `RayClient` changes.

- **Update READMEs** — ALM and FLEURS READMEs updated to document `RayClient` usage and show full execution patterns (including `RayClient`) in composability examples.

[...]

Issue #2118, opened 2026-06-29
[Docs epic] Close post-v1.2.0 documentation gaps for 26.07


lbliii:
## Goal

Close every actionable documentation gap introduced after [NeMo Curator v1.2.0](<https://github.com/NVIDIA-NeMo/Curator/releases/tag/v1.2.0>), and make the next release's docs/release notes complete enough for users to adopt new features and migrate breaking behavior.

## Audit boundary and method

* **Primary boundary:** commits reachable from `main` but absent from the v1.2.0 tag, reviewed through PR [#2116](<https://github.com/NVIDIA-NeMo/Curator/issues/2116>) (merged 2026-06-29 19:34:32 UTC)
* **Original inventory:** 67 PRs merged after the v1.2.0 publication timestamp
* **Supplemental code-freeze audit:** 22 additional release-relevant PRs identified by Sara that merged after the 26.04 code freeze but before the release publication timestamp
* **Evidence reviewed:** PR titles/bodies, changed file paths, `git v1.2.0..main`, and the current `fern/versions/main/pages/` documentation tree
* **Decision rule:** create focused docs work for user-facing features, public API/configuration changes, behavior changes, and migration hazards; bundle smaller release-note and reference gaps instead of creating noise

> The publication timestamp alone was insufficient: it omitted release-relevant commits that missed v1.2.0 despite merging before the release was published. The supplemental disposition is tracked in NVIDIA-NeMo/Curator#2144, NVIDIA-NeMo/Curator#2145, and NVIDIA-NeMo/Curator#2146.

## Planned documentation workstreams

- [ ] NVIDIA-NeMo/Curator#2120 — Publish 26.07 release notes and a migration checklist
- [X] NVIDIA-NeMo/Curator#2119 — Document the Nemotron-CLIMB curation recipe
- [ ] NVIDIA-NeMo/Curator#2123 — Document audio tagging, quality metrics, normalization, and LLM PnC
- [ ] NVIDIA-NeMo/Curator#2122 — Document long-form audio cutting for ALM pretraining
- [ ] NVIDIA-NeMo/Curator#2121 — Document the Nemotron OCR synthetic-data pipeline
- [ ] NVIDIA-NeMo/Curator#2125 — Document the video caption-quality evaluation workflow
- [ ] NVIDIA-NeMo/Curator#2124 — Replace legacy resumability guidance and document deterministic task identity
[...]

PR #2136, opened 2026-06-29
[codex] docs: add long-form audio cutting guide

### LINKED ISSUE #2122: [Docs] Document long-form audio cutting for ALM pretraining
## Context

Child of #2118. #1898 added a new long-form diarized-audio cutting pipeline, with follow-up tutorial/notebook stabilization in #1835. It currently has a repository README but no first-class Fern page.

## APIs and workflow
- `ReadLongFormManifestStage`
- `OverlapFilterStage`
- `SnippetCutPlannerStage`
- `SnippetExtractionStage`
- `SnippetManifestWriterStage`
- `PretrainMetricsAggregatorStage`
- `prepare_audio_pretrain_outputs`, `finalize_audio_pretrain_outputs`, and `build_audio_pretrain_pipeline`

## Requirements
- Define the input JSONL/segment schema and the final snippet-manifest schema.
- Explain overlap removal, gap/duration planning, timestamp shifting/clamping, mono conversion, resampling, and metadata preservation.
- Document dry-run sizing, shard cleanup/merge behavior, metrics output, and backend compatibility.
- Include CLI and Python examples plus production cautions for output paths and stale shards.

## Acceptance criteria
- [ ] New Fern concept/tutorial content is added to the audio hierarchy
- [ ] Input/output schemas include realistic examples
- [ ] All important defaults and path semantics are documented
- [ ] Dry-run and resume/retry guidance is included
- [ ] The ALM tutorial and audio landing pages link to it
- [ ] Commands are tested against current code and Fern checks pass

Related PRs: #1898, #1835

lbliii:
## Summary

- add a first-class Fern tutorial for cutting long-form diarized and transcribed audio into ALM pretraining snippets
- document the seven-stage pipeline, greedy packing and overlap rules, repetition filtering, mono/resampling behavior, and manifest/tar reconciliation
- provide realistic input, output, metrics, and field-transformation schemas
- include copy-pasteable dry-run, full extraction, and Python API examples
- document every CLI default, path-resolution mode, tokenizer requirement, backend behavior, storage consideration, and important failure mode
- explain driver-side shard preparation/finalization, partial-output recovery, stale-shard cleanup, and the lack of checkpoint resume
- add navigation and cross-links from the audio landing page, tutorial index, ALM tutorial, data builder, and overlap-filtering reference

## Why

The workflow introduced by #1898 and stabilized by #1835 was discoverable only through `tutorials/audio/audio_pretrain/README.md`. Users had no Fern guidance for preparing the source schema, choosing path semantics, sizing a run with `--dry-run`, interpreting the output archive and metrics, or safely operating the multi-worker shard lifecycle.

This page also distinguishes physical snippet extraction from the existing ALM metadata-window tutorial so users can choose the correct workflow.

## Validation

- `fern check` — 0 errors (103 existing warnings)
- `fern docs broken-links` — no errors in changed pages; 22 pre-existing API-reference errors elsewhere
- documented full-run argument set parsed successfully against `_build_parser()`
- focused `tests/stages/audio/alm/pretrain` suite — 123 passed
- `python -m py_compile` for the tutorial runner and all pretraining pipeline modules
- `git diff --cached --check`

The repository enforces Linux at import time. Focused tests were run on this macOS host by bypassing only that package-level guard and excluding the unrelated session-wide Ray cluster fixture; stage code and declared `audio_cpu` dependencies were otherwise used directly.

Closes #2122


lbliii:
@sarahyurick @yqwangustc @mohammadaaftabv friendly ping on this docs PR for the #2118 release-docs workstream — when you have a moment, could you take a look at the long-form audio cutting guide?

Docs preview: https://nvidia-preview-codex-docs-long-form-audio-cutting.docs.buildwithfern.com/nemo/curator

Thanks!

yqwangustc (approved):
Thank you!

fern/versions/main/pages/curate-audio/tutorials/audio-pretrain.mdx
```diff
@@ -0,0 +1,393 @@
...
+  --output-dir /data/pretrain/preview \
+  --output-manifest /data/pretrain/preview/snippets.jsonl \
+  --output-audio-tar-path /data/pretrain/preview/snippets.tar \
+  --metrics-path /data/pretrain/preview/metrics.json \
+  --tokenizer-path openai/whisper-large-v3 \
+  --max-duration-sec 30 \
+  --dry-run
+```
+
+The preview manifest's `audio_filepath` still contains the tar member name that a real run would create, but `snippets.tar` does not exist. Preview duration is the planned `end - start`; a real run records the resampled frame-count duration, which can differ by at most one target-rate frame.
```
mohammadaaftabv, line 136: Dry-run can reconcile the new preview manifest against a stale tar.

This sentence holde true only when `--output-audio-tar-path` did not already exist. `prepare_audio_pretrain_outputs()` removes only `*.shard-*` files, and dry-run extraction creates no new tar shard. The CLI nevertheless always calls `finalize_audio_pretrain_outputs()`: `_merge_tar_shards()` finds no new shards and preserves any previous final tar, after which `_reconcile_manifest_with_tar()` filters the newly generated preview manifest against that stale archive.

Reusing an output path from an earlier full run can therefore silently remove preview rows whose member names are absent, or validate matching rows against old audio.

Please make finalization dry-run-aware so it still merges manifest/metrics shards but skips tar merging and reconciliation, or reject an existing tar path in dry-run mode. Please also add a regression test covering a dry run with a pre-existing tar. At minimum, the guide should explicitly require a fresh, nonexistent tar path.

fern/versions/main/pages/curate-audio/tutorials/audio-pretrain.mdx
```diff
@@ -0,0 +1,393 @@
...
+        {"word": "meeting.", "start": 18.1, "end": 18.8}
+      ]
+    }
+  ]
+}
+```
+
+| Field | Requirement | Behavior |
+| --- | --- | --- |
+| `id` | Required, nonempty, unique | Forms snippet IDs and per-source metrics keys. Missing, empty, and duplicate IDs are skipped with a warning. Numeric IDs are converted to strings. |
```
mohammadaaftabv: Source-unique IDs can collide after tar-key normalization.

This table says IDs need only be unique in their original form, but the reader checks uniqueness before `make_snippet_id()` replaces `.`, `/`, and `\` with `_`. Distinct documented-valid IDs such as `session.1` and `session_1` therefore both produce `session_1-0_000-30_000` when their cut times match.

That gives two manifest rows the same `audio_filepath` and creates duplicate tar-member names. Final reconciliation indexes members by name, so one duplicate wins in that index while both rows can survive and subsequently resolve to the same audio payload.

Please validate uniqueness after applying the same source-ID normalization used by `make_snippet_id()`, or use a collision-free escaping/hash scheme. Please add regression coverage for `session.1` versus `session_1` and path-separator variants, and update this requirement to state that IDs must remain unique after tar-key normalization.

fern/versions/main/pages/curate-audio/tutorials/audio-pretrain.mdx
```diff
@@ -0,0 +1,393 @@
...
+| `audio_filepath` | Required by default | Resolved according to `--audio-path-resolution`. Change the key with `--audio-filepath-key`. |
+| `segments` | Required for useful output | Each segment needs numeric `start` and `end`. Input order does not matter; filtering sorts by `(start, end)`. |
```
mohammadaaftabv: A missing `segments` key fails validation; it does not merely produce no useful output.

`ReadLongFormManifestStage` checks `id` and the audio path but still emits rows without `segments`. The next stage declares `segments` in `OverlapFilterStage.inputs()`, so the base `process_batch()` raises `ValueError` before `process()` can treat the value as an empty list. A single such row can therefore fail the pipeline.

This should distinguish a missing key from an explicitly empty list: `segments: []` passes validation and safely produces a zero-output stub for metrics.

Please document `segments` as runtime-required. It would also be better to validate it in the manifest reader so malformed rows can be rejected with their JSONL line number.

```suggestion
| `segments` | Required | The key must be present; a missing key fails `OverlapFilterStage` input validation. An empty list is accepted and produces zero snippets. Each segment needs numeric `start` and `end`; input order does not matter because filtering sorts by `(start, end)`. |

File: fern/versions/main/pages/curate-audio/tutorials/audio-pretrain.mdx [hunk 4]
```diff
@@ -0,0 +1,393 @@
...
+  "output_total_duration_sec": 23840.4,
+  "dropped": {
+    "empty": 20,
+    "overlap": 180,
+    "too_long": 4,
+    "too_short": 12,
+    "no_text": 8,
+    "repetition": 6,
+    "missing_audio": 1,
+    "corrupted_audio": 0
```
mohammadaaftabv: The example presents sparse reconciliation counters as though zero-valued keys are always emitted.

The worker summary does not initialize `missing_audio` or `corrupted_audio`. `_patch_metrics_post_reconcile()` adds each key only when its corresponding count is nonzero; the tests explicitly assert that `corrupted_audio` is absent when only missing members were dropped, and vice versa.

With the values shown here (`missing_audio: 1`, `corrupted_audio: 0`), the actual JSON contains `missing_audio` but omits `corrupted_audio`. This example may lead consumers to index the key directly and receive a `KeyError`.

Please either remove the zero-valued key and document that these counters are sparse (consumers should use `.get(key, 0)`), or change the runtime schema to always initialize both counters and update the tests accordingly. Line 368 should also clarify that each key appears only when its respective reconciliation count is nonzero.

mohammadaaftabv (changes requested):

lbliii:
Thanks for the detailed review. This PR is docs-only (we handle documentation, not the stage code), so I've updated the guide to accurately document the current behavior and added the "at minimum" guidance you called out. The underlying behavior changes are flagged below as follow-up engineering work.

### Fixed in docs (commit b20eb2cd8)

- **ID collision after normalization** — The `id` field now states IDs must be unique *after tar-key normalization*, and explains that `make_snippet_id` replaces `.`, `/`, and `\` with `_`, so `session.1` and `session_1` collide into the same tar member name.
- **Missing `segments` key** — Marked `segments` as **Required**: a missing key fails `OverlapFilterStage` input validation (can fail the pipeline), while an empty list `[]` is accepted and yields zero snippets. Applied your suggested wording.
- **Sparse reconciliation counters** — Removed the misleading `"corrupted_audio": 0` from the example metrics JSON, documented that `missing_audio`/`corrupted_audio` are written only when nonzero, and recommended defensive access (`dropped.get("corrupted_audio", 0)`) to avoid `KeyError`.
- **Dry-run vs. stale tar** — Added a warning requiring a fresh, nonexistent `--output-audio-tar-path` for every dry run, explaining that finalization otherwise reconciles the preview manifest against a stale archive (silent row removal / validation against unrelated audio).

The greptile Python-API `audio_filepath_key` and histogram 30–60 bin notes were already addressed in the prior commit.

### Suggested engineering follow-ups (out of scope here — code, not docs)

These change stage behavior and belong in a code PR against `nemo_curator/stages/audio/alm/pretrain/`:

1. **Dry-run-aware finalization** — Have `finalize_audio_pretrain_outputs` merge manifest/metrics shards but skip tar merge + `_reconcile_manifest_with_tar` in dry-run mode (or reject an existing tar path in dry-run). Add a regression test for a dry run against a pre-existing tar.
2. **Uniqueness after normalization** — Validate ID uniqueness *after* applying the same `make_snippet_id` normalization (or use a collision-free escaping/hash scheme). Add regression coverage for `session.1` vs `session_1` and path-separator variants.
3. **Reader-side `segments` validation** — Reject rows with a missing `segments` key in `ReadLongFormManifestStage` with their JSONL line number, rather than deferring the failure to `OverlapFilterStage`.
4. **Always-initialize counters (optional)** — Initialize `missing_audio`/`corrupted_audio` to `0` in the runtime schema and update the tests, if a dense schema is preferred over the current sparse one.

Happy to open a tracking issue for these if that's useful.

lbliii:
ok to test b20eb2cd8e9ec2b5eeb88f9a8e356780328888fb

lbliii:
ok to test dcfcc1ff5de5cf0def96a79da425fa20ac0dc23a

lbliii:
@mohammadaaftabv thanks again for the detailed review — this is now ready for another pass. Since the PR is docs-only, I updated the guide to accurately describe the current behavior and added the "at minimum" guidance you flagged (commit dcfcc1ff5). Your five inline points (dry-run vs. stale tar reconciliation, ID collision after `make_snippet_id()` normalization, missing `segments` validation, and the sparse reconciliation counters) are all addressed in the docs, with the underlying stage-behavior items called out as follow-up engineering work. Could you re-review and, if it looks good, clear the change request? @yqwangustc has already approved.

mohammadaaftabv:
> @mohammadaaftabv thanks again for the detailed review — this is now ready for another pass. Since the PR is docs-only, I updated the guide to accurately describe the current behavior and added the "at minimum" guidance you flagged (commit [dcfcc1f](https://github.com/NVIDIA-NeMo/Curator/commit/dcfcc1ff5de5cf0def96a79da425fa20ac0dc23a)). Your five inline points (dry-run vs. stale tar reconciliation, ID collision after `make_snippet_id()` normalization, missing `segments` validation, and the sparse reconciliation counters) are all addressed in the docs, with the underlying stage-behavior items called out as follow-up engineering work. Could you re-review and, if it looks good, clear the change request? @yqwangustc has already approved.

Thanks for addressing the feedback. The guide now accurately documents all four edge cases. Runtime improvements can be handled as follow-up engineering work.

mohammadaaftabv (approved):

lbliii:
ok to test 16f662223921eb444604b39f799d69987e2d6233

PR #2156, opened 2026-07-02
fix: validate pyarrow columns in stage input checks

### LINKED ISSUE #2151: ProcessingStage input validation fails when the input task is DocumentBatch and the data type is pyarrow
**Describe the bug**

The check in validate_input per stage: 
```
        for attr in required_data_attrs:
            if not hasattr(task.data, attr):
                missing_data_attrs.append(attr)
``` 
w.r.t required attirs in the data object fails for pyarrow tables since columns don't pass this check

Please list *minimal* steps or code snippet for us to be able to reproduce the bug.

A helpful guide on on how to craft a minimal bug report http://matthewrocklin.com/blog/work/2018/02/28/minimal-bug-reports.

**Expected behavior**

**Additional context**

nightcityblade:
## Description
closes #2151

- teach `ProcessingStage.validate_input()` to recognize column names on table-like task data instead of relying only on `hasattr()`
- cover the regression with focused `DocumentBatch` tests using PyArrow tables

## Usage
```python
from nemo_curator.tasks import DocumentBatch
import pyarrow as pa

batch = DocumentBatch(task_id="batch", dataset_name="dataset", data=pa.table({"text": ["hello"]}))
assert stage.validate_input(batch)
```
## Checklist

Validation run locally:
- `python3 -m ruff check nemo_curator/stages/base.py tests/stages/common/test_base.py`
- `python3 -m pytest --noconftest tests/stages/common/test_base.py -q` *(blocked in this environment: `pyarrow` is not installed; the full test setup also requires `ray` via `tests/conftest.py`)*


nemo_curator/stages/base.py
```diff
@@ -202,6 +202,23 @@ def validate_input(self, task: Task) -> bool:
 
         return not missing_top_level_attrs and not missing_data_attrs
 
+    @staticmethod
+    def _has_data_attr(data: object, attr: str) -> bool:
+        """Return whether a task data object exposes the requested field/column."""
+        if hasattr(data, attr):
+            return True
+
+        if hasattr(data, "column_names"):
```
sarahyurick: Should it look more like
```suggestion
        if isinstance(data, pa.Table) and hasattr(data, "column_names"):
```
?

nemo_curator/stages/base.py
```diff
@@ -202,6 +202,23 @@ def validate_input(self, task: Task) -> bool:
...
+            return attr in data.column_names
+
+        if hasattr(data, "columns"):
```
sarahyurick:
```suggestion
        if isinstance(data, pd.DataFrame) and hasattr(data, "columns"):
```
?

nemo_curator/stages/base.py
```diff
@@ -202,6 +202,23 @@ def validate_input(self, task: Task) -> bool:
...
+            return attr in data.columns
+
+        if hasattr(data, "schema") and hasattr(data.schema, "names"):
```
sarahyurick:
```suggestion
        if isinstance(data, pa.Table) and hasattr(data, "schema") and hasattr(data.schema, "names"):
```
?

nemo_curator/stages/base.py
```diff
@@ -202,6 +202,23 @@ def validate_input(self, task: Task) -> bool:
...
+            return attr in data.schema.names
```
sarahyurick: Is this block supposed to cover the empty table case? Like what conditions would you expect for it to get all the way down here?

tests/stages/common/test_base.py
```diff
@@ -328,6 +329,38 @@ def test_num_workers_override_accepts_none_as_explicit_override(self):
...
+
+    def inputs(self) -> tuple[list[str], list[str]]:
+        return [], ["text"]
+
+
+class TestProcessingStageValidateInput:
+    def test_validate_input_accepts_pyarrow_columns(self):
+        stage = RequiredColumnStage()
+        batch = DocumentBatch(
+            task_id="batch",
```
sarahyurick: Make sure you have the latest version of main:
```suggestion
```
we no longer pass `task_id` directly to create a `DocumentBatch`.

tests/stages/common/test_base.py
```diff
@@ -328,6 +329,38 @@ def test_num_workers_override_accepts_none_as_explicit_override(self):
...
+            dataset_name="dataset",
+            data=pa.table({"text": ["hello"]}),
+        )
+
+        assert stage.validate_input(batch) is True
+
+    def test_validate_input_rejects_missing_pyarrow_columns(self):
+        stage = RequiredColumnStage()
+        batch = DocumentBatch(
+            task_id="batch",
```
sarahyurick:
```suggestion
```

tests/stages/common/test_base.py
```diff
@@ -265,6 +266,38 @@ def process(self, task: MockTask) -> MockTask:
...
+            dataset_name="dataset",
+            data=pa.table({"other": ["hello"]}),
+        )
+
+        assert stage.validate_input(batch) is False
```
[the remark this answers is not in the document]
sarahyurick: +1 let's try to make sure the test coverage here reaches every possible block of `_has_data_attr` please.

sarahyurick (changes requested):

nightcityblade:
Pushed a follow-up commit for the review notes. I narrowed the fallback checks to concrete  /  cases, updated the  construction to match current , and added coverage for the pandas path plus the explicit-attribute/unknown-attribute branches.\n\nLocal verification: All checks passed! passed. I also attempted the focused test module, but the local environment here is missing repo test dependencies ( from , plus package deps needed to import  end-to-end), so I wasn’t able to run that pytest target locally before pushing.

nightcityblade:
Pushed a follow-up commit for the review notes. I narrowed the fallback checks to concrete `pa.Table` / `pd.DataFrame` cases, updated the `DocumentBatch` construction to match current `main`, and added coverage for the pandas path plus the explicit-attribute/unknown-attribute branches.

Local verification: `python3 -m ruff check nemo_curator/stages/base.py tests/stages/common/test_base.py` passed. I also attempted the focused test module, but the local environment here is missing repo test dependencies (`ray` from `tests/conftest.py`, plus package deps needed to import `nemo_curator` end-to-end), so I wasn’t able to run that pytest target locally before pushing.

nemo_curator/stages/base.py
sarahyurick: Bumping this comment. @nightcityblade should this condition be removed or is there a case where the function could enter this block?

nightcityblade:
Addressed the latest review note: I removed the unreachable  fallback and pushed a follow-up commit, since for  we already return through the  branch.\n\nLocal verification:  passed. I also retried 
no tests ran in 0.00s, but the local environment here still fails at import time because  is not installed via .

nightcityblade:
Addressed the latest review note: I removed the unreachable schema.names fallback and pushed a follow-up commit, since for pa.Table we already return through the column_names branch.

Local verification: python3 -m ruff check nemo_curator/stages/base.py tests/stages/common/test_base.py passed. I also retried pytest tests/stages/common/test_base.py -q, but the local environment here still fails at import time because ray is not installed via tests/conftest.py.

nightcityblade:
Pushed a follow-up force-push to add the missing DCO sign-off to the latest commit, so the DCO check should be able to clear on the updated head.

PR #2170, opened 2026-07-07
feat(audio): agent-ready foundation — StageContract mixin + residency resolver

shubhamNvidia:
## Summary
Shared helpers imported by the audio stage modules.

- `_agent_ready.py` — `AgentReady` mixin + `StageContract`/`IOSpec`/`Gates`/`Role`: a read-only, additive discovery contract exposed via `describe()`. No runtime behavior change.
- `_residency.py` — input-residency resolver (`resolve_audio`, `resolve_audio_path`, `produce_audio_filepath`, `cleanup_temp_files`) unifying file/waveform/auto audio input handling and temp-file lifecycle.
- `common.py` — shared audio helpers (`ensure_mono`, `ensure_waveform_2d`) used by the stage modules.

## Notes for reviewers
- Additive and backward compatible: new stage knobs default to prior behavior; `describe()` is read-only and does not change execution.

### DIFF
--- a/nemo_curator/stages/audio/common.py
+++ b/nemo_curator/stages/audio/common.py
@@ -25,6 +25,7 @@ from fsspec.core import url_to_fs
 from loguru import logger
 
 from nemo_curator.backends.base import NodeInfo, WorkerMetadata
+from nemo_curator.stages.audio._agent_ready import AgentReady, Gates, IOSpec, StageContract
 from nemo_curator.stages.base import CompositeStage, ProcessingStage
 from nemo_curator.stages.file_partitioning import FilePartitioningStage
 from nemo_curator.tasks import AudioTask, EmptyTask, FileGroupTask
@@ -41,7 +42,7 @@ def get_audio_duration(audio_filepath: str) -> float:
 
 
 @dataclass
-class GetAudioDurationStage(ProcessingStage[AudioTask, AudioTask]):
+class GetAudioDurationStage(AgentReady, ProcessingStage[AudioTask, AudioTask]):
     """Compute audio duration from the file at *audio_filepath_key* and
     store the result under *duration_key*.
 
@@ -65,6 +66,12 @@ class GetAudioDurationStage(ProcessingStage[AudioTask, AudioTask]):
     def outputs(self) -> tuple[list[str], list[str]]:
         return [], [self.duration_key]
 
+    def describe(self) -> StageContract:
+        return StageContract(
+            reads=IOSpec(data_keys=[self.audio_filepath_key], accepts=["file"]),
+            writes=IOSpec(data_keys=[self.duration_key]),
+        )
+
     def process(self, task: AudioTask) -> AudioTask:
         t0 = time.perf_counter()
         audio_filepath = task.data[self.audio_filepath_key]
@@ -74,7 +81,7 @@ class GetAudioDurationStage(ProcessingStage[AudioTask, AudioTask]):
         return task
 
 
-class PreserveByValueStage(ProcessingStage[AudioTask, AudioTask]):
+class PreserveByValueStage(AgentReady, ProcessingStage[AudioTask, AudioTask]):
     """Filter entries by comparing *input_value_key* against *target_value*.
 
     Returns ``None`` from ``process()`` to drop entries that fail the
@@ -87,6 +94,7 @@ class PreserveByValueStage(ProcessingStage[AudioTask, AudioTask]):
     """
 
     name: str = "PreserveByValueStage"
+    BATCH_ONLY = True  # process() raises; only process_batch is implemented (agent-discovery hint)
 
     def __init__(
         self,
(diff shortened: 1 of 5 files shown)


nemo_curator/stages/audio/common.py
```diff
@@ -65,6 +66,12 @@ def inputs(self) -> tuple[list[str], list[str]]:
     def outputs(self) -> tuple[list[str], list[str]]:
         return [], [self.duration_key]
 
+    def describe(self) -> StageContract:
+        return StageContract(
+            reads=IOSpec(data_keys=[self.audio_filepath_key], accepts=["file"]),
+            writes=IOSpec(data_keys=[self.duration_key]),
+        )
```
sarahyurick, line 73: Why not just have `AgentReady` read the `inputs` and `outputs` instead of having to repeat code here?

nemo_curator/stages/audio/common.py
```diff
@@ -177,9 +192,16 @@ def ray_stage_spec(self) -> dict[str, Any]:
     def num_workers(self) -> int | None:
         return 1
 
+    def describe(self) -> StageContract:
+        return StageContract(
+            writes=IOSpec(data_keys=["audio_filepath"]),
+            cardinality="1:N fan-out",
```
sarahyurick, line 198: Similar comment here, `AgentReady` can just check the `ray_stage_spec`. Instead of having to manually do it for every audio stage.

nemo_curator/stages/audio/common.py
shubhamNvidia: Fair, the keys here do duplicate inputs()/outputs(). describe() is kept separate because StageContract expresses things those hooks can't: residency (accepts=["file"|"waveform"]), alternative input shapes (reads_one_of), per-segment writes (segment_data_keys), cardinality, and gates. Also inputs()/outputs() default to ([], []) across the framework, so most stages don't populate them reliably — for this stage the only thing beyond them is accepts=["file"].

nemo_curator/stages/audio/common.py
shubhamNvidia: ray_stage_spec is documented as Ray-Data-backend-only, and is_fanout_stage is just a boolean; it can't express the other cardinalities the contract needs (filter, N:1 aggregate, 1:1 nested-list), so it isn't a complete source for cardinality. That's why it's stated explicitly in describe(), which is the single backend-agnostic place that captures the full shape.

mohammadaaftabv (changes requested):
Could we include one representative, end-to-end audio stage migration in this PR to demonstrate how this foundation is actually consumed?

`GetAudioDurationStage.describe()` shows how a contract is declared, but the PR still leaves it unclear how the complete path is intended to work: which contract fields are inferred versus manually authored, how an agent or registry consumes the contract, and how the new residency helpers participate in execution.

I’m not asking to migrate every audio stage here. One small vertical slice would be enough—preferably an existing stage that can accept either `audio_filepath` or `waveform`. The example should show:

1. The stage inheriting `AgentReady`.
2. Its `StageContract`, including reads/writes, accepted audio forms, dispatch, gates, and configurable keys.
3. How the supported discovery path obtains that contract.
4. How the consumer uses the contract to choose file versus waveform execution.
5. A focused test covering both input forms and any temporary-file cleanup.

This would give later audio-stage PRs a canonical implementation to copy and would validate that `_agent_ready.py` and `_residency.py` work together as intended, rather than merging two foundational APIs whose integration is demonstrated only in later PRs. If the registry/orchestrator must remain outside this PR, a minimal test consumer or documented pseudocode for that boundary would still make the intended contract much clearer.

shubhamNvidia:
@mohammadaaftabv 
I pushed the vertical slice to this PR. MonoConversionStage is now migrated here (dependency-light: soundfile/torch only), along with tests/stages/audio/test_agent_ready_mono_slice.py and a minimal in-test consumer. It exercises the full path we discussed:

Contract: the stage inherits AgentReady and its describe() declares reads/writes (IOSpec), accepts=["file","waveform"], cardinality, gates (writes_to_disk when write_to_disk=True), and the configurable *_key fields.
Residency in execution: process() calls resolve_audio(item, residency=self.input_residency, ...), which returns the in-memory waveform when present and otherwise loads the file — that's the file-vs-waveform branch. The temp-WAV path (resolve_audio_path + register_temp/cleanup_temp_files) is covered too.
Consumer/discovery boundary: choose_input_form() in the test reads stage.describe(), inspects reads.accepts, and picks execution accordingly — standing in for the registry, with a short pseudocode block showing the real build_contract(stage) path.
Test: the same stage is run on audio_filepath and on waveform+sample_rate (asserting identical results), plus a case asserting the temp file is created and then removed. Green on GPU nodes.
Dispatch: test_dispatch_is_per_item asserts the stage is not BATCH_ONLY (per-item process() is usable); the concrete dispatch field is registry-derived.
On inferred vs. authored contract fields: describe() authors only the shape — reads/writes/reads_one_of, accepts, cardinality, gates. The rest is inferred by build_contract/static_contract in the orchestrator PR: params (from the constructor fields), key_roles (from the *_key names), and dispatch/batch_only (from batch support). The test's pseudocode documents that boundary so the intended contract is clear without merging the orchestrator early.

This gives later audio-stage PRs a reference implementation to copy and validates that _agent_ready.py and _residency.py work together within this PR. (I also moved mono_conversion.py out of the preprocessing PR #2172 so it isn't duplicated.)

shubhamNvidia:

PR #2171, opened 2026-07-07
feat(audio-io): make IO stages agent-ready (contract + serialization-safe sink)

shubhamNvidia:
## Summary
Makes the audio IO stages agent-ready (they now implement `describe()`) and hardens the document sink.

### `io/convert.py` (`AudioToDocumentStage`) — N:1 sink, AudioTask → DocumentBatch
- Now configurable via constructor: `keep_keys` (whitelist of columns to emit; `None` = all, prior behavior), `drop_keys` (blacklist), `serialize_segments` + `segments_key`.
- `segments` (a list of per-segment dicts that may embed waveform tensors) was previously dropped wholesale. With `serialize_segments=True` it is kept but recursively cleaned by the new `_sanitize_nested` helper, which walks nested dicts/lists and strips `torch.Tensor` / audio-blob values (using a `_DROP_VALUE` sentinel so a legitimate `None` is preserved).
- `describe()` sets `Gates(sanitizes_output=True)` to mark this as the tensor-safe boundary before a JSON/parquet writer.

### `io/extract_segments.py` (`SegmentExtractionStage`)
- Adds `describe()` with OR-shaped reads covering the extraction combos (timestamp-based vs diarization-based). Marked `BATCH_ONLY`: extraction runs per batch and appends to a shared metadata file, so `process()` raises and `process_batch()` does the work.

## Notes for reviewers
- Backward compatible: `keep_keys=None` emits all keys as before; `segments` still dropped unless `serialize_segments=True`.
- Worth a look: `_sanitize` / `_sanitize_nested` in `convert.py` (recursive tensor stripping) and the combo auto-detection in `extract_segments.py`.

### DIFF
--- a/nemo_curator/stages/audio/_agent_ready.py
+++ b/nemo_curator/stages/audio/_agent_ready.py
@@ -1,0 +1,318 @@
+# Copyright (c) 2026, NVIDIA CORPORATION.  All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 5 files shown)


shubhamNvidia:

PR #2172, opened 2026-07-07
feat(audio-preprocessing): agent-ready mono-conversion + segment concatenation

shubhamNvidia:
## Summary
Makes the preprocessing stages agent-ready and lets mono-conversion accept in-memory audio.

### `preprocessing/mono_conversion.py`
- Loads audio through the shared residency resolver, so it accepts either an in-memory `waveform`+`sample_rate` or an `audio_filepath` (`input_residency` = file / waveform / auto).
- Optional `write_to_disk` (write the mono WAV to a temp file) and `update_audio_filepath` (repoint `audio_filepath`, preserving the original under `original_audio_filepath`). Always annotates `is_mono`, `duration`, `num_samples`.
- Adds `describe()` / `StageContract`.

### `preprocessing/concatenation.py` (`SegmentConcatenationStage`) — N:1
- Concatenates a task's `segments` into one waveform and emits `segment_mappings` metadata: for each segment, its `[concat_start_ms, concat_end_ms]` position in the joined audio and the corresponding `[original_start_ms, original_end_ms]`. This is what lets the timestamp mapper translate results back to the original timeline.

## Notes for reviewers
- Backward compatible at defaults (`write_to_disk=False`, `update_audio_filepath=False` → same in-memory result as before).
- Residency default `"auto"` prefers an in-memory `waveform` over re-reading the file when both are present; pass `input_residency="file"` to force a reload.
- Worth a look: the `segment_mappings` construction (ms math, inter-segment offsets) in `concatenation.py`.

## Test plan
- [ ] `pytest tests/stages/audio/preprocessing -q`

### DIFF
--- a/nemo_curator/stages/audio/_agent_ready.py
+++ b/nemo_curator/stages/audio/_agent_ready.py
@@ -1,0 +1,318 @@
+# Copyright (c) 2026, NVIDIA CORPORATION.  All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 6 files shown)


shubhamNvidia:

PR #2173, opened 2026-07-07
feat(audio-postprocessing): agent-ready timestamp mapper (refined-span + multi-speaker fixes)

shubhamNvidia:
## Summary
Makes `TimestampMapperStage` agent-ready and fixes two timestamp-correctness bugs.

### What the stage does
Translates per-segment times measured on concatenated/processed audio back to the original source timeline, using the `segment_mappings` metadata produced upstream (see `_translate_to_original`).

### Fixes
- Refined-span preference ("Failure 1"): when a segment carries refined `start_ms`/`end_ms` (e.g. from a VAD pass) AND coarser `diar_segments`, the refined values are now mapped through and preferred; diarization spans are used only as a fallback when refined times are absent.
- Multi-speaker: each speaker's segment maps to its own distinct window instead of collapsing to a shared/union span (no cross-speaker assignment).

### Contract
- `describe()` declares OR-shaped reads (`start_ms`+`end_ms` | `diar_segments` | `duration`) and `metadata_reads=[segment_mappings]`.

## Notes for reviewers
- Behavior change to be aware of: the stage sets `preserves_upstream_keys=False` and REPLACES `task.data`, keeping only a `passthrough_keys` whitelist (defaults keep filter scores + speaker metadata like `speaker_id`). Non-whitelisted keys (e.g. `text`, `segment_num`) are dropped — set `passthrough_keys` to retain them.
- Worth a look: `_translate_to_original` (overlap/offset math, ms units) and the refined-vs-diar branch.

## Test plan
- [ ] `pytest tests/stages/audio/postprocessing -q`
- [ ] `pytest tests/stages/audio -q -k timestamp_mapper`

### DIFF
--- a/nemo_curator/stages/audio/_agent_ready.py
+++ b/nemo_curator/stages/audio/_agent_ready.py
@@ -1,0 +1,318 @@
+# Copyright (c) 2026, NVIDIA CORPORATION.  All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


shubhamNvidia:

PR #2174, opened 2026-07-07
feat(audio-filtering): agent-ready band/sigmos/utmos filters (residency + filter/annotate)

shubhamNvidia:
## Summary
Makes the quality/bandwidth filters agent-ready, unifies their audio loading, and adds an annotate mode.

### All three (`utmos.py`, `sigmos.py`, `band.py`)
- Now subclass `AgentReady` with a `describe()` contract, and gain configurable data keys (`audio_filepath_key` / `waveform_key` / `sample_rate_key` / `segments_key` / score keys) plus `input_residency`.
- New `mode` (`task` / `segments` / `auto`) selects whether to score the top-level audio or each item in a nested `segments` list (`auto` = segments if present).
- New `action` (`filter` / `annotate`): `filter` (default) drops items that fail the threshold; `annotate` keeps every item and only writes the score/prediction.
- utmos/sigmos now load audio via the shared `resolve_audio` resolver instead of their own inline loaders — behavior preserved for the default path, including the "waveform present but no `sample_rate`" (skip) and "file load error → return None" cases.

## Notes for reviewers
- Backward compatible: `action="filter"`, `mode="auto"`; thresholds and score keys unchanged by default.
- Residency default `"auto"` prefers in-memory waveform over file when both exist.
- Worth a look: `band.py` catches `(OSError, RuntimeError)` on load (previously broader) — confirm that is the intended failure surface.

## Test plan
- [ ] `pytest tests/stages/audio/filtering -q`

### DIFF
--- a/nemo_curator/stages/audio/_agent_ready.py
+++ b/nemo_curator/stages/audio/_agent_ready.py
@@ -1,0 +1,318 @@
+# Copyright (c) 2026, NVIDIA CORPORATION.  All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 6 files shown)


shubhamNvidia:

PR #2175, opened 2026-07-07
feat(audio-metrics): agent-ready WER/SQUIM/bandwidth metrics (contract hygiene)

shubhamNvidia:
## Summary
Makes the metrics stages agent-ready and corrects the WER read contract.

- `metrics/wer.py` (`ComputeWERStage`) — `describe()` now declares BOTH supported input shapes (it previously under-declared): per-segment WER over a `segments` list, OR top-level WER over `hypothesis`+`reference` text keys on the record. This matches what `validate_input` / `process` already accept. No metric-math change.
- `metrics/squim.py` (`TorchSquimQualityMetricsStage`), `metrics/bandwidth.py` (`BandwidthEstimationStage`) — add `describe()` declaring filepath-based reads (plus `segments` / `duration` where used) and writing results under the `metrics` dict. `squim` is `BATCH_ONLY` (GPU/batched — `process()` raises, `process_batch()` runs).

## Notes for reviewers
- These are annotate-only (they compute metrics, they don't filter); backward compatible defaults.
- Worth a look: that each `describe()` matches the keys the stage actually reads/writes.

## Test plan
- [ ] `pytest tests/stages/audio/metrics -q`

### DIFF
--- a/nemo_curator/stages/audio/_agent_ready.py
+++ b/nemo_curator/stages/audio/_agent_ready.py
@@ -1,0 +1,318 @@
+# Copyright (c) 2026, NVIDIA CORPORATION.  All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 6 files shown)


shubhamNvidia:

PR #2176, opened 2026-07-07
feat(audio-inference): agent-ready ASR/VAD/diarization (residency + fan-out provenance fix)

shubhamNvidia:
## Summary
Makes the ASR/VAD/diarization stages agent-ready, routes their audio loading through the residency resolver, and fixes fan-out provenance.

### Common changes
- Each stage loads its input path via `resolve_audio_path` (accepts a file, or an in-memory waveform materialized to a temp WAV) and cleans up temp files in a `finally` block. `input_residency="file"` by default → same behavior as before. Each adds `describe()`; batched stages (e.g. ASR) are `BATCH_ONLY`.

### Fan-out provenance fix (pyannote, sortformer, whisperx_vad)
- These emit one child record per detected segment/speaker. Each child now resolves `original_file` through a fallback chain — configured `original_file_key` → `audio_filepath_key` → canonical `audio_filepath` → legacy `resampled_audio_filepath` → `"unknown"` — instead of collapsing to `"unknown"` when audio arrived under the canonical key. Children also get a distinct `segment_num` and per-segment timing (`start`/`end`, `start_ms`/`end_ms`, `duration`).
- WhisperX specifically drops its raw internal `segments` key from children (a list of tuples) so it doesn't collide with the semantic `segments` role and break dict-shaped consumers.
- sortformer additionally avoids pinning a materialized temp path onto its children.

## Notes for reviewers
- Worth a look: the `_segment_child_data` / `_fanout_segments` helpers in `pyannote.py`, `sortformer.py`, `whisperx_vad.py` (provenance fallback + temp handling) — these are near-identical across the three.

## Test plan
- [ ] `pytest tests/stages/audio/inference -q`

### DIFF
--- a/nemo_curator/stages/audio/_agent_ready.py
+++ b/nemo_curator/stages/audio/_agent_ready.py
@@ -1,0 +1,318 @@
+# Copyright (c) 2026, NVIDIA CORPORATION.  All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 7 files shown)


shubhamNvidia:

PR #2177, opened 2026-07-07
feat(audio-segmentation): agent-ready speaker-separation + VAD (provenance fix)

shubhamNvidia:
## Summary
Makes the segmentation stages agent-ready and fixes speaker-separation provenance.

### `segmentation/speaker_separation.py` (1:N fan-out) — provenance fix
- Splits one multi-speaker input into one record per speaker. Previously each child inherited the parent `audio_filepath` (the full mixed file) and carried no source provenance. Now each child drops the parent file path and snapshots `original_file` (from `original_file` / `audio_filepath`) before emission, and carries a distinct `speaker_id` plus per-speaker `waveform` / `diar_segments`.

### `segmentation/vad_segmentation.py`
- Residency-aware VAD (accepts waveform or file), with `fan-out` (one task per segment) or `nested` (all segments under one key) modes, configurable keys, and `describe()`.

## Notes for reviewers
- Backward compatible defaults; residency default `"auto"` prefers in-memory waveform.
- Worth a look: `_build_speaker_tasks` in `speaker_separation.py` (which parent keys are dropped vs inherited).

## Test plan
- [ ] `pytest tests/stages/audio/segmentation -q`

### DIFF
--- a/nemo_curator/stages/audio/_agent_ready.py
+++ b/nemo_curator/stages/audio/_agent_ready.py
@@ -1,0 +1,318 @@
+# Copyright (c) 2026, NVIDIA CORPORATION.  All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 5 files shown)


shubhamNvidia:

PR #2178, opened 2026-07-07
feat(audio-tagging): agent-ready tagging stages (resample temp-path fix + configurable keys)

shubhamNvidia:
## Summary
Makes the tagging stages agent-ready and fixes temp-path safety in resampling.

### `tagging/resample_audio.py` — temp-path fix
- When materializing/producing audio, temp files now use unique `mkstemp` paths registered via `register_temp` and removed by `cleanup_temp_files`, so parallel workers / repeated calls no longer collide on a shared temp path. Residency-aware input; optional in-memory output (`write_to_disk`), with the original path preserved when not updating.

### Others (contract + configurable keys, logic unchanged)
- `nemo_asr_align.py`, `merge_alignment_diarization.py`, `prepare_module_segments.py`, `split.py`, `text/itn.py`, `text/chinese_conversion.py` — add `describe()` / `StageContract` and replace hardcoded data-key strings with configurable fields.

## Notes for reviewers
- Backward compatible defaults. `NeMoASRAlignerStage.process()` delegates to `process_batch([task])` (works per item, so intentionally no `BATCH_ONLY`).
- Worth a look: the temp-file lifecycle in `resample_audio.py` (creation + cleanup on both success and error).

## Test plan
- [ ] `pytest tests/stages/audio/tagging -q`

### DIFF
--- a/nemo_curator/stages/audio/_agent_ready.py
+++ b/nemo_curator/stages/audio/_agent_ready.py
@@ -1,0 +1,318 @@
+# Copyright (c) 2026, NVIDIA CORPORATION.  All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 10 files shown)


shubhamNvidia:

PR #2182, opened 2026-07-07
Audio conversation pipeline

Ssofja:
## Description

This PR adds an end-to-end **multi-speaker conversation data-generation pipeline** for audio, wiring together four audio stages into a single runnable tutorial: LLM conversation generation → TTS synthesis → forced alignment → conversation merging.

It integrates three previously separate audio feature branches and adds the glue needed to run them as one pipeline:

- **`vLLMInference` stage** (`nemo_curator/stages/audio/llm/vllm_inference.py`): generates structured multi-turn conversations from topic prompts. It delegates to the existing `nemo_curator.models.VLLMModel` (already on `main`) for engine management/inference rather than creating `vllm.LLM` directly, and emits one `AudioTask` per conversation turn (shared `conversation_id`, `speaker`, `utterance`, `overlap`). Includes JSON-schema validation with batch retry and per-turn expansion.
- **`ChatterboxTTSStage`** (from the `chatterbox_tts` work): synthesizes per-turn audio with voice cloning.
- **`MFAAlignmentStage`** (from the `montreal_forced_aligner` work): Montreal Forced Aligner producing word-level TextGrid → RTTM/CTM.
- **`MergeConversationSDPStage`** (from the `merging_audios` work): merges per-turn audio into multi-speaker conversations (per-speaker WAVs, multichannel, mixed mono, combined RTTM/CTM/seglst).

Supporting changes:

- Registered `MergeConversationSDPStage` and `vLLMInference` in the audio package lazy exports (`nemo_curator/stages/audio/__init__.py`) so both are discoverable via `nemo_curator.stages.audio` without eagerly importing heavy optional deps.
- Added the `tutorials/audio/data-generation/` tutorial: `main.py` (two-phase runner), `pipeline.yaml` (Hydra config), `README.md`, plus 2/3/4-speaker prompt templates and sample topic manifests. Registered the tutorial in `tutorials/audio/README.md`.
- Reuses the existing `vllm` optional extra and the `audio_common` extra (which already carries `chatterbox-tts` and `praatio`); no new dependency declarations were required.

The pipeline runs in two phases: Phase 1 (GPU) runs `ManifestReader → vLLMInference → ChatterboxTTSStage → MFAAlignmentStage` via the standard executor (one `AudioTask` per turn); Phase 2 (CPU) groups turns by `conversation_id`, runs `MergeConversationSDPStage` per conversation, and writes the final manifest.

## Usage

Install the audio + vLLM extras and MFA models, then run the tutorial from the repo root:

```bash
# Install (GPU)
uv sync --extra audio_cuda12 --extra vllm && source .venv/bin/activate

# Download MFA models (one-time)
mfa model download acoustic english_us_arpa
mfa model download dictionary english_us_arpa
mfa model download g2p english_us_arpa

# Run the 2-speaker dialog pipeline
python tutorials/audio/data-generation/main.py \
    --config-path . \
    --config-name pipeline \
    input_manifest=tutorials/audio/data-generation/topics/topics_dialog.jsonl \
    output_dir=/data/conversations \
    reference_voices_dataset=/data/reference_voices \
    prompt_file=tutorials/audio/data-generation/prompts/dialog_prompt.yaml
```

Using the `vLLMInference` stage directly in a pipeline:

```python
from nemo_curator.stages.audio import (
    ChatterboxTTSStage,
    MFAAlignmentStage,
    MergeConversationSDPStage,
    vLLMInference,
)
from nemo_curator.stages.audio.common import ManifestReader

llm = vLLMInference(
    prompt_file="tutorials/audio/data-generation/prompts/dialog_prompt.yaml",
    model={"model": "Qwen/Qwen2.5-7B-Instruct", "max_model_len": 4096, "max_tokens": 4096},
)

pipeline = Pipeline(name="conversation_generation")
pipeline.add_stage(ManifestReader(manifest_path="topics/topics_dialog.jsonl"))
pipeline.add_stage(llm)
pipeline.add_stage(ChatterboxTTSStage(
    output_audio_dir="/data/out/audio",
    reference_voices_dataset="/data/reference_voices",
))
pipeline.add_stage(MFAAlignmentStage(output_dir="/data/out/alignment", text_key="utterance"))
# MergeConversationSDPStage requires all turns of a conversation in one batch;
# see tutorials/audio/data-generation/main.py for the two-phase runner.
```

## Checklist

  - Existing stage unit tests pass (`tests/stages/audio/tts`, `tests/stages/audio/alignment`, `tests/stages/audio/merging` — 58 tests). The `vLLMInference` wrapper and the tutorial runner were verified via import/compile smoke checks; a full end-to-end run requires a GPU plus vLLM/Chatterbox/MFA models and was not run in CI.
  - Added `tutorials/audio/data-generation/README.md` and registered the tutorial in `tutorials/audio/README.md` (tutorial selection, data availability, and system dependency tables).


pyproject.toml
```diff
@@ -112,12 +112,14 @@ audio_common = [
...
     "scipy",
     "pydub>=0.25.1",
+    "praatio>=6.0",
     "transformers",
     "accelerate",
     "nemo_text_processing; platform_machine == 'x86_64' and platform_system != 'Darwin'",  # pulls pynini, which ships x86_64-only wheels and fails to build on aarch64
     "opencc-python-reimplemented",
     "pyannote-audio>=4.0.0; platform_machine == 'x86_64' and platform_system != 'Darwin'",
     "whisperx>=3.8.4; platform_machine == 'x86_64' and platform_system != 'Darwin'",
+    "chatterbox-tts>=0.1.4; platform_machine == 'x86_64' and platform_system != 'Darwin'",
```
mohammadaaftabv, line 122: `pyproject.toml:105-132,336` adds `chatterbox-tts>=0.1.4` to `audio_common`, which is inherited by both audio extras, while Curator requires `transformers>=4.56.0,<5.0`. `uv lock --check` has no solution: Chatterbox 0.1.4–0.1.6 pins Transformers 4.46.3 and 0.1.7 pins 5.2.0. `uv.lock` was not updated and contains neither new package. This is also the exact failure in the current GitHub pre-commit job, so unrelated CPU/GPU audio installs are blocked before tests can run.

Please move Chatterbox into a dedicated compatible runtime/extra or use a compatible pinned release/fork, reconcile the supported Transformers range, regenerate `uv.lock`, and validate clean `audio_cpu`, `audio_cuda12`, and `all` installs for every supported Python split. This repeats the unresolved dependency finding from source PR 1976 (`discussion_r3718274906`).

nemo_curator/stages/audio/alignment/mfa_alignment.py
```diff
@@ -0,0 +1,615 @@
...
+        if self.create_rttm:
+            self._rttm_dir.mkdir(parents=True, exist_ok=True)
+        if self.create_ctm:
+            self._ctm_dir.mkdir(parents=True, exist_ok=True)
+
+    def process(self, task: AudioTask) -> AudioTask:
+        msg = "MFAAlignmentStage only supports process_batch"
+        raise NotImplementedError(msg)
+
+    def process_batch(self, tasks: list[AudioTask]) -> list[AudioTask]:  # noqa: C901, PLR0912
```
mohammadaaftabv, line 220: nemo_curator/stages/audio/alignment/mfa_alignment.py:220-223` uses `if not tasks`, but `nemo_curator/backends/ray_data/adapter.py:61-71` passes Ray's `batch["item"]` numpy array directly. The audio developer guide explicitly requires `len(tasks) == 0` for this reason. The default alignment CLI combines `--backend ray_data` with `--batch-size 256` (`tutorials/audio/alignment/pipeline.py:197-222`), so the documented quick start reaches this on its first multi-row batch. A two-element ndarray reproduced `ValueError: The truth value of an array with more than one element is ambiguous`.

Replace the empty guard with `if len(tasks) == 0` and add a Ray Data regression with at least two tasks. The same invalid guard appears in the three other new batch implementations (`vllm_inference.py:261-270`, `chatterbox_tts.py:566-574`, and `merge_conversation.py:115-128`) and should be fixed together.

nemo_curator/stages/audio/alignment/mfa_alignment.py
```diff
@@ -0,0 +1,615 @@
...
+                    f"Empty text for {audio_filepath} "
+                    f"(key={self.text_key!r})"
+                )
+                raise ValueError(msg)
+            audio_path = Path(audio_filepath)
+            if not audio_path.exists():
+                msg = f"Audio file not found: {audio_path}"
+                raise FileNotFoundError(msg)
+
+            file_stem = audio_path.stem
```
mohammadaaftabv, line 243: `nemo_curator/stages/audio/alignment/mfa_alignment.py:243-255` disambiguates duplicate stems only inside the current batch, while `:302-316` and the fallback path write persistent `rttms/{stem}.rttm` and `ctms/{stem}.ctm`. `/a/foo.wav` and `/b/foo.wav` in separate batches therefore use the same artifact paths. In a two-batch reproduction, both returned tasks named the same RTTM/CTM, and after batch two the first task's path contained batch two's output.

Derive one stable artifact identity from the full source identity/content (or a framework-owned source ID), use it consistently for corpus, TextGrid, RTTM, and CTM names, and test duplicate basenames both within and across batches/workers.

nemo_curator/stages/audio/merging/merge_conversation.py
```diff
@@ -0,0 +1,774 @@
...
+            return []
+
+        speaker_buffers: dict[str, np.ndarray] = {}
+        current_position = 0
+        total_length = 0
+        sample_rate: int | None = None
+        actual_overlaps: list[float] = []
+        sorted_speakers: list[str] = []
+
+        for i, entry in enumerate(conversation_entries):
```
mohammadaaftabv, line 490: nemo_curator/stages/audio/merging/merge_conversation.py:490-514` only warns when a turn's sample rate differs from the first one. It then treats the raw arrays as if they shared one sample clock (`:519-560`) and writes every channel at the first rate (`:579-594`). One valid one-second 8 kHz turn followed by one valid one-second 16 kHz turn produced an 8 kHz output lasting **3.0 seconds**, not 2.0 seconds; the reverse order produces 1.5 seconds. RTTM/CTM placement is consequently wrong too.

Require an explicit merge sample rate and resample every turn before any sample/timeline calculation, or reject mismatches before writing anything. Add both rate orderings to the merge tests.

nemo_curator/stages/audio/llm/vllm_inference.py
```diff
@@ -0,0 +1,316 @@
...
+        return " ".join(cleaned.split()).strip()
+
+    def _is_valid_text(self, text: str) -> bool:
+        """Check if text has actual speakable content after cleaning."""
+        cleaned = self._clean_text(text)
+        if not cleaned or len(cleaned) < _MIN_CLEAN_TEXT_LEN:
+            return False
+        return any(c.isalpha() for c in cleaned)
+
+    def _turns_are_valid(self, turns: list[dict]) -> bool:
```
mohammadaaftabv, line 133: `nemo_curator/stages/audio/llm/vllm_inference.py:133-143` accepts any non-empty speaker string. The merge stage then embeds it verbatim in paths at `nemo_curator/stages/audio/merging/merge_conversation.py:515-581,655-665,713-724`. With a speaker value such as `/tmp/.../escaped`, `conv_dir / f"{speaker}.wav"` becomes that absolute path; the reproduction created the file outside `output_conversations_dir`. The reusable merge stage also uses an input `conversation_id` directly as the directory component at `:375-377`.

Treat both identifiers as untrusted labels: map them to generated safe filenames (while retaining the original labels in metadata), reject path separators/traversal, and assert every resolved output remains beneath the configured root before opening it. Add absolute, `..`, slash, and platform-separator tests. This carries over the unresolved source-PR-2031 path finding (`discussion_r3306150744`).

nemo_curator/stages/audio/merging/merge_conversation.py
```diff
@@ -0,0 +1,774 @@
...
+    # ------------------------------------------------------------------
+
+    def _merge_conversation(  # noqa: C901, PLR0912, PLR0915
+        self,
+        conversation_id: str,
+        turns: list[dict],
+    ) -> dict | None:
+        """Merge all turns into per-speaker outputs under a conversation folder."""
+
+        valid_turns: list[dict] = []
```
mohammadaaftabv, line 357: `nemo_curator/stages/audio/merging/merge_conversation.py:357-377` silently removes missing turns and reuses an existing conversation directory. The RTTM/CTM writers do nothing when the current run has no corresponding inputs, but result assembly at `:434-445` publishes any old `all.rttm`/`all.ctm` that happens to exist. I ran the same conversation once with `old_speaker` RTTM and again with no RTTM; the second task still returned the first run's path and contents. An existing-but-unreadable turn is similarly skipped inside `:490-569`, allowing a declared two-speaker result with only one audio channel (and, with the installed SoundFile version, the later `sf.info` can instead escape as an uncaught exception).

Build each conversation in a fresh temporary directory, validate every required turn and artifact, and atomically replace the final directory only after success. Clear absent optional outputs and construct the manifest from artifacts created by this invocation, not `.exists()` checks against a reused directory. Missing/corrupt required turns should fail or explicitly mark the entire conversation incomplete rather than silently changing its semantics.

nemo_curator/stages/audio/tts/chatterbox_tts.py
```diff
@@ -0,0 +1,628 @@
...
+            return wav
+        current_db = 20 * torch.log10(rms + 1e-8)
+        gain = 10 ** ((self.normalize_level - current_db) / 20)
+        normalised = wav * gain
+        peak = torch.max(torch.abs(normalised))
+        if peak > 1.0:
+            normalised = normalised / peak * 0.99
+        return normalised
+
+    @staticmethod
```
mohammadaaftabv, line 551: `nemo_curator/stages/audio/tts/chatterbox_tts.py:551-559` hashes only conversation, speaker, text, and reference ID. It omits model/revision, language, reference content and preprocessing, effective exaggeration, all sampling parameters, normalization, and sample rate. `:595-608` trusts any existing WAV, discards the rate returned by `sf.read`, and divides its frames by the current configured rate. A 24 kHz one-second cache entry reused under a 16 kHz configuration skipped generation and reported 1.5 seconds while remaining a one-second 24 kHz file.

Hash a canonical, versioned manifest of every output-affecting setting plus the model/revision and reference-content identity; write a sidecar and audio atomically; validate both on hits; and calculate duration using the actual file rate. Add cross-language, cross-rate, changed-reference, changed-normalization, and changed-sampling regressions. This is the unresolved PR-1976 cache finding (`discussion_r3718296036`).

nemo_curator/stages/audio/tts/chatterbox_tts.py
```diff
@@ -0,0 +1,628 @@
...
+        """Get consistent exaggeration for a conversation (random range support)."""
+        if self.exaggeration_range is None:
+            return self.exaggeration
+
+        if conversation_id not in self.conversation_exaggeration:
+            lo, hi = self.exaggeration_range
+            self.conversation_exaggeration[conversation_id] = self._rng.uniform(lo, hi)
+        return self.conversation_exaggeration[conversation_id]
+
+    def _generate_turn_audio(
```
mohammadaaftabv, line 508: nemo_curator/stages/audio/tts/chatterbox_tts.py:508-532,599-608` never asks the model for its native sample rate and never resamples. It simply writes the returned sample array with `self.sample_rate` in the WAV header and divides by that value. A model-free reproduction returning 24,000 native 24 kHz samples under `sample_rate=12000` created a 12 kHz file reported as two seconds, changing playback speed and pitch.

Use the model-reported native rate, then explicitly resample to the requested rate before writing, or remove the configurable rate and write at the model's actual rate. Verify the header, duration, and signal timing for a non-default rate. This repeats PR 1976's unresolved `discussion_r3718298128` finding.

nemo_curator/stages/audio/tts/chatterbox_tts.py
```diff
@@ -0,0 +1,628 @@
...
+        self, text: str, reference_wav: str, conversation_id: str
+    ) -> np.ndarray:
```
mohammadaaftabv, line 510: nemo_curator/stages/audio/tts/chatterbox_tts.py:508-535` catches every inference exception, including GPU OOM, and returns two seconds of zeros. `:595-613` writes that under the ordinary success-cache name and emits normal success metadata. The test at `tests/stages/audio/tts/test_chatterbox_tts.py:358-369` codifies this behavior. A reproduction showed the zero WAV persisted and a second run never called a now-healthy generator.

Propagate an exception or emit an explicit failed task/status, never cache fabricated silence as successful model output, and test that a transient failure is retried. If a fallback asset is required, it must be opt-in, visibly labelled, excluded from the success cache, and filterable downstream. This repeats source PR 1976's `discussion_r3305614605` finding.

nemo_curator/stages/audio/tts/chatterbox_tts.py
```diff
@@ -0,0 +1,628 @@
...
+        else:
+            self.exaggeration_range = None
+            self.exaggeration = float(exaggeration)
+
+        self.model = None
+        self.reference_wavs_list: list[str] | None = None
+        self._reference_layout: str = "wavs"
+        self._speaker_audio_map: dict[str, list[str]] = {}
+
+        self.speaker_to_reference: dict[str, str] = {}
```
mohammadaaftabv, line 178: `nemo_curator/stages/audio/tts/chatterbox_tts.py:178-184,465-505` stores reference and exaggeration assignments in actor-local dictionaries backed by an unseeded RNG. With one GPU and `batch_size=1` per actor, Xenna/Ray can work-steal different turns from the same conversation across actor copies. Two stage instances reproduced different reference IDs (`dialog/b` and `dialog/a`) for the same `(conversation_id, speaker)`, contradicting the stage and tutorial promise that voices remain consistent.

Make assignment stateless and deterministic from conversation ID, speaker, dataset/revision, and an explicit seed; pre-plan assignments before GPU fan-out; or partition complete conversations to one worker. Cover multiple actors, out-of-order turns, and restarts. This is the still-unresolved PR-1976 finding `discussion_r3718281009`.

nemo_curator/stages/audio/llm/vllm_inference.py
mohammadaaftabv, line 133: `nemo_curator/stages/audio/llm/vllm_inference.py:133-164` performs a few hand checks rather than enforcing the schemas shipped in the prompt assets. It accepted an object with a top-level extra field, per-turn extra fields, only one repeated speaker, and overlap `999`. A numeric speaker reaches `.strip()` and raises `AttributeError`, escaping the intended invalid-generation retry path. The prompt assets themselves disagree about legal overlap ranges (`dialog_prompt.yaml:42,48-50`; `fourperson_prompt.yaml:49-52`).

Use one actual JSON Schema/Pydantic contract as the source for both structured generation and validation, including exact speaker membership/count, additional-properties rejection, finite bounded overlaps, and string types. Convert all validation failures into the existing per-response retry/failure path and test malformed types, booleans, NaN/non-finite values, wrong speakers, extras, and each prompt variant.

nemo_curator/stages/audio/llm/vllm_inference.py
```diff
@@ -0,0 +1,316 @@
...
+
+        if self.prompt_file:
+            with open(self.prompt_file) as f:
+                self.prompt_data = yaml.safe_load(f)
+        else:
+            self.prompt_data = None
+
+        self.tokenizer = None
+
+    def generate_conversation_id(self, turns: list[dict]) -> str:
```
mohammadaaftabv, line 107: `nemo_curator/stages/audio/llm/vllm_inference.py:107-112` hashes an unframed concatenation of only speaker and utterance. It omits overlap, topic, source identity, and turn boundaries. Distinct generated conversations with the same words but overlaps `0` and `0.7` receive the same ID; duplicate inputs or repeated model outputs do too. `tutorials/audio/data-generation/main.py:87-95` then combines every such turn into a single conversation group.

Hash canonical JSON containing the complete turn representation and a stable source/topic identity (or a deliberate per-input generation ID), and test identical text with different timing and duplicate source rows.

nemo_curator/stages/audio/tts/chatterbox_tts.py
```diff
@@ -0,0 +1,628 @@
...
+                    )
+                    sf.write(audio_path, audio_data, self.sample_rate)
+            except OSError as e:
+                logger.error(f"File I/O failed for task {task.task_id}: {e}")
+                output_tasks.append(task)
+                continue
+
+            duration = len(audio_data) / self.sample_rate
+
+            out_data = dict(data)
```
mohammadaaftabv, line 610: nemo_curator/stages/audio/tts/chatterbox_tts.py:610-620` constructs a new `AudioTask` with only data, task ID, and dataset, resetting `_metadata`, `_stage_perf`, and `_source_id`. Its “preserves metadata” test at `tests/stages/audio/tts/test_chatterbox_tts.py:295-315` checks ordinary data fields, not those framework fields. The vLLM fan-out at `vllm_inference.py:293-306` and merge fan-in at `merge_conversation.py:144-150` repeat the pattern.

Mutate and return the original task for one-to-one audio processors, as the audio stage contract prescribes. For fan-out/fan-in, add or use a framework helper that explicitly propagates provenance/lineage, and assert the real `_metadata`, `_stage_perf`, and `_source_id` fields. This repeats PR 1976's unresolved `discussion_r3718309325` finding.

nemo_curator/stages/audio/llm/vllm_inference.py
```diff
@@ -0,0 +1,316 @@
...
+            Supports ``$topic`` substitution from each entry's ``topic`` field.
+        model: Either a :class:`VLLMModel` instance or a dict whose keys
+            match ``VLLMModel.__init__`` parameters (``model``,
+            ``max_model_len``, ``tensor_parallel_size``, etc.).
+        apply_chat_template: Kwargs forwarded to
+            ``tokenizer.apply_chat_template()``.
+    """
+
+    name = "vLLMInference"
+    resources = Resources(gpus=1)
```
mohammadaaftabv, line 57: `nemo_curator/stages/audio/llm/vllm_inference.py:57-63,84-94` allowlists and forwards `disable_dual_chunk_attention`, but the current `VLLMModel.__init__` has no such argument; construction reproduced `TypeError`. The same allowlist accepts `tensor_parallel_size>1` while the stage always declares exactly one GPU, so a valid-looking multi-GPU config is scheduled with insufficient resources.

Validate against the real public `VLLMModel` signature, remove or implement the nonexistent knob, and derive stage GPU resources from tensor parallelism (or reject mismatches). Add constructor/config tests for every exposed key and multi-GPU scheduling.

nemo_curator/stages/audio/merging/merge_conversation.py
```diff
@@ -0,0 +1,774 @@
...
+    local_offset: float = 0.0
+
+
+_MIN_RTTM_FIELDS = 5
+_MIN_CTM_FIELDS = 5
+_MIN_RANDOMIZE_PAUSE = 0.5
+_PEAK_NORMALIZE_THRESHOLD = 0.99
+
+
+class MergeConversationSDPStage(ProcessingStage[AudioTask, AudioTask]):
```
mohammadaaftabv, line 61: nemo_curator/stages/audio/merging/merge_conversation.py:61-150` is an N→1 keyed aggregation but inherits empty `inputs()`/`outputs()` and `is_resumable=True`. Ordinary executors do not guarantee that one batch contains one complete conversation; mixed IDs only log a warning at `:130-139` and are merged anyway. The tutorial works around this by collecting all phase-1 output and invoking `.setup()`/`.process_batch()` directly (`tutorials/audio/data-generation/main.py:98-139`), bypassing the adapter's task-ID, performance, failure, and resumability machinery.

Implement reusable keyed fan-in/grouping in the framework (or a proper non-resumable grouping stage), reject mixed/incomplete groups, declare complete stage contracts, and run merge through an executor. Mark this stage non-resumable until N→1 lineage/checkpoint attribution exists.

nemo_curator/stages/audio/alignment/mfa_alignment.py
```diff
@@ -0,0 +1,615 @@
...
+
+    def outputs(self) -> tuple[list[str], list[str]]:
+        data_keys = ["textgrid_filepath"]
+        if self.create_rttm:
+            data_keys.append("rttm_filepath")
+        if self.create_ctm:
+            data_keys.append("ctm_filepath")
+        return [], data_keys
+
+    def xenna_stage_spec(self) -> dict[str, Any]:
```
mohammadaaftabv, line 163: `nemo_curator/stages/audio/alignment/mfa_alignment.py:163-166` caps one worker per node only through `xenna_stage_spec`; the tutorial's default Ray Data path has no corresponding constraint. `num_jobs` is passed to MFA at `:371-386` without matching CPU reservation, `subprocess.run` at `:421-427` has no timeout, and `_setup_local_mfa` at `:589-612` reuses a hostname-only cache without validating source/model identity or an atomic completeness marker. One wedged MFA call can therefore monopolize alignment indefinitely, while reused node caches can silently contain the wrong models.

Either make Ray Data unsupported explicitly or enforce the same per-node lifecycle there; reserve CPUs for `num_jobs`; add a positive configurable timeout with process-group termination; and key, lock, atomically populate, and validate the local model cache by resolved source and requested models. Add multi-worker, different-source, interrupted-copy, and timeout tests. These repeat unresolved PR-1977 feedback (`discussion_r3390343876`, `discussion_r3718249875`, and `discussion_r3718246239`).

nemo_curator/stages/audio/llm/vllm_inference.py
```diff
@@ -0,0 +1,316 @@
+# Copyright (c) 2026, NVIDIA CORPORATION.  All rights reserved.
```
mohammadaaftabv, line 1: There is no test file for the new 316-line `vLLMInference`, no test for the 204-line two-phase runner, and no YAML composition or end-to-end contract test. The PR description explicitly says these paths received only import/compile smoke checks. That gap allowed strict-schema, invalid-key, grouping, provenance, Ray Data, and resource-contract failures onto the primary path, while the contribution guide expects new code to be covered.

Add focused vLLM unit tests (configuration, tokenizer lifecycle, validation/retry, fan-out and provenance), runner/config tests (Hydra composition, stage selection, backend creation, grouping and failure propagation), and at least one dependency-light integration through generated turns → mocked TTS/MFA → merge/writer. Add the reproduced failure cases above before treating the pipeline as covered.

tutorials/audio/alignment/pipeline.py
```diff
@@ -0,0 +1,231 @@
+# Copyright (c) 2026, NVIDIA CORPORATION.  All rights reserved.
```
mohammadaaftabv, line 1: `tutorials/audio/tts/pipeline.py:48-106,170-197`, `tutorials/audio/tts/run.py:45-93`, `tutorials/audio/alignment/pipeline.py:41-109`, and `tutorials/audio/alignment/run.py:32-75` duplicate executor selection, pipeline construction, lifecycle, and CLI behavior. Their YAMLs use the older private `processors:` shape, so they cannot use current main's canonical `nemo_curator/config/run.py` `stages:` path. This is 608 lines of redundant runner code before the extra 204-line data-generation runner.

Keep README + `pipeline.yaml`, convert the YAMLs to `stages:` with backend/execution mode/resources/batch settings, use the audio `ManifestWriterStage`, and invoke the generic runner as the current `qwen_omni_inprocess` tutorial does. The data-generation fan-in requirement should become a reusable framework facility rather than a third tutorial-specific orchestration implementation.

This file is unnecessary check out https://github.com/NVIDIA-NeMo/Curator/tree/main/tutorials/audio/qwen_omni_inprocess

  --config-path ../../tutorials/audio/qwen_omni_inprocess \
  --config-name pipeline \
  manifest_path=tests/fixtures/audio/tagging/sample_input.jsonl \
  output_path=/tmp/qwen_omni_output.jsonl \
  workspace_dir=/tmp/qwen_omni_workspace \
  default_language=en

mohammadaaftabv (changes requested):

PR #2183, opened 2026-07-07
Update Wikipedia URL generation and download

sarahyurick:


praateekmahajan (approved):
IIUC are we getting rate limited and therefore we need this User Agent thing?

sarahyurick:
@praateekmahajan yes, exactly!

PR #2187, opened 2026-07-08
fix(audio): benchmark tagging on pre-staged AMI meetings

mohammadaaftabv:
## Summary

Upgrades the audio-tagging benchmark from a basic smoke test into a reliable release-quality signal.

- Uses representative, long, multi-speaker AMI meeting audio.
- Prepares data and models in advance for consistent nightly runs.
- Fails on missing, incomplete, or low-quality output instead of reporting false success.
- Covers both normal and doubled workloads, using the standard merged benchmark configuration.

## Validation

Fresh local GPU runs passed end to end:

- Normal: 3/3 rows and 219 complete tagged segments.
- Repeat: 6/6 rows and 435 complete tagged segments.
- All benchmark requirements, focused tests, lint, formatting, and compilation checks passed.


mohammadaaftabv:
/ok to test https://github.com/NVIDIA-NeMo/Curator/pull/2187/commits/2293e77d54078c6ecde4539c69ea0d2e98ec35d0

rlratzel (commented):
This is great. I really like the idea of formalizing data prep by adding it to a YAML for the tool to process. It also makes it more self-documenting and eliminates surprises by making it obvious that certain benchmarks have data setup steps.

One thing I think would be better is if the new `data_setups` entry was read through the existing config reading code, rather than adding a new/separate `--data-setup-config` arg and function.  I think this should be a minor change, and would make for fewer args/functions/docs to maintain. I think the `nightly-data-setup.yaml` can continue to be a separate file too if that helps, meaning the tool would be called with 
`--config nightly-benchmark.yaml --config nightly-data-setup.yaml`
when data setup needs to happen.

This would also give us the flexibility to create custom `config.yaml` files that combine everything needed (path definitions, data prep, benchmark entries) into one file, which can be useful when other tooling (i.e. our gitlab nightly runner) generates a config at runtime.

rlratzel (commented):
This is great. I really like the idea of formalizing data prep by adding it to a YAML for the tool to process. It also makes it more self-documenting and eliminates surprises by making it obvious that certain benchmarks have data setup steps.

One thing I think would be better is if the new `data_setups` entry was read through the existing config reading code, rather than adding a new/separate `--data-setup-config` arg and function.  I think this should be a minor change, and would make for fewer args/functions/docs to maintain. I think the `nightly-data-setup.yaml` can continue to be a separate file too if that helps, meaning the tool would be called with 
`--config nightly-benchmark.yaml --config nightly-data-setup.yaml`
when data setup needs to happen.

This would also give us the flexibility to create custom `config.yaml` files that combine everything needed (path definitions, data prep, benchmark entries) into one file, which can be useful when other tooling (i.e. our gitlab nightly runner) generates a config at runtime.

benchmarking/data_prep/prepare_audio_tagging_data.py
```diff
@@ -0,0 +1,473 @@
...
+
+    logger.remove()
+    logger.add(sys.stderr, level="INFO")
+
+    if args.verify_only:
+        logger.info(f"Verifying staged audio-tagging dataset at: {output_path}")
+        logger.info(f"Verifying staged audio-tagging model at: {model_output_path}")
+        return 0 if verify_dataset(output_path) and verify_model(model_output_path) else 1
+
+    dataset_ready = _has_dataset_artifacts(output_path) and verify_dataset(output_path)
```
rlratzel: It seems like you shouldn't need `_has_dataset_artifacts` since `verify_dataset` seems to do that check and more, but maybe I'm missing something.

Same for `_has_model_artifacts`.

rlratzel (approved):
The new changes LGTM.
I had one question about a data prep script, but that shouldn't hold up approval.

benchmarking/data_prep/prepare_audio_tagging_data.py
mohammadaaftabv: Fixed it Rick. Thanks.

mohammadaaftabv:
/ok to test https://github.com/NVIDIA-NeMo/Curator/pull/2187/commits/be6c1060de0938390d13213ab90a23e15e7e6fb9

mohammadaaftabv:
/ok to test https://github.com/NVIDIA-NeMo/Curator/pull/2187/commits/5dfd7f3b464ae77a0eb3412f5f3ddf95d2775d23

PR #2232, opened 2026-07-21
Interleaved Lance Reader

suiyoubi:
## Summary

  Adds an `InterleavedLanceReader` for row-wise interleaved multimodal Lance datasets.

  The reader builds on Curator's existing `LancePartitioningStage` and
  `LanceReaderStage`, then splits each resulting Arrow table into validated
  `InterleavedBatch` outputs while preserving row order and `sample_id`
  document boundaries.

  ## Changes

  - Added `nemo_curator.stages.interleaved.lance.InterleavedLanceReader`
  - Added `InterleavedLanceReaderStage`
  - Reuses `LanceReaderStage.read_task()` for:
    - Dataset and scanner configuration
    - Pinned Lance versions
    - Blob materialization
    - Optional Lance metadata columns
  - Supports document-aware output batching with:
    - `max_batch_bytes`
    - `max_batch_rows`
  - Supports:
    - `fragment_ids`
    - `fragments_per_partition`
    - Field projection and Lance read options
  - Preserves source row order and never splits consecutive rows belonging
    to the same `sample_id`
  - Validates required `InterleavedBatch` fields and rejects null `sample_id`
    values
  - Exports the reader through the interleaved stage package

  ## Tests

  Added unit coverage for:

  - Required field validation
  - Reader decomposition/config propagation
  - Invalid batch/output limits
  - Row-count based splitting without splitting `sample_id`
  - Streaming reads where one sample spans scanner batches
  - `max_output_rows` behavior without emitting partial samples
  - Lance metadata columns
  - Rejection of unsupported streaming blob columns

### DIFF
--- a/nemo_curator/core/utils.py
+++ b/nemo_curator/core/utils.py
@@ -274,16 +274,3 @@ def split_table_by_group(
     starts = [0, *split_points]
     ends = [*split_points, num_rows]
     return [table.slice(start, end - start) for start, end in zip(starts, ends, strict=True)]
-
-
-def split_table_by_group_max_bytes(
-    table: pa.Table,
-    group_column: str,
-    max_batch_bytes: int | None,
-) -> list[pa.Table]:
-    """Backward-compatible wrapper for byte-limited interleaved batching."""
-    return split_table_by_group(
-        table,
-        group_column,
-        max_batch_bytes=max_batch_bytes,
-    )
--- a/nemo_curator/core/utils.py
+++ b/nemo_curator/core/utils.py
@@ -214,54 +214,76 @@ def init_cluster(  # noqa: PLR0913
     return proc
 
 
+def split_table_by_group(
+    table: pa.Table,
+    group_column: str,
+    *,
+    max_batch_bytes: int | None = None,
+    max_batch_rows: int | None = None,
+) -> list[pa.Table]:
+    """Split an Arrow table without reordering or splitting consecutive groups.
+
+    Rows for each group must be consecutive. If a single group exceeds a batch
+    limit, it is still emitted as one chunk.
+    """
+    for name, value in (("max_batch_bytes", max_batch_bytes), ("max_batch_rows", max_batch_rows)):
+        if value is not None and value <= 0:
+            msg = f"{name} must be > 0, got {value}"
+            raise ValueError(msg)
+
+    if group_column not in table.column_names:
+        msg = f"Group column '{group_column}' not found in table"
+        raise ValueError(msg)
+    if table.num_rows == 0:
+        return [table]
+
+    chunked = table[group_column]
+    groups = chunked.combine_chunks() if len(chunked.chunks) > 1 else chunked.chunks[0]
+    if pc.any(pc.is_null(groups)).as_py():
+        msg = f"Group column '{group_column}' contains null values"
+        raise ValueError(msg)
+    if max_batch_bytes is None and max_batch_rows is None:
+        return [table]
+
+    num_rows = table.num_rows
+    group_change = pc.not_equal(groups.slice(1), groups.slice(0, num_rows - 1))
+    group_ends = [*(point + 1 for point in pc.indices_nonzero(group_change).to_pylist()), num_rows]
+
+    split_points: list[int] = []
+    chunk_start = 0
+    chunk_bytes = 0.0
+    chunk_rows = 0
+    avg_bytes_per_row = table.nbytes / num_rows
+    for group_end in group_ends:
+        group_start = chunk_start + chunk_rows
+        group_rows = group_end - group_start
+        group_bytes = group_rows * avg_bytes_per_row
+        exceeds_batch_limit = chunk_rows > 0 and (
+            (max_batch_bytes is not None and chunk_bytes + group_bytes > max_batch_bytes)
+            or (max_batch_rows is not None and chunk_rows + group_rows > max_batch_rows)
+        )
+        if exceeds_batch_limit:
+            split_points.append(group_start)
+            chunk_start = group_start
+            chunk_bytes = 0.0
+            chunk_rows = 0
+
+        chunk_bytes += group_bytes
+        chunk_rows += group_rows
+
+    starts = [0, *split_points]
+    ends = [*split_points, num_rows]
+    return [table.slice(start, end - start) for start, end in zip(starts, ends, strict=True)]
+
+
 def split_table_by_group_max_bytes(
     table: pa.Table,
     group_column: str,
     max_batch_bytes: int | None,
 ) -> list[pa.Table]:
-    """Split an Arrow table by approximate byte size without splitting group rows.
-
-    Each unique value in ``group_column`` is kept in a single output table.
-    If a single group exceeds ``max_batch_bytes``, it is still emitted as one chunk.
-
-    Note: null values in ``group_column`` are grouped together (consecutive
-    nulls are not split).  Callers should ensure the column is non-nullable
-    or handle nulls upstream.
-    """
-    if max_batch_bytes is None or table.num_rows == 0:
-        return [table]
-    if max_batch_bytes <= 0:
-        msg = f"max_batch_bytes must be > 0, got {max_batch_bytes}"
-        raise ValueError(msg)
-    if group_column not in table.column_names:
-        msg = f"Group column '{group_column}' not found in table"
-        raise ValueError(msg)
-
-    sort_indices = pc.sort_indices(table, sort_keys=[(group_column, "ascending")])
-    table = table.take(sort_indices)
-    col = table[group_column]
-    n = table.num_rows
-
-    if n <= 1:
-        return [table]
-
-    ne = pc.not_equal(col.slice(1), col.slice(0, n - 1))
-    split_points = pc.indices_nonzero(ne).to_pylist()
-    group_starts = [0, *(p + 1 for p in split_points)]
-    group_ends = [*(p + 1 for p in split_points), n]
-
-    avg_bytes_per_row = table.nbytes / n
-    chunk_split_indices: list[int] = []
-    chunk_bytes = 0.0
-    for i, (gs, ge) in enumerate(zip(group_starts, group_ends, strict=True)):
-        group_bytes = (ge - gs) * avg_bytes_per_row
-        if i > 0 and chunk_bytes > 0 and (chunk_bytes + group_bytes > max_batch_bytes):
-            chunk_split_indices.append(gs)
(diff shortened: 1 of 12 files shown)


nemo_curator/stages/interleaved/lance/reader.py
```diff
@@ -0,0 +1,362 @@
...
+        flushed_rows = chunk.num_rows
+        acc.clear()
+
+    acc.pending_groups.append(group)
+    acc.pending_bytes += group_bytes
+    acc.pending_rows += group_rows
+    return flushed_rows
+
+
+def _tables_from_record_batches(  # noqa: C901
```
VibhuJawa: codex: This overlaps with `split_table_by_group_max_bytes`, which is already used by the Parquet and WebDataset interleaved readers. Can we consolidate these into one shared interleaved batching utility that accepts tables or record batches, preserves row order, and supports byte, row, and output limits?

nemo_curator/stages/interleaved/lance/reader.py
```diff
@@ -0,0 +1,362 @@
...
+                "reader_output_splits": float(len(splits)),
+                "reader_output_rows": float(rows),
+                "reader_output_bytes": float(bytes_),
+            }
+        )
+        if not batches:
+            return []
+        return batches if len(batches) > 1 else batches[0]
+
+    def _stream_read_task(self, task: LanceReadTask) -> tuple[list[pa.Table], dict[str, Any]]:
```
VibhuJawa: codex: This duplicates most of `LanceReaderStage.read_task` and requires importing the private `_pop_dataset_kwargs` helper. Can we extract scanner construction into a reusable `LanceReaderStage` method that owns dataset options, pinned version, fragments, requested columns, and metadata-column options?

nemo_curator/stages/interleaved/lance/reader.py
```diff
@@ -0,0 +1,362 @@
...
+                max_batch_rows=self.max_batch_rows,
+                max_output_rows=self.max_output_rows,
+            )
+            metadata = _reader_metadata(
+                output.metadata if output.metadata is not None else task._metadata,
+                streaming_read=False,
+                stopped_early=stopped_early,
+            )
+
+        batches = [
```
VibhuJawa, line 74: codex: Existing interleaved readers share schema reconciliation through `BaseInterleavedReader._align_output`. Can we expose that as a shared interleaved utility and use it here so Lance, Parquet, and WebDataset readers apply the same reserved-column typing?

nemo_curator/stages/interleaved/lance/reader.py
```diff
@@ -0,0 +1,362 @@
...
+            max_batch_bytes=self.max_batch_bytes,
+            max_batch_rows=self.max_batch_rows,
+            max_output_rows=self.max_output_rows,
+        )
+        if not splits:
+            self._validate_result(task, pa.Table.from_pylist([]))
+        if self.include_lance_metadata:
+            splits = [add_lance_metadata_columns(split) for split in splits]
+
+        metadata = {
```
VibhuJawa: codex: This recreates the metadata already assembled by `LanceReaderStage`. Can we centralize the common Lance task metadata there and have this reader add only `streaming_read` and `stopped_early`?

nemo_curator/stages/interleaved/__init__.py
```diff
@@ -12,10 +12,17 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from nemo_curator.stages.interleaved.lance import InterleavedLanceReader, InterleavedLanceReaderStage
```
VibhuJawa: codex: Can we reuse the lazy export pattern from the text I/O reader package introduced in #2111? That keeps optional Lance imports consistent and avoids maintaining another package-export approach.

VibhuJawa (commented):
codex: Consolidation opportunities with the existing Lance and interleaved reader infrastructure.

nemo_curator/stages/interleaved/io/__init__.py
```diff
@@ -15,8 +15,10 @@
 from nemo_curator.stages.interleaved.io.reader import InterleavedParquetReader, InterleavedWebdatasetReader
 from nemo_curator.stages.interleaved.io.writers.tabular import InterleavedParquetWriterStage
 from nemo_curator.stages.interleaved.io.writers.webdataset import InterleavedWebdatasetWriterStage
+from nemo_curator.stages.interleaved.lance import InterleavedLanceReader
```
VibhuJawa: codex: **[P1] This makes the optional Lance dependency mandatory for all interleaved I/O imports.** Both package initializers now eagerly import the Lance reader, which eventually imports `lance`, even though Lance remains an optional extra and is not included in `interleaved_cpu`. Importing the existing Parquet or WebDataset APIs will therefore fail for users without the Lance extra. Please expose the Lance symbols lazily, following the text I/O reader pattern from #2111.

VibhuJawa (commented):
codex: Optional dependency import boundary.

nemo_curator/stages/interleaved/lance/reader.py
suiyoubi: makes sense, but it appears to me that existing utility sorts by group column; and this Lance path preserves row order and supports streaming carry-over, row limits, and output limits. 
I think it needs a careful shared utility refactor with tests across Parquet/WebDataset/Lance. @VibhuJawa  do you think this is necessary to merge ?

VibhuJawa: Actually, I am still confused about the value of the streaming path. `_tables_from_record_batches()` collects every output chunk in a list, and `process()` returns them only after scanning finishes. Therefore, it does not provide bounded-memory streaming unless `max_output_rows` stops the scan early.

What concrete workload requires this path instead of reading the task normally and splitting the resulting table afterward?

For example:

```python
output = self.read_task(task, dict(self.read_kwargs or {}), self.fields)
table = align_interleaved_table(output.data)

splits = split_table_by_group(
    table,
    "sample_id",
    max_batch_bytes=self.max_batch_bytes,
    max_batch_rows=self.max_batch_rows,
)
```

Like, what are we trying to achieve by adding this code ?

suiyoubi: The original motivation for this path was to avoid read_task() calling `scanner.to_table()` and constructing one complete fragment-size Arrow table before any document-aware splitting occurs. 

But based on my experiments with mint1t, I think this might not be necessary, so I'll refactor to use your recommendation 👍

nemo_curator/core/utils.py
```diff
@@ -214,54 +214,76 @@ def init_cluster(  # noqa: PLR0913
...
+    ends = [*split_points, num_rows]
+    return [table.slice(start, end - start) for start, end in zip(starts, ends, strict=True)]
+
+
+def split_table_by_group_max_bytes(
+    table: pa.Table,
+    group_column: str,
+    max_batch_bytes: int | None,
+) -> list[pa.Table]:
+    """Backward-compatible wrapper for byte-limited interleaved batching."""
```
VibhuJawa: This function `split_table_by_group`  looks good, dont think we should add the compatibility shim , so can get rid of `split_table_by_group_max_bytes(`

VibhuJawa (changes requested):
Mostly looks good, added a comment on compatibility shim.

tests/stages/interleaved/test_interleaved_lance_reader.py
```diff
@@ -0,0 +1,181 @@
...
+
+
+def _tables(result: InterleavedBatch | list[InterleavedBatch]) -> list[pa.Table]:
+    batches = result if isinstance(result, list) else [result]
+    return [batch.to_pyarrow() for batch in batches]
+
+
+def test_interleaved_lance_reader_validates_required_fields() -> None:
+    with pytest.raises(ValueError, match="omit required columns"):
+        InterleavedLanceReader(path="example.lance", fields=["sample_id"]).decompose()
```
VibhuJawa: codex: Can we test the owning stage directly and verify that the error identifies every missing field? Please rename this to `test_interleaved_lance_reader_stage_reports_missing_required_fields.


Something like:

```
def test_interleaved_lance_reader_stage_reports_missing_required_fields() -> None:
    fields = ["sample_id"]
    missing = sorted(InterleavedBatch.REQUIRED_COLUMNS - set(fields))

    with pytest.raises(ValueError) as exc_info:
        InterleavedLanceReaderStage(fields=fields)

    assert str(exc_info.value) == (
        f"InterleavedLanceReaderStage fields are missing required columns: {missing}"
    )

```

tests/stages/interleaved/test_interleaved_lance_reader.py
```diff
@@ -0,0 +1,181 @@
...
+    ).process(task)
+
+    tables = _tables(result)
+    assert [table["sample_id"].combine_chunks().to_pylist() for table in tables] == [
+        ["doc-a", "doc-a"],
+        ["doc-b", "doc-b"],
+        ["doc-c", "doc-c"],
+    ]
+    batches = result if isinstance(result, list) else [result]
+    assert all(batch._stage_perf is not task._stage_perf for batch in batches)
```
VibhuJawa, line 139: codex: Since `task._stage_perf` is empty, this would still pass if the reader discarded its contents and created unrelated empty lists. Can we seed the task with a performance entry and assert that every output preserves it while owning a distinct list?

tests/stages/interleaved/test_interleaved_lance_reader.py
```diff
@@ -0,0 +1,181 @@
...
+    assert result._metadata["lance"]["version"] == task.version
+    assert result._metadata["lance"]["fragment_ids"] == task.data
+    assert result._metadata["lance"]["has_stable_row_ids"] is False
+
+
+def test_interleaved_lance_reader_honors_top_level_version_read_kwarg(tmp_path: Path) -> None:
+    dataset_path = tmp_path / "version.lance"
+    rows = [_row("doc-a", 0, "text", "a0"), _row("doc-a", 1, "image")]
+    _write_interleaved_dataset(dataset_path, rows)
+    version = lance.dataset(str(dataset_path)).version
```
VibhuJawa: codex: The requested version is also the current dataset version, so this test passes even if `read_kwargs["version"]` is ignored. 

Can we write distinguishable `old` and `new` dataset versions, request the old version through `InterleavedLanceReader`, and assert both `task.version == old_version` and that only the old rows are returned? This would also match the version-history coverage in `test_lance_reader_uses_task_version_over_read_kwargs`.

VibhuJawa (approved):
LGTM, thanks for the work on this

PR #2239, opened 2026-07-22
Allow the inputs for a stage to support more than one input task type

### LINKED ISSUE #2244: Support more than 1 task type for stage inputs
Is your feature request related to a problem? Please describe.
Currently many stages typically work with a single type of task as input for eg: DocumentBatch or VideoTask.
Even if a stage accepts different task types often they may have the same input/output requirements ["data"]["text"].
However with more cases of stages being applicable cross modality and cross task there might be cases where the input and output spec vary depending on the task type.

For eg: MinHashStage can be extended to operate on DocumentBatch in addition to FilegroupTask. However the input requirements for fgtask is (["data"], []) however for DocumentBatch it would be (["Data"], ["text","id"]). Which breaks validate_inputs in many cases or forces the stages to include the least amount of validation requirements to allow multiple types to work.

Describe the solution you'd like
One solution might be to allow inputs to be of type Dict{Class: tuple[list, list]} where each key is the task implementation.

Describe alternatives you've considered
Alternative is for each stage with this requirement to override the validate_input/validate_output stage to handle the custom branching logic.

Additional context

ayushdg:
## Description

Some stages probably work with more than one type of task.  <br>For eg: FileGroupTask or DocumentBatch or InterleavedTask

The current paradigm for inputs is that it supports only `tuple(list, list)`. However deeding on the input task type the validation requirements may vary.

FileGroupTask may require only a data attribute while DocumentBatch may also require a text attribute.  <br>This PR extends the base Stage class to now support a dict mapping the input task type to expect validation attribute checks so that more than input task can be validated without sacrificing checks. Fixes NVIDIA-NeMo/Curator#2244, Fixes NVIDIA-NeMo/Curator#2244

## Usage

```python
class MyStage(ProcessingStage[FileGroupTask | DocumentBatch, FileGroupTask]):
    def inputs(self):
        return {
            FileGroupTask: (["data"], [])
            DocumentBatch: (["data"], ["text", "id"])
    }
```

## Checklist

- [X] I am familiar with the [Contributing Guide](<https://github.com/NVIDIA-NeMo/Curator/blob/main/CONTRIBUTING.md>).


sarahyurick:
Can we close this in favor of https://github.com/NVIDIA-NeMo/Curator/pull/2243?

ayushdg:
> Can we close this in favor of #2243?

In the interest of simplifying reviews this PR just changes the core logic to support more than one input type and #2243 builds on top of this (with a smaller diff once this is merged) that uses to extend minhash. 
I split it up for ease of review but happy to just have #2243 as the source of truth for both if that works

nemo_curator/stages/base.py
```diff
@@ -217,6 +224,42 @@ def validate_input(self, task: Task) -> bool:
...
+        if isinstance(input_specs, tuple):
+            return self._validate_input_spec(input_specs, "inputs()")
+        if not isinstance(input_specs, dict):
+            msg = (
+                f"Stage {self.name} inputs() must return an input spec tuple "
+                "or dict of task type to input spec"
+            )
+            raise TypeError(msg)
+
+        for candidate_type in type(task).mro():
```
praateekmahajan, line 247: Can we add a comment here describing what this mro does?

praateekmahajan (approved):

PR #2243, opened 2026-07-22
Document batch minhash input spec

ayushdg:
## Description

MinHash currently is FileGroupTask in and out. However for many use cases it might make sense for it to support generic DocumentBatch tasks with the id_Generator already run. This helps for cases where we need to run some Cpu based preprocessing on the text before minhashing.

Replaces #2155 and builds on top of #2239 
## Usage

```python

Pipeline(
   JsonlReader(generate_id=True),
    MinHashStage(...)
)
```

## Checklist

<details>

<summary>/</summary>

Closes NMCUR-186

</details>


tests/pipelines/test_pipelines.py
```diff
@@ -97,6 +97,54 @@ def test_raises_when_ray_serve_active_with_xenna_and_gpu_stages() -> None:
...
+    assert "    Required attributes: data" in description
+    assert "    Required columns: values" in description
+    assert "Error getting stage info" not in description
+
+
+def test_describe_skips_empty_dict_input_specs() -> None:
+    description = Pipeline(name="test", stages=[_EmptyDictInputStage()]).describe()
+
+    assert "  Inputs:" not in description
+    assert "    _SimpleTask:" not in description
```
sarahyurick, line 144: Why do we not want this to include `_SimpleTask`?

tests/stages/deduplication/fuzzy/test_minhash_stage.py
```diff
@@ -402,3 +407,140 @@ def test_setup_idempotency(self, tmp_path: Path) -> None:
...
+        with suppress(Exception):
+            kill_id_generator_actor()
+
+        stage = MinHashStage(output_path=str(tmp_path / "no_actor"), text_field="text", pool=False)
+        stage.setup()
+        assert stage.id_generator is None
+
+    @pytest.mark.usefixtures("ray_client_with_id_generator")
+    def test_read_format_none_ok_for_document_batch(self, batch_dataframe: pd.DataFrame, tmp_path: Path) -> None:
+        """read_format=None is fine for DocumentBatch input (no file reading happens)."""
```
sarahyurick: Should it have a warning about being ignored?

tests/pipelines/test_pipelines.py
ayushdg: This matches the old behavior where if a stage's inputs returned `([ ], [ ])` we would not render any required attributes or columns in the `describe` output like this:
```
print(Pipeline("test", stages=[_NoopStage()]).describe())
Pipeline: test
Description: No description provided
Stages: 1

Stage 1: noop
  Resources: 1.0 CPUs
  Batch size: 1
```

sarahyurick: Ok I just wanted to be sure, like if a stage is explicitly including `return {_SimpleTask: ([], [])}` then maybe it should be because the user is expecting it to verify the task type?

tests/stages/deduplication/fuzzy/test_minhash_stage.py
ayushdg: Added a warning if read_format is set but we pass in a doc batch

ayushdg:
/ok to test [a209308](https://github.com/NVIDIA-NeMo/Curator/pull/2243/commits/a2093083bfc5d13c0d34221d6a7f2f78579000ae)

praateekmahajan (approved):

nemo_curator/stages/deduplication/fuzzy/minhash.py
```diff
@@ -283,42 +292,45 @@ def setup(self, _worker_metadata: "WorkerMetadata | None" = None) -> None:
...
+            FileGroupTask: (["data"], []),
+            DocumentBatch: (["data"], [CURATOR_DEDUP_ID_STR, self.text_field]),
+        }
 
     def outputs(self) -> tuple[list[str], list[str]]:
         """Define outputs - produces FileGroupTask with minhash files."""
         return (["data"], [])
 
-    def process(self, task: FileGroupTask) -> FileGroupTask:
+    def process(self, task: FileGroupTask | DocumentBatch) -> FileGroupTask:
```
praateekmahajan, line 313: based on the benchmarks it seems like the GPU is being used after the benchmark too, which makes this stage risky to be chained with another, let's ensure we have a teardown() or soemthing to cleanup the stage

benchmarking/scripts/minhash_benchmark.py
```diff
@@ -0,0 +1,351 @@
...
+    except json.JSONDecodeError as error:
+        msg = f"Expected a JSON object, got invalid JSON: {error}"
+        raise argparse.ArgumentTypeError(msg) from error
+    if not isinstance(parsed, dict):
+        msg = "Expected a JSON object"
+        raise argparse.ArgumentTypeError(msg)
+    return parsed
+
+
+def _dataset_ratio(value: str) -> float:
```
sarahyurick, line 60: Should this be in a util file instead?

sarahyurick (commented):
LGTM! Left a minor comment.

PR #2257, opened 2026-07-27
Audio: add Qwen3-ASR adapter

mohammadaaftabv:
## Summary

Add Qwen3-ASR as a second implementation of Curator's shared `ASRAdapter` contract.

The PR keeps both adapters:

- `QwenASRAdapter` for `Qwen/Qwen3-ASR-0.6B`, using qwen-asr's vLLM-only `Qwen3ASRModel.LLM` path.
- The existing `QwenOmniASRAdapter`, Qwen3-Omni pipeline, dependency contract, tests, and vLLM behavior remain functionally unchanged.

There is no standalone Qwen-ASR runner. Manifest transcription uses the generic `ASRStage` adapter path.

## Qwen3-ASR adapter

- Implements the shared lifecycle names and signatures used by Qwen-Omni: `download_weights_on_node`, `load_model`, `unload_model`, and `transcribe_batch`.
- Constructs `Qwen3ASRModel.LLM` with the same engine arguments as `nkoluguri/integration-test`: model, GPU-memory utilization, inference batch size, max new tokens, remote-code trust, eager execution, prefix caching, and `xxhash`.
- Contains no Transformers `.from_pretrained` model loading or inference implementation.
- Includes the same scoped `transformers.check_model_inputs` compatibility shim as the nkoluguri reference because qwen-asr 0.0.6 still invokes the decorator-factory form.
- Maps one Curator adapter batch to one qwen-asr `transcribe` call while preserving input order and skipped positions.
- Forwards `revision` to weight download and vLLM construction.
- Exposes a copied `vllm_kwargs` dictionary; adapter-owned engine arguments cannot be overridden through it.
- Requires exactly one integer GPU per ASR actor and validates integer/resource knobs.
- Cleans up after partial model-load failure and marks empty/whitespace/None output as skipped.
- Returns detected language through `ASRResult.extras`.

`audio_cuda12` pins `qwen-asr==0.0.6`. Qwen-ASR installations compose it with Curator's shared `vllm` extra. The dependency intentionally does not use `qwen-asr[vllm]`, whose vLLM 0.14.0 pin conflicts with Curator's Qwen-Omni-compatible vLLM 0.22.0 stack.

## Generic adapter extras

`ASRStage` can persist arbitrary adapter metadata without introducing dynamic top-level columns:

- `ASRResult.extras` remains one generic adapter-owned dictionary.
- Set `ASRStage.extras_key` to write a shallow copy under one nested manifest field.
- `extras_key=None` is the default, preserving the existing Qwen-Omni output contract and existing ASRStage callers.
- Qwen-ASR configures `extras_key: asr_extras`, producing `"asr_extras": {"detected_language": "English"}`.
- Empty extras remove stale metadata for processed rows.
- Empty names and collisions with prediction, `_skipme`, or `additional_notes` columns are rejected.
- The nested dictionary is declared through `outputs()`, so executor/schema planning remains deterministic.

## Shared stage, tests, and tutorial

- Adds `tutorials/audio/qwen_asr/` using `ManifestReader -> ResampleAudioStage -> ASRStage(QwenASRAdapter) -> ManifestWriterStage`.
- Uses SoundFile to decode resampled PCM WAVs in the shared ASR stage.
- Uses Qwen's 30-code language order:
  `zh, en, yue, ar, de, fr, es, pt, id, it, ko, ru, th, vi, ja, tr, hi, ms, nl, sv, da, fi, pl, cs, fil, fa, el, hu, mk, ro`.
- Reuses `tests/fixtures/audio/qwen_omni/audio_1_5s_16khz_mono.wav`; this PR adds no duplicate WAV fixture.
- Keeps the GPU tests in `tests/models/asr/test_qwen_asr.py` under `@pytest.mark.gpu` and adds the `audio_qwen_asr` GPU test group.
- Uses `tests/fixtures/audio/tagging/sample_input.jsonl` as the documented two-file smoke input.
- Adds the tutorial-only `vllm_kwargs.max_model_len=8192` override for 12 GB GPUs; the adapter's production/reference default is unchanged.
- Makes no change to `benchmarking/README.md`.

Shared `validate_vllm_kwargs` and `merge_vllm_kwargs` helpers now live in `nemo_curator.utils.vllm_utils`. Both Qwen adapters describe their actual caller-owned constructor kwargs to `merge_vllm_kwargs`, and the duplicate adapter-local `_RESERVED_VLLM_KWARGS` constants are removed. Protected-key collision checking now occurs once, inside the shared merge helper when `load_model()` assembles engine kwargs; the adapters no longer repeat validation in `__post_init__`. Qwen-Omni's protected-key behavior, engine arguments, and output contract remain unchanged.

## Current-main integration

- PR head: `4c8ac6f6604ec71ea5919ee014e4137ee1d0e560`
- Integrated `main`: `09ae28c41cfb7704d66eb4d6cbffc33fc6f05fe3`
- History: three DCO-signed commits on current main

### Current head

- 173 focused Qwen-ASR, Qwen-Omni, generic ASRStage, tutorial-contract, configuration, lazy-import, and shared-vLLM utility tests passed; 2 GPU tests were deselected.
- Exact Qwen-ASR and Qwen-Omni constructor-call tests pass after the shared kwargs merge refactor.
- Ruff formatting and lint checks passed.
- `uv lock --check` passed with the current-main-required uv 0.12.2.
- `git diff --check` passed.
- GPU-test-group coverage passed for all 46 GPU test files.
- GitHub DCO passes for all three signed commits.

### Local GPU and tutorial smoke

The earlier head `ddfb74cdc423dec0ffe64dae3e05fcf210b14617` passed:

- Real one-GPU vLLM smoke on an RTX 3080 Ti using qwen-asr 0.0.6, vLLM 0.22.0, and the existing Qwen-Omni WAV fixture.
- The documented Ray Data tutorial command on both bundled OPUS inputs: 2/2 nonempty transcripts, no skips, and `asr_extras.detected_language == "English"`.

The current head has not repeated that GPU smoke after the main rebase and kwargs-assembly refactor. The exact constructor tests confirm that the refactor preserves the engine arguments used by both Qwen adapters.

### Latest paired Qwen3-ASR parity

The latest paired run compared target commit `ddfb74cdc423dec0ffe64dae3e05fcf210b14617` with `nkoluguri/integration-test` commit `fff467c8d2e3632f9a96c96ec34cd0e6614503da`, using the same frozen eight-row English cohort (569.486 seconds), one node, and one L40 per arm.

- Target run: `1dc99ea4-4b0f-438d-832d-5d902384453b`
- Reference run: `0443d025-95c7-42f8-a68b-ee510c9ce0e1`
- Both independently submitted arms and all tasks succeeded on their first attempt.
- 8/8 expected identities and nonempty predictions in each arm.
- 8/8 target `asr_extras.detected_language` values matched reference `asr_language`; all were `English`.
- No missing, unexpected, duplicate, skipped, malformed, empty, forbidden-payload, or published-WAV cases.
- Mean normalized token similarity: `0.9699208113126543` (gate: at least `0.95`).
- Rows with normalized similarity at least `0.95`: `7/8` (gate: at least `7/8`).
- Normalized exact matches: `2/8`; raw exact matches: `1/8`.
- Lowest row similarity: `0.8516129032258064`.

The target used qwen-asr 0.0.6 with Curator's vLLM 0.22.0 stack; the remote-exact reference used qwen-asr 0.0.6 with its pinned vLLM 0.14.0 stack. Timing is therefore a canary observation, not a performance comparison.

Since the parity-tested head, the branch was rebased onto current `main` and the host-side vLLM kwargs validation/assembly was centralized. The actual Qwen-ASR engine values, input contract, output mapping, and inference path are unchanged; live parity has not been rerun after this review refactor.

## Diff

- 18 files changed
- 1,490 insertions / 32 deletions

### DIFF
--- a/nemo_curator/models/asr/qwen_asr.py
+++ b/nemo_curator/models/asr/qwen_asr.py
@@ -35,7 +35,7 @@ from huggingface_hub import snapshot_download
 from loguru import logger
 
 from nemo_curator.models.asr.base import ASRResult
-from nemo_curator.utils.vllm_utils import merge_vllm_kwargs, validate_vllm_kwargs
+from nemo_curator.utils.vllm_utils import merge_vllm_kwargs
 
 _DEFAULT_QWEN3_ASR_MODEL = "Qwen/Qwen3-ASR-0.6B"
 
@@ -129,11 +129,6 @@ class QwenASRAdapter:
             )
             raise ValueError(msg)
         self.vllm_kwargs = deepcopy(dict(self.vllm_kwargs))
-        validate_vllm_kwargs(
-            self.vllm_kwargs,
-            self._model_owned_vllm_kwargs(),
-            owner_description="adapter-owned arguments",
-        )
 
     def _model_owned_vllm_kwargs(self) -> dict[str, Any]:
         """Return the qwen-asr constructor arguments owned by this adapter."""
--- a/nemo_curator/models/asr/qwen_asr.py
+++ b/nemo_curator/models/asr/qwen_asr.py
@@ -23,14 +23,12 @@ results back to ``ASRResult`` positions.
 from __future__ import annotations
 
 import gc
-import inspect
 from copy import deepcopy
 from dataclasses import dataclass, field
 from typing import Any
 
 import numpy as np
 import torch
-import transformers
 from huggingface_hub import snapshot_download
 from loguru import logger
 
@@ -52,30 +50,6 @@ def _qwen_asr_model_cls() -> Any:  # noqa: ANN401
         msg = "QwenASRAdapter requires the audio_cuda12 and vllm extras: uv sync --extra audio_cuda12 --extra vllm"
         raise ImportError(msg) from exc
     return Qwen3ASRModel
-
-
-def _patch_transformers_compat() -> None:
-    """Accept qwen-asr's decorator-factory syntax on newer Transformers.
-
-    This matches the compatibility patch used by the nkoluguri Qwen-ASR
-    reference. Newer Transformers exposes ``check_model_inputs`` as a plain
-    decorator, while qwen-asr 0.0.6 still invokes it with parentheses.
-    """
-    try:
-        original = getattr(transformers, "check_model_inputs", None)
-        if original is None:
-            return
-        parameters = list(inspect.signature(original).parameters.values())
-        if parameters and parameters[0].name == "func":
-
-            def compat_check_model_inputs(*args: Any) -> Any:  # noqa: ANN401
-                if args and callable(args[0]):
-                    return original(args[0])
-                return original
-
-            transformers.check_model_inputs = compat_check_model_inputs
-    except Exception:  # noqa: BLE001, S110
-        pass
 
 
 @dataclass
@@ -171,7 +145,6 @@ class QwenASRAdapter:
         )
         if model_kwargs["revision"] is None:
             del model_kwargs["revision"]
-        _patch_transformers_compat()
         try:
             self._model = _qwen_asr_model_cls().LLM(**model_kwargs)
         except Exception:
--- a/nemo_curator/models/asr/qwen_asr.py
+++ b/nemo_curator/models/asr/qwen_asr.py
@@ -1,0 +1,262 @@
+# Copyright (c) 2026, NVIDIA CORPORATION.  All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Qwen3-ASR vLLM implementation of the shared ASR adapter.
+
+This uses the same ``Qwen3ASRModel.LLM`` construction and vLLM engine settings
+as the nkoluguri reference. ``ASRStage`` owns mono conversion and resampling;
+the adapter hands one prepared batch to one ``transcribe`` call and maps
+results back to ``ASRResult`` positions.
+"""
+
+from __future__ import annotations
+
+import gc
+import inspect
+from copy import deepcopy
+from dataclasses import dataclass, field
+from typing import Any
+
+import numpy as np
+import torch
+import transformers
+from huggingface_hub import snapshot_download
+from loguru import logger
+
+from nemo_curator.models.asr.base import ASRResult
+from nemo_curator.utils.vllm_utils import validate_vllm_kwargs
+
+_DEFAULT_QWEN3_ASR_MODEL = "Qwen/Qwen3-ASR-0.6B"
+
+# Qwen's audio processor needs >=200 samples for STFT padding. 1600 samples
+# (100 ms at 16 kHz) is a conservative floor that also matches the Qwen-Omni
+# preprocessing path.
+_MIN_SAMPLES = 1600
+
+_RESERVED_VLLM_KWARGS = frozenset(
+    {
+        "model",
+        "revision",
+        "gpu_memory_utilization",
+        "max_inference_batch_size",
+        "max_new_tokens",
+        "trust_remote_code",
+        "enforce_eager",
+        "enable_prefix_caching",
+        "prefix_caching_hash_algo",
+    }
+)
+
+
+def _qwen_asr_model_cls() -> Any:  # noqa: ANN401
+    try:
+        from qwen_asr import Qwen3ASRModel
+    except ImportError as exc:
+        msg = "QwenASRAdapter requires the audio_cuda12 and vllm extras: uv sync --extra audio_cuda12 --extra vllm"
+        raise ImportError(msg) from exc
+    return Qwen3ASRModel
+
+
+def _patch_transformers_compat() -> None:
+    """Accept qwen-asr's decorator-factory syntax on newer Transformers.
+
+    This matches the compatibility patch used by the nkoluguri Qwen-ASR
+    reference. Newer Transformers exposes ``check_model_inputs`` as a plain
+    decorator, while qwen-asr 0.0.6 still invokes it with parentheses.
+    """
+    try:
+        original = getattr(transformers, "check_model_inputs", None)
+        if original is None:
+            return
+        parameters = list(inspect.signature(original).parameters.values())
+        if parameters and parameters[0].name == "func":
+
+            def compat_check_model_inputs(*args: Any) -> Any:  # noqa: ANN401
+                if args and callable(args[0]):
+                    return original(args[0])
+                return original
+
+            transformers.check_model_inputs = compat_check_model_inputs
+    except Exception:  # noqa: BLE001, S110
+        pass
+
+
+@dataclass
+class QwenASRAdapter:
+    """Run vLLM-backed Qwen3-ASR over Curator waveform items.
+
+    Every valid item in one adapter call goes to a single ``transcribe`` call,
+    so the caller's batch boundary is the model's batch boundary.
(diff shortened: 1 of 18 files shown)


nemo_curator/models/asr/qwen_asr.py
```diff
@@ -0,0 +1,270 @@
...
+from nemo_curator.models.asr.base import ASRResult
+
+_DEFAULT_QWEN3_ASR_MODEL = "Qwen/Qwen3-ASR-0.6B"
+
+# Qwen's audio processor needs >=200 samples for STFT padding. 1600 samples
+# (100 ms at 16 kHz) is a conservative floor that also matches the Qwen-Omni
+# preprocessing path.
+_MIN_SAMPLES = 1600
+
+_RESERVED_VLLM_KWARGS = frozenset(
```
sarahyurick: Since this is not specific to Qwen ASR can we move it to a more generic utils file, and find out if there are other stages that can use it?

nemo_curator/models/asr/qwen_asr.py
```diff
@@ -0,0 +1,270 @@
...
+        self.vllm_kwargs = deepcopy(dict(self.vllm_kwargs))
+        reserved_vllm_kwargs = sorted(_RESERVED_VLLM_KWARGS.intersection(self.vllm_kwargs))
+        if reserved_vllm_kwargs:
+            msg = f"vllm_kwargs cannot override adapter-owned arguments: {', '.join(reserved_vllm_kwargs)}"
+            raise ValueError(msg)
+
+    @classmethod
+    def download_weights_on_node(cls, model_id: str, revision: str | None = None) -> None:
+        """Populate the local Hugging Face cache without allocating a GPU."""
+        from huggingface_hub import snapshot_download
```
sarahyurick: Can we move most if not all of these imports to the top of the file please?

nemo_curator/models/asr/qwen_asr.py
```diff
@@ -0,0 +1,270 @@
...
+            raise ValueError(msg)
+        return waveform
+
+    def transcribe_batch(self, items: list[dict[str, Any]]) -> list[ASRResult]:
+        """Transcribe one adapter call while preserving input order."""
+        if not items:
+            return []
+        if self._model is None:
+            msg = "QwenASRAdapter is not initialized; call load_model() first"
+            raise RuntimeError(msg)
```
sarahyurick: Let's remove this block, if the user is running it correctly in a pipeline then it should never reach.

tests/models/asr/test_qwen_asr.py
```diff
@@ -0,0 +1,407 @@
...
+
+def _adapter_with_model(model: MagicMock) -> QwenASRAdapter:
+    adapter = QwenASRAdapter()
+    adapter._model = model
+    return adapter
+
+
+# ----------------------------------------------------------------------
+# Contract conformance and construction validation
+# ----------------------------------------------------------------------
```
sarahyurick: Can we remove the claude comments please.

tests/models/asr/test_qwen_asr_gpu.py
sarahyurick, line 1: The GPU tests can be with the other tests in `test_qwen_asr` with `@pytest.mark.gpu`.

tutorials/audio/qwen_asr/README.md
```diff
@@ -0,0 +1,142 @@
...
+There is no Qwen-specific inference stage or tutorial runner. Model lifecycle,
+audio loading and normalization, language routing, batching, skip handling,
+and output assembly stay in the shared ASR stage; the adapter only owns the
+Qwen3-ASR model call.
+
+## Requirements
+
+- x86_64 Linux with one CUDA GPU per ASR actor
+- `ffmpeg`
+- Audio files accessible from the machine running the pipeline
```
sarahyurick, line 18: Can you provide a sample Hugging Face dataset or similar to be used with the tutorial?

pyproject.toml
```diff
@@ -129,6 +129,14 @@ audio_cpu = [
     "onnxruntime>=1.20.1,<1.24",
 ]
 
+# Qwen3-ASR adapter; compose qwen-asr with Curator's shared vLLM stack. Do not
+# use qwen-asr[vllm], whose vLLM 0.14.0 pin conflicts with Curator's current
+# Qwen-Omni-compatible vLLM version.
+audio_qwen_asr = [
```
sarahyurick: Any reason why we can't just add this to `audio_cuda12`?

tutorials/audio/qwen_asr/README.md
mohammadaaftabv: Hey Sarah we already provide sample audio files for quick run by users here in the readme at tests/fixtures/audio/tagging/sample_input.jsonl
Line 39 of same readme.

virajkarandikar (commented):
lgtm. vllm kwargs can be moved to utils.

nemo_curator/models/asr/qwen_asr.py
```diff
@@ -0,0 +1,249 @@
...
+def _qwen_asr_model_cls() -> Any:  # noqa: ANN401
+    try:
+        from qwen_asr import Qwen3ASRModel
+    except ImportError as exc:
+        msg = "QwenASRAdapter requires the audio_cuda12 and vllm extras: uv sync --extra audio_cuda12 --extra vllm"
+        raise ImportError(msg) from exc
+    return Qwen3ASRModel
+
+
+def _patch_transformers_compat() -> None:
```
sarahyurick: Instead of patching should we just use the actual compatible Transformers version? We could do it via the stage's `runtime_env` maybe?

tests/models/asr/test_qwen_asr.py
```diff
@@ -0,0 +1,463 @@
...
+    adapter = QwenASRAdapter(vllm_kwargs=vllm_kwargs)
+    vllm_kwargs["compilation_config"]["cudagraph_mode"] = "FULL"
+
+    assert adapter.vllm_kwargs == {"compilation_config": {"cudagraph_mode": "NONE"}}
+
+
+@pytest.mark.parametrize(
+    "reserved_key",
+    [
+        "model",
```
sarahyurick: Can this list be imported from somewhere?

tests/utils/test_vllm_utils.py
```diff
@@ -24,7 +24,58 @@
...
-from nemo_curator.utils.vllm_utils import pick_free_port, resolve_local_model_path
+from nemo_curator.utils.vllm_utils import (
+    merge_vllm_kwargs,
+    pick_free_port,
+    resolve_local_model_path,
+    validate_vllm_kwargs,
+)
+
+
+class TestValidateVllmKwargs:
```
sarahyurick: Nit but can we not have 2 separate test classes for validate/merge please.

nemo_curator/models/asr/qwen_asr.py
```diff
@@ -0,0 +1,222 @@
...
+# (100 ms at 16 kHz) is a conservative floor that also matches the Qwen-Omni
+# preprocessing path.
+_MIN_SAMPLES = 1600
+
+
+def _qwen_asr_model_cls() -> Any:  # noqa: ANN401
+    try:
+        from qwen_asr import Qwen3ASRModel
+    except ImportError as exc:
+        msg = "QwenASRAdapter requires the audio_cuda12 and vllm extras: uv sync --extra audio_cuda12 --extra vllm"
```
sarahyurick, line 50: Not sure if my previous comment was lost but does this need a specific transformers version that is different from what we have pinned for Curator?

mohammadaaftabv: No separate Transformers version is needed. Curator’s audio_common requires transformers>=4.57.6,<5.0, and qwen-asr==0.0.6 narrows that to the compatible transformers==4.57.6.
To clarify the error message, this PR did not create the vllm extra. audio_cuda12 installs Qwen-ASR and its Transformers dependency, while Curator’s existing vllm extra installs the shared vLLM backend used by this adapter.

This PR does not add a new transformers version at all.

mohammadaaftabv:
/ok to test https://github.com/NVIDIA-NeMo/Curator/pull/2257/commits/f5d42dffc19f8033f7cf82e370f3b17dfe418591

sarahyurick (approved):
Thanks!

PR #2258, opened 2026-07-27
Audio: add rule-based ASR hallucination filter

mohammadaaftabv:
## Description

Adds `WhisperHallucinationStage`, a CPU-only transcript filter for common ASR hallucination patterns. No model, GPU, or network is used; the stage applies four lexical checks to text produced by an upstream ASR stage.

The branch is rebased onto current `main` at `223f771`.

## Detection checks

Any single hit flags the row.

| Check | Behavior |
| --- | --- |
| Repeated n-grams | Flags when the unique-word ratio is at or below `unique_words_threshold`. |
| Long word | Flags on the absolute `long_word_threshold` or the relative `long_word_rel_threshold`. |
| Phrase match | Matches the cleaned transcript against the bundled common-hallucination phrase corpus. |
| High char rate | Flags when word characters per second exceed `max_char_rate`. |

For Finnish, Hungarian, Estonian, German, Dutch, Danish, Swedish, and Norwegian, the relative long-word check is skipped. The absolute `long_word_threshold` is now the single threshold for every language; there is no separate `agglutinative_long_word_threshold`.

The individual detector functions and `setup()` are AST-identical to `nithinraok/Curator:nkoluguri/integration-test` at `2963bf3`. `process()` retains the reference self-setup fallback. As a simple CPU-only audio stage, it follows current Curator convention and defines only direct single-task processing.

## YAML configuration and example

The runnable example is in `tutorials/audio/whisper_hallucination/pipeline.yaml`. From the Curator repository root:

```bash
  --config-path ../../tutorials/audio/whisper_hallucination \
  --config-name pipeline \
  manifest_path=tests/fixtures/audio/text_filtering/whisper_hallucination_input.jsonl \
  output_path=/tmp/whisper_hallucination_output.jsonl
```

The YAML exposes:

- `unique_words_threshold`
- `long_word_threshold`
- `long_word_rel_threshold`
- `max_char_rate`
- `overwrite`
- `recovery_value`

`_PREFIX_MATCH_MIN_LEN = 8` remains a `ClassVar`, not a YAML parameter. Lowering this safety boundary can make common short phrases such as `you`, `Ja`, or `Да` prefix-match unrelated transcripts.

The bundled `phrases.txt` is byte-identical to `tutorials/audio/granary_v2_postprocessing/phrases.txt` on the reference branch:

```text
SHA-256 34ba2fcd7756f193e80ba4ac34a6b5db0dab92adeb0750beb796b1bf57f6bc42
size    1535 bytes
```

The pre-commit exclusions are scoped only to this frozen fixture because the reference bytes intentionally contain trailing spaces and no final newline.

## Flag ownership and recovery

The stage writes `"Hallucination:{name}"` only when the row is unflagged or already has a hallucination flag, so a different filter's reason is preserved.

With `overwrite=true`, clean second-pass text clears an existing hallucination flag. `recovery_value` is a caller-selected note:

```yaml
overwrite: true
recovery_value: "Recovered:ASR"
```

For a clean row previously marked `Hallucination:PrimaryASR`, the result is:

- `_skipme: ""`
- `additional_notes.WhisperHallucination: "recovered:asr"`

An empty `recovery_value` records `"recovered"`.

## Validation

- 39 focused tests pass.
- The refreshed paired local comparison at `e0edecb` produced byte-identical target/reference JSONL across all 21 frozen rows with zero failures.
- The stage keeps no per-worker processed/flagged aggregate counters.
- Pre-commit passes for all changed files.
- Detector helpers plus `setup()` are AST-identical to the reference implementation.
- Phrase files match under both `cmp` and SHA-256.
- The exact documented command completed a local Ray Data run over the checked-in two-row fixture: the clean row passed and `Thank you.` was flagged with `phrase_match`.

## Checklist

- [x] I am familiar with the Contributing Guide.
- [x] New and existing tests cover the changes.
- [x] The example YAML and executable command are included.


nemo_curator/stages/audio/text_filtering/whisper_hallucination.py
```diff
@@ -0,0 +1,220 @@
...
+    .. code-block:: bash
+
+        python nemo_curator/config/run.py \\
+            --config-path ../../tutorials/audio/whisper_hallucination \\
+            --config-name pipeline \\
+            manifest_path=tests/fixtures/audio/text_filtering/whisper_hallucination_input.jsonl \\
+            output_path=/tmp/whisper_hallucination_output.jsonl
+    """
+
+    common_hall_file: str = ""
```
sarahyurick:
```suggestion
    common_hall_file: str
```

nemo_curator/stages/audio/text_filtering/whisper_hallucination.py
```diff
@@ -0,0 +1,220 @@
...
+            return True
+        if skip_relative:
+            return False
+        if len(lengths) > 1 and lengths[-2] > 0:
+            return (lengths[-1] - lengths[-2]) / lengths[-2] >= self.long_word_rel_threshold
+        return False
+
+    def _frequent_single_word(self, text: str) -> bool:
+        cleaned = text.strip().replace(".", "").replace("?", "").replace("!", "")
+        if cleaned in self._phrases:
```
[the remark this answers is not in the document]
sarahyurick: @mohammadaaftabv wdyt?

nemo_curator/stages/audio/text_filtering/whisper_hallucination.py
```diff
@@ -0,0 +1,220 @@
...
+            return True
+        return any(
+            len(phrase) >= self._PREFIX_MATCH_MIN_LEN and cleaned.startswith(phrase) for phrase in self._phrases
+        )
```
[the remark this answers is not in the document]
sarahyurick: @mohammadaaftabv wdyt?

nemo_curator/stages/audio/text_filtering/whisper_hallucination.py
```diff
@@ -0,0 +1,220 @@
...
+        chars = sum(len(w) for w in words)
+        return chars / duration > self.max_char_rate
+
+    def _is_agglutinative(self, task: AudioTask) -> bool:
+        lang = str(task.data.get(self.language_key, "")).lower().strip()
+        return lang in AGGLUTINATIVE_COMPOUNDING_LANGS
+
+    def process(self, task: AudioTask) -> AudioTask:
+        if not self._setup_called:
+            logger.warning(
```
sarahyurick: Let's remove this block. Setup will always be called if the user is running the pipeline correctly.

tests/stages/audio/text_filtering/test_whisper_hallucination.py
```diff
@@ -0,0 +1,342 @@
+# Copyright (c) 2025, NVIDIA CORPORATION.  All rights reserved.
```
sarahyurick:
```suggestion
# Copyright (c) 2026, NVIDIA CORPORATION.  All rights reserved.
```

tests/stages/audio/text_filtering/test_whisper_hallucination.py
```diff
@@ -0,0 +1,342 @@
...
+def test_stage_declares_its_io_contract(tmp_path: Path) -> None:
+    stage = _make_stage(tmp_path, [])
+    required_attrs, required_columns = stage.inputs()
+    _required_out, optional_out = stage.outputs()
+    assert required_attrs == []
+    assert required_columns == [_TEXT_KEY, _SKIP_KEY, "duration"]
+    assert {_SKIP_KEY, "additional_notes"} <= set(optional_out)
+
+
+def test_stage_is_cpu_only(tmp_path: Path) -> None:
```
sarahyurick: I think some of these tests are overkill.

.pre-commit-config.yaml
```diff
@@ -32,9 +32,10 @@ repos:
       - id: check-yaml
       - id: detect-private-key
       - id: end-of-file-fixer
-        exclude: docs/
+        exclude: 'docs/|^tutorials/audio/whisper_hallucination/phrases\.txt$'
       - id: requirements-txt-fixer
       - id: trailing-whitespace
+        exclude: '^tutorials/audio/whisper_hallucination/phrases\.txt$'
```
sarahyurick: Can we remove these changes? I would rather the txt file not have trailing whitespaces, etc.

nemo_curator/stages/audio/text_filtering/whisper_hallucination.py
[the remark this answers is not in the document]
mohammadaaftabv: @nithinraok, could you also clarify the intended case and punctuation normalization for the hallucination phrase corpus?

The current `[nkoluguri/integration-test](https://github.com/nithinraok/Curator/blob/nkoluguri/integration-test)` implementation:

- stores corpus entries using only `line.strip()`;
- remains case-sensitive;
- removes `.`, `?`, and `!` from the transcript;
- does not remove commas or hyphens.

As a result:

- `Thank you.` matches `Thank you`;
- `thank you` does not match `Thank you`;
- `THANK YOU` does not match `Thank you`;
- `Merci,` does not match `Merci`;
- `Thank-you` does not match `Thank you`.

One proposed alternative is to normalize both corpus phrases and transcripts by:

1. replacing Unicode punctuation with spaces;
2. applying Unicode-aware `casefold()`;
3. collapsing whitespace;
4. requiring equality of the complete normalized transcript.

That alternative would make all of these equivalent:

- `Thank you`
- `thank you`
- `THANK YOU!`
- `Thank-you`

It would also make `Merci,` match `Merci`.

However, it changes the reference behavior and reduces the 101 raw entries to 99 unique normalized phrases because some entries collapse to the same normalized value.

Could you advise whether the case-sensitive and limited-`.?!` cleanup in the reference is deliberate based on observed Whisper output? Should this PR preserve exact reference parity, or should both implementations move to broader normalization?

For now, the PR preserves the reference behavior. The cleaned and reference text files load the same 101 stripped phrases, and fresh local execution shows 100% phrase-decision parity.

nemo_curator/stages/audio/text_filtering/whisper_hallucination.py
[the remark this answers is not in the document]
mohammadaaftabv: @nithinraok, could you clarify the intended phrase-prefix behavior based on the data and rationale behind `[nkoluguri/integration-test](https://github.com/nithinraok/Curator/blob/nkoluguri/integration-test)`?

The current reference implementation first checks exact phrase membership and then permits prefix matching when the corpus phrase is at least eight characters:

```python
cleaned = text.strip().replace(".", "").replace("?", "").replace("!", "")
if cleaned in self._phrases:
    return True
return any(
    len(phrase) >= self._PREFIX_MATCH_MIN_LEN
    and cleaned.startswith(phrase)
    for phrase in self._phrases
)

This means the following all match the corpus entry Thank you:
Thank you
Thank you.
Thank you for your time today.
The review concern is that the third example may be a valid continuation rather than a Whisper hallucination.
Could you advise which behavior is intended?
Preserve the current reference behavior:
exact matches are flagged;
longer transcripts beginning with a phrase of at least eight characters are also flagged.

Move to exact full-transcript matching:
Thank you and Thank you. are flagged;
Thank you for your time today. passes.

For now, the PR has been restored to match nkoluguri/integration-test exactly, and local execution shows phrase-decision parity. I would like your guidance before intentionally changing that reference behavior.

File: nemo_curator/stages/audio/text_filtering/whisper_hallucination.py (hunk 2 above)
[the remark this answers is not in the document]
Comment (nithinraok) [reply nemo_curator/stages/audio/text_filtering/whisper_hallucination.py]: >does not remove commas or hyphens.

Hi Aaftab, nice catch. lets also consider `commas or hyphens` and also update in  `nkoluguri/integration-test` branch

File: nemo_curator/stages/audio/text_filtering/whisper_hallucination.py (hunk 3 above)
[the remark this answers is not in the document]
Comment (nithinraok) [reply nemo_curator/stages/audio/text_filtering/whisper_hallucination.py:171]: `Thank you` and words followed by it are the most common hallucinated sentences, hence we decided to skip them.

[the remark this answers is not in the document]
Comment (nithinraok) [reply nemo_curator/stages/audio/text_filtering/whisper_hallucination.py:171]: Like, `Thank you, please subscribe to our channel`

**Comment by mohammadaaftabv** [comment]
/ok to test https://github.com/NVIDIA-NeMo/Curator/pull/2258/changes/e7137234b2595e657fce4f3db3808669becf64d5

File: nemo_curator/stages/audio/text_filtering/whisper_hallucination.py (hunk 2 above)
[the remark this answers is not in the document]
Comment (mohammadaaftabv) [reply nemo_curator/stages/audio/text_filtering/whisper_hallucination.py]: @nithinraok , following up on the normalization discussion here: while implementing the requested comma/hyphen handling, I found one remaining asymmetry for ., ?, and !.
The current PR normalizes incoming transcripts with:

text.strip().replace(".", "").replace("?", "").replace("!", "").replace(",", "")

but corpus entries are normalized with only:

text.strip().replace(",", "")

Consequently, these bundled corpus phrases remain unreachable:

Това е първият въпрос. is stored with the final ., while the incoming candidate loses it.
¿Y qué tal? is stored with the final ?, while the incoming candidate loses it.

Direct execution currently allows both transcripts to pass instead of flagging them as phrase_match.
I propose the following constrained corpus-side change:
@staticmethod
def _normalize_corpus_phrase(text: str) -> str:
    return (
        text.strip()
        .replace(".", "")
        .replace("?", "")
        .replace("!", "")
        .replace(",", "")
    )
This is intentionally narrower than the broader normalization originally discussed:
no casefold();
no Unicode-wide punctuation replacement;
no internal whitespace collapsing;
no change to prefix matching;
corpus hyphens remain preserved.

Preserving corpus hyphens is important because normalizing Fifty-four to Fifty four would make the prefix rule incorrectly flag valid text such as Fifty four dollars. The incoming side can continue checking both the original hyphenated transcript and a hyphen-to-space candidate, so Fifty-four and Thank-you remain handled without broadening corpus prefixes.
I would add regression coverage showing:
Това е първият въпрос. → flagged;
¿Y qué tal? → flagged;
Bonjour, je m'appelle Marie → still flagged;
Fifty-four → still flagged;
Fifty four dollars → still passes;
lowercase/case variants remain case-sensitive.

Would you approve this narrow .?!, corpus normalization, and should we apply the same change to nkoluguri/integration-test? Or should the PR retain the current reference behavior even though those two bundled phrases cannot be matched?

File: nemo_curator/stages/audio/text_filtering/whisper_hallucination.py (hunk 1 above)
Comment (sarahyurick) [reply nemo_curator/stages/audio/text_filtering/whisper_hallucination.py]: Bumping this.

File: nemo_curator/stages/audio/text_filtering/whisper_hallucination.py [hunk 8]
```diff
@@ -0,0 +1,255 @@
...
+    _setup_called: bool = field(default=False, init=False, repr=False)
+    _n_processed: int = field(default=0, init=False, repr=False)
+    _n_flagged: int = field(default=0, init=False, repr=False)
+
+    # Long phrases such as "Thank you" intentionally match their common
+    # Whisper-generated continuations.
+    _PREFIX_MATCH_MIN_LEN: int = 8
+
+    def __post_init__(self) -> None:
+        if not self.common_hall_file:
```
sarahyurick: Same here, please remove this check.

nemo_curator/stages/audio/text_filtering/whisper_hallucination.py
```diff
@@ -0,0 +1,255 @@
...
+        elif self.overwrite and was_flagged:
+            task.data[self.skip_me_key] = ""
+            recovery_note = self.recovery_value.lower() if self.recovery_value else "recovered"
+            self._set_note(task.data, recovery_note)
+        else:
+            self._set_note(task.data, "passed")
+        return task
+
+    def process(self, task: AudioTask) -> AudioTask:
+        if not self._setup_called:
```
sarahyurick: We don't need this check.

nemo_curator/stages/audio/text_filtering/whisper_hallucination.py
```diff
@@ -0,0 +1,255 @@
...
+            logger.warning(
+                f"WhisperHallucinationStage ({self.name}): setup() was not called before process(). "
+                "Calling setup() now — check that your executor invokes setup() on each worker."
+            )
+            self.setup()
```
sarahyurick: We should never call setup in the process function.

nemo_curator/stages/audio/text_filtering/whisper_hallucination.py
```diff
@@ -0,0 +1,255 @@
...
+        return self._process_single(task)
+
+    def process_batch(self, tasks: list[AudioTask]) -> list[AudioTask]:
+        if not self._setup_called:
+            logger.warning(
+                f"WhisperHallucinationStage ({self.name}): setup() was not called before process_batch(). "
+                "Calling setup() now — check that your executor invokes setup() on each worker."
+            )
+            self.setup()
```
sarahyurick: Same feedback as above.

sarahyurick (changes requested):

virajkarandikar (approved):

nemo_curator/stages/audio/text_filtering/whisper_hallucination.py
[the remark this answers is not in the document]
nithinraok: looks good to me, make same changes to `nkoluguri/integration-test`

nithinraok (approved):
LGTM

nemo_curator/stages/audio/text_filtering/whisper_hallucination.py
```diff
@@ -0,0 +1,236 @@
...
+        elif self.overwrite and was_flagged:
+            task.data[self.skip_me_key] = ""
+            recovery_note = self.recovery_value.lower() if self.recovery_value else "recovered"
+            self._set_note(task.data, recovery_note)
+        else:
+            self._set_note(task.data, "passed")
+        return task
+
+    def process(self, task: AudioTask) -> AudioTask:
+        return self._process_single(task)
```
sarahyurick: Similar comments as before but should not need `process` implementation since you are already overriding `process_batch`.

sarahyurick (approved):
Thanks!

mohammadaaftabv:
/ok to test https://github.com/NVIDIA-NeMo/Curator/pull/2258/commits/78a295e3986bc438e4068853e6b5887ca7b2f289

PR #2260, opened 2026-07-27
Audio: add sound event detection inference stage

mohammadaaftabv:
## What this does

Adds a generic `SEDInferenceStage` that runs sound-event detection over audio and emits framewise class probabilities.

This PR contains only the SED producer. Event thresholding, interval construction, filtering, and event-to-task fan-out remain in #2261.

## Stage / adapter split

The implementation follows the same boundary as the generic ASR stage:

- `nemo_curator/stages/audio/inference/base.py` owns behavior shared by adapter-backed audio stages: lazy Hydra adapter resolution, resource-to-physical-GPU conversion, load-failure cleanup, worker teardown, file-versus-waveform input declaration, and SoundFile decoding.
- `nemo_curator/stages/audio/inference/asr/stage.py` and `nemo_curator/stages/audio/inference/sed.py` inherit that shared infrastructure and retain only task/model-specific preparation, inference, and output assembly.
- Worker sizing uses the existing `ProcessingStage.with_(num_workers=...)` mechanism. SED no longer carries a separate `num_workers_override` field or `num_workers()` implementation.
- `nemo_curator/models/sed/base.py` defines the canonical `SEDAdapter` / `SEDResult` contract.
- `nemo_curator/models/sed/panns.py` owns PANNs checkpoint loading, device selection, model-specific padding, one batched inference call, and temporal result metadata.
- `adapter_target` remains lazy and YAML-selectable, so another runtime can replace PANNs without changing the stage's task schema.
- CNN14 and its registry live under `nemo_curator/models/sed/`; no model implementation remains under the stages package.

## Runnable YAML

A complete example is included at `tutorials/audio/sed/pipeline.yaml`. Each input JSONL row contains an audio path:

```json
{"audio_filepath": "/absolute/path/to/audio.wav"}
```

Run from the Curator repository root:

```bash
uv run --extra audio_sed python nemo_curator/config/run.py \
  --config-path ../../tutorials/audio/sed \
  --config-name pipeline \
  manifest_path=/absolute/path/to/input.jsonl \
  'checkpoint_path=/absolute/path/to/Cnn14_DecisionLevelMax_mAP\=0.385.pth' \
  output_dir=/absolute/path/to/sed_output
```

[...]

PR #2261, opened 2026-07-28
Audio: add sound event detection postprocessing

mohammadaaftabv:
## What this does

Adds `SEDPostprocessingStage`, which turns a framewise sound-event probability matrix into a list of timed events by thresholding, smoothing, and merging. Events are labelled from the AudioSet ontology and can be emitted either per class (`electric_guitar`) or collapsed to superclasses (`speech`, `music`, `animal`, and several noise categories).

This was split out of #2260, which was too large to review in one pass. This is the **consumer** half of sound event detection; #2260 is the producer. **The two are independent and can be reviewed and merged in either order.**

## Why it is independent

The stage never imports a model, and it imports nothing from the inference side. Its input is three task data keys:

| Key | Meaning |
| --- | --- |
| `_sed_framewise` | `[frame, class]` probability matrix |
| `sed_fps` | frame rate, used to convert frame indices to times |
| `sed_valid_frames` | how many leading frames are real audio rather than batch padding |

Anything that can write those keys can drive this stage. An NPZ sidecar path is accepted as a fallback, so a pipeline can resume from probabilities computed in an earlier run. The matrix is dropped from task data once events are derived, since it is far larger than the events it produces.

Everything here is pure numpy on CPU, so **no new dependency is required** and `pyproject.toml` is untouched.

## Attribution

`sed_utils.py` is ported from the internal `hifi_granary` project and is credited as such in its module docstring. Its class groupings reference the [AudioSet ontology](https://research.google.com/audioset/), published by Google under CC BY 4.0, from which only integer class indices and label strings are reproduced.

No third-party source code is vendored in this PR. The vendored PANNs/CNN14 code that needs legal sign-off is entirely in #2260, so that review does not block this half.

## Tests

The source branch shipped no tests for any of this. Fifty-two are added, all passing:

- segment extraction, hysteresis, merging, aggregation modes, and event geometry for the pure-numpy helpers
- threshold, duration, and superclass behavior for the stage, over both the in-memory and NPZ input paths
- exact stage I/O, public package export, inherited batch processing, and CPU resource behavior
- clean-interpreter import without Torch/TorchLibrosa and smoothing fallback when SciPy is unavailable
- superclass groups are disjoint and every grouped class index has a name

```
52 passed
```

`ruff check`, `ruff format --check`, and the file-scoped pre-commit hooks are clean. Importing `SEDPostprocessingStage` from its feature package pulls in neither `torch` nor `torchlibrosa`, which is asserted in a clean interpreter rather than assumed. The stage also reuses `ProcessingStage.process_batch()` from `main` instead of carrying an identical list-comprehension override.

## Reviewer notes

- #2260 is now present on `main`; these tests intentionally stay focused on this consumer stage and drive it with the producer's canonical in-memory and NPZ contracts.
- After merging current `main`, this PR no longer edits the shared top-level audio registry. `SEDPostprocessingStage` remains importable from its postprocessing package, eliminating the former arbitrary-order conflict with #2260.
- `min_duration_sec` defaults to 0.3s, which drops very short events; set it to 0 to see everything the threshold produces.


mohammadaaftabv:
@nithinraok, could you please review and approve this intentional divergence from your `integration-test` reference?

Both the original PR implementation and the reference contained the same legacy hard-coded AudioSet indices. However, several of those indices do not match the canonical 527-class ordering used by the PANNs checkpoint. For example:

| Index | Legacy mapping | Canonical label |
|---:|---|---|
| 27 | laughter | Singing |
| 288 | rain | Water |
| 350 | dishes_pots_pans | Engine knocking |
| 420 | gunshot | Jackhammer |
| 493 | static | Rumble |

I updated PR #2261 so its curated groups use the canonical PANNs indices—for example, `16` for laughter, `289` for rain, `364` for dishes/pots/pans, `427` for gunshot, and `515` for static.

Could you confirm that you approve treating the upstream AudioSet/PANNs index table as authoritative and allowing PR #2261 to diverge from the current reference for these mappings?

References:

- [`class_labels_indices.csv`](https://github.com/qiuqiangkong/audioset_tagging_cnn/blob/master/metadata/class_labels_indices.csv)
- [Current reference implementation](https://github.com/nithinraok/Curator/blob/1e2d639946f28dfd289a0a9456480385d4480c28/nemo_curator/stages/audio/postprocessing/sed_utils.py)
- [Corrected PR implementation](https://github.com/mohammadaaftabv/Curator/blob/4c43fecdcf8ae8f9713dc58a85eeca0b1d6185d7/nemo_curator/stages/audio/postprocessing/sed_utils.py)

virajkarandikar (approved):

nune-tadevosyan (approved):

VibhuJawa (approved):
LGTM

mohammadaaftabv:
/ok to test https://github.com/NVIDIA-NeMo/Curator/pull/2261/commits/65f8194657cf5d04990be327df4b04b91b41787f

PR #2266, opened 2026-07-28
feat(audio): select the best ASR prediction

mohammadaaftabv:
## Summary
- Add a standalone stage that selects a stable transcript from primary, fallback, or reference text.
- Handle unsupported-language routing, recovered fallbacks, cross-model agreement, reference recovery, forced-reference selection, and the Qwen-Omni short-audio ground-truth policy.
- Match the relevant selector behavior from `nithinraok/Curator:nkoluguri/integration-test` at `2963bf3a23d807ff26a80cf80aeb1a33fd0f7172` (selector blob `e9517ad7dc150eb42c70f8c55d28470f12d0b33b`, last changed by `8469455fbd18de74928f21e1cc241358fc8d9e08`).
- Adapt that behavior to current main's `AudioTask` and `ProcessingStage` contracts:
  - keep the model-generic `fallback_text_key` and `primary_fallback_agreement_wer` names;
  - accept `asr_text_key` as a compatibility alias;
  - retain a lightweight local word-distance helper, with the reference's two-decimal WER precision, instead of importing the heavyweight NeMo ASR metric stack;
  - declare the notes output and clear stale skip/agreement audit state.
- Rebase onto NVIDIA-NeMo/Curator `main` at `0b9e6e56999720795470b24cc9a1d47f934b7a62`.

This branch contains only prediction-selection code and focused tests.

## Reference parity
- Loaded the fetched reference selector locally and compared its decisions with this branch across 589,824 deterministic scenarios.
- 237,132 scenarios matched exactly.
- 352,692 differences were fully explained by the intentional main-contract fixes: stale skip cleanup, stale agreement cleanup, ignoring this stage's prior recovery note, and rejecting an unsupported primary when no usable fallback exists.
- Zero unexplained mismatches remained.
- Added pinned golden vectors for precedence and fallthrough behavior without making CI depend on a second Git branch.

## Validation
- 20 focused tests passed with the repository's normal Ray-backed pytest fixture.
- Ruff check passed.
- Ruff format check passed.
- Changed-file pre-commit hooks passed.
- `git diff --check` passed.

### DIFF
--- a/tests/stages/audio/text_filtering/test_select_best_prediction.py
+++ b/tests/stages/audio/text_filtering/test_select_best_prediction.py
@@ -35,7 +35,8 @@ def test_uses_recovery_prediction_after_hallucination_recheck() -> None:
     assert task.data["_skipme"] == "Hallucination"
 
 
-def test_falls_back_when_primary_language_is_unsupported() -> None:
+def test_selects_supported_fallback_and_clears_primary_language_skip() -> None:
+    """The primary adapter was skipped; a separate supported model produced the fallback."""
     task = AudioTask(
         data={
             "primary_model_prediction": "",
@@ -49,7 +50,11 @@ def test_falls_back_when_primary_language_is_unsupported() -> None:
 
     assert task.data["best_prediction"] == "bonjour"
     assert task.data["best_prediction_source"] == "fallback"
-    assert task.data["_skipme"] == "language_not_supported"
+    assert task.data["_skipme"] == ""
+    assert task.data["additional_notes"] == {
+        "primary_model_prediction": "lang_not_supported:fr",
+        "SelectBestPrediction": "used fallback (primary lang unsupported)",
+    }
 
 
 def test_unsupported_primary_without_fallback_matches_reference_fallthrough() -> None:
--- a/tests/stages/audio/text_filtering/test_select_best_prediction.py
+++ b/tests/stages/audio/text_filtering/test_select_best_prediction.py
@@ -12,6 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# ruff: noqa: INP001
+
 from copy import deepcopy
 
 from nemo_curator.stages.audio.text_filtering import SelectBestPredictionStage
--- a/tests/stages/audio/text_filtering/test_select_best_prediction.py
+++ b/tests/stages/audio/text_filtering/test_select_best_prediction.py
@@ -1,0 +1,361 @@
+# Copyright (c) 2026, NVIDIA CORPORATION.  All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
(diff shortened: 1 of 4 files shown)


tests/stages/audio/text_filtering/test_select_best_prediction.py
```diff
@@ -0,0 +1,495 @@
...
+            "_skipme": "language_not_supported",
+            "additional_notes": {"primary_model_prediction": "lang_not_supported:fr"},
+        }
+    )
+
+    SelectBestPredictionStage().process(task)
+
+    assert task.data["best_prediction"] == "bonjour"
+    assert task.data["best_prediction_source"] == "fallback"
+    assert task.data["_skipme"] == "language_not_supported"
```
virajkarandikar: now that valid transcript is updated, should skipme get reset ?

mohammadaaftabv: Once a valid fallback transcript is selected, the existing language_not_supported skip reason applies only to the rejected primary and becomes stale. Leaving it set also causes downstream text-filtering stages to skip the valid fallback.
However, nithinraok/Curator:nkoluguri/integration-test@a1df5ec9 preserves _skipme on this path. The proposed fix clears _skipme only when the primary is unsupported and a valid fallback is selected.
@nkoluguri, can you confirm that this intentional divergence from the reference is approved?
Thanks

nithinraok: Good catch, if we select fallback model prediction, we need to reset skipme and only add this to additional_notes. 

But why would we in first place run inference with a model whose language is not supported?

mohammadaaftabv: Thanks. 
Changed code to reset skipme when fallback model prediction accepted.

The primary model is not invoked for an unsupported language—ASRStage filters unsupported rows before calling the adapter and records lang_not_supported. 

This test represents that skipped primary row plus a transcript from a separate fallback model that supports the language. I renamed and documented the test to make that setup explicit. Per your confirmation, selecting the valid fallback now clears _skipme while retaining the primary-language diagnosis in additional_notes

nithinraok: >The primary model is not invoked for an unsupported language—ASRStage filters unsupported rows before calling the adapter and records lang_not_supported.

Exactly, my question was why do we allow this even in first place? Why would the model be primary if it doesn;t support that language?

mohammadaaftabv: This test case is to handle incoming data that has a row skip_me set as language_not_supported. **_ Each block on curator main is built to work both within and outside granary v2 pipeline._** So, when we use granary v2 pipeline and included ASR stage, we should never encounter this issue. But, if someone runs ASR inference using different model (like outside curator) that does not filter out unsupported rows, then also this stage should work as intended and the test makes sure of that.

Fundamental of any given audio stage in curator main is that it should work correctly either in a given pipeline or when run as an independent entity with correct input data structure.

nithinraok: I will approve this, but this is not per expectation.

nithinraok (approved):
LGTM

virajkarandikar (approved):

mohammadaaftabv:
/ok to test https://github.com/NVIDIA-NeMo/Curator/pull/2266/commits/6acc04e780d23b75932dcb62df384da47a9225b9

mohammadaaftabv:
/ok to test https://github.com/NVIDIA-NeMo/Curator/commit/6acc04e780d23b75932dcb62df384da47a9225b9

abhinavg4 (approved):
All files changes inside Audio. Approved by Curator policy.

PR #2267, opened 2026-07-28
feat(audio): add transcript normalization stages

mohammadaaftabv:
## Summary
- Add ordered YAML-driven regex substitution for ASR transcripts (`pred_text` → `text`) with strict setup-time rule validation, whitespace normalization, skip preservation, and empty-after-cleaning handling.
- Add language-aware concatenation of ASR-spelled abbreviations with contraction, particle, and Unicode plural handling plus normalization notes.
- Export both stages through `nemo_curator.stages.audio.text_filtering` and the lazy `nemo_curator.stages.audio` API.
- Use the current-main stage lifecycle with inline `process()` and inherited batch processing.

## Validation
- 79 focused tests pass.
- A 492,274-case generated abbreviation probe is idempotent and preserves character sequences apart from intended space removal.
- Target/reference differential checks cover abbreviation text and metadata, regex rules, and the default chained stages.
- Ruff and pre-commit pass.

Includes and supersedes #2268.

### DIFF
--- a/nemo_curator/stages/audio/text_filtering/abbreviation_concat.py
+++ b/nemo_curator/stages/audio/text_filtering/abbreviation_concat.py
@@ -114,13 +114,15 @@ def _join_match(match: re.Match[str], particles: frozenset[str]) -> str:  # noqa
     if any(len(part) >= _MULTI_CHAR_LEN for part in parts):
         return original
 
-    # Keep trailing words such as "Is", "As", and "Os" separate while
-    # still allowing consonant plural suffixes such as "Xs".
+    # Keep trailing words such as "Is", "As", and "Os" separate. The
+    # sequence "A P Is" is normalized to the plural abbreviation "APIs".
     tail = ""
+    is_api_plural = parts == ["A", "P", "Is"]
     if (
         len(parts) >= _MIN_PARTS
         and len(parts[-1]) == _PLURAL_SUFFIX_LEN
         and parts[-1][1] == "s"
+        and not is_api_plural
         and (parts[-1][0] in _VOWELS or not parts[-1][0].isupper())
     ):
         tail = " " + parts.pop()
@@ -181,7 +183,13 @@ def concat_abbreviations(text: str, language: str = "en") -> tuple[str, list[str
         else:
             joined = _join_match(match, particles)
         if joined != raw:
-            abbreviation = joined.strip().rstrip("’s").rstrip("’s")  # noqa: RUF001
+            abbreviation = joined.strip()
+            raw_parts = raw.split(" ")
+            last_part = raw_parts[-1]
+            if len(last_part) == _PLURAL_SUFFIX_LEN and last_part[0].islower() and last_part[1] == "s":
+                abbreviation = abbreviation.removesuffix(last_part).rstrip()
+            elif not (raw == "A P Is" and abbreviation == "APIs"):
+                abbreviation = abbreviation.rstrip("’s").rstrip("’s")  # noqa: RUF001
             if abbreviation:
                 found.append(abbreviation)
         return joined
--- a/nemo_curator/stages/audio/text_filtering/abbreviation_concat.py
+++ b/nemo_curator/stages/audio/text_filtering/abbreviation_concat.py
@@ -1,0 +1,224 @@
+# Copyright (c) 2026, NVIDIA CORPORATION.  All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from __future__ import annotations
+
+import functools
+import re
+from dataclasses import dataclass, field
+from typing import Any
+
+from loguru import logger
+
+from nemo_curator.stages.base import ProcessingStage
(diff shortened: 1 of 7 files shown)


mohammadaaftabv:
@nithinraok, could you please confirm which of the following remaining behavioural differences from `nkoluguri/integration-test` should be kept?

Comparison anchors

PR head: 397749323a99ed3591b5107bff3509958e9a6441
Reference: nithinraok/Curator:nkoluguri/integration-test at a1df5ec915fb62d37dda780b790e9ceb153b68da

1. Contraction boundaries also differ:

   | Input | PR | Reference |
   |---|---|---|
   | `I'm A B` | `I'm AB` | `I'mAB` |
   | `A B I'm` | `AB I'm` | `ABI'm` |
   | `A B I’m` | `AB I’m` | `AB’m` |

2. **Rejected matches and edge particles are preserved instead of deleting tokens.**

   | Input | PR | Reference |
   |---|---|---|
   | `A A Is` | `A A Is` | `A A` |
   | `a a B` | `a aB` | `aB` |
   | `B a a` | `Ba a` | `Ba` |

3. **An ASCII `s` after an uppercase non-ASCII letter is accepted as an abbreviation plural.** This is also the outstanding Greptile concern because the ASCII-only vowel check classifies uppercase accented vowels as consonants.

   | Input | PR | Reference |
   |---|---|---|
   | `А Б Вs` | `АБВs` | `АБ Вs` |
   | `Ä Ö Üs` | `ÄÖÜs` | `ÄÖ Üs` |
   | `A Ös` | `AÖs` | `A Ös` |
   | `Α Β Γs` | `ΑΒΓs` | `ΑΒ Γs` |

4. **Language resolution is more permissive.** A missing or falsey `source_lang` falls back to English, and supplied language codes are stripped and lowercased. The reference requires the key and uses its value unchanged.

5. **Regex YAML validation is strict.** The document must be a list of mappings; `pattern` and `repl` must be strings; and `count` must be a non-boolean, non-negative integer.

6. **Missing or non-string regex input stabilizes the output schema.** The PR materializes an empty output field when it is absent; the reference raises `KeyError` for a missing input and returns a non-string row without creating the output.


Please confirm whether all six differences are acceptable, or identify the item numbers that should be aligned with the reference.

tests/stages/audio/text_filtering/test_abbreviation_concat.py
```diff
@@ -0,0 +1,133 @@
...
+
+    assert result == expected
+
+
+@pytest.mark.parametrize(
+    ("text", "expected"),
+    [
+        ("A P I and G P U", ["API", "GPU"]),
+        ("a A P I", ["a API"]),
+        ("A P Is", ["AP I"]),
```
virajkarandikar: should expected be  `APIs` ?

mohammadaaftabv: APIs is the more natural expectation here. However, the current test was matching nithinraok/Curator:nkoluguri/integration-test.
The reference treats terminal Is as a separate word, producing AP Is, and reports it using replaced.strip().rstrip("’s").rstrip("’s"), which produces AP I because rstrip() treats its argument as a character set rather than an exact suffix.
@nkoluguri, can you confirm whether we should intentionally diverge from the reference and make both the normalized text and reported abbreviation APIs, or retain the reference behavior?

nithinraok: It should be `APIs` as correctly pointed by Viraj.

nemo_curator/stages/audio/text_filtering/abbreviation_concat.py
```diff
@@ -0,0 +1,228 @@
...
+    name: str = "AbbreviationConcat"
+    resources: Resources = field(default_factory=lambda: Resources(cpus=1.0))
+
+    def inputs(self) -> tuple[list[str], list[str]]:
+        return [], [self.text_key]
+
+    def outputs(self) -> tuple[list[str], list[str]]:
+        return [], [self.output_text_key, self.notes_key]
+
+    def process(self, task: AudioTask) -> AudioTask:
```
nune-tadevosyan, line 219: Do we want to have process_batch as well?

nemo_curator/stages/audio/text_filtering/regex_substitution.py
```diff
@@ -0,0 +1,114 @@
...
+            result = re.sub(
+                rule["pattern"],
+                rule["repl"],
+                result,
+                count=rule.get("count", 0),
+            )
+        result = re.sub(r"\s+", " ", result).strip()
+        task.data[self.output_text_key] = result
+        if not result and had_text:
+            task.data[self.skip_me_key] = "Empty after regex cleaning"
```
nune-tadevosyan, line 113: Can we check is _skipme flag was already there and not to overwrite it?

nune-tadevosyan (changes requested):

tests/stages/audio/text_filtering/test_abbreviation_concat.py
nithinraok: CC: @nune-tadevosyan

nune-tadevosyan (approved):

tests/stages/audio/text_filtering/test_abbreviation_concat.py
nune-tadevosyan: Yes it should be `APIs`

virajkarandikar (approved):

mohammadaaftabv:
/ok to test https://github.com/NVIDIA-NeMo/Curator/pull/2267/commits/a55b89ab4c96cb46cf7c8b20c09b42d1ff185e76

abhinavg4 (approved):
All files changes inside Audio. Approved by Curator policy.

mohammadaaftabv:
/ok to test https://github.com/NVIDIA-NeMo/Curator/pull/2267/commits/47d7a3812d2e8c60e9bd233c343846e21284b6a3

PR #2268, opened 2026-07-28
Audio: concatenate spelled abbreviations

mohammadaaftabv:
## Summary
- Add the independent abbreviation concatenation stage used after transcript normalization.
- Join safe spaced-letter sequences across supported Latin/Cyrillic/Greek language patterns while protecting particles, repeated letters, contractions, and skipped rows.
- Grounded in `nkoluguri/integration-test` at `1f1e770b429d47292a4133dd3ab15e46b1cf8ecc`, refactored into a reusable main-compatible stage.
- Reuse `ProcessingStage.process_batch()` from main rather than duplicating its per-task loop.

This branch starts directly from `main` and contains only abbreviation concatenation.

## Validation
- 6 focused tests passed.
- Ruff check and format passed.


mohammadaaftabv:
Folded into #2267, which now contains both regex transcript normalization and abbreviation concatenation rebased onto current `main`. Closing this PR as superseded.

PR #2281, opened 2026-07-30
Support GlotLID language and script identification

### LINKED ISSUE #2276: Support GlotLID for language/script identification.
FineWeb2 and a few other datasets use `GlotLID` for lang identification in favor of the more common languid file shared by fastext: https://fasttext.cc/docs/en/language-identification.html
https://github.com/cisnlp/GlotLID

It claims to cover more languages and also annotate by `lang_script` combo which can be useful in some cases.

We already have FastextStages and that this can inherit from since it's a fastext based model.

**Additional context**

ccyyy1023:
## Summary

* preserve complete FastText labels after removing the `__label__` prefix, including GlotLID language-script labels such as `eng_Latn`
* support case-insensitive filtering by either language (`eng`) or exact language-script pair (`eng_Latn`)
* document GlotLID usage and add focused regression tests

## Why

`FastTextLangId` previously extracted only the final two characters of a model label. For a GlotLID prediction such as `__label__eng_Latn`, this produced `TN` and discarded both the language and script information.

The updated parsing preserves the model label and lets users choose between language-level and language-script-level filtering while retaining the existing confidence cutoff behavior.

## Validation

* `pytest -q tests/stages/text/filters/fasttext/test_fasttext_filters.py` — 10 passed
* `pytest -q tests/stages/text/modules/test_filters.py tests/stages/text/filters/fasttext/test_fasttext_filters.py` — 85 passed, 1 skipped
* `pytest -m 'not gpu' -q tests/stages/text` — 780 passed, 5 skipped, 53 deselected, 1 xfailed
* `ruff check` and `ruff format --check` on the changed Python files

Fixes NVIDIA-NeMo/Curator#2276
Fixes NMCUR-189


nemo_curator/stages/text/filters/fasttext/fasttext_filters.py
```diff
@@ -56,12 +56,20 @@ def keep_document(self, score: float) -> bool:
...
+    match. Matching is case-insensitive.
+    """
+
     def __init__(self, model_path: str | None = None, min_langid_score: float = 0.3, lang: str | None = None):
         if model_path is None:
             msg = "Must provide a valid path to a FastText model to identify languages with this filter"
             raise ValueError(msg)
         self._model_path = model_path
-        self._lang_code = lang.upper() if lang else None
+        self._lang_code = lang.casefold() if lang else None
```
sarahyurick:
```suggestion
        self._lang_code = lang if lang else None
```

how about we change the behavior slightly. When we are doing the comparison we can use `casefold` but the `lang_code` that is returned in the dataframe does not need the `casefold`.

nemo_curator/stages/text/filters/fasttext/fasttext_filters.py
```diff
@@ -73,14 +81,16 @@ def model_check_or_download(self) -> None:
...
         # See setup() function in modules/filter.py
         model = self._fasttext_langid_model
 
         pp = text.strip().replace("\n", " ")
         label, score = model.predict([pp], k=1)
         score = score[0][0].item()
-        lang_code = label[0][0][-2:].upper()
+        lang_code = label[0][0].removeprefix("__label__")
+        if "_" not in lang_code:
+            lang_code = lang_code.upper()
```
sarahyurick: Similar comment as above, instead of using the underscore to decide about case sensitivity we can just not do it at all. And just use `casefold` for the comparison. Make sense?

tests/stages/text/filters/fasttext/__init__.py
sarahyurick, line 1: No copyright on an empty file please.

tests/stages/text/filters/fasttext/test_fasttext_filters.py
sarahyurick, line 1: We already have mocked fasttext tests in https://github.com/NVIDIA-NeMo/Curator/blob/main/tests/stages/text/modules/test_filters.py . I think it makes sense to move them to this new file, can you do that?

tests/stages/text/filters/__init__.py
sarahyurick, line 1: No copyright on an empty file please.

sarahyurick (changes requested):

ccyyy1023:
Hi @sarahyurick, I’ve addressed all of your review comments in fa849dbd:

- preserved the model label’s original casing and use `casefold()` only for comparisons
- removed the copyright headers from the empty `__init__.py` files
- moved the mocked FastText tests into the dedicated FastText test module
- updated the related documentation

The targeted tests pass (87 passed, 1 skipped), and I also ran the full non-GPU text test suite, with the environment-related cases rerun successfully.

Could you please take another look when you have a chance?

nemo_curator/stages/text/filters/fasttext/fasttext_filters.py
```diff
@@ -73,14 +81,14 @@ def model_check_or_download(self) -> None:
...
-    def score_document(self, text: str) -> list[float | str]:
+    def score_document(self, text: str) -> str:
         # See setup() function in modules/filter.py
         model = self._fasttext_langid_model
 
         pp = text.strip().replace("\n", " ")
         label, score = model.predict([pp], k=1)
         score = score[0][0].item()
-        lang_code = label[0][0][-2:].upper()
+        lang_code = label[0][0].removeprefix("__label__")
```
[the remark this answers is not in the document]
sarahyurick: Yes, this is fine and being documented.

sarahyurick (approved):
Thank you @ccyyy1023 !

PR #2283, opened 2026-07-30
[benchmarking] Default nightly runs to 8 GPUs

rlratzel:
## Summary
- Change the global nightly benchmark Ray default from 4 GPUs to 8 GPUs.
- Remove per-entry Ray overrides from the audio readspeech benchmarks so they inherit the new default.
- Update README examples to show the new 8-GPU default.

## Testing
- `git diff --check`
- Parsed `benchmarking/nightly-benchmark.yaml` and confirmed global `ray.num_gpus` is 8.
- Effective enabled entry counts by `num_gpus`: 13 entries use 0 GPUs, 2 entries use 1 GPU, and 26 entries use 8 GPUs.
- Confirmed no enabled entries explicitly set `num_gpus: 4`.
- pre-commit hooks passed during commit.

Closes NMCUR-198


benchmarking/nightly-benchmark.yaml
```diff
@@ -890,7 +884,7 @@ entries:
     requirements:
       # ensure the total number of documents processed is correct
       - metric: num_clips_generated
-        exact_value: 1400
+        min_value: 1400
```
praateekmahajan: keep this to exact_value and increase it because we're now using 4k input files

benchmarking/nightly-benchmark.yaml
```diff
@@ -890,7 +884,7 @@ entries:
...
       - metric: throughput_clips_per_sec
         min_value: 4.0
```
praateekmahajan: let's change this because now we have more gpus, so throughput should be faster

benchmarking/nightly-benchmark.yaml
```diff
@@ -914,7 +908,7 @@ entries:
     requirements:
       # ensure the total number of documents processed is correct
       - metric: num_clips_generated
-        exact_value: 1400
+        min_value: 1400
```
praateekmahajan: keep it to exact too but increase to actual value

benchmarking/nightly-benchmark.yaml
```diff
@@ -914,7 +908,7 @@ entries:
...
       - metric: throughput_clips_per_sec
         min_value: 2.0
```
praateekmahajan: change the throughput number to be closer to what we observe given we have more gpus

benchmarking/nightly-benchmark.yaml
```diff
@@ -939,7 +933,7 @@ entries:
     requirements:
       # ensure the total number of documents processed is correct
       - metric: num_clips_generated
-        exact_value: 1400
+        min_value: 1400
```
praateekmahajan: keep it to exact

benchmarking/nightly-benchmark.yaml
```diff
@@ -964,7 +958,7 @@ entries:
...
```
praateekmahajan: keep it exact

benchmarking/nightly-benchmark.yaml
```diff
@@ -991,7 +985,7 @@ entries:
     requirements:
       # ensure the total number of documents processed is correct
       - metric: num_clips_generated
-        exact_value: 377
+        min_value: 377
```
praateekmahajan: exact

benchmarking/nightly-benchmark.yaml
```diff
@@ -991,7 +985,7 @@ entries:
...
       - metric: throughput_clips_per_sec
         min_value: 0.25
```
praateekmahajan: update throughput

benchmarking/nightly-benchmark.yaml
```diff
@@ -980,9 +974,9 @@ entries:
       --generate-captions
       --no-generate-embeddings
       --enhance-captions
-      --video-limit=256
+      --video-limit=512
       --transcode-encoder=libopenh264
-    timeout_s: 1800
+    timeout_s: 2400
```
praateekmahajan: why is timeout increasing because we've added 2x compute and 2x the video.. so it should've run the same

benchmarking/nightly-benchmark.yaml
```diff
@@ -1034,10 +1028,10 @@ entries:
       --model-path={model_weights_path}/text_model_weights/hf_cache/models--openai--gpt-oss-20b/snapshots/6cee5e81ee83917806bbde320786a8fb61efebee
       --executor=ray_data
       --engine-kwargs='{"tensor_parallel_size": 1}'
-      --autoscaling-config='{"min_replicas": 4, "max_replicas": 4}'
-    timeout_s: 2700  # 45 min for Dynamo runtime_env setup plus 4-replica gpt-oss-20b weights load
+      --autoscaling-config='{"min_replicas": 8, "max_replicas": 8}'
+    timeout_s: 3600  # 60 min for Dynamo runtime_env setup plus 8-replica gpt-oss-20b weights load
```
praateekmahajan: did we actually see runtime_env causing such a huge spike?

benchmarking/nightly-benchmark.yaml
```diff
@@ -1053,7 +1047,7 @@ entries:
       - metric: throughput_rows_per_sec
         min_value: 2.9
```
praateekmahajan: change this metric

benchmarking/nightly-benchmark.yaml
```diff
@@ -1108,7 +1102,7 @@ entries:
     requirements:
       # ensure the total number of documents processed is correct
       - metric: num_clips_generated
-        exact_value: 113
+        min_value: 113
```
praateekmahajan: exact

benchmarking/nightly-benchmark.yaml
```diff
@@ -1250,7 +1245,7 @@ entries:
       - metric: throughput_audio_hours_per_hour
         min_value: 1.0
 
-  # The repeat entry exercises ~2.50 source-audio hours while reading EOS once.
+  # The repeat entry exercises ~10 source-audio hours while reading EOS once.
```
praateekmahajan, line 1226: stop repeating entries, create future ticket for this... it should test a real production workload

benchmarking/nightly-benchmark.yaml
```diff
@@ -1282,30 +1278,30 @@ entries:
         ping_on_failure:
           - U06J49AK7BQ  # Sushmitha Deva
     ray:
-      num_cpus: 16
-      num_gpus: 1
+      num_cpus: 64
+      num_gpus: 8
```
praateekmahajan: take it out, use the default above

benchmarking/nightly-benchmark.yaml
```diff
@@ -1282,30 +1278,30 @@ entries:
...
       enable_object_spilling: false
     requirements:
       - metric: is_success
         exact_value: true
```
praateekmahajan: remove this

benchmarking/nightly-benchmark.yaml
```diff
@@ -1282,30 +1278,30 @@ entries:
...
       - metric: input_output_row_count_match
         exact_value: true
```
praateekmahajan: remove

benchmarking/nightly-benchmark.yaml
```diff
@@ -697,6 +697,7 @@ entries:
       --split=train
       --wer-threshold=5.5
       --gpus=1
+      --asr-stage-num-workers=8
```
praateekmahajan: Don't think we need this since xenna/ray should scale to 8 available gpus automatically

benchmarking/nightly-benchmark.yaml
```diff
@@ -725,6 +726,7 @@ entries:
...
```
praateekmahajan: Don't think we need this since xenna/ray should scale to 8 available gpus automatically

benchmarking/nightly-benchmark.yaml
```diff
@@ -385,7 +385,7 @@ entries:
 
     requirements:
       - metric: num_documents_processed
-        exact_value: 312458951
+        exact_value: 168568490
```
praateekmahajan: i believe this is because semdedup outputs incorrect num_documents_processed which are sensitive to number of workers... let's create a linear ticket for this and maybe either fix it in this pr (if it's easy enough) or in a followup

praateekmahajan (changes requested):
this is great, seems good to merge after addressing reviews on the comments so far, but let's make sure as a fast-follow 
1. each entry is 10-15 minutes
2. video entries use 4k videos but they should be uniformly sampled rather than alphabetical  (codex can figure this out, with minor code change, but let's get @weijiac0619 / @suiyoubi to review it)
3. any audio entry should not be using repeat and actually doing scale testing on production data (@mohammadaaftabv)

benchmarking/nightly-benchmark.yaml
rlratzel: Agreed, I think this was more of a conservative change to allow the benchmark to finish in case it scaled poorly. It's been reverted in the next commit.

benchmarking/nightly-benchmark.yaml
rlratzel: This was another “8 replicas may take longer to start/load" conservative change. Reverted in the next commit.

benchmarking/scripts/semdedup_identification_benchmark.py
```diff
@@ -74,12 +76,16 @@ def run_semdedup_identification_benchmark(  # noqa: PLR0913
     Path(output_path).mkdir(parents=True, exist_ok=True)
     Path(cache_path).mkdir(parents=True, exist_ok=True)
 
-    logger.info("Starting semantic duplicate identification benchmark")
-    run_start_time = time.perf_counter()
+    input_files = load_dataset_files(
```
rlratzel, line 78: @praateekmahajan do you prefer the run time to include the time taken for loading the dataset files?

benchmarking/nightly-benchmark.yaml
rlratzel: I think a recent commit should have addressed this, but please review the latest changes to `semdedup_identification_benchmark.py` to see if it's what you were thinking.

rlratzel:
> this is great, seems good to merge after addressing reviews on the comments so far, but let's make sure as a fast-follow
> 
>     1. each entry is 10-15 minutes
> 
>     2. video entries use 4k videos but they should be uniformly sampled rather than alphabetical  (codex can figure this out, with minor code change, but let's get @weijiac0619 / @suiyoubi to review it)
> 
>     3. any audio entry should not be using repeat and actually doing scale testing on production data (@mohammadaaftabv)

I've created the following issues in Linear:

[NMCUR-248](https://linear.app/nvidia/issue/NMCUR-248/increase-benchmark-workloads-so-entries-run-roughly-10-15-minutes)
[NMCUR-249](https://linear.app/nvidia/issue/NMCUR-249/uniformly-sample-video-benchmark-inputs-instead-of-alphabetical)
[NMCUR-250](https://linear.app/nvidia/issue/NMCUR-250/replace-repeated-audio-tagging-workload-with-real-production-scale)
[NMCUR-251](https://linear.app/nvidia/issue/NMCUR-251/investigate-video-embedding-8-gpu-underutilization-and-throughput)

benchmarking/scripts/task_metrics.py
```diff
@@ -0,0 +1,63 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from collections.abc import Mapping, Sequence
+from typing import Protocol
+
+
+class _StagePerfLike(Protocol):
```
praateekmahajan: We have this file https://github.com/NVIDIA-NeMo/Curator/blob/main/nemo_curator/tasks/utils.py

Can you confirm this isn't repetitive and let's reuse what we can, at the least let's move this code there and add a corresponding test inside test_utils.py

benchmarking/scripts/semdedup_identification_benchmark.py
praateekmahajan: after is good!

benchmarking/scripts/task_metrics.py
rlratzel: Thanks for pointing that out…I'll TAL.

benchmarking/nightly-benchmark.yaml
rlratzel: Created [NMCUR-250](https://linear.app/nvidia/issue/NMCUR-250/replace-repeated-audio-tagging-workload-with-real-production-scale)

benchmarking/scripts/task_metrics.py
rlratzel: We still needed the ability to get unique stage stats, but it's been refactored to move only that utility into `TaskPerfUtils`, update the existing tests, and remove `task_metrics.py` since it's no longer needed.  Thanks!

benchmarking/scripts/semdedup_identification_benchmark.py
```diff
@@ -92,13 +97,23 @@ def run_semdedup_identification_benchmark(  # noqa: PLR0913
...
+
     # Run the workflow, extract metrics from the WorkflowRunResult object
     executor_obj = setup_executor(executor)
     workflow_run_result = workflow.run(pairwise_executor=executor_obj)
 
     run_time_taken = time.perf_counter() - run_start_time
     task_metrics = TaskPerfUtils.aggregate_task_metrics(workflow_run_result)
 
+    num_documents_processed = int(
+        TaskPerfUtils.get_unique_stage_stat(
```
praateekmahajan: Sorry @rlratzel I saw your fix, but I realised maybe a more central fix would be better.
I did that in a branch I was working on here https://github.com/praateekmahajan/Curator/commit/e00f74e1905ae1b864cec64c98f8eda24a2cc2fd

praateekmahajan (approved):
LGTM

benchmarking/scripts/semdedup_identification_benchmark.py
rlratzel: ok thanks. I just cherry-picked those change into the latest commit here.

benchmarking/nightly-benchmark.yaml
```diff
@@ -172,7 +172,7 @@ gpu_mem_use_warning_threshold: 0.01
 # Global ray settings inherited by all entries. Per-entry ray sections override these values.
 ray:
   num_cpus: 64
```
praateekmahajan: Sorry can you also change this to 128 cpus?

PR #2293, opened 2026-08-04
feat(ray-data): add scheduler observability

praateekmahajan:
Adds an opt-in Ray 2.56.1 runtime shim that writes Ray Data scheduling decisions to session logs without requiring a patched wheel.

Tests: 26 Ray Data backend tests passed in Docker.

PR #2295, opened 2026-08-05
feat(agents): Ray Data scheduler reference in AGENTS.md (nmcur-223)

weijiac0619:
## Summary

- Adds a root `AGENTS.md` with Curator's core abstractions and repository-wide guidance for optional dependencies, Fern docs, durable comments, code reuse, traceable and realistic tests, Conventional Commits, and pre-commit checks.
- Adds `nemo_curator/backends/ray_data/AGENTS.md` with a log-first workflow for diagnosing and tuning Ray Data pipelines.
- Explains Curator's execution model—one opaque `Task` row per Ray block—and the implications for task sizing, batching, and Ray block-level optimizations.
- Documents the scheduler signals and tuning controls that matter for resource admission, downstream backpressure, actor autoscaling, concurrency, metadata fetching, and operator fusion.
- Uses Ray 2.57.0 as the supported baseline and calls out version-sensitive defaults and private APIs that must be rechecked when Ray changes.

## Validation

- Repository pre-commit hooks passed for the documentation commits.
- `git diff --check` passed.
- No runtime tests were run because this PR changes agent documentation only.

## Dependency

The structured scheduler log events described here depend on the diagnostics shim in #2293; the remaining guidance is standalone.


AGENTS.md
```diff
@@ -0,0 +1,357 @@
...
+   to borrow from upstream's share when the downstream budget would fall below
+   its minimum scheduling requirement.
+
+3. **Budget per operator:** `remaining_reserved + share_of_shared_pool`. An
+   operator that has consumed beyond its reserved floor reduces the remaining
+   shared pool for other operators on that cycle.
+
+The reservation system can be disabled entirely with
+`DataContext.get_current().op_resource_reservation_enabled = False`
+(`RAY_DATA_ENABLE_OP_RESOURCE_RESERVATION`).
```
praateekmahajan: IMO we shouldn't recommend this since this is a deprecated codepath inside Ray code

AGENTS.md
```diff
@@ -0,0 +1,357 @@
...
+events. The shim is **opt-in**: set `NEMO_CURATOR_RAY_DATA_DIAGNOSTICS=1` (or `true`,
+`yes`, `on`) before starting the pipeline. Without this env var the shim is silently
+disabled and no events are written. See `nemo_curator/backends/ray_data/diagnostics.py`.
+
+When enabled, events are routed through the `ray.data` logger to:
+
+```
+$RAY_TEMP_DIR/session_latest/logs/ray-data/ray-data.log
+```
+where `RAY_TEMP_DIR` defaults to `~/.ray` unless overridden by `_temp_dir` in `ray.init()`.
```
praateekmahajan: instead of ray.init() we should say `RayClient / SlurmRayClient` since that's how Curator pipelines do.. so maybe `RayClient / SlurmRayClient / ray.init()`

AGENTS.md
```diff
@@ -0,0 +1,357 @@
...
+### 5. Stage fusion requires a TaskPool upstream — ActorPool→ActorPool never fuses
+
+From `FuseOperators._can_fuse()`, the upstream operator must be a
+`TaskPoolMapOperator`. `ActorPoolMapOperator → ActorPoolMapOperator` is not a
+supported fusion pattern. Fusion is possible for:
+- `TaskPool → TaskPool` (when remote args match)
+- `TaskPool → ActorPool` (upstream task feeds into an actor stage)
+
+When fusion occurs, the two operators collapse into one, reducing
+`num_eligible_ops` and increasing `reserved_per_op` for all remaining operators.
```
praateekmahajan: nit : reducing x and increasing Y for all remaining operators which can lead to Z

AGENTS.md
```diff
@@ -0,0 +1,357 @@
...
+Default `map_batches` with a plain function (not a callable class) uses TaskPool
+and can participate in fusion. Callable classes default to `ActorPoolStrategy(min_size=1, max_size=None)`
+automatically, which is ActorPool and cannot be fused as an upstream stage.
```
praateekmahajan: Again we should make this Curator specific that the adapter decides if something is Task or Actor and can be overridden. Task as long no setup / state (like model weights) is needed

AGENTS.md
```diff
@@ -0,0 +1,357 @@
...
+even though it returns the CPU to the shared pool.
+
+---
+
+## Tuning knobs
+
+| Knob | Where to set | What it controls |
+|---|---|---|
+| `op_resource_reservation_ratio` | `DataContext.get_current().op_resource_reservation_ratio` or `RAY_DATA_OP_RESERVATION_RATIO` | Fraction of resources reserved per operator vs. shared pool (default `0.5`) |
+| `op_resource_reservation_enabled` | `DataContext.get_current().op_resource_reservation_enabled` or `RAY_DATA_ENABLE_OP_RESOURCE_RESERVATION` | Enable/disable the reservation system entirely (default `True`) |
```
praateekmahajan: I don't think this should be tuned

AGENTS.md
```diff
@@ -0,0 +1,357 @@
...
+| `actor_pool_max_upscaling_delta` | `DataContext.get_current().autoscaling_config.actor_pool_max_upscaling_delta` or `RAY_DATA_DEFAULT_ACTOR_POOL_MAX_UPSCALING_DELTA` | Max actors added per scaling tick (default `1`) |
+| `min_size` / `max_size` | `ActorPoolStrategy(min_size=N, max_size=N)` | Actor pool lower and upper bounds |
+| `initial_size` | `ActorPoolStrategy(min_size=m, max_size=n, initial_size=k)` | Pre-create `k` actors at startup; defaults to `min_size` if not specified (must satisfy `min_size <= initial_size <= max_size`) |
+| `compute` | `map_batches(..., compute=ActorPoolStrategy(...))` | Compute strategy; preferred over the deprecated `concurrency` argument |
+| `concurrency` | `map_batches(..., concurrency=N)` or `map_batches(..., concurrency=(min, max))` | **Deprecated since Ray 2.51** — use `compute=` instead. For callable classes: int → fixed pool, tuple → (min, max) or (min, max, initial). For plain functions: int → TaskPool size. |
```
praateekmahajan:
```suggestion
```

Don't think we need to mention compute (decided by curator's adapter) or concurrency (deprecated).. so what we can recommend is whether a stage is actor vs task

AGENTS.md
```diff
@@ -0,0 +1,357 @@
...
+CPU stays available — but the redistribution depends on operator position in the
+topology due to the reversed iteration order in `update_budgets`. The net effect
+is not simply "invisible": it dilutes the per-op `default_reserved` calculation
+even though it returns the CPU to the shared pool.
+
+---
+
+## Tuning knobs
+
+| Knob | Where to set | What it controls |
```
praateekmahajan: In the where to set column given the end user never interacts with raw RayData code we should point them to set how to do it in Curator code like modify Resources or RayStageSpecKeys

AGENTS.md
```diff
@@ -0,0 +1,357 @@
...
+| `max_concurrency` | `@ray.remote(max_concurrency=N)` on actor class, or `map_batches(..., max_concurrency=N)` kwarg | Max concurrent tasks per actor; affects the utilization denominator (default `1`) |
+| `max_tasks_in_flight_per_actor` | `ActorPoolStrategy(max_tasks_in_flight_per_actor=N)` or `DataContext.get_current().max_tasks_in_flight_per_actor` | Max tasks queued per actor (default `2 × max_concurrency`); affects utilization and task pipelining |
+| `downstream_capacity_backpressure_ratio` | `DataContext.get_current().downstream_capacity_backpressure_ratio` or `RAY_DATA_DOWNSTREAM_CAPACITY_BACKPRESSURE_RATIO` | Queue/capacity ratio threshold for downstream capacity backpressure (default `10.0`) |
+| `target_max_block_size` | `DataContext.get_current().target_max_block_size` | Maximum block size in bytes (default `128 * 1024 * 1024` = 128MB). Smaller blocks reduce peak object store usage between stages but increase scheduling overhead. Critical for pipelines with large per-item payloads (video frames, embeddings). |
```
praateekmahajan: I don't think this affects curator at all, since we pass Tasks and batch_size

AGENTS.md
```diff
@@ -0,0 +1,357 @@
...
+| Object store memory | `ray.init(object_store_memory=N)` | Total object store memory cluster-wide; affects pending output budgets |
```
praateekmahajan: We should have for object store threshold, by default ray data reserves 0.5, but we might want more or less.
Also we should show document how heap memory plays a role here w.r.t object store

AGENTS.md
```diff
@@ -0,0 +1,357 @@
...
+### 4. Scale-up rate is bounded by `WAIT_FOR_TASK_COMPLETION_TIMEOUT_S = 0.1s`
+
+The scheduling loop blocks on `ray.wait()` for up to 100ms per iteration.
+With `actor_pool_max_upscaling_delta=1` (default), at most one actor is added
+per iteration. Maximum scale-up rate is therefore 1 actor per 100ms. Ramping
+from 1 to 96 actors takes at minimum ~9.5s of scheduling time (in practice
+~20–40s with overhead). This is why a fast upstream stage can reach its budget
+cap well before a downstream stage receives its first item.
+
+### 5. Stage fusion requires a TaskPool upstream — ActorPool→ActorPool never fuses
```
praateekmahajan: Let's also add a note for Task vs Actor autoscaling differences

AGENTS.md
```diff
@@ -0,0 +1,357 @@
...
+This document describes Ray Data's scheduling and resource allocation behaviors
+as they apply to NeMo Curator pipelines. It is reference material for agents
+diagnosing performance or underutilization issues — not a prescriptive tuning
+guide, since the right decisions depend on pipeline shape and infrastructure.
+
+**Ray version:** Behaviors, defaults, env var names, and class names were verified
+against Ray **2.56.0** source (the version available in the dev environment).
+The NeMo Curator diagnostics shim (`diagnostics.py`) is version-locked to
+**2.56.1** and will not install on other versions — if any internal APIs changed
+between 2.56.0 and 2.56.1, the documented behavior may differ slightly from
```
praateekmahajan: Let's only focus on 2.56.1 and not 2.56.0 since Ray is pinned to 2.56.1

AGENTS.md
```diff
@@ -0,0 +1,357 @@
...
+operator's reserved and shared CPU allocation to 0. It never consumes CPU
+budget.
+
+However, it IS included in `num_eligible_ops`, which shrinks
+`default_reserved = total * ratio / num_eligible_ops` for all operators. Because
+the 0-cpu op takes nothing from either the reserved or shared pool, the extra
+CPU stays available — but the redistribution depends on operator position in the
+topology due to the reversed iteration order in `update_budgets`. The net effect
+is not simply "invisible": it dilutes the per-op `default_reserved` calculation
+even though it returns the CPU to the shared pool.
```
praateekmahajan: Let's add a note if we have pipeline where we have only 1 stage that needs GPUs and a cluster with 8 gpus, that we're better off dedicating all 8 gpus to that cluster. We can do so using size=8 or we can even do initial_size=8 however with initial_size the autoscaler will start to downscale again until input becomes available to it based on the downscaling threshold

AGENTS.md
```diff
@@ -0,0 +1,357 @@
...
+| `configured_ratio` | The `backpressure_capacity_ratio` threshold |
+| `utilized_object_store_budget_fraction` | Object store utilization fraction (must exceed 0.5 for this policy to fire) |
+
+---
+
+## General rules
+
+These are structural invariants that hold regardless of pipeline shape, derived
+directly from the scheduler source. An agent can rely on them when reasoning
+about a specific pipeline.
```
praateekmahajan: Let's also add a section on memory (object store / ray data object store utilization / heap memory)

praateekmahajan (dismissed):

AGENTS.md
weijiac0619: good catch

AGENTS.md
```diff
@@ -0,0 +1,41 @@
+# AGENTS.md — NeMo Curator
+
+NeMo Curator is a scalable library for preparing multimodal datasets. Pipelines
+are composed of `ProcessingStage` objects executed by a backend (Ray Data, Xenna,
+or Slurm) over streams of `Task` objects.
```
sarahyurick: Why is Slurm being grouped with Ray Data and Xenna?

AGENTS.md
```diff
@@ -0,0 +1,41 @@
...
+| `Task` | `nemo_curator/tasks/` | Data item flowing through the pipeline |
+| `Pipeline` | `nemo_curator/pipeline/pipeline.py` | Ordered sequence of stages executed by a backend |
+| `RayClient` / `SlurmRayClient` | `nemo_curator/core/client.py` | Cluster connection and Ray init |
+
+## Key rules
+
+- **Optional extras**: feature families are behind extras in `pyproject.toml`.
+  Do not make heavyweight dependencies unconditional.
+- **Fern docs**: user-facing documentation lives in `fern/`, not `docs/`. Edit
+  MDX files there; do not add docs to the `docs/` directory.
```
sarahyurick, line 24: The docs directory should be completed deleted now no?

nemo_curator/backends/ray_data/AGENTS.md
```diff
@@ -0,0 +1,340 @@
+# Ray Data Backend — Agent Tuning Guide
+
+Use this guide when diagnosing or tuning a NeMo Curator pipeline running on the
+Ray Data backend. Start from measured scheduler behavior; do not apply a setting
+only because it helped another pipeline.
+
+**Supported baseline:** Curator requires Ray 2.57.0 or later. The diagnostic shim
+in `diagnostics.py` targets exactly Ray 2.57.0 unless the installed Ray version
```
sarahyurick, line 8: This might be hard to maintain since we are continuously bumping versions. Like I don't think this file should always have to be updated every time we do a Ray bump?

nemo_curator/backends/ray_data/AGENTS.md
```diff
@@ -0,0 +1,340 @@
...
+collected back to the driver.
+
+### Task stages and actor stages
+
+`RayDataStageAdapter` applies every stage with `map_batches` and chooses the compute
+strategy as follows:
+
+| Stage | Curator selection | Worker behavior |
+|---|---|---|
+| Actor | Overrides `setup()`, or requests both CPU and GPU; can be forced with `IS_ACTOR_STAGE=True` | Persistent state and `ActorPoolStrategy`; fixed or autoscaling pool |
```
sarahyurick, line 47: Just wanted to check is this actually true that it will always be an actor stage when setup is overridden? I thought that was only for certain stages but maybe I am wrong.

nemo_curator/backends/ray_data/AGENTS.md
praateekmahajan: I see where you're coming from, however the issues is this contains some Ray Data internals which might change version to version and the diagnostics.py is indeed version bound. IOW I think we must change this (let's try that for a release or two and see if it's too much or if this AGENTS.md is even valuable)

nemo_curator/backends/ray_data/AGENTS.md
praateekmahajan: Yes this is true https://github.com/NVIDIA-NeMo/Curator/blob/3cab11281e3ab4c72b79ff98b047dddb54f3dc44/nemo_curator/backends/ray_data/utils.py#L67

AGENTS.md
praateekmahajan: Bad slop sorry!

AGENTS.md
praateekmahajan: Question for @lbliii I guess

AGENTS.md
```diff
@@ -0,0 +1,41 @@
...
+- **Reuse before adding**: search Curator for existing implementations,
+  utilities, and patterns before writing new code; reuse or extend them when
+  they fit.
+- **Tests**: keep unit-test files in a one-to-one mapping with source files so
+  coverage is easy to trace. Test user-visible behavior and real integration
+  boundaries; avoid tests that merely enumerate defaults or mock away the
+  dependency being validated. Prefer the narrowest scope that provides
+  confidence: unit, then integration, then GPU. GPU tests must be registered
+  separately.
+- **Commits and PRs**: use Conventional Commits style for commit messages and
```
ayushdg, line 36: Should we also add signing/signoff?

AGENTS.md
```diff
@@ -0,0 +1,41 @@
...
+|---|---|---|
+| `ProcessingStage` | `nemo_curator/stages/base.py` | Unit of work: defines `process_batch()`, `resources`, and optional `setup()` for stateful stages |
+| `Resources` | `nemo_curator/stages/resources.py` | Per-stage CPU/GPU requirements (`cpus`, `gpus`, `gpu_memory_gb`) |
+| `Task` | `nemo_curator/tasks/` | Data item flowing through the pipeline |
+| `Pipeline` | `nemo_curator/pipeline/pipeline.py` | Ordered sequence of stages executed by a backend |
+| `RayClient` / `SlurmRayClient` | `nemo_curator/core/client.py` | Cluster connection and Ray init |
+
+## Key rules
+
+- **Optional extras**: feature families are behind extras in `pyproject.toml`.
```
ayushdg, line 21: Another one to use the venv where Curator is installed rather than current env. 

idk if there's a strong pref but agents also try to do uv run or python prefix hackery

ayushdg (commented):
non blocking comments

AGENTS.md
sarahyurick: Hi, I don't see any new commits yet?

AGENTS.md
praateekmahajan: I added pre-commit for that reason since pre-commit catches signoff right? (I also wanted to add ruff / signoff but realised pre-commit captures both)

sarahyurick (approved):
Thanks!

PR #2302, opened 2026-08-11
Increase text benchmark workloads

sarahyurick:
Summary of changes:

| Benchmark | Legacy (s) | New (s) | Change |
|---|---|---|---|
| arxiv_e2e_pipeline_raydata | 388.5 | 714.8 | +1.8x |
| arxiv_e2e_pipeline_xenna | 376.2 | 582.9 | +1.5x |
| domain_classification_raydata | 369.4 | 809.9 | +2.2x |
| domain_classification_xenna | 277.4 | 800.8 | +2.9x |
| embedding_generation_raydata | 311.7 | 899.2 | +2.9x |
| embedding_generation_xenna | 213.1 | 767.8 | +3.6x |
| exact_dedup_identification | 367.7 | 365.5 | ~same |
| fasttext_filter_raydata | 48.4 | 913.6 | +18.9x |
| fasttext_filter_xenna | 30.6 | 751.1 | +24.5x |
| modifier_raydata | 21.7 | 746.6 | +34.4x |
| modifier_xenna | 28.5 | 854.0 | +30.0x |
| ndd_dynamo_dp8 | 106.0 | 1389.9 | +13.1x |
| ndd_ray_serve_dp8 | 107.1 | 1360.8 | +12.7x |
| score_filter_raydata | 48.3 | 881.7 | +18.3x |
| score_filter_xenna | 107.5 | 1053.0 | +9.8x |
| semdedup_identification_xenna | 181.8 | - | split into 2 |
| semdedup_identification_xenna_dataset_size_ratio | - | 277.9 | new |
| semdedup_identification_xenna_fit_data_fraction | - | 1424.6 | new |

Stability analysis:

- `arxiv_e2e_pipeline_raydata`: CPU stabilizes quickly (~60s) and holds steady, but the GPU classifier stages never fully saturate - GPU count fluctuates between 2-4 the entire run because the CPU pipeline can't feed them fast enough to keep all 4 actors busy. Needs more data or faster CPU stages to drive the GPUs consistently.
- `arxiv_e2e_pipeline_xenna`: GPUs lock in at 3 actors after ~240s and hold steady for the remaining 360s. CPU usage swings more but that's just batch size variation, not reallocation.
- `domain_classification_raydata`: GPUs ramp up steadily over ~250s as Ray Data scales classifier actors one at a time until all 8 are loaded, then stay rock solid for the remaining ~540s. Good steady state.
- `domain_classification_xenna`: All 8 GPUs loaded within 60s and stay there for 92% of the run. CPU fluctuates but that's just reader/tokenizer workers adjusting to keep the GPU fed - the allocation itself is stable.
- `embedding_generation_raydata`: Similar to `domain_classification` - GPUs ramp over ~250s then hold steady for ~580s with very little fluctuation. Good steady state.
- `embedding_generation_xenna`: All 8 GPUs loaded within 60s and stable for 92% of the run. Clean.
- `fasttext_filter_raydata`: Fast ramp (~80s) to ~96 CPUs, holds well for most of the run, then drops to ~70 CPUs in the final ~400s as workers drain. No mid-run instability - the drop is just wind-down starting earlier than expected.
- `fasttext_filter_xenna`: Very stable after ~240s ramp with only small CPU fluctuations (~2-3%) across 540s of steady state. One of the cleanest benchmarks.
- `modifier_raydata`: Ramp takes ~460s because Ray Data scales from ~96 to 128 CPUs halfway through - a genuine mid-run scale-up as the pipeline discovers it can use more workers. Only ~280s of true steady state; could benefit from more data.
- `modifier_xenna`: Stable after 240s ramp with small fluctuations (~4%) for 600s. Good.
- `ndd_dynamo_dp8`: Essentially perfectly stable - 32 CPUs pinned from the first 10 seconds and held for ~1200s straight. Throughput fluctuates somewhat due to the LLM server response times, but resource allocation never changes.
- `ndd_ray_serve_dp8`: Same story as Dynamo - 32 CPUs from the start, held for the entire run. Ray Serve shows slightly more throughput variance than Dynamo but resource allocation is equally stable.
- `score_filter_raydata`: Fast ramp (~130s), then 128 CPUs held solidly for ~740s with very little fluctuation (~4%). One of the most stable benchmarks.
- `score_filter_xenna`: Stable after 240s ramp with 840s of steady state. Slightly more fluctuation than the Ray Data version (~12%) because Xenna's 30 `ScoreFilter` stages cause minor ongoing worker rebalancing, but nothing dramatic.
- `semdedup_dataset_size_ratio`: Too short (~2 min total) to draw any conclusions - it barely completes one autoscale cycle before finishing.
- `semdedup_fit_fraction`: GPUs pin to 8 actors within 60s and stay there for 93% of the run. CPU varies more reflecting the phase transition from similarity computation to dedup identification, but the GPU allocation is solid.


benchmarking/scripts/ndd_benchmark.py
```diff
@@ -305,30 +222,13 @@ def run_ndd_benchmark(  # noqa: PLR0915
...
             inference_server.stop()
 
-    # -- Post-run: extract metrics from _stage_perf ----------------------
-    input_row_count = int(
-        TaskPerfUtils.get_aggregated_stage_stat(output_tasks, "DataDesignerStage", "custom.num_input_records")
-    )
     output_row_count = int(
         TaskPerfUtils.get_aggregated_stage_stat(output_tasks, "DataDesignerStage", "custom.num_output_records")
     )
-    input_tokens_median_per_record = float(
```
praateekmahajan, line 315: Is there no way to get input/ouput tokens, because that'll be something to compare bw dynamo vs ray serve

benchmarking/nightly-benchmark.yaml
```diff
@@ -385,11 +373,50 @@ entries:
...
+        ping_on_failure:
+          - U07JL5K0L10  # Praateek Mahajan
+          - U022P8CDX40  # Sarah Yurick
+
+    requirements:
+      - metric: num_documents_processed
+        exact_value: 329774475
+      - metric: num_duplicates
+        min_value: 10000000
+        max_value: 11000000
```
praateekmahajan, line 432: how come the variance is different bw this and the previous run?

sarahyurick: It is on the 1M scale vs the 10M scale, no strong reason for the differences between the max and min values. I can make the ranges smaller if you prefer.

benchmarking/scripts/ndd_benchmark.py
sarahyurick: I had deleted them while testing but let me see if we can re-add. Agree it would be nice to keep these.

praateekmahajan (approved):
thankyou!!

PR #2308, opened 2026-08-13
docs: add Hugging Face Jobs variant of the Slurm array tutorial

davanstrien:
## Description

Runs the `tutorials/slurm` array workflow on [Hugging Face Jobs](https://huggingface.co/docs/huggingface_hub/guides/jobs) with `array_pipeline.py` and `retry_array.py` unmodified: the env-driven array sharding contract (`NEMO_CURATOR_SLURM_ARRAY_*`, #2059) is scheduler-agnostic, so only the scheduler-side plumbing changes. An HF bucket mounted in every Job plays the shared-filesystem role; completion manifests land there, so retry discovery works unchanged.

Adds `tutorials/hf_jobs/`, i.e.  two new files, a README, and a one-line cross-reference from the Slurm README:

- `run_shard.py` — per-shard shim, the analog of `submit_array.sh`. Installs `nemo-curator==1.3.0` from PyPI via PEP 723 metadata and fetches `array_pipeline.py` at the matching `v1.3.0` tag (sha256-verified), since `hf jobs uv run` uploads a single file. Absorbs three Jobs-container quirks (documented in the README) so the pipeline file stays clean.
- `launch_array.sh` — fires one independent Job per shard; same env contract, same retry semantics as `submit_array.sh`.

The array pattern's deterministic shard assignment plus durable completion manifests make it portable to any scheduler that can run K independent containers against shared storage so this tutorial demonstrates that portability and is a way to try the array workflow before (or without) provisioning Slurm.

**Testing** I ran two-wave run of the tutorial as written, using its own sample-data snippet to test the tutorial e2e. 

I also ran this approach at larger scale using `FineWebEduClassifier` over ~400k FineWeb rows on GPU shards, including a retry wave that recovered a lost shard.  Output here: https://huggingface.co/datasets/davanstrien/fineweb-edu-showcase.

## Usage

```bash
# put JSONL files on a bucket (see README for a sample-data generator)
hf buckets create <user>/<bucket>
hf buckets sync ./my-jsonl-dir hf://buckets/<user>/<bucket>/tutorial/input

# launch 3 shards, one Job each
export BUCKET=hf://buckets/<user>/<bucket>
bash tutorials/hf_jobs/launch_array.sh 3

# discover + relaunch incomplete shards (retry_array.py, unmodified)
bash tutorials/hf_jobs/launch_array.sh 3 2
```

## Checklist

- [x] New or Existing tests cover these changes. (Docs/tutorial-only — no package code touched; the tutorial itself was run end-to-end from this branch, see Testing above.)


davanstrien:
sorry for the delay @meatybobby. I think I've addressed the issues Greptile was flagging now!

PR #2310, opened 2026-08-13
Update video benchmark

meatybobby:
## Description
Update video benchmark in pipeline

## Usage

```python
```
## Checklist


benchmarking/nightly-benchmark.yaml
```diff
@@ -1049,7 +1049,7 @@ entries:
       --generate-captions
       --no-generate-embeddings
       --enhance-captions
-      --video-limit=512
+      --video-limit=256
```
praateekmahajan: why are we reducing this? lets keep it the same

benchmarking/nightly-benchmark.yaml
```diff
@@ -1022,7 +1022,7 @@ entries:
...
```
praateekmahajan: why are we reducing this? lets keep it the same

benchmarking/nightly-benchmark.yaml
meatybobby: It may be timeout over new sample videos.

praateekmahajan: let's keep it to 512 and increase timeout to 30 mins or so

praateekmahajan (approved):

PR #2312, opened 2026-08-13
Increase Image benchmark pipeline duration to ~10min

suiyoubi:
## Summary

Increase the `image_curation_raydata` and `image_curation_xenna` workloads so their benchmark runtimes target approximately 10–15 minutes.

This PR intentionally changes only workload sizing, dataset selection, and the corresponding benchmark requirements. Ray Data performance tuning and Xenna/Ray Data throughput parity are out of scope and will be handled separately.

## Changes

- Use the fixed EOS WebDataset:
  - `image_curation/mint1t_unique_jpeg_1m_v1`
  - Contains 1 million unique JPEG images in deterministic 1,000-image shards.
- Process:
  - Ray Data: 375 shards / 375,000 images
  - Xenna: 850 shards / 850,000 images
- Add `--input-partition-limit` to select a deterministic subset without duplicating work.
- Correct `num_images_processed` collection:
  - The writer returns output files, so counting `len(task.data)` counted files rather than images.
  - The benchmark now uses embedding-stage item metrics to count processed images.
- Report `num_images_written` and `num_output_files`.
- Update benchmark requirements:
  - Ray Data:
    - `num_images_processed == 375000`
    - `throughput_images_per_sec >= 350`
  - Xenna:
    - `num_images_processed == 850000`
    - `throughput_images_per_sec >= 900`
- Keep the existing 1,500-second timeouts. The measured/projected runtimes fit within them, so there is no evidence that a timeout increase is needed.
- Do not add executor-specific worker counts, stage fusion, GPU allocation, or other performance-tuning arguments.

## Rationale

The original MSCOCO workload contains only 3,800 images and completed too quickly:

| Entry | Before runtime |
|---|---:|
| `image_curation_raydata` | 321 s / 5.35 min |
| `image_curation_xenna` | 268 s / 4.47 min |

The larger dataset provides a realistic workload using unique images instead of synthetic repetition.

Ray Data and Xenna currently have different measured throughput, so they require different image counts to reach approximately the same runtime. Matching their throughput is a separate performance investigation and is intentionally not addressed here.

## Local validation

- `ruff check` passed
- `ruff format --check` passed
- Python compilation passed
- Benchmark YAML/session parsing passed
- 49 focused unit tests passed


benchmarking/nightly-benchmark.yaml
```diff
@@ -669,27 +633,30 @@ entries:
...
     script: image_pipeline_benchmark.py
     args: >-
       --benchmark-results-path={session_entry_dir}
-      --input-wds-dataset-dir={dataset:mscoco,wds}
+      --input-wds-dataset-dir={dataset:image_curation_mint1t,wds}
       --output-dataset-dir={session_entry_dir}/scratch/output
       --model-dir={dataset:mscoco_model_weights,files}
       --executor=xenna
       --tar-files-per-partition=1
+      --input-partition-limit=850
```
sarahyurick, line 700: So Xenna is 2-3x faster?

sarahyurick (commented):
LGTM, thanks.

benchmarking/nightly-benchmark.yaml
praateekmahajan: We should keep both ray data and xenna to run on the same dataset. It's fine if Ray Data is 2-3x slower.

praateekmahajan (dismissed):
let's keep xenna/ray data to run on the same dataset size

benchmarking/nightly-benchmark.yaml
suiyoubi: Yes Xenna is much faster. Will need to optimize the raydata on this in a follow-up PR.

benchmarking/nightly-benchmark.yaml
```diff
@@ -623,13 +623,15 @@ entries:
...
-      --input-partition-limit=375
+      --input-partition-limit=850
       --batch-size=1000
       --embedding-batch-size=500
       --aesthetic-batch-size=500
       --aesthetic-threshold=0.9
       --verbose
-    timeout_s: 1500
+    # The 375k-image calibration took 614.84s end-to-end; scaling to 850k
+    # projects to about 23 minutes, so allow measured-variance headroom.
```
praateekmahajan: Let's remove this comment since its super local to this commit

praateekmahajan (approved):

PR #2317, opened 2026-08-17
feat(text): bound vLLM embedding generation

praateekmahajan:
## Description

**Main Change** - Add `model_inference_batch_size` support for vLLM too. Otherwise it generates a gigantic tokenized column for all texts which can result in CPU OOM. Now we only tokenize what we feed into the LLM. The output returned is a pyarrow table (earlier it was a pd.DataFrame with a embedding column which was list[float] i.e. fp64. Now we do fp32.

Also related to OOM: we don't now persist text column always

**Other changes**
1. Use `create_vllm_llm_with_retry` inside setup
2. Add `metadata_fields` for embedding (removed text as the default column that gets preserved)

Linear: [NMCUR-240](https://linear.app/nvidia/issue/NMCUR-240)

## Validation

- 27 focused embedder, utility, and workflow-configuration tests passed
- 12 semantic-dedup workflow tests passed across Xenna and Ray Data
- Pre-commit passed

## Benchmark comparison

Compared [nightly baseline](http://rratzel-ws1:5050/run-viewer?dir=login-eos-nightly%3A%2Flustre%2Ffsw%2Fcoreai_dlalgo_ci%2Fcurator_ci%2Fresults%2Fmain&run=nightly-2026_08_20__06_39_17_UTC) (`ae713b5e`) with the [PR run](http://rratzel-ws1:5050/run-viewer?dir=login-eos-adhoc%3A%2Flustre%2Ffsw%2Fcoreai_dlalgo_ci%2Fcurator_ci%2Fresults%2Fmain-mirror&run=pr-2317-2026_08_21__00_38_20_UTC-878cf463) (`878cf463`). Both entries succeeded and processed the same 5,807,132 documents. Time values are seconds; delta is `(PR - nightly) / nightly`, so negative time deltas are improvements.

| Executor | Runtime metric | Nightly | PR | Delta |
|---|---|---:|---:|---:|
| Ray Data | `time_taken_s` | 901.925 | 847.189 | -6.07% |
| Ray Data | `throughput_docs_per_sec` | 6,438.60 | 6,854.59 | +6.46% |
| Ray Data | vLLM `process_time_mean` | 2.962 | 2.850 | -3.79% |
| Ray Data | custom `tokenization_time_mean` | 0.388 | 0.543 | +39.78% |
| Ray Data | custom `vllm_embedding_time_mean` | 2.417 | 2.461 | +1.81% |
| Xenna | `time_taken_s` | 768.562 | 743.618 | -3.25% |
| Xenna | `throughput_docs_per_sec` | 7,555.84 | 7,809.30 | +3.35% |
| Xenna | vLLM `process_time_mean` | 3.031 | 2.910 | -4.00% |
| Xenna | custom `tokenization_time_mean` | 0.413 | 0.560 | +35.44% |
| Xenna | custom `vllm_embedding_time_mean` | 2.454 | 2.510 | +2.31% |

Overall end-to-end runtime improved by 6.1% on Ray Data and 3.2% on Xenna, while embedding-stage `process_time_mean` improved by about 4% for both executors. The custom tokenization mean increased, but its overlap with vLLM inference keeps the aggregate stage time lower. This is a single-run comparison on different EOS hosts (`eos0288` before and `eos0183` after), so modest differences should be treated as directional.

### DIFF
--- a/nemo_curator/stages/text/embedders/vllm.py
+++ b/nemo_curator/stages/text/embedders/vllm.py
@@ -17,7 +17,7 @@ from __future__ import annotations
 import gc
 import time
 from concurrent.futures import ThreadPoolExecutor
-from typing import TYPE_CHECKING, Any, cast
+from typing import TYPE_CHECKING, Any
 
 import numpy as np
 import pyarrow as pa
@@ -184,13 +184,11 @@ class VLLMEmbeddingModelStage(ProcessingStage[DocumentBatch, DocumentBatch]):
 
         from vllm.inputs import TokensPrompt
 
-        tokenizer = cast("AutoTokenizer", self.tokenizer)
-        model = cast("LLM", self.model)
         t0 = time.perf_counter()
-        tokenized_data = tokenizer.batch_encode_plus(
+        tokenized_data = self.tokenizer.batch_encode_plus(
             input_data,
             truncation=True,
-            max_length=model.model_config.max_model_len,
+            max_length=self.model.model_config.max_model_len,
         )
         prompts = [TokensPrompt(prompt_token_ids=ids) for ids in tokenized_data.input_ids]
         return prompts, time.perf_counter() - t0
@@ -238,9 +236,8 @@ class VLLMEmbeddingModelStage(ProcessingStage[DocumentBatch, DocumentBatch]):
             yield pending_offset, pending_chunk_size, input_data, tokenization_time
 
     def _embed_chunk(self, input_data: list[Any], expected_size: int) -> tuple[np.ndarray, dict[str, float]]:
-        model = cast("LLM", self.model)
         t0 = time.perf_counter()
-        vllm_output = model.embed(
+        vllm_output = self.model.embed(
             input_data,
             tokenization_kwargs={"truncate_prompt_tokens": -1},
             use_tqdm=self.verbose,
@@ -296,7 +293,7 @@ class VLLMEmbeddingModelStage(ProcessingStage[DocumentBatch, DocumentBatch]):
             embedding_matrix[offset : offset + chunk_size] = chunk_embedding_matrix
             del chunk_embedding_matrix
 
-        return cast("np.ndarray", embedding_matrix), {
+        return embedding_matrix, {
             "tokenization_time": tokenization_time,
             "vllm_embedding_time": vllm_embedding_time,
             "input_tokens": input_tokens,
--- a/nemo_curator/stages/text/embedders/vllm.py
+++ b/nemo_curator/stages/text/embedders/vllm.py
@@ -67,7 +67,7 @@ class VLLMEmbeddingModelStage(ProcessingStage[DocumentBatch, DocumentBatch]):
         verbose: bool = False,
         *,
         metadata_fields: list[str] | None = None,
-        model_inference_batch_size: int = 10_000,
+        model_inference_batch_size: int | None = 8192,
     ):
         self.model_identifier = model_identifier
         self.vllm_init_kwargs = vllm_init_kwargs or {}
@@ -75,13 +75,13 @@ class VLLMEmbeddingModelStage(ProcessingStage[DocumentBatch, DocumentBatch]):
         self.text_field = text_field
         self.pretokenize = pretokenize
         self.embedding_field = embedding_field
-        self.metadata_fields = list(dict.fromkeys(metadata_fields)) if metadata_fields is not None else None
-        if (
+        self.metadata_fields = list(dict.fromkeys(metadata_fields or []))
+        if model_inference_batch_size is not None and (
             isinstance(model_inference_batch_size, bool)
             or not isinstance(model_inference_batch_size, int)
             or model_inference_batch_size <= 0
         ):
-            msg = f"model_inference_batch_size must be a positive integer, got {model_inference_batch_size}"
+            msg = f"model_inference_batch_size must be a positive integer or None, got {model_inference_batch_size}"
             raise ValueError(msg)
         self.model_inference_batch_size = model_inference_batch_size
         self.max_chars = max_chars
@@ -104,7 +104,7 @@ class VLLMEmbeddingModelStage(ProcessingStage[DocumentBatch, DocumentBatch]):
         return ["data"], [self.text_field]
 
     def outputs(self) -> tuple[list[str], list[str]]:
-        output_fields = list(self.metadata_fields) if self.metadata_fields is not None else [self.text_field]
+        output_fields = list(self.metadata_fields)
         if self.embedding_field not in output_fields:
             output_fields.append(self.embedding_field)
         return ["data"], output_fields
@@ -205,10 +205,17 @@ class VLLMEmbeddingModelStage(ProcessingStage[DocumentBatch, DocumentBatch]):
     def _iter_prepared_chunks(
         self, text_column: pa.ChunkedArray, num_rows: int
     ) -> Iterator[tuple[int, int, list[Any], float]]:
-        """Prepare one chunk ahead while the caller embeds the current chunk."""
+        """Prepare one chunk ahead while the caller embeds the current chunk.
+
+        The sole executor worker tokenizes only the next chunk. ``Future.result``
+        blocks until that work is ready, so this generator does not poll or spin.
+        While the caller embeds a yielded chunk on the GPU, the worker prepares
+        the following chunk on the CPU.
+        """
+        inference_batch_size = self.model_inference_batch_size or num_rows
         chunk_specs = iter(
-            (offset, min(self.model_inference_batch_size, num_rows - offset))
-            for offset in range(0, num_rows, self.model_inference_batch_size)
+            (offset, min(inference_batch_size, num_rows - offset))
+            for offset in range(0, num_rows, inference_batch_size)
         )
         pending_offset, pending_chunk_size = next(chunk_specs)
 
@@ -219,26 +226,22 @@ class VLLMEmbeddingModelStage(ProcessingStage[DocumentBatch, DocumentBatch]):
                 pending_offset,
                 pending_chunk_size,
             )
-            while True:
+            for next_offset, next_chunk_size in chunk_specs:
                 input_data, tokenization_time = pending_input.result()
-                try:
-                    next_offset, next_chunk_size = next(chunk_specs)
-                except StopIteration:
-                    next_input = None
-                else:
-                    next_input = executor.submit(
-                        self._prepare_input_chunk,
-                        text_column,
-                        next_offset,
-                        next_chunk_size,
-                    )
+                next_input = executor.submit(
+                    self._prepare_input_chunk,
+                    text_column,
+                    next_offset,
+                    next_chunk_size,
+                )
 
                 yield pending_offset, pending_chunk_size, input_data, tokenization_time
-                if next_input is None:
-                    break
                 pending_offset = next_offset
                 pending_chunk_size = next_chunk_size
                 pending_input = next_input
+
+            input_data, tokenization_time = pending_input.result()
+            yield pending_offset, pending_chunk_size, input_data, tokenization_time
 
     def _embed_chunk(self, input_data: list[Any], expected_size: int) -> tuple[np.ndarray, dict[str, float]]:
         if self.model is None:
@@ -274,44 +277,35 @@ class VLLMEmbeddingModelStage(ProcessingStage[DocumentBatch, DocumentBatch]):
             msg = f"Input batch is missing required text field {self.text_field!r}"
             raise ValueError(msg)
 
-        metadata_fields = self.metadata_fields if self.metadata_fields is not None else input_table.column_names
-        missing_fields = [field for field in metadata_fields if field not in input_table.column_names]
+        missing_fields = [field for field in self.metadata_fields if field not in input_table.column_names]
         if missing_fields:
             msg = f"Input batch is missing metadata fields: {missing_fields}"
             raise ValueError(msg)
-        return input_table.select(metadata_fields)
+        return input_table.select(self.metadata_fields)
 
     def _collect_embeddings(self, text_column: pa.ChunkedArray, num_rows: int) -> tuple[np.ndarray, dict[str, float]]:
         """Embed bounded chunks and assemble one ordered float32 matrix."""
-        embedding_matrix: np.ndarray | None = None
-        tokenization_time = 0.0
-        vllm_embedding_time = 0.0
-        input_tokens = 0.0
+        prepared_chunks = iter(self._iter_prepared_chunks(text_column, num_rows))
+        first_offset, first_chunk_size, first_input_data, tokenization_time = next(prepared_chunks)
+        first_embedding_matrix, first_metrics = self._embed_chunk(first_input_data, first_chunk_size)
+        embedding_matrix = np.empty(
+            (num_rows, first_embedding_matrix.shape[1]),
+            dtype=np.float32,
+        )
+        embedding_matrix[first_offset : first_offset + first_chunk_size] = first_embedding_matrix
+        del first_embedding_matrix
 
-        for offset, chunk_size, input_data, chunk_tokenization_time in self._iter_prepared_chunks(
-            text_column, num_rows
-        ):
+        vllm_embedding_time = first_metrics["vllm_embedding_time"]
+        input_tokens = first_metrics["input_tokens"]
+
+        for offset, chunk_size, input_data, chunk_tokenization_time in prepared_chunks:
             tokenization_time += chunk_tokenization_time
             chunk_embedding_matrix, chunk_metrics = self._embed_chunk(input_data, chunk_size)
             vllm_embedding_time += chunk_metrics["vllm_embedding_time"]
             input_tokens += chunk_metrics["input_tokens"]
-            if embedding_matrix is None:
-                embedding_matrix = np.empty(
-                    (num_rows, chunk_embedding_matrix.shape[1]),
-                    dtype=np.float32,
-                )
-            elif embedding_matrix.shape[1] != chunk_embedding_matrix.shape[1]:
-                msg = (
-                    f"vLLM embedding dimension changed from {embedding_matrix.shape[1]} "
-                    f"to {chunk_embedding_matrix.shape[1]}"
-                )
-                raise ValueError(msg)
             embedding_matrix[offset : offset + chunk_size] = chunk_embedding_matrix
             del chunk_embedding_matrix
 
-        if embedding_matrix is None:
-            msg = "vLLM did not return embeddings for a non-empty document batch"
-            raise RuntimeError(msg)
         return embedding_matrix, {
             "tokenization_time": tokenization_time,
             "vllm_embedding_time": vllm_embedding_time,
--- a/nemo_curator/stages/text/embedders/vllm.py
+++ b/nemo_curator/stages/text/embedders/vllm.py
@@ -67,6 +67,7 @@ class VLLMEmbeddingModelStage(ProcessingStage[DocumentBatch, DocumentBatch]):
         *,
         metadata_fields: list[str] | None = None,
         model_inference_batch_size: int | None = 8192,
+        # Keep float32 when feeding semantic dedup; cuDF cannot read nested Float16 Parquet as numeric list values.
         embedding_output_dtype: Literal["float16", "float32", "float64"] = "float32",
     ):
         self.model_identifier = model_identifier
--- a/nemo_curator/stages/text/embedders/vllm.py
+++ b/nemo_curator/stages/text/embedders/vllm.py
@@ -50,6 +50,7 @@ if TYPE_CHECKING:
     from nemo_curator.backends.base import NodeInfo, WorkerMetadata
 
 _VLLM_INSTALL_HINT = "vLLM is required for VLLMEmbeddingModelStage. Install with: pip install nemo_curator[vllm]"
+_MAX_LIST_ARRAY_VALUES = np.iinfo(np.int32).max
 
 
 class VLLMEmbeddingModelStage(ProcessingStage[DocumentBatch, DocumentBatch]):
@@ -197,9 +198,7 @@ class VLLMEmbeddingModelStage(ProcessingStage[DocumentBatch, DocumentBatch]):
         prompts = [TokensPrompt(prompt_token_ids=ids) for ids in tokenized_data.input_ids]
         return prompts, time.perf_counter() - t0
 
-    def _iter_prepared_chunks(
-        self, text_column: pa.ChunkedArray, num_rows: int
-    ) -> Iterator[tuple[int, int, list[Any], float]]:
+    def _iter_prepared_chunks(self, text_column: pa.ChunkedArray, num_rows: int) -> Iterator[tuple[list[Any], float]]:
         """Prepare one chunk ahead while the caller embeds the current chunk.
 
         One worker is intentional: this is a single-producer prefetch pipeline,
@@ -231,13 +230,13 @@ class VLLMEmbeddingModelStage(ProcessingStage[DocumentBatch, DocumentBatch]):
                     next_chunk_size,
                 )
 
-                yield pending_offset, pending_chunk_size, input_data, tokenization_time
+                yield input_data, tokenization_time
                 pending_offset = next_offset
                 pending_chunk_size = next_chunk_size
                 pending_input = next_input
 
             input_data, tokenization_time = pending_input.result()
-            yield pending_offset, pending_chunk_size, input_data, tokenization_time
+            yield input_data, tokenization_time
 
     def _embed_chunk(self, input_data: list[Any]) -> tuple[np.ndarray, dict[str, float]]:
         t0 = time.perf_counter()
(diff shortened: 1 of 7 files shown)


tests/stages/text/embedders/test_vllm.py
```diff
@@ -81,17 +131,41 @@ def test_custom_initialization(self) -> None:
...
         assert stage.verbose is True
 
         assert stage.inputs() == (["data"], ["content"])
-        assert stage.outputs() == (["data"], ["content", "emb"])
+        assert stage.outputs() == (["data"], ["id", "content", "emb"])
 
         assert stage.resources.gpus == 1
         assert stage.resources.cpus == 1
 
+    def test_new_options_preserve_existing_positional_arguments(self) -> None:
```
praateekmahajan: remove this test

nemo_curator/stages/text/embedders/vllm.py
```diff
@@ -54,13 +65,25 @@ def __init__(  # noqa: PLR0913
         cache_dir: str | None = None,
         hf_token: str | None = None,
         verbose: bool = False,
+        *,
+        metadata_fields: list[str] | None = None,
+        model_inference_batch_size: int = 10_000,
```
praateekmahajan: It can be None too in which case it just pre-embeds everything

praateekmahajan: Let the default also be 2**x i.e. 8192 or 16,384

nemo_curator/stages/text/embedders/vllm.py
```diff
@@ -81,7 +104,10 @@ def inputs(self) -> tuple[list[str], list[str]]:
         return ["data"], [self.text_field]
 
     def outputs(self) -> tuple[list[str], list[str]]:
-        return ["data"], [self.text_field, self.embedding_field]
+        output_fields = list(self.metadata_fields) if self.metadata_fields is not None else [self.text_field]
```
praateekmahajan: Let's look through other dedup algorithms for consistency and see if it's called meteadata fields or output fields, because in semdedup case these are the fields we'd like to have in our output. And if None is specified I don't think default should be text, since it takes too much memory

nemo_curator/stages/text/embedders/vllm.py
```diff
@@ -149,43 +175,186 @@ def setup(self, worker_metadata: WorkerMetadata | None = None) -> None:  # noqa:
...
+        )
+        if chunk_embedding_matrix.ndim != _EMBEDDING_MATRIX_NDIM:
+            msg = f"Expected a two-dimensional embedding matrix, got shape {chunk_embedding_matrix.shape}"
+            raise ValueError(msg)
+        return chunk_embedding_matrix, {
+            "vllm_embedding_time": elapsed,
+            "input_tokens": float(sum(len(output.prompt_token_ids) for output in vllm_output)),
+        }
+
+    def _select_output_table(self, input_table: pa.Table) -> pa.Table:
```
praateekmahajan, line 258: If metadata_fields is None, how will this work?

nemo_curator/stages/text/embedders/vllm.py
```diff
@@ -149,43 +175,186 @@ def setup(self, worker_metadata: WorkerMetadata | None = None) -> None:  # noqa:
...
+            tokenization_time += chunk_tokenization_time
+            chunk_embedding_matrix, chunk_metrics = self._embed_chunk(input_data, chunk_size)
+            vllm_embedding_time += chunk_metrics["vllm_embedding_time"]
+            input_tokens += chunk_metrics["input_tokens"]
+            if embedding_matrix is None:
+                embedding_matrix = np.empty(
+                    (num_rows, chunk_embedding_matrix.shape[1]),
+                    dtype=np.float32,
+                )
+            elif embedding_matrix.shape[1] != chunk_embedding_matrix.shape[1]:
```
praateekmahajan: When will this ever happen?

nemo_curator/stages/text/embedders/vllm.py
```diff
@@ -149,43 +175,186 @@ def setup(self, worker_metadata: WorkerMetadata | None = None) -> None:  # noqa:
...
+        input_tokens = 0.0
+
+        for offset, chunk_size, input_data, chunk_tokenization_time in self._iter_prepared_chunks(
+            text_column, num_rows
+        ):
+            tokenization_time += chunk_tokenization_time
+            chunk_embedding_matrix, chunk_metrics = self._embed_chunk(input_data, chunk_size)
+            vllm_embedding_time += chunk_metrics["vllm_embedding_time"]
+            input_tokens += chunk_metrics["input_tokens"]
+            if embedding_matrix is None:
```
praateekmahajan: Instead of creating this in for loop can we create this outside?

nemo_curator/stages/text/embedders/vllm.py
```diff
@@ -149,43 +175,186 @@ def setup(self, worker_metadata: WorkerMetadata | None = None) -> None:  # noqa:
...
+        pending_offset, pending_chunk_size = next(chunk_specs)
+
+        with ThreadPoolExecutor(max_workers=1, thread_name_prefix="vllm-tokenizer") as executor:
+            pending_input: Future[tuple[list[Any], float]] = executor.submit(
+                self._prepare_input_chunk,
+                text_column,
+                pending_offset,
+                pending_chunk_size,
+            )
+            while True:
```
praateekmahajan: Won't this while True without any sleep be super expensive? I'm not sure

nemo_curator/stages/text/embedders/vllm.py
```diff
@@ -149,43 +175,186 @@ def setup(self, worker_metadata: WorkerMetadata | None = None) -> None:  # noqa:
...
+        self, text_column: pa.ChunkedArray, num_rows: int
+    ) -> Iterator[tuple[int, int, list[Any], float]]:
+        """Prepare one chunk ahead while the caller embeds the current chunk."""
+        chunk_specs = iter(
+            (offset, min(self.model_inference_batch_size, num_rows - offset))
+            for offset in range(0, num_rows, self.model_inference_batch_size)
+        )
+        pending_offset, pending_chunk_size = next(chunk_specs)
+
+        with ThreadPoolExecutor(max_workers=1, thread_name_prefix="vllm-tokenizer") as executor:
```
praateekmahajan, line 217: Explain how threadpoolexecutor works with 1 worker only?

tests/stages/text/embedders/test_vllm.py
```diff
@@ -144,6 +218,317 @@ def __init__(self, model: str, **kwargs: Any) -> None:  # noqa: ANN401
...
+            stage.process(batch)
+
+    def test_process_rejects_uninitialized_tokenizer(self) -> None:
+        stage = VLLMEmbeddingModelStage(model_identifier=TEST_MODEL, pretokenize=True)
+        stage.model = _FakeEmbeddingModel()  # type: ignore[assignment]
+        batch = DocumentBatch(dataset_name="test_dataset", data=pa.table({"text": ["0"]}))
+
+        with pytest.raises(ValueError, match="Tokenizer is not initialized"):
+            stage.process(batch)
+
```
praateekmahajan: All these are redundant tests

tests/stages/text/embedders/test_vllm.py
```diff
@@ -144,6 +218,317 @@ def __init__(self, model: str, **kwargs: Any) -> None:  # noqa: ANN401
...
+        with pytest.raises(ValueError, match="embedding dimension changed from 2 to 3"):
+            stage.process(DocumentBatch(dataset_name="test_dataset", data=pa.table({"text": ["0", "1"]})))
+
+    def test_process_replaces_selected_embedding_field_in_place(self) -> None:
+        stage = VLLMEmbeddingModelStage(
+            model_identifier=TEST_MODEL,
+            pretokenize=False,
+            metadata_fields=["id", "embeddings", "text"],
+        )
+        stage.model = _FakeEmbeddingModel()  # type: ignore[assignment]
```
praateekmahajan: delete these two tests

tests/stages/text/embedders/test_vllm.py
```diff
@@ -144,6 +218,317 @@ def __init__(self, model: str, **kwargs: Any) -> None:  # noqa: ANN401
...
+                    use_tqdm=use_tqdm,
+                )[:-1]
+
+        stage = VLLMEmbeddingModelStage(model_identifier=TEST_MODEL, pretokenize=False)
+        stage.model = _ShortOutputModel()  # type: ignore[assignment]
+
+        with pytest.raises(ValueError, match="returned 1 embeddings for a 2-row input chunk"):
+            stage.process(DocumentBatch(dataset_name="test_dataset", data=pa.table({"text": ["0", "1"]})))
+
+    def test_process_rejects_non_matrix_embeddings(self) -> None:
```
praateekmahajan: delete this too

tests/stages/text/embedders/test_vllm.py
```diff
@@ -144,6 +218,317 @@ def __init__(self, model: str, **kwargs: Any) -> None:  # noqa: ANN401
...
+
+        result = stage.process(batch)
+        metrics = stage._consume_custom_metrics()
+
+        assert result.num_items == 5
+        assert metrics["tokenization_time"] == 0.75
+        assert metrics["input_tokens"] == 10
+        assert metrics["vllm_embedding_time"] == 1.5
+
+    def test_process_rejects_wrong_vllm_output_count(self) -> None:
```
praateekmahajan: delete this too

tests/stages/text/embedders/test_vllm.py
```diff
@@ -144,6 +218,317 @@ def __init__(self, model: str, **kwargs: Any) -> None:  # noqa: ANN401
...
+            model_identifier=TEST_MODEL,
+            pretokenize=False,
+            metadata_fields=metadata_fields,
+        )
+        stage.model = SimpleNamespace()
+
+        with pytest.raises(ValueError, match=message):
+            stage.process(DocumentBatch(dataset_name="test_dataset", data=table))
+
+    def test_process_reports_aggregate_chunk_metrics(self) -> None:
```
praateekmahajan: combine this with above

tests/stages/text/embedders/test_vllm.py
```diff
@@ -144,6 +218,317 @@ def __init__(self, model: str, **kwargs: Any) -> None:  # noqa: ANN401
...
+        assert third_chunk_prepared.is_set()
+
+    @pytest.mark.parametrize(
+        ("table", "metadata_fields", "message"),
+        [
+            (pa.table({"id": [1]}), None, "missing required text field"),
+            (pa.table({"text": ["first"], "id": [1]}), ["id", "source"], "missing metadata fields"),
+        ],
+    )
+    def test_process_rejects_missing_fields(
```
praateekmahajan: combine this test with other tests for reject, that says test_process_rejects_invalid_configurations and then in the same test try different things

tests/stages/text/embedders/test_vllm.py
```diff
@@ -144,6 +218,317 @@ def __init__(self, model: str, **kwargs: Any) -> None:  # noqa: ANN401
...
+    def test_process_rejects_uninitialized_tokenizer(self) -> None:
+        stage = VLLMEmbeddingModelStage(model_identifier=TEST_MODEL, pretokenize=True)
+        stage.model = _FakeEmbeddingModel()  # type: ignore[assignment]
+        batch = DocumentBatch(dataset_name="test_dataset", data=pa.table({"text": ["0"]}))
+
+        with pytest.raises(ValueError, match="Tokenizer is not initialized"):
+            stage.process(batch)
+
+    @pytest.mark.parametrize("pretokenize", [True, False])
+    def test_process_chunks_rows_in_both_pretokenization_modes(self, pretokenize: bool) -> None:
```
praateekmahajan: Let's not use FakeEmbeddingModel or Tokenizer and actually use the real model for this test.. in fact let's try to combine test below like test_process_prepares_exactly_one_chunk_ahead / test_process_preserves_configured_arrow_columns_without_pandas / test_process_chunks_rows_in_both_pretokenization_modes / test_process_reports_aggregate_chunk_metrics


The goal of test shouldn't be to increase lines of code, but just simply test correctness and functionality..


Let's for instance start a documentbatch with 5 rows, and break it such that we use 2 and 3 rows, confirm somehow that it did batching indeed, and then it gets aggregate metrics and such.. and confirm it's output matches if it had never done batching and is same as processing all 5 rows correctly

nemo_curator/stages/text/embedders/vllm.py
```diff
@@ -274,44 +277,35 @@ def _select_output_table(self, input_table: pa.Table) -> pa.Table:
...
+        embedding_matrix[first_offset : first_offset + first_chunk_size] = first_embedding_matrix
+        del first_embedding_matrix
 
-        for offset, chunk_size, input_data, chunk_tokenization_time in self._iter_prepared_chunks(
-            text_column, num_rows
-        ):
+        vllm_embedding_time = first_metrics["vllm_embedding_time"]
+        input_tokens = first_metrics["input_tokens"]
+
+        for offset, chunk_size, input_data, chunk_tokenization_time in prepared_chunks:
```
praateekmahajan: This is bad as it replicates embed_chunk call and assigned of embedding matrix.., move it back inside the loop

nemo_curator/stages/text/embedders/vllm.py
```diff
@@ -54,13 +65,25 @@ def __init__(  # noqa: PLR0913
...
+        model_inference_batch_size: int | None = 8192,
     ):
         self.model_identifier = model_identifier
         self.vllm_init_kwargs = vllm_init_kwargs or {}
 
         self.text_field = text_field
         self.pretokenize = pretokenize
         self.embedding_field = embedding_field
+        self.metadata_fields = list(dict.fromkeys(metadata_fields or []))
+        if model_inference_batch_size is not None and (
```
sarahyurick: Nit but I don't think most of this check is necessary. We should have the typehint be `int` only and we can keep the check that it is not <= 0.

nemo_curator/stages/text/embedders/vllm.py
```diff
@@ -149,43 +175,179 @@ def setup(self, worker_metadata: WorkerMetadata | None = None) -> None:  # noqa:
...
+        self, text_column: pa.ChunkedArray, offset: int, chunk_size: int
+    ) -> tuple[list[Any], float]:
+        input_data = text_column.slice(offset, chunk_size).to_pylist()
         if self.max_chars is not None:
-            df[self.text_field] = df[self.text_field].str.slice(0, self.max_chars)
-        input_data = df[self.text_field].tolist()
-        metrics = {}
+            input_data = [text[: self.max_chars] if text is not None else None for text in input_data]
+        if not self.pretokenize:
+            return input_data, 0.0
```
sarahyurick, line 188: Sorry what is 0.0 here?

nemo_curator/stages/text/embedders/vllm.py
```diff
@@ -149,43 +175,179 @@ def setup(self, worker_metadata: WorkerMetadata | None = None) -> None:  # noqa:
...
 
-        if self.pretokenize:
-            from vllm.inputs import TokensPrompt
+        if self.tokenizer is None:
+            msg = "Tokenizer is not initialized. Please call setup() before processing or set pretokenize to False."
+            raise ValueError(msg)
+        if self.model is None:
+            msg = "vLLM model is not initialized. Please call setup() before processing."
```
sarahyurick: I don't think we should include these. I think agents like to add these checks a lot for some reason. As long as the user is building a Curator pipeline and running it that way then these blocks should never be true.

nemo_curator/stages/text/embedders/vllm.py
sarahyurick: Oh I see, it's perf.

nemo_curator/stages/text/embedders/vllm.py
```diff
@@ -149,43 +175,179 @@ def setup(self, worker_metadata: WorkerMetadata | None = None) -> None:  # noqa:
...
+        """
+        inference_batch_size = self.model_inference_batch_size or num_rows
+        chunk_specs = iter(
+            (offset, min(inference_batch_size, num_rows - offset))
+            for offset in range(0, num_rows, inference_batch_size)
+        )
+        pending_offset, pending_chunk_size = next(chunk_specs)
+
+        with ThreadPoolExecutor(max_workers=1, thread_name_prefix="vllm-tokenizer") as executor:
+            pending_input: Future[tuple[list[Any], float]] = executor.submit(
```
sarahyurick, line 218: Could you add a brief explanation why this needs the `ThreadPoolExecutor`?

nemo_curator/stages/text/embedders/vllm.py
```diff
@@ -149,43 +175,179 @@ def setup(self, worker_metadata: WorkerMetadata | None = None) -> None:  # noqa:
...
+                pending_offset = next_offset
+                pending_chunk_size = next_chunk_size
+                pending_input = next_input
+
+            input_data, tokenization_time = pending_input.result()
+            yield pending_offset, pending_chunk_size, input_data, tokenization_time
+
+    def _embed_chunk(self, input_data: list[Any], expected_size: int) -> tuple[np.ndarray, dict[str, float]]:
+        if self.model is None:
+            msg = "vLLM model is not initialized. Please call setup() before processing."
```
sarahyurick: Same comment as above.

nemo_curator/stages/text/embedders/vllm.py
```diff
@@ -149,43 +175,179 @@ def setup(self, worker_metadata: WorkerMetadata | None = None) -> None:  # noqa:
...
         vllm_output = self.model.embed(
             input_data,
             tokenization_kwargs={"truncate_prompt_tokens": -1},
             use_tqdm=self.verbose,
         )
-        metrics["vllm_embedding_time"] = time.perf_counter() - t0
+        elapsed = time.perf_counter() - t0
+        if len(vllm_output) != expected_size:
+            msg = f"vLLM returned {len(vllm_output)} embeddings for a {expected_size}-row input chunk"
+            raise ValueError(msg)
```
sarahyurick: I am curious does it actually run into this case sometimes?

nemo_curator/stages/text/embedders/vllm.py
```diff
@@ -149,43 +175,179 @@ def setup(self, worker_metadata: WorkerMetadata | None = None) -> None:  # noqa:
...
+                embedding_matrix.shape[1],
+                dtype=np.int64,
+            )
+        )
+        return pa.ListArray.from_arrays(embedding_offsets, embedding_values)
 
-        df[self.embedding_field] = [e.outputs.embedding for e in vllm_output]
+    def process(self, batch: DocumentBatch) -> DocumentBatch:
+        if self.model is None:
+            msg = "vLLM model is not initialized. Please call setup() before processing."
```
sarahyurick: Similar comment as above.

nemo_curator/stages/text/embedders/vllm.py
```diff
@@ -149,43 +175,179 @@ def setup(self, worker_metadata: WorkerMetadata | None = None) -> None:  # noqa:
...
+            raise ValueError(msg)
+        input_table = batch.to_pyarrow()
+        if input_table.num_rows == 0:
+            msg = "Cannot generate embeddings for an empty document batch"
```
sarahyurick: Now that we have added `NoneTask` we should try to do a sweep where no more stages return empty DataFrames, etc. so that we don't need these types of checks anymore. Not saying it should be addressed in this PR but it reminded me of this.

sarahyurick (commented):
Thanks, generally looks good to me, left a bunch of mostly minor comments.

nemo_curator/stages/text/embedders/vllm.py
VibhuJawa: Yup, plus one to remove all these guards.

nemo_curator/stages/text/embedders/vllm.py
```diff
@@ -149,43 +175,179 @@ def setup(self, worker_metadata: WorkerMetadata | None = None) -> None:  # noqa:
...
+        elapsed = time.perf_counter() - t0
+        if len(vllm_output) != expected_size:
+            msg = f"vLLM returned {len(vllm_output)} embeddings for a {expected_size}-row input chunk"
+            raise ValueError(msg)
+
+        chunk_embedding_matrix = np.asarray(
+            [output.outputs.embedding for output in vllm_output],
+            dtype=np.float32,
+        )
+        if chunk_embedding_matrix.ndim != _EMBEDDING_MATRIX_NDIM:
```
VibhuJawa: Should we just compare with 2.  Like, do we expect `3` dimentional arrays here ? 

Or not even compare at all.

nemo_curator/stages/text/embedders/vllm.py
VibhuJawa: Yup,  i am confused too about a thread pool executor with `max_workers=1` ? 

Random Nit: I look at this code and i feel like, we should use like a data loaders instead somehow instead of trying to handle this ourselves. I feel like a lot of code can be offloaded to dataloaders ?

nemo_curator/stages/text/embedders/vllm.py
```diff
@@ -149,43 +175,179 @@ def setup(self, worker_metadata: WorkerMetadata | None = None) -> None:  # noqa:
...
+            msg = f"Expected a two-dimensional embedding matrix, got shape {chunk_embedding_matrix.shape}"
+            raise ValueError(msg)
+        return chunk_embedding_matrix, {
+            "vllm_embedding_time": elapsed,
+            "input_tokens": float(sum(len(output.prompt_token_ids) for output in vllm_output)),
```
VibhuJawa: nit: Why is this float ?

VibhuJawa (commented):
Took a look, mostly nits.

nemo_curator/stages/text/embedders/vllm.py
praateekmahajan: Good question on why `max_workers=1`... 
Here is how I understand it, the main/caller thread is doing two things 
1. submits the next chunk of rows for tokenization a thread to tokenize (having a future) 
2. after receiving the current chunk, calling vLLM

Historically, I have typically used `workers > 1`, but that is intended to run the same task concurrently (i.e. do the thing "faster"). Here we don't want to multiple chunks to be tokenized concurrently, but rather the next chunk to be tokenized while current is being processed by vLLM.

In simplified Python, it looks like this
```python

def generate_rows():
  with ThreadPoolExecutor(max_workers=1) as executor:
 
      # submit the first request
	  next_chunk = executor.submit(prepare, chunk_0)
     
      for chunk_spec in remaining_chunks:
          # wait for the current chunk
          current_chunk = next_chunk.result()
          # submit a future to fetch the next chunk
          next_chunk = executor.submit(prepare, chunk_spec)

          # return current_chunk to caller, while next_chunk is being fetched
          yield current_chunk

for rows in generate_rows():
  vllm_generate_embeddings(rows)
```
Therefore, there are already two concurrent execution context, the caller running vLLM and the executor's one worker preparing the next chunk. 

Regarding why not DataLoader, I agree the pattern seems overly similar, I think it might be an overkill here, as wed need a custom `IterableDataset` / `coallate_fn`.

nemo_curator/stages/text/embedders/vllm.py
praateekmahajan: I agree, probably better to do that. I don't know how I can do that here, maybe I'll remove the check for now and let it do runtime error than maintain this code

nemo_curator/stages/text/embedders/vllm.py
praateekmahajan: I want `None / 0` to work as use all rows in a single batch, since that was the previous behavior and still works if DocumentBatch is small enough

nemo_curator/stages/text/embedders/vllm.py
sarahyurick: Curious was `batch_encode_plus` slow enough that you noticed a decent difference? Like I can see how the overlap can be nice, but it kind of just reminds me of when vLLM embedding support was initially being added to Curator, and why we didn't just support a CPU tokenization stage with it instead of having it as all 1 stage?

praateekmahajan: `batch_encode_plus` can take 1-2s for a normal sized batch (with 32k context len).. so those 1-2 we lose GPU activity.
Long ago when I added vLLM embedder, I actually did consider CPU tokenization stage, except back then `vllm.embed(text)` was the most performant. However after vllm 0.19+ text path degenerated so we had to move to tokenize and then embed. 
Having now done that at scale, I believe even the separate tokenization stage is risky, because if you have a document batch with 100k documents, the tokenization stage will hold all that in memory which can be really large, whereas this approach strictly tokenizes what's fed into the model therefore conserving CPU memory (a LOT).

sarahyurick: I see, thanks! Sounds good to me.

nemo_curator/stages/text/embedders/vllm.py
```diff
@@ -149,43 +172,155 @@ def setup(self, worker_metadata: WorkerMetadata | None = None) -> None:  # noqa:
...
         vllm_output = self.model.embed(
             input_data,
             tokenization_kwargs={"truncate_prompt_tokens": -1},
             use_tqdm=self.verbose,
         )
-        metrics["vllm_embedding_time"] = time.perf_counter() - t0
+        elapsed = time.perf_counter() - t0
+        chunk_embedding_matrix = np.asarray(
+            [output.outputs.embedding for output in vllm_output],
+            dtype=np.float32,
```
sarahyurick: Last question should we make these dtype casts configurable?

sarahyurick (approved):
LGTM, left one minor minor comment.

nemo_curator/stages/text/embedders/vllm.py
praateekmahajan: Good call, will add that!

nemo_curator/stages/text/embedders/vllm.py
```diff
@@ -149,43 +176,159 @@ def setup(self, worker_metadata: WorkerMetadata | None = None) -> None:  # noqa:
...
+            embedding_matrix.reshape(-1),
+            type=pa.from_numpy_dtype(embedding_matrix.dtype),
+            from_pandas=False,
+        )
+        embedding_offsets = pa.array(
+            np.arange(
+                0,
+                (embedding_matrix.shape[0] + 1) * embedding_matrix.shape[1],
+                embedding_matrix.shape[1],
+                dtype=np.int64,
```
sarahyurick: Does this one matter at all?

VibhuJawa (approved):
LGTM

nemo_curator/stages/text/embedders/vllm.py
praateekmahajan: That was a really good catch 🙏 

Before since it was pandas, we didn't have the 2bn list-elements limit, since we moved to pyarrow now, we would've had that limit (iow if embedding_dim=1024, our document batch couldn't have had more than 2mn rows; so i then ended up moving it to ChunkedArray)

This issue isn't really an issue for SemDedup since we're anyway constrained that a single file MUST contain less than `2bn/EMBEDDING_DIM` number of rows, however I can imagine there might be some embedding use case out there where we do want to support this

PR #2318, opened 2026-08-17
feat(semdedup): bound pairwise precision and memory

praateekmahajan:
## Summary

Implements [NMCUR-318](https://linear.app/nvidia/issue/NMCUR-318):

- bounds Pairwise memory to a reusable `N x B` workspace controlled by positive `pairwise_batch_size` (default `1024`);
- compares each ranked row only with earlier-ranked neighbors and deterministically chooses the earliest row on ties;
- preserves arbitrary ranking columns and the `id`, `max_id`, and `cosine_sim_score` output contract;
- supports independent KMeans embedding storage and Pairwise compute precision while keeping FP32/FP32 as the default;
- retains FP16 through multiplication and reduction, promoting scores to FP32 only for cuDF output;
- gives each Pairwise actor a 1 GiB growing RMM pool shared by cuDF, CuPy, and Torch;
- reports footer scan, read, rank, precision conversion, compute, and write timings;
- documents the ranked-neighbor and duplicate-removal behavior with a complete five-row `X @ X.T` example.

`IdentifyDuplicatesStage` combines ten one-cluster Pairwise tasks into one output and now retains all ten Pairwise performance records. The generic fan-in provenance issue is tracked separately in [NMCUR-457](https://linear.app/nvidia/issue/NMCUR-457).

The ranked-neighbor sidecar remains an extension point only. This PR does not implement NMCUR-242 or NMCUR-319.

## Benchmarks

The final FP32 validation at `9c5549ae` succeeded for both SemDedup entries:

| Documents | Workflow | KMeans | Pairwise | Duplicates |
| ---: | ---: | ---: | ---: | ---: |
| 65,844,782 | 175.0s | 120.5s | 54.5s | 1,154,805 |
| 329,774,475 | 758.2s | 355.7s | 402.5s | 10,819,461 |

Compared with the prior RMM run, the final implementation improved:

| Documents | End-to-end speedup | Pairwise speedup |
| ---: | ---: | ---: |
| 65,844,782 | 15.8% | 29.9% |
| 329,774,475 | 9.0% | 13.9% |

[Open the run-viewer comparison](http://rratzel-ws1:5050/compare?dir_a=login-eos-adhoc%3A%2Flustre%2Ffsw%2Fcoreai_dlalgo_ci%2Fcurator_ci%2Fresults%2Fmain-mirror&dir_b=login-eos-adhoc%3A%2Flustre%2Ffsw%2Fcoreai_dlalgo_ci%2Fcurator_ci%2Fresults%2Fmain-mirror&run_a=pr-2318-2026_09_04__22_37_45_UTC-fb6bbdbb&run_b=pr-2318-2026_09_08__19_43_50_UTC-9c5549ae).

The earlier 329,774,475-document precision experiment at `5e5b8248` showed:

| Configuration | Workflow | Pairwise | Duplicates | Retained artifacts |
| --- | ---: | ---: | ---: | ---: |
| September 4 main nightly | 1122.8s | 740.6s | 10,819,452 | — |
| KMeans FP32 → Pairwise FP32 | 786.9s | 423.0s | 10,819,441 | 1.025 TB |
| KMeans FP32 → Pairwise FP16 | 539.6s | 184.9s | 10,823,919 | 1.024 TB |
| KMeans FP16 → Pairwise FP16 | 497.0s | 154.3s | 10,823,956 | 498.7 GB |

FP16 remains opt-in pending duplicate-ID analysis; committed benchmark entries use the FP32 defaults.

Pairwise phase timings are cumulative actor-seconds across concurrent GPUs, while `pairwise_time` is wall-clock time. The final run contains exactly one Pairwise record for each of the 1,000 clusters.

## Validation

Using `nvcr.io/nvidian/nemo-curator:nightly-2026-09-04` with GPUs 2 and 3 only:

- Pairwise focused tests at `6d76839a`: 27 passed, 1 skipped
- IdentifyDuplicates and workflow regression tests at `9c5549ae`: 46 passed
- RMM allocator smoke, Ruff, formatting, and pre-commit: passed


benchmarking/scripts/semdedup_identification_benchmark.py
```diff
@@ -46,8 +60,10 @@ def run_semdedup_identification_benchmark(  # noqa: PLR0913
     input_filetype: str = "parquet",
     eps: float = 0.01,
     which_to_keep: str = "hard",
-    pairwise_batch_size: int = 1024,
+    pairwise_batch_size: int | str = 1024,
```
praateekmahajan: make default `auto`

benchmarking/scripts/semdedup_pairwise_benchmark.py
```diff
@@ -0,0 +1,205 @@
+# Copyright (c) 2026, NVIDIA CORPORATION.  All rights reserved.
```
praateekmahajan: We can remove this, since the nightly yaml of fitting on 0.2 data but runs pairwise on all the data so we get the metric from there

benchmarking/scripts/semdedup_identification_benchmark.py
```diff
@@ -195,7 +215,21 @@ def main() -> int:
...
-        "--pairwise-batch-size", type=int, default=1024, help="Batch size for pairwise similarity computation"
+        "--pairwise-compute-dtype",
+        choices=["auto", "float16", "float32"],
+        default="auto",
+        help="Multiplication precision used by Pairwise",
+    )
+    parser.add_argument(
+        "--pairwise-batch-size",
+        type=_parse_pairwise_batch_size,
+        default=1024,
```
praateekmahajan, line 258: keep default as auto and add it to yaml too

nemo_curator/stages/deduplication/semantic/pairwise.py
```diff
@@ -31,44 +34,225 @@
...
+class _PairwiseSimilarityResult:
+    """Internal result object that can grow ranked-neighbor data later."""
+
+    max_similarity: "cp.ndarray"
+    max_indices: "cp.ndarray"
+    batch_size: int
+    compute_dtype: "torch.dtype"
+
+
+def _release_cached_gpu_memory() -> None:
```
praateekmahajan: Can you move these do deduplication/utils.py since this might be used there too? I also think minhash in one of it's stages uses this..

nemo_curator/stages/deduplication/semantic/pairwise.py
```diff
@@ -31,44 +34,225 @@
...
+        dtype=cluster_reps.dtype,
+        device=cluster_reps.device,
+    )
+    invalid_batch_triangle = torch.ones(
+        (resolved_batch_size, resolved_batch_size),
+        dtype=torch.bool,
+        device=cluster_reps.device,
+    ).tril_()
+
+    for start_idx in range(0, num_rows, resolved_batch_size):
```
praateekmahajan: Let's add comment describing what we do here, and similarly above for why `tril_` exists and so on.. keep in mind that person reading the code is coming from a similar mind-frame of `A @ A.T` so they need to be able to follow it

nemo_curator/stages/deduplication/semantic/pairwise.py
```diff
@@ -31,44 +34,225 @@
...
-        max_values_and_indices = torch.max(triu_sim_matrix, dim=0)
-        max_similarity[start_idx:end_idx] = max_values_and_indices[0]
-        max_indices[start_idx:end_idx] = max_values_and_indices[1]
-
-    if device == "cuda":
-        return cp.asarray(max_similarity), cp.asarray(max_indices)
-    else:
-        # convert to numpy arrays
-        return max_similarity.numpy(), max_indices.numpy()
+    result = _pairwise_cosine_similarity_batched(cluster_reps, batch_size, compute_dtype)
```
praateekmahajan: This function just calls the one above, so what's the point splitting it into two?

nemo_curator/stages/deduplication/semantic/pairwise.py
```diff
@@ -116,6 +314,18 @@ def __init__(  # noqa: PLR0913
...
+        """Process one cluster and release cached allocations on every exit path."""
+        try:
+            return self._process(task)
+        except BaseException as exc:
+            # An exception's traceback otherwise keeps the unwound _process
+            # frame—and its large GPU objects—alive until after this finalizer.
+            traceback.clear_frames(exc.__traceback__)
+            raise
+        finally:
+            _release_cached_gpu_memory()
```
praateekmahajan: do we need this? why do other GPU stages not need this?

nemo_curator/stages/deduplication/semantic/pairwise.py
```diff
@@ -226,20 +446,40 @@ def process(self, task: FileGroupTask) -> FileGroupTask:
...
+                    "pairwise_num_rows": num_rows,
+                    "pairwise_input_file_count": len(task.data),
+                    "pairwise_file_group_count": len(file_groups),
+                    "pairwise_batch_size": similarity_result.batch_size,
+                    "pairwise_batch_size_auto": float(self.pairwise_batch_size == "auto"),
+                    "pairwise_compute_dtype_bytes": torch.empty(
+                        (), dtype=similarity_result.compute_dtype
+                    ).element_size(),
+                    "pairwise_gpu_free_before_similarity_bytes": free_memory_before_similarity,
+                    **profiler.metrics,
```
praateekmahajan: What metrics are these?

nemo_curator/stages/deduplication/semantic/pairwise.py
```diff
@@ -226,20 +446,40 @@ def process(self, task: FileGroupTask) -> FileGroupTask:
...
+        profiler.end("total", total_started)
+        profiler.finish()
+        if self.profile:
+            self._log_metrics(
+                {
+                    "pairwise_num_rows": num_rows,
+                    "pairwise_input_file_count": len(task.data),
+                    "pairwise_file_group_count": len(file_groups),
+                    "pairwise_batch_size": similarity_result.batch_size,
+                    "pairwise_batch_size_auto": float(self.pairwise_batch_size == "auto"),
```
praateekmahajan: We should also log the batch size determined by auto

tests/benchmarking/test_semdedup_identification_benchmark.py
```diff
@@ -0,0 +1,147 @@
+# Copyright (c) 2026, NVIDIA CORPORATION.  All rights reserved.
```
praateekmahajan: remove this test

tests/stages/deduplication/semantic/test_pairwise.py
```diff
@@ -29,15 +31,95 @@
...
         PairwiseStage,
         pairwise_cosine_similarity_batched,
     )
     from nemo_curator.stages.deduplication.semantic.pairwise_io import ClusterWiseFilePartitioningStage
     from nemo_curator.stages.deduplication.semantic.ranking import RankingStrategy
+    from nemo_curator.stages.text.embedders.utils import create_list_series_from_1d_or_2d_ar
     from nemo_curator.tasks import FileGroupTask
 
+with suppress(ImportError):
+    from nemo_curator.stages.deduplication.semantic.utils import decode_embedding_array
```
praateekmahajan: Should be tested inside semantic/test_utils.py .. always maintain 1-1 relationship of source file and test file

tests/stages/text/deduplication/test_semantic.py
```diff
@@ -147,6 +148,63 @@ def fail_pipeline_run(self, executor) -> None:  # noqa: ANN001, ARG001
...
+    with pytest.raises(ValueError, match=message):
+        TextSemanticDeduplicationWorkflow(input_path="/input", output_path=str(tmp_path), **kwargs)
+
+
+@pytest.mark.gpu
+def test_new_text_workflow_options_preserve_existing_positional_order() -> None:
+    parameters = list(inspect.signature(TextSemanticDeduplicationWorkflow).parameters)
+
+    assert parameters.index("pairwise_compute_dtype") > parameters.index("clear_output")
+    assert parameters.index("pairwise_profile") > parameters.index("clear_output")
```
praateekmahajan: no need for all these tests, try to fold them into an existing test above

tests/stages/deduplication/semantic/test_pairwise.py
```diff
@@ -378,9 +937,23 @@ def test_decompose(self) -> None:
...
         assert stages[1].embedding_field == "embedding"
         assert stages[1].output_path == "/output/path"
-        assert stages[1].pairwise_batch_size == 1024
+        assert stages[1].pairwise_batch_size == "auto"
+        assert stages[1].input_embedding_dtype == "float16"
+        assert stages[1].compute_dtype == "float32"
+        assert stages[1].profile is True
         assert stages[1].verbose is True
 
+    def test_new_options_preserve_existing_positional_order(self) -> None:
```
praateekmahajan: this test is very local, and against what AGENTS.md recommends

tests/stages/deduplication/semantic/test_workflow.py
```diff
@@ -30,6 +31,59 @@
 # Suppress GPU-related import errors when running pytest -m "not gpu"
 with suppress(ImportError):
     from nemo_curator.stages.deduplication.semantic import SemanticDeduplicationWorkflow
+    from nemo_curator.stages.deduplication.semantic import workflow as workflow_module
+
+
+@pytest.mark.gpu
+def test_pairwise_precision_and_profile_options_reach_composite_stage(
```
praateekmahajan: very local test

tests/stages/deduplication/semantic/test_workflow.py
```diff
@@ -30,6 +31,59 @@
...
+
+    pairwise_stage = captured_stages[0]
+    assert pairwise_stage.input_embedding_dtype == "auto"
+    assert pairwise_stage.compute_dtype == "float32"
+    assert pairwise_stage.pairwise_batch_size == "auto"
+    assert pairwise_stage.profile is True
+
+
+@pytest.mark.gpu
+@pytest.mark.parametrize(
```
praateekmahajan: local test

tests/stages/text/deduplication/test_semantic.py
```diff
@@ -147,6 +148,63 @@ def fail_pipeline_run(self, executor) -> None:  # noqa: ANN001, ARG001
         workflow._run_embedding_generation(executor=object())
 
 
+@pytest.mark.gpu
+def test_pairwise_precision_options_reach_semantic_workflow(
```
praateekmahajan: very local tests in these files too

tests/stages/deduplication/semantic/test_pairwise.py
```diff
@@ -88,11 +170,448 @@ def test_pairwise_cosine_similarity_batched_rand_array(self, batch_size: int) ->
         )
         np.testing.assert_array_equal(max_indices_ref.tolist(), max_indices_test.tolist())
 
+    @pytest.mark.parametrize("batch_size", [1, 2, 3, 4])
```
praateekmahajan: can we follow tests already in the file and see if they work.. those were human written and need to be followed, and ideally a similar pattern or we just make minor modification there to also tests these.. we don't need 100 granular small tests

tests/stages/deduplication/semantic/test_pairwise.py
```diff
@@ -88,11 +170,448 @@ def test_pairwise_cosine_similarity_batched_rand_array(self, batch_size: int) ->
...
+
+        with pytest.raises(ValueError, match="Unsupported compute_dtype"):
+            pairwise_cosine_similarity_batched(embeddings, compute_dtype="bfloat16")
+
 
 @pytest.mark.gpu
 class TestPairwiseCosineSimilarityStage:
     """Test cases for PairwiseCosineSimilarityStage."""
 
+    @staticmethod
```
praateekmahajan: bunch of very verbose tests below

nemo_curator/stages/deduplication/gpu_utils.py
```diff
@@ -37,3 +43,37 @@ def get_device_free_memory() -> int | None:
         return pynvml.nvmlDeviceGetMemoryInfo(handle).free
     except pynvml.NVMLError:
         return None
+
+
+def release_cached_gpu_memory() -> None:
```
praateekmahajan: @ayushdg can you review this, to see if it's not sloppy? I'm wondering if we can then also add this in the minhash teardown part

tests/stages/deduplication/semantic/test_pairwise.py
```diff
@@ -69,32 +70,82 @@ def test_pairwise_cosine_similarity_batched(self, batch_size: int) -> None:
...
-            atol=1e-5,
-        )
-        np.testing.assert_array_equal(max_indices_ref.tolist(), max_indices_test.tolist())
+        assert fp16_scores.dtype == cp.float32
+        assert fp32_scores.dtype == cp.float32
+        np.testing.assert_array_equal(fp16_indices.tolist(), fp32_indices.tolist())
+        np.testing.assert_allclose(fp16_scores.tolist(), fp32_scores.tolist(), rtol=2e-3, atol=2e-3)
+
+    @pytest.mark.parametrize("batch_size", [1, 2])
+    def test_negative_similarity_is_not_replaced_by_masked_zero(self, batch_size: int) -> None:
```
praateekmahajan, line 138: Is this ever possible because KMeans performs normalization before feeding into KMeans and then writes those embeddings out

nemo_curator/stages/deduplication/semantic/pairwise.py
```diff
@@ -24,51 +25,188 @@
...
+
+PairwiseBatchSize = int | Literal["auto"]
+PairwiseComputeDtype = Literal["auto", "float16", "float32"]
+
+_AUTO_BATCH_MEMORY_FRACTION = 0.8
+_AUTO_BATCH_OVERHEAD_FACTOR = 1.25
+_AUTO_BATCH_ALIGNMENT = 256
+
+
+class _PairwiseProfiler:
```
praateekmahajan: What value is this adding? Do we need it? if not let's get rid of it

nemo_curator/stages/deduplication/semantic/pairwise.py
```diff
@@ -24,51 +25,188 @@
...
     cluster_reps: "torch.Tensor",
-    batch_size: int = 1024,
+    batch_size: PairwiseBatchSize = 1024,
+    compute_dtype: PairwiseComputeDtype = "auto",
 ) -> tuple["cp.ndarray", "cp.ndarray"] | tuple[np.ndarray, np.ndarray]:
+    """Return each row's best earlier-ranked neighbor and similarity.
+
+    The first row maps to itself with score 0. For later rows, ties are resolved
+    deterministically in favor of the earliest-ranked candidate. Multiplication
+    uses ``compute_dtype`` while returned similarity scores are always FP32.
```
praateekmahajan: If for some reason the input dtype is FP32, then is there a need for similarity to be FP32? the stage before this KMeans, needs FP32 numbers however it can write out FP16 numbers for storage reasons.. and once it writes out fp16 then we should ensure that the math is also fp16 instead of promoting it to fp32 unless it makes sense?

tests/stages/deduplication/semantic/test_pairwise.py
```diff
@@ -69,32 +70,82 @@ def test_pairwise_cosine_similarity_batched(self, batch_size: int) -> None:
...
+        expected_similarity[0] = 0.0
+        expected_indices[0] = 0
+
+        max_similarity, max_indices = pairwise_cosine_similarity_batched(rand_arr, batch_size=batch_size)
+
+        np.testing.assert_allclose(max_similarity.tolist(), expected_similarity.tolist(), rtol=1e-5, atol=1e-5)
+        np.testing.assert_array_equal(max_indices.tolist(), expected_indices.tolist())
+
+    def test_fp16_and_fp32_agree_on_non_tie_neighbor_ids(self) -> None:
+        embeddings = torch.tensor(
```
praateekmahajan, line 121: describe this test.. such as we haev 4 rows a,b,c,d who are similar to each other with bla bla bla.. and we expect bla bla bla

nemo_curator/stages/deduplication/gpu_utils.py
```diff
@@ -37,3 +43,37 @@ def get_device_free_memory() -> int | None:
...
+    """Best-effort release of Python and GPU allocator caches."""
```
praateekmahajan: Should minhash's stage teardown also call this?

benchmarking/scripts/semdedup_identification_benchmark.py
```diff
@@ -65,7 +81,9 @@ def run_semdedup_identification_benchmark(  # noqa: PLR0913
         input_filetype: Input file type ("parquet" or "jsonl")
         eps: Epsilon value for duplicate identification threshold (cosine_sim >= 1-eps)
         which_to_keep: Strategy for ranking within clusters ("hard", "easy", "random")
-        pairwise_batch_size: Batch size for pairwise similarity computation
+        pairwise_compute_dtype: Multiplication precision used by Pairwise
```
ayushdg, line 69: nit: Should we rename this to precision?

benchmarking/nightly-benchmark.yaml
```diff
@@ -643,7 +644,8 @@ entries:
       --input-filetype=parquet
       --eps=0.01
       --which-to-keep=hard
-      --pairwise-batch-size=1024
+      --pairwise-batch-size=auto
+      --pairwise-profile
```
ayushdg: nit: Could you make the dtype precision being set to auto explicit here as well?

nemo_curator/stages/deduplication/semantic/pairwise.py
```diff
@@ -24,51 +25,204 @@
...
+
+
+def _resolve_pairwise_batch_size(cluster_reps: "torch.Tensor", batch_size: PairwiseBatchSize) -> int:
+    """Resolve a fixed or memory-derived workspace width for ``cluster_reps``."""
+    validate_pairwise_batch_size(batch_size)
+    num_rows = cluster_reps.shape[0]
+    if batch_size != "auto":
+        return min(batch_size, num_rows)
+
+    free_memory, _ = torch.cuda.mem_get_info(cluster_reps.device)
```
ayushdg: To look

nemo_curator/stages/deduplication/semantic/pairwise.py
```diff
@@ -24,51 +25,204 @@
...
+    if _estimate_pairwise_workspace_bytes(cluster_reps, 1) * _AUTO_BATCH_OVERHEAD_FACTOR > memory_budget:
+        msg = (
+            "Insufficient free GPU memory for one Pairwise similarity column: "
+            f"free_memory={free_memory}, num_rows={num_rows}, embedding_width={cluster_reps.shape[1]}, "
+            f"compute_dtype={cluster_reps.dtype}"
+        )
+        raise MemoryError(msg)
+
+    low, high = 1, num_rows
+    while low < high:
```
ayushdg: This estimate is quadratic in batch_size, so we could solve directly for the value instead of binary-searching [1, num_rows]. That said, the binary search is only O(log(num_rows)) and may be preferable for this and to keep this logic independent of the estimator’s exact formula. Is that the intent?

nemo_curator/stages/deduplication/semantic/pairwise.py
```diff
@@ -24,51 +25,204 @@
...
+    max_batch_size = low
+    if max_batch_size >= _AUTO_BATCH_ALIGNMENT:
+        max_batch_size = max_batch_size // _AUTO_BATCH_ALIGNMENT * _AUTO_BATCH_ALIGNMENT
+    return min(max_batch_size, num_rows)
 
 
 def pairwise_cosine_similarity_batched(
     cluster_reps: "torch.Tensor",
-    batch_size: int = 1024,
+    batch_size: PairwiseBatchSize = 1024,
```
ayushdg: default auto or 1024?

nemo_curator/stages/deduplication/semantic/pairwise.py
```diff
@@ -80,11 +234,14 @@ def __init__(  # noqa: PLR0913
         embedding_field: str,
         output_path: str,
         ranking_strategy: RankingStrategy,
-        pairwise_batch_size: int = 1024,
+        pairwise_batch_size: PairwiseBatchSize = 1024,
```
ayushdg: same question as above

nemo_curator/stages/deduplication/semantic/pairwise.py
```diff
@@ -80,11 +234,14 @@ def __init__(  # noqa: PLR0913
...
         verbose: bool = False,
         embedding_dim: int | None = None,
         read_kwargs: dict[str, Any] | None = None,
         write_kwargs: dict[str, Any] | None = None,
+        input_embedding_dtype: EmbeddingStorageDtype = "auto",
```
ayushdg: nAre there cases where this needs to differ from the compute dtype?

nemo_curator/stages/deduplication/semantic/pairwise.py
```diff
@@ -127,7 +307,11 @@ def process(self, task: FileGroupTask) -> FileGroupTask:
             msg = "centroid_id not found in task metadata"
             raise ValueError(msg)
 
-        t1 = time.perf_counter()
+        if self.profile:
+            torch.cuda.synchronize()
```
ayushdg: Maybe this line is obvious to torch users but can you explain why this is needed?

nemo_curator/stages/deduplication/semantic/pairwise.py
```diff
@@ -213,7 +379,29 @@ def process(self, task: FileGroupTask) -> FileGroupTask:
         ids = ranked_metadata_df[self.id_field]
 
         # Compute pairwise similarities
-        max_similarity, max_indices = pairwise_cosine_similarity_batched(cluster_embeddings, self.pairwise_batch_size)
+        resolved_compute_dtype = _resolve_compute_dtype(cluster_embeddings, self.compute_dtype)
+        cluster_embeddings = cluster_embeddings.to(device="cuda", dtype=resolved_compute_dtype)
```
ayushdg: any plans to make this more complex? iiuc "auto" just picked based on embedding dtype. So we'll never end up with stuff like bf16 etc which might be faster. (not sure if bf16 is even allowed)

nemo_curator/stages/deduplication/gpu_utils.py
```diff
@@ -37,3 +40,61 @@ def get_device_free_memory() -> int | None:
...
+        import rmm
+
+        resource = rmm.mr.get_current_device_resource()
+        release = getattr(resource, "release", None)
+        if callable(release):
+            release()
+        elif type(resource).__name__ in {"CudaAsyncMemoryResource", "CudaAsyncViewMemoryResource"}:
+            import torch
+
+            torch.cuda.synchronize()
```
ayushdg: not sure if this works as expected. Async memory can be created with a pool that won't be cleared up by torch

nemo_curator/stages/deduplication/gpu_utils.py
```diff
@@ -37,3 +40,61 @@ def get_device_free_memory() -> int | None:
         return pynvml.nvmlDeviceGetMemoryInfo(handle).free
     except pynvml.NVMLError:
         return None
+
+
+def release_cached_gpu_memory(*, reset_owned_rmm_pool: bool = False) -> None:
```
ayushdg: Not an expert on memory stuff but I know there's a way to ask cupy and possibly torch to re-use the same rmm pool being used by cudf. Not sure if we use that or if that needs to be handled separately.

benchmarking/nightly-benchmark.yaml
ayushdg: Given the synchronize calls is there a huge perf penalty to enabling this? I can see it being necessary for benchmarking nightly, but we might slip if these numbers are used for benchmarking numbers that are shared broadly.

ayushdg (commented):
At a high level the concept makes sense to me. Just had high level questions & few nits

nemo_curator/stages/deduplication/semantic/pairwise.py
praateekmahajan: Removed it, no need for auto as it doesn't really affect perf

nemo_curator/stages/deduplication/semantic/pairwise.py
praateekmahajan: Removed this

benchmarking/nightly-benchmark.yaml
praateekmahajan: Removed auto completely

nemo_curator/stages/deduplication/semantic/pairwise.py
praateekmahajan: Removed auto and default is same as before

benchmarking/scripts/semdedup_identification_benchmark.py
praateekmahajan: Can do it but I think it could become ambiguous. 

KMeans outputs FP32 or 16 and this one can perform matmul in FP32 or 16, both of which are "precision" args.

nemo_curator/stages/deduplication/semantic/pairwise.py
praateekmahajan: Good call, removed it.

benchmarking/nightly-benchmark.yaml
praateekmahajan: Got rid of the profile arg to keep API simpler, but I haven't benchmarked torch.synchronize cost separately. However the performance gain is still large enough from main that the synchronize call can stay

nemo_curator/stages/deduplication/semantic/pairwise.py
```diff
@@ -117,7 +238,35 @@ def __init__(  # noqa: PLR0913
         self.name = "PairwiseCosineSimilarityStage"
         self.resources = Resources(cpus=1.0, gpus=1.0)
 
-    def process(self, task: FileGroupTask) -> FileGroupTask:  # noqa: PLR0915
+    def setup(self, _: "WorkerMetadata | None" = None) -> None:
+        """Make Torch allocate from the RMM resource already used by cuDF and CuPy."""
+        torch.cuda.memory.change_current_allocator(rmm_torch_allocator)
+
+    def process(self, task: FileGroupTask) -> FileGroupTask:
```
sarahyurick, line 245: Nit but do we still need `process` if we are doing `process_batch` now?

nemo_curator/stages/deduplication/semantic/pairwise.py
```diff
@@ -117,7 +238,35 @@ def __init__(  # noqa: PLR0913
...
+        """Process one cluster, dropping failed frames so RMM can reclaim their allocations."""
+        try:
+            return self._process(task)
+        except BaseException as exc:
+            # An exception's traceback otherwise keeps the unwound _process
+            # frame—and its large GPU objects—alive until after this finalizer.
+            traceback.clear_frames(exc.__traceback__)
```
sarahyurick, line 252: Sorry what is the point of this?

nemo_curator/stages/deduplication/semantic/pairwise.py
```diff
@@ -210,16 +373,31 @@ def process(self, task: FileGroupTask) -> FileGroupTask:  # noqa: PLR0915
...
+        torch.cuda.synchronize()
         rank_time = time.perf_counter() - rank_start
 
         ids = ranked_metadata_df[self.id_field]
 
-        # Compute pairwise similarities
+        conversion_start = time.perf_counter()
+        resolved_compute_dtype = _resolve_compute_dtype(cluster_embeddings, self.compute_dtype)
+        storage_was_converted = cluster_embeddings.dtype != resolved_compute_dtype
+        cluster_embeddings = cluster_embeddings.to(dtype=resolved_compute_dtype)
```
sarahyurick, line 385: So you are doing `_resolve_compute_dtype` and `to(dtype=...)` here and then again inside `pairwise_cosine_similarity_batched`? Are they the same thing each time?

nemo_curator/stages/deduplication/semantic/pairwise.py
```diff
@@ -38,42 +39,157 @@
...
-    TODO: In future we can estimate memory requirement and calculate batch size dynamically.
+    compute_dtype: PairwiseComputeDtype = "float32",
+) -> tuple["cp.ndarray", "cp.ndarray"]:
+    """Return each ranked row's most similar earlier row and its similarity.
+
+    The input order is a preservation preference: row ``A`` is preferred over
+    ``B``, ``B`` over ``C``, and so on. For example, consider these normalized
+    four-dimensional embeddings::
+
+        X = A [1.00, 0.00, 0.00, 0.00]
```
sarahyurick, line 101: This seems like a nice example, is it something we should consider explaining in the docs too?

nemo_curator/stages/deduplication/semantic/pairwise.py
praateekmahajan: Great catch, this was a bug because of fan-in in the subsequent stage

nemo_curator/stages/deduplication/semantic/pairwise.py
praateekmahajan: Added to docs, not sure if it's in a great place though @lbliii

nemo_curator/stages/deduplication/gpu_utils.py
praateekmahajan: Great call out, added a comment in the PR description regarding the same, about what we tried and what we learnt.

nemo_curator/stages/deduplication/semantic/pairwise.py
praateekmahajan: Fixed

nemo_curator/stages/deduplication/semantic/pairwise.py
```diff
@@ -38,42 +39,162 @@
...
+        msg = f"Pairwise requires float16 or float32 embeddings, got {cluster_reps.dtype}"
+        raise TypeError(msg)
+    if cluster_reps.dtype == torch.float16 and compute_dtype == "float32":
+        msg = "float16 embeddings must use compute_dtype='auto' or 'float16'"
+        raise ValueError(msg)
+    if compute_dtype == "float16":
+        return torch.float16
+    if compute_dtype == "float32":
+        return torch.float32
+    return cluster_reps.dtype
```
sarahyurick, line 80: Nit: it seems to be calling these functions a bunch, just want to make sure this many `validate` checks are intended?

nemo_curator/stages/deduplication/semantic/pairwise.py
```diff
@@ -117,8 +243,19 @@ def __init__(  # noqa: PLR0913
...
         self.resources = Resources(cpus=1.0, gpus=1.0)
 
+    def setup(self, _: "WorkerMetadata | None" = None) -> None:
+        """Make cuDF, CuPy, and Torch share a growing RMM pool."""
+        self._rmm_memory_resource = rmm.mr.PoolMemoryResource(
+            rmm.mr.CudaMemoryResource(),
+            initial_pool_size="1GB",
+        )
+        rmm.mr.set_current_device_resource(self._rmm_memory_resource)
+        torch.cuda.memory.change_current_allocator(rmm_torch_allocator)
```
VibhuJawa, line 248: @praateekmahajan , just to be sure,  if we use Pool Allocator it will mean that it might grow to the whole GPU's memory so if some one to set `gpus<1.0` it will blow up.

sarahyurick (approved):

nemo_curator/stages/deduplication/semantic/pairwise.py
```diff
@@ -210,16 +361,29 @@ def process(self, task: FileGroupTask) -> FileGroupTask:  # noqa: PLR0915
             raise RuntimeError(msg)
         del embeddings, first_embeddings, embedding_arrays, frames
         cluster_embeddings = torch.as_tensor(ranked_embeddings, device="cuda")
+        # Finish queued reordering before closing the rank timing interval.
+        torch.cuda.synchronize()
```
VibhuJawa, line 360: If we dont care about timing indivisual things , we can remove this

VibhuJawa (commented):
Mostly looks good to me , had a minor feedback on GPU memory pool .

nemo_curator/stages/deduplication/semantic/pairwise.py
praateekmahajan: I don't see any one setting this to < 1gpu, because that implies the user is confident that each cluster would fit on <1 gpu and the other cluster which will also get scheduled on that GPU potentially will also be able to fit in that remaining fraction.

nemo_curator/stages/deduplication/semantic/pairwise.py
praateekmahajan: We do want the individual timings so that we can see which specific part got faster / slower

PR #2323, opened 2026-08-19
[benchmarking] Add benchmark SKU override config

rlratzel:
## Summary
- rename the canonical full-suite benchmark config from `nightly-benchmark.yaml` to `benchmarks.yaml`
- add `4xGB200-64CPU.yaml` as an override for 4-GPU/64-CPU GB200 benchmark runs
- document how to layer SKU overrides and add config merge tests for the default and 4xGB200-64CPU configs

## Notes
- `benchmarks.yaml` remains a complete runnable config calibrated for the default 8xH100 nightly environment
- `4xGB200-64CPU.yaml` keeps timeout values within the existing 3h59m job-runtime ceiling, overrides explicit 8-worker/8-replica settings where needed, sets the object store to 400 GiB, and uses historical 4-GPU video throughput thresholds from the pre-8GPU config history
- other performance requirements are inherited until 4xGB200-64CPU-specific baselines are measured

## Testing
- `python -m py_compile tests/benchmarking/runner/test_benchmark_configs.py benchmarking/scripts/alm_pipeline_benchmark.py`
- directly invoked the new config test functions because local `pytest` startup imports `ray` from `tests/conftest.py`, which is not installed in this environment
- `git diff --check`

### DIFF
--- a/benchmarking/4xGB200-64CPU.yaml
+++ b/benchmarking/4xGB200-64CPU.yaml
@@ -1,11 +1,11 @@
-# 4xGB200-ptyche benchmark overrides.
+# 4xGB200-64CPU benchmark overrides.
 #
-# Layer this file after benchmarks.yaml for a 4-GPU, 64-CPU GB200/Ptyche environment.
+# Layer this file after benchmarks.yaml for a 4-GPU, 64-CPU GB200 environment.
 # This override halves the default Ray CPU/GPU resources, increases timeout
 # values within the existing 3h59m job-runtime ceiling, and adjusts explicit
 # GPU-stage worker counts from 8 to 4. Known historical 4-GPU video throughput
 # requirements are included here; other performance requirements are inherited
-# from benchmarks.yaml until GB200/Ptyche-specific baselines are measured and added.
+# from benchmarks.yaml until 4xGB200-64CPU-specific baselines are measured and added.
 #
 # NDD entries retain their current names for continuity, but this file overrides
 # their serving replica count from 8 to 4. Their current 8x performance thresholds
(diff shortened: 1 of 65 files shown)


benchmarking/4xGB200-64CPU.yaml
```diff
@@ -0,0 +1,242 @@
+# 4xGB200-64CPU benchmark overrides.
+#
+# Layer this file after benchmarks.yaml for a 4-GPU, 64-CPU GB200 environment.
+# This override halves the default Ray CPU/GPU resources, increases timeout
+# values within the existing 3h59m job-runtime ceiling, and adjusts explicit
+# GPU-stage worker counts from 8 to 4. Known historical 4-GPU video throughput
+# requirements are included here; other performance requirements are inherited
+# from benchmarks.yaml until 4xGB200-64CPU-specific baselines are measured and added.
+#
+# NDD entries retain their current names for continuity, but this file overrides
```
sarahyurick: Feel free to rename them to `ndd_dynamo` and `ndd_ray_serve`. As long as the replicas are being exposed in this file and the `benchmarks.yaml` file then I think it should be fine.

benchmarking/4xGB200-64CPU.yaml
```diff
@@ -0,0 +1,242 @@
+# 4xGB200-64CPU benchmark overrides.
```
praateekmahajan, line 1: Should we also split the current nightly.yaml to nightly.yaml and 8*H100.yaml?

rlratzel: Thanks for thinking about this, I'm looking for some feedback on my approach here:
I renamed the "base" file to `benchmarks.yaml` since "nightly-*" didn't seem as accurate anymore, because it was being used as a definition of the full-suite for any type of run (nightly, ad-hoc, etc.).  I also wanted it to be useful as-is, so users wouldn't be burdened with having to understand that it needs overrides. `benchmarks.yaml` will work on the cluster we use for nightlies and most ad-hoc runs - basically, it's still backwards compatible with the old `nightly-benchmark.yaml` use cases.

We could have a whole family of override files though, including 8xH100, but I was worried that users wouldn't like having to use `--config benchmark.yaml --config 8xH100.yaml` for their runs.  Although, at the same time, I don't like how it's not obvious that H100 runs are supported by default and everything else needs an override file.  I don't feel too strongly, but I should mention that it might be tricky to generate a meaningful error message when users forget to specify an override file - they'd just get strange errors or worse: defaults that allow benchmarks to pass when they shouldn't.

sarahyurick (approved):

PR #2324, opened 2026-08-20
Add LLM-as-a-judge to eval suite and bump NDD version

sarahyurick:
Things to consider:

- [x] Agent skills file for how to write the YAML and Jinja files
- [x] How to handle empty examples
- [x] How to truncate according to token limits (opened a Linear issue to track it as a follow-up: https://linear.app/nvidia/issue/NMCUR-402/enhance-llm-judge-eval-to-filtermodify-by-token-count-limits)
- [x] Multiple LLMs doing the same judging task
- [x] Instructions about LLMs to use as judges
- [x] Instructions for Slurm array partitioning
- [x] Instructions for resumability and retry support. Like with my small test run, there were 192 input records but 189 output records. Likely an NDD refusal or error for some reason. Do we care about getting every single judgement?
- [x] Test NDD's `LLMJudgeColumnConfig.multi_modal_context` (e.g., `ImageContext`)
- [x] vLLM shutdown issue when I Ctrl+C a run (not in scope for this PR)


eval/llm_judge/generic_text_judge.py
```diff
@@ -0,0 +1,541 @@
...
+from nemo_curator.stages.synthetic.nemo_data_designer import DataDesignerStage
+from nemo_curator.stages.text.filters import Filter, ScoreFilter
+from nemo_curator.stages.text.io.reader import JsonlReader, ParquetReader
+from nemo_curator.stages.text.io.writer import JsonlWriter, ParquetWriter
+from nemo_curator.tasks import DocumentBatch
+
+
+DataFormat = Literal["jsonl", "parquet"]
+FilterOperator = Literal["eq", "ne", "gt", "gte", "lt", "lte", "in", "not_in"]
+_FILTER_OPERATORS = {"eq", "ne", "gt", "gte", "lt", "lte", "in", "not_in"}
```
sarahyurick: I think we have some operator typing in audio somewhere... if yes, we should make it a generic util (if it is not already).

eval/llm_judge/generic_text_judge.py
```diff
@@ -0,0 +1,541 @@
...
+
+
+class _LoggingLanguageFilter(ScoreFilter):
```
sarahyurick: Not at all needed, just added for my own curiosity. I didn't check if our `ScoreFilter`, etc. classes have more verbose logging?

eval/llm_judge/generic_text_judge.py
```diff
@@ -0,0 +1,541 @@
...
+    if not isinstance(config, dict):
+        raise ValueError(f"Judge config must contain a mapping: {path}")
+    return config
+
+
+def _read_template(path: str, *, config_path: Path) -> str:
+    template_path = Path(path)
+    if not template_path.is_absolute():
+        template_path = config_path.parent / template_path
+    return template_path.read_text(encoding="utf-8")
```
sarahyurick: Again these should be utils, if they are not already.

eval/llm_judge/generic_text_judge.py
```diff
@@ -0,0 +1,541 @@
...
+    return models
+
+
+def _load_execution_stages(config: dict[str, Any]) -> tuple[str, dict[str, Any], list[dict[str, Any]]]:
+    """Load judge groups and the requested one- or many-NDD-stage execution mode."""
+    execution = config.get("execution", {})
+    if not isinstance(execution, dict):
+        raise ValueError("'execution' must be a mapping when provided.")
+    mode = execution.get("mode", "single_stage")
+    if mode not in {"single_stage", "multi_stage"}:
```
sarahyurick: I am not sure if splitting up each judge task into its own NDD stage is worth it, or if it should all stay within the same NDD stage. Both are possible. Not sure it matters though.

eval/llm_judge/run_llm_judge.py
```diff
@@ -0,0 +1,541 @@
...
+    output_path: str,
+    output_format: DataFormat,
+    judge_stages: list[
+        tuple[str, dd.DataDesignerConfigBuilder, list[dd.ModelProvider], dict[str, Any] | None, list[dict[str, Any]]]
+    ],
+    language_filter_stage: ScoreFilter | None,
+    files_per_partition: int | None,
+) -> Pipeline:
+    """Build one streaming Curator reader → optional language gate → NDD stages → filters → writer pipeline."""
+    # TODO: Add an optional TokenLengthFilter stage before NDD stages so prompts
```
sarahyurick, line 304: Big TODO. Right now I just do string truncation in the jinja file. But we should decide the best course of action.

Also, since some extractors might return null while others return content, we should decide whether those examples are evaluated or not? Right now I force the empty string and still give it to the judge.

eval/llm_judge/README.md
```diff
@@ -0,0 +1,436 @@
+# Generic text LLM judge
+
+Use this example when you have JSONL or Parquet records and want an LLM to evaluate fields in each record with your own prompt and rubric. You supply the input data, a YAML configuration, and one or more small Jinja prompt files.
+
+The runner starts a local Curator Dynamo/vLLM `InferenceServer`, runs NeMo Data Designer (NDD) judge columns, and writes the original records plus judge output. The bundled example compares `raw_text`, `justext_text`, and `trafilatura_text`. Copy and adapt it to judge a PDF parser result against raw PDF text, assess whether two documents are semantic duplicates, or evaluate another text-processing result.
```
sarahyurick: Very specific example to the data generated by `cc_extract.py`. But everything is supposed to be generic as long as the user provides the input path, YAML, and Jinja file(s).

pyproject.toml
```diff
@@ -285,7 +285,7 @@ interleaved_cuda12 = [
 
 # Synthetic Data Generation (SDG) Dependencies
 sdg_cpu = [
-    "data-designer==0.5.5",
+    "data-designer==0.9.1",
```
sarahyurick, line 288: Using this version for my tests. It is only like 2 commits behind NDD main right now.

eval/llm_judge/cc_extract_example/text_extraction_qwen_judge.yaml
```diff
@@ -0,0 +1,94 @@
+# User-owned evaluation semantics: replace prompts, rubrics, models, and execution mode for another task.
+models:
+  - alias: judge
```
sarahyurick, line 6: It can handle as many models as we want to serve (assuming the GPU math works out). Each should have a unique alias.

eval/llm_judge/examples/text_extraction_judge.yaml
```diff
@@ -0,0 +1,94 @@
+# User-owned evaluation semantics: replace prompts, rubrics, models, and execution mode for another task.
+models:
+  - alias: judge
+    model: /raid/syurick/hf_cache/Qwen3.8-27B
```
sarahyurick: Oops.
```suggestion
    model: /path/to/your/hf_cache/Qwen3.8-27B
```

eval/llm_judge/cc_extract_example/text_extraction_qwen_judge.yaml
```diff
@@ -0,0 +1,94 @@
...
+      judges:
+        - name: qwen3_8_27b_text_extraction_judgment
+          model_alias: judge
+          system_prompt_path: text_extraction_system.jinja
+          prompt_path: text_extraction_prompt.jinja
+          with_trace: last_message
+          extract_reasoning_content: false
+          scores:
+            - name: best_extraction
+              description: Select the candidate that is most useful as clean document text.
```
sarahyurick, line 44: I had Codex generate all the Jinja, scoring, etc. prompts. Just a POC for now.

eval/llm_judge/generic_text_judge.py
```diff
@@ -0,0 +1,541 @@
...
+        raise ValueError("execution.stages must be a non-empty list of mappings.")
+
+    stage_names: set[str] = set()
+    judge_names: set[str] = set()
+    for stage in stages:
+        if not stage.get("name"):
+            raise ValueError("Every execution stage requires a unique 'name'.")
+        name = str(stage["name"])
+        if name in stage_names:
+            raise ValueError(f"Execution stage names must be unique; {name!r} appears more than once.")
```
sarahyurick: Sorry for the slop. We shouldn't need all of these checks.

eval/llm_judge/generic_text_judge.py
```diff
@@ -0,0 +1,541 @@
...
+                raise ValueError(f"Every judge in execution stage {name!r} requires a unique 'name'.")
+            if str(judge_name) in judge_names:
+                raise ValueError(f"Judge names must be unique; {judge_name!r} appears more than once.")
+            judge_names.add(str(judge_name))
+        if stage.get("runtime_env") is not None and not isinstance(stage["runtime_env"], dict):
+            raise ValueError(f"execution stage {name!r}.runtime_env must be a mapping.")
+    return str(mode), execution, stages
+
+
+def _validate_filters(
```
sarahyurick: More slop, sorry. Maybe some could be useful but most looks like it is not.

eval/llm_judge/generic_text_judge.py
```diff
@@ -0,0 +1,541 @@
...
+                inference_parameters=dd.ChatCompletionInferenceParams(**model.get("inference_parameters", {})),
+            )
+            for model in models
+        ]
+    )
+
+    judge_names: set[str] = set()
+    for judge in judges:
+        if not judge.get("name") or not judge.get("prompt_path") or not judge.get("scores"):
+            raise ValueError("Every judge requires 'name', 'prompt_path', and non-empty 'scores'.")
```
sarahyurick: Same here. For the most part we should remove a lot of checks. Will do it in future commits.

eval/llm_judge/run_llm_judge.py
```diff
@@ -0,0 +1,482 @@
...
+        template_path = config_path.parent / template_path
+    return template_path.read_text(encoding="utf-8")
+
+
+def _place_filters(config: dict[str, object], stages: list[dict[str, object]]) -> list[list[dict[str, object]]]:
+    """Place top-level filters after the NDD stage that produces their judge column."""
+    producer_stage_by_judge = {
+        str(judge["name"]): index for index, stage in enumerate(stages) for judge in stage["judges"]
+    }
+    stage_filters = [list(stage.get("filters", [])) for stage in stages]
```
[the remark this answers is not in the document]
sarahyurick: Resolved in [c59e534](https://github.com/NVIDIA-NeMo/Curator/pull/2324/commits/c59e53435ddd7f2633cb65b3b53cdf12dea29eb1).

eval/llm_judge/generic_text_judge.py
sarahyurick: Keeping both for now. I think `single_stage` is more in-line with existing Curator + NDD patterns, but `multi_stage` is nice if each individual judges need their own `runtime_env` and if we want to use resumability between judging stages.

eval/llm_judge/run_llm_judge.py
sarahyurick: Opened a Linear issue, since this PR is already fairly large: https://linear.app/nvidia/issue/NMCUR-402/enhance-llm-judge-eval-to-filtermodify-by-token-count-limits.

pyproject.toml
sarahyurick: cc @ayushdg if you have any opinions about these changes?

pyproject.toml
```diff
@@ -351,13 +351,14 @@ constraint-dependencies = [
...
-    "fsspec>=2025.3.0", # required for nemo-toolkit[asr]==2.7.2 (fsspec==2024.12.0), override due to data-designer-engine==0.5.5 requires >=2025.3.0
+    "fsspec>=2025.3.0", # Overrides nemo-toolkit[asr]'s fsspec==2024.12.0 pin; data-designer-engine==0.9.1 requires >=2025.3.0
     "apex; sys_platform == 'never'",
     "distance; sys_platform == 'never'",
-    "huggingface-hub>=0.34,<1.0", # Override huggingface-hub, transformers and data-designer require two different versions of hugging-face hub
+    "huggingface-hub>=0.34,<1.0", # Overrides Transformers' <1 cap and data-designer==0.9.1's >=1.0.1 requirement
     "kaldiio;  sys_platform == 'never'",
     "levenshtein;  sys_platform == 'never'",
     "numba==0.65.0", # Override RAPIDS/legacy caps for the inference image; vLLM 0.22 requires numba 0.65.0
+    "packaging>=25", # Required by data-designer==0.9.1; overrides Lightning's packaging<25 cap from nemo-toolkit[asr]
```
ayushdg, line 361: Let's add a followup ticket to bump asr. I think some of these overrides can be avoided.

pyproject.toml
```diff
@@ -366,8 +367,9 @@ override-dependencies = [
     "torchcodec~=0.11.0; platform_machine == 'x86_64' and platform_system != 'Darwin'", # TorchCodec 0.11 matches torch 2.11; TorchCodec does not declare a torch dependency, so enforce the ABI-compatible series across direct and transitive requirements; satisfies pyannote-audio's >=0.7.0 floor; x86_64-only since aarch64 lacks wheels
     "nixl-cu12>=0.10.0; platform_system == 'Linux' and (platform_machine == 'x86_64' or platform_machine == 'aarch64')",  # Use the CUDA 12 backend on both supported Linux architectures.
     "xgrammar>=0.1.32", # Override vllm's ==0.1.29 pin to address CVE GHSA-7rgv-gqhr-fxg3 (DoS via multi-layer nesting)
-    "sqlfluff>=4.2.0", # Address CVE-2026-46373/46374 (parser DoS); overrides data-designer-engine==0.5.5 sqlfluff<4 cap
-    "pandas>=3.0.0, <= 3.0.4" # Override data-designers constraint of <3
+    "sqlfluff>=4.2.0", # Address CVE-2026-46373/46374 (parser DoS); data-designer-engine==0.9.1 supports sqlfluff>=4.1,<5
```
ayushdg: We can probably get rid of this and do `uv lock --upgrade-package sqlfluff` to have uv pick up a newer version rather than needing an override now.

uv.lock
```diff
@@ -1306,45 +1320,45 @@ toml = [
 
 [[package]]
 name = "cryptography"
-version = "50.0.0"
+version = "49.0.0"
```
ayushdg: Can we also override cryptography. Looks like NDD adds a pin but should work with newer versions. It's just following older conventions of keeping an upper pin: https://github.com/NVIDIA-NeMo/DataDesigner/pull/821/changes. 
`>=50` is recommended for https://github.com/NVIDIA-NeMo/Curator/security/dependabot/217.

We can also open an issue in NDD to remove the upper pin on cryptography

uv.lock
```diff
@@ -2864,14 +2905,14 @@ wheels = [
 
 [[package]]
 name = "gitpython"
-version = "3.1.58"
+version = "3.1.57"
```
ayushdg: Do we know if a specific package is downgrading this?

ayushdg (commented):
Minor comments around dependencies. Thanks a lot for the bump!

uv.lock
sarahyurick: Not sure. Looks like the latest is 3.1.60. I can do `uv lock --upgrade-package GitPython`.

eval/llm_judge/run_llm_judge.py
```diff
@@ -0,0 +1,494 @@
...
+    )
+    for filter_config, filter_stage_index in filters:
+        judge_name = str(filter_config["judge"])
+        score_name = str(filter_config["score"])
+        if judge_name not in judge_scores:
+            msg = f"Filter refers to unknown judge output column {judge_name!r}."
+            raise ValueError(msg)
+        if score_name not in judge_scores[judge_name]:
+            msg = f"Filter refers to unknown score {score_name!r} on judge {judge_name!r}."
+            raise ValueError(msg)
```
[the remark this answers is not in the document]
sarahyurick: Fixed.

pyproject.toml
sarahyurick: Opened a GitHub issue: https://github.com/NVIDIA-NeMo/Speech/issues/16144 and a Linear issue: https://linear.app/nvidia/issue/NMCUR-404/relax-the-lightning240-constraint-for-nemo-speech-asr.

eval/llm_judge/generic_text_judge.py
sarahyurick: There is something similar in audio but it is not directly compatible with this.

eval/llm_judge/README.md
```diff
@@ -0,0 +1,259 @@
...
+  --output-format jsonl
+```
+
+The bundled [text_extraction_qwen_gemma_judges.yaml](cc_extract_example/text_extraction_qwen_gemma_judges.yaml) runs the same extraction rubrics with both Qwen and Gemma. Use it when you want to compare model agreement; update the model paths and serving settings for your hardware before running it.
+
+Use `--checkpoint-path output/judge_checkpoint` to write Curator checkpoint metadata to a durable location. It is useful for normal pipeline recovery, but you should still inspect input and output counts after a run.
+
+## Input and output
+
+The runner does not require a fixed text schema. A prompt can reference any fields present in an input JSONL or Parquet row. Keep a stable identifier such as `document_id` or `track_id` when you need to join results to another dataset.
```
VibhuJawa: What is a track_id ?

eval/llm_judge/README.md
```diff
@@ -0,0 +1,259 @@
...
+```json
+{
+  "document_id": "article-42",
+  "raw_text": "Example site | Subscribe | The article begins here ...",
+  "justext_text": "The article begins here ...",
+  "trafilatura_text": "The article begins here, with an extra footer ..."
+}
+```
+
+NDD adds one top-level column for each judge. A judge named `extraction_quality` with a `quality` score produces a result shaped like this:
```
VibhuJawa, line 44: Quick question: Is each judge 1 LLM call or multiple LLM calls ?

eval/llm_judge/README.md
```diff
@@ -0,0 +1,259 @@
...
+Use a HuggingFace-format model vLLM can load, either local weights or a repo id. Some architectures require `trust_remote_code: true` in `engine_kwargs`, and architecture support varies by the installed Dynamo/vLLM version, so a serving failure can mean the version needs updating rather than that the model is unusable. The model has to fit the GPUs it's given (`tensor_parallel_size` per replica × `num_replicas`), and `max_model_len` has to cover the full rendered prompt plus `max_tokens`.
+
+To download the Qwen judge model to a local path referenced by `models[].model`, for example:
+
+```bash
+hf download Qwen/Qwen3.8-27B \
+  --local-dir /path/to/Qwen3.8-27B
+```
+
+A model that serves fine can still be a bad judge, and the only way to know is to run it. The prompt asks the model in plain language to answer with one of a fixed set of values, like "answer with exactly one of: yes, no, unclear." The pipeline checks that answer against the schema afterward. It doesn't stop the model from answering wrong in the first place — it just drops any row where the answer doesn't match. So a model that doesn't follow instructions well shows up as missing rows, not wrong scores.
```
VibhuJawa, line 151: The thing that helped me identify bad judges if we  swap_candidates and results change/hold. 

I wonder if we should make that a first class citizen ?

eval/llm_judge/README.md
```diff
@@ -0,0 +1,259 @@
...
+    score: quality
+    operator: gte
+    value: 4
+```
+
+The example supports `eq`, `ne`, `gt`, `gte`, `lt`, `lte`, `in`, and `not_in`. Multiple filters use AND semantics. Top-level filters are placed immediately after the stage that produces their judge column; a filter can also be placed under a specific execution stage when you deliberately need it later. Before Ray or the model server starts, the runner checks that every filter refers to a configured judge column and score.
+
+## Analyzing results
+
+Running the same rubric through multiple LLMs turns judge agreement into a signal, not just a sanity check: where the models agree, the record is likely easy and the score can be trusted with less scrutiny; where they disagree, look into why before trusting the rubric or filter at scale. A disagreement can mean the record is genuinely ambiguous or hard to score — evidence for a human-in-the-loop or an `unresolved`-style rubric option — or it can mean the prompt or rubric wording is too vague or underspecified for a model to apply consistently, which calls for tightening the prompt rather than trusting either score.
```
VibhuJawa, line 204: This is also my lived experience

eval/llm_judge/README.md
```diff
@@ -0,0 +1,259 @@
...
+
+`max_model_len` limits total request context, while `max_tokens` limits only the completion. Increasing `max_model_len` consumes KV-cache capacity; it does not make oversized raw documents safe. `max_num_seqs` should be at least the intended in-flight load, but increasing it by itself does not improve throughput.
+
+For the optional FastText language gate, provide `--language`, `--fasttext-langid-model-path`, and optionally `--min-langid-score` and `--language-text-field`. Omitting `--language` skips the stage and does not require FastText.
+
+Press `Ctrl-C` once to cancel a local Dynamo run and allow normal Ray and inference-server cleanup to finish. If cancellation interrupts cleanup, inspect remaining model-server subprocesses before starting another run.
```
VibhuJawa: Dont think this is  how we/our customers run things. Probably good to clean it up. We should try to keep this llm judge relevant.

eval/llm_judge/cc_extract_example/text_extraction_disagreement_prompt.jinja
```diff
@@ -0,0 +1,16 @@
...
+{{ (raw_text or "")[:12000] }}
+</raw_text>
+
+<justext_text>
+{{ (justext_text or "")[:8000] }}
+</justext_text>
+
+<trafilatura_text>
+{{ (trafilatura_text or "")[:8000] }}
+</trafilatura_text>
```
VibhuJawa: Just adding a todo for us to add `few-shot calibration examples`

eval/llm_judge/LLM_JUDGE_CONFIG_SKILL.md
```diff
@@ -0,0 +1,232 @@
...
+optional filters on the results. Prefer adapting an existing working YAML/Jinja
+pair in this repo as a starting point rather than writing from scratch, then
+change only what the task requires.
+
+Before writing anything, pin down:
+1. The actual input row schema (field names, which fields can be `null`).
+2. The decision the judge must make and what evidence it may use.
+3. The output contract: judge/score names and option values that downstream
+   code depends on.
+4. Available serving resources (GPUs, model).
```
VibhuJawa: 5. The evaluation mode: pointwise (each judge assigns a score and we compare scores), pairwise(compare a vs b ), or both, including explicit tie
   and insufficient-evidence semantics.

6. The  policy for pairwise evaluations: use
   anonymous candidate labels, evaluate both A/B and B/A when position bias
   matters, and preserve the mapping back to the original candidates.

8. The aggregation and escalation policy for repeated or multi-model
   judgments, including what constitutes disagreement or an unresolved result.

eval/llm_judge/LLM_JUDGE_CONFIG_SKILL.md
```diff
@@ -0,0 +1,232 @@
...
+(e.g. `unclear`) for categorical ones. Use bare numeric keys. Quote string
+labels that YAML would otherwise coerce to another type, such as `"yes"`,
+`"no"`, `"true"`, `"false"`, `"on"`, `"off"`, and `"null"`, so
+`yaml.safe_load` does not convert them to booleans or null. Keep one score's
+option values deliberately typed and stable — downstream filters and
+analysis scripts compare against them.
+
+## Writing the Jinja prompt
+
+- `{{ field_name }}` renders a value from the current input row. Only
```
VibhuJawa: - For pairwise comparisons, use neutral labels such as `candidate_a` and
  `candidate_b`. Do not expose identity unless their
  identity is legitimately part of the evaluation criterion. Source labels can
  influence the judge independently of candidate quality.

- When position bias could affect the result, render two mirrored prompts:
  A/B and B/A. Preserve the mapping from each anonymous position to the
  original candidate. If the two judgments do not identify the same underlying
  winner, mark the result as position-inconsistent rather than silently
  averaging it or converting it to an ordinary tie.

- Test prompts with controlled counterfactuals before scaling. Useful fixtures
  include the same content with candidates swapped, source names removed, superficial formatting changed, or nested instructions inserted into untrusted content. Change one factor per fixture  so failures can be attributed to a specific sensitivity.

eval/llm_judge/README.md
```diff
@@ -0,0 +1,259 @@
...
+```jinja
+The first judge gave content fidelity: {{ extraction_quality.content_fidelity.score }}
+```
+
+NDD detects that dependency and runs the first judge before the dependent one. Omitting `.score` inserts the complete structured result, including its reasoning.
+
+## YAML configuration
+
+The YAML has two main sections: `models` describes what Dynamo/vLLM serves, and `execution.stages` describes the judge columns to run.
+
```
VibhuJawa, line 94: Example `yaml`

eval/llm_judge/README.md
```diff
@@ -0,0 +1,259 @@
...
+models:
+  - alias: judge
+    model: YOUR_JUDGE_MODEL
+    served_model_name: YOUR_JUDGE_MODEL
+    dynamo_model:
+      num_replicas: 1
+      mode: aggregated
+      engine_kwargs:
+        tensor_parallel_size: 1
+        max_model_len: 32768
```
VibhuJawa, line 105: I wonder if by default we should make this -1 or support model's max inherent length .  Like 32k is a runtime optimization. 
 
```suggestion
        max_model_len: -1
```

eval/llm_judge/run_llm_judge.py
```diff
@@ -0,0 +1,504 @@
...
+DataFormat = Literal["jsonl", "parquet"]
+FilterOperator = Literal["eq", "ne", "gt", "gte", "lt", "lte", "in", "not_in"]
+
+
+def _load_yaml(path: Path) -> dict[str, object]:
+    with path.open(encoding="utf-8") as file:
+        config = yaml.safe_load(file)
+    if not isinstance(config, dict):
+        msg = f"Judge config must contain a mapping: {path}"
+        raise TypeError(msg)
```
VibhuJawa, line 58: Should we move this to utils or something ? Feels like a util for us to upstream

eval/llm_judge/run_llm_judge.py
```diff
@@ -0,0 +1,504 @@
...
+
+def _read_template(path: str, *, config_path: Path) -> str:
+    template_path = Path(path)
+    if not template_path.is_absolute():
+        template_path = config_path.parent / template_path
+    return template_path.read_text(encoding="utf-8")
+
+
+def _place_filters(config: dict[str, object], stages: list[dict[str, object]]) -> list[list[dict[str, object]]]:
+    """Place top-level filters after the NDD stage that produces their judge column."""
```
VibhuJawa, line 70: For my understanding: What is the use of place filters produces their judge column

eval/llm_judge/run_llm_judge.py
```diff
@@ -0,0 +1,504 @@
...
+
+def _get_num_workers(config: dict[str, object], *, owner: str) -> int | None:
+    """Return an optional fixed Ray worker count for one NDD stage."""
+    num_workers = config.get("num_workers")
+    if num_workers is None:
+        return None
+    if isinstance(num_workers, bool) or not isinstance(num_workers, int) or num_workers <= 0:
+        msg = f"{owner} must be a positive integer."
+        raise ValueError(msg)
+    return num_workers
```
VibhuJawa: Maybe move to a common util or something ?

eval/llm_judge/run_llm_judge.py
```diff
@@ -0,0 +1,504 @@
...
+    return ScoreFilter(
+        filter_obj=FastTextLangId(
+            model_path=model_path,
+            min_langid_score=min_score,
+            lang=language,
+        ),
+        text_field=text_field,
+        verbose=True,
+    ).with_(name="fasttext_language_filter")
+
```
VibhuJawa, line 192: I get the motivation for this but i wonder if we want it as a generic feature here ?

eval/llm_judge/run_llm_judge.py
```diff
@@ -0,0 +1,504 @@
...
+                        [filter_config for filters in stage_filters for filter_config in filters],
+                    )
+                ],
+                language_filter_stage=language_filter_stage,
+                files_per_partition=args.files_per_partition,
+            )
+            pipeline.run(executor=RayDataExecutor(), checkpoint_path=args.checkpoint_path)
+        else:
+            judge_stages = []
+            for stage, filters_after_stage in zip(configured_stages, stage_filters, strict=True):
```
VibhuJawa: What is the intuition b/w using 1 stage vs multiple stages , do we expect perf to be better b/w  them ?

VibhuJawa (commented):
Still understanding the workflow. Left an initial review .  I wonder if we should do pairwise comparisons as a first pass vs score based judges ?

eval/llm_judge/README.md
sarahyurick: It was a column in a dataset I was using haha. Can remove it here.

eval/llm_judge/README.md
sarahyurick: Each `judges` entry (under `execution.stages.name`) in the YAML file is 1 LLM call per row (no matter how many `scores` it has).

- `cc_extract_example/text_extraction_qwen_judge.yaml`: 2 `judges` = 2 calls/row
- `cc_extract_example/text_extraction_qwen_gemma_judges.yaml`: 4 `judges` = 4 calls/row

eval/llm_judge/README.md
sarahyurick: Sure, I can add guidance about that.

I think building runner support for it (e.g., a `--swap-check` flag) would be a bit more overhead than it is worth (the Python script would have to guess or be told which prompt fields to swap, etc.). Whereas with the existing implementation, it is extremely easy to just edit the YAML file by duplicating the judge and pointing its prompt at a version with `candidate_a`/`candidate_b` swapped (or add a second Jinja template with fields reversed). WDYT?

eval/llm_judge/LLM_JUDGE_CONFIG_SKILL.md
sarahyurick: Good call outs, I will update the example YAML/Jinja files to follow these rules too.

eval/llm_judge/README.md
sarahyurick: I think Dynamo/vLLM can already infer it. The reason I have it explicitly listed is so that the user can have a general idea/knowledge that they should do character or token truncation.

I am not sure about -1 but I think if the line was removed entirely from `engine_kwargs` (`max_model_len=None`) then it would work already.

eval/llm_judge/run_llm_judge.py
sarahyurick: Sure I can move it to a separate file. Although I don't think it would be very relevant for any other pipelines right now, since this would be our first example of using a YAML for `InferenceServer` things.

eval/llm_judge/run_llm_judge.py
sarahyurick: The idea is to place `Filter` stages immediately after the judge stage that is producing the relevant column. For example, if judge 1 produces `extraction_quality` and you filter on it, placing the filter right after stage 1 means rows that fail the filter get dropped before judge 2 ever runs on them, saving that judge's LLM calls on rows you were going to discard anyway.

Although I am not so sure I want to keep all the filtering logic anyway. Like realistically, the user will probably want to inspect the "bad" columns themselves and keep the judging information, rather than just throwing it away. WDYT?

eval/llm_judge/run_llm_judge.py
sarahyurick: Can you explain more what your suggestion is? IMO it is nice for doing language-specific testing, although I can understand it isn't applicable to every use case. It is mostly supposed to be a "nice-to-have."

eval/llm_judge/run_llm_judge.py
sarahyurick: Initially I was seeing better perf by using 1 stage, but the advantage of using multiple stages is that it enables resumability between each Data Designer stage.

I can try to do more rigorous benchmarks though. Like if multi stage can get the same perf as single stage, then we should only support multi stage.

eval/llm_judge/run_llm_judge.py
sarahyurick: I think I will remove the function actually. It is an unnecessary check IMO.

eval/llm_judge/README.md
VibhuJawa: Yeah, we can have 2 yaml files

eval/llm_judge/run_llm_judge.py
sarahyurick: ~~Hmm I ran another comparison benchmark and still found single stage to be faster. So I am thinking for now to just keep both, so that we can still enable resumability with multi stage.~~

Actually, the `single_stage` vs `multi_stage` comparison I was looking at isn't totally valid. `single_stage` took 67m but only wrote 411 rows; `multi_stage` took 80m and wrote 1553 rows from the same eligible pool (409 of the 411 overlap). So `single_stage` "won" by doing ~1/4 the work.

Root cause (from Codex): both judges use the same model, so their calls share one concurrency pool and one 60-second per-request timeout (`inference_parameters.timeout` defaults to 60s in NDD's `factory.py`). In `single_stage`, both judges' tasks (`judge1`: 3 scored fields with reasoning, much slower; `judge2`: 1 field, fast) get dispatched concurrently into the same actor's queue since they're independent columns with no dependency between them. Mixing a heavy and light workload in one bounded-concurrency queue inflates tail latency - enough of `judge1`'s calls blew past 60s that the run spiraled into timeouts and never recovered (confirmed: "degraded performance" warnings hit 90–100% and then the run went silent for its last ~6 minutes). `multi_stage` splits the two judges into separate actor pools, so each queue is homogeneous and never hit this.

So actually, I am thinking I will just support `multi_stage`.

nemo_curator/eval/llm_judge/workflow.py
```diff
@@ -0,0 +1,449 @@
+# Copyright (c) 2026, NVIDIA CORPORATION.  All rights reserved.
```
praateekmahajan, line 1: Nit, looks like we are adding an `eval.llm_judge` module to Curator. Historically we've never had `run*` scripts insde our module (an equivalent analogy in my mind would be `run_fuzzy_dedup`, and such a thing doesn't exist. It typically exists in tutorials. Is there a reason we have this here?

praateekmahajan: Another question, if there are no new "stages" being added, or Curator functionality but rather "how we use Curator" shouldn't this be a tutorial / recipe rather than a new module?

nemo_curator/eval/llm_judge/LLM_JUDGE_CONFIG_SKILL.md
```diff
@@ -0,0 +1,215 @@
...
+
+Writes configs for `LLMJudgeWorkflow` (see `llm_judge_workflow.py` in this
+directory, and `tutorials/eval/llm_judge/README.md` for the full runner
+documentation). Every judge config is three things: one or
+more Jinja prompt files, a YAML file wiring models to prompts to rubrics, and
+optional filters on the results. Prefer adapting an existing working YAML/Jinja
+pair in this repo as a starting point rather than writing from scratch, then
+change only what the task requires.
+
+Before writing anything, pin down:
```
VibhuJawa, line 16: Thanks for updating, these are much cleaner now

nemo_curator/eval/llm_judge/workflow.py
```diff
@@ -0,0 +1,504 @@
...
+
+def _read_template(path: str, *, config_path: Path) -> str:
+    template_path = Path(path)
+    if not template_path.is_absolute():
+        template_path = config_path.parent / template_path
+    return template_path.read_text(encoding="utf-8")
+
+
+def _place_filters(config: dict[str, object], stages: list[dict[str, object]]) -> list[list[dict[str, object]]]:
+    """Place top-level filters after the NDD stage that produces their judge column."""
```
VibhuJawa: Sorry, I missed this . We can keep it for now

nemo_curator/eval/llm_judge/workflow.py
```diff
@@ -0,0 +1,504 @@
...
+    return ScoreFilter(
+        filter_obj=FastTextLangId(
+            model_path=model_path,
+            min_langid_score=min_score,
+            lang=language,
+        ),
+        text_field=text_field,
+        verbose=True,
+    ).with_(name="fasttext_language_filter")
+
```
VibhuJawa: Yeah, we can have it for now as an optional nice to have feature. We can keep in the PR

tutorials/eval/llm_judge/run_llm_judge.py
```diff
@@ -0,0 +1,138 @@
...
+import sys
+from pathlib import Path
+
+# `eval` is a repo-root package, not installed with nemo-curator, so it is only
+# importable when the repo root is on sys.path. Running this script directly
+# (`python run_llm_judge.py` or `python path/to/run_llm_judge.py`) does not put
+# the repo root on sys.path — only the script's own directory is added
+# automatically — so add it explicitly before importing from `eval`.
+sys.path.insert(0, str(Path(__file__).resolve().parents[3]))
+
```
VibhuJawa: Oof, i thought this had already happened and is importable .  We can move out the `from eval.llm_judge import LLMJudgeWorkflow` to another workspace, its fine. 

We can move the video eval to tutorial (in a follow up PR) . Should not block your work.  Sorry for the back and forth

nemo_curator/eval/llm_judge/workflow.py
```diff
@@ -0,0 +1,441 @@
...
+    )
+    writer = JsonlWriter(path=output_path) if output_format == "jsonl" else ParquetWriter(path=output_path)
+    processing_stages = []
+    for stage_name, config_builder, model_providers, runtime_env, num_workers, stage_filters in judge_stages:
+        processing_stages.append(
+            DataDesignerStage(config_builder=config_builder, model_providers=model_providers).with_(
+                name=f"ndd_{stage_name}", runtime_env=runtime_env, num_workers=num_workers
+            )
+        )
+        processing_stages.extend(_build_filter_stages(stage_filters, name_prefix=f"judge_filter_{stage_name}"))
```
VibhuJawa, line 322: codex: vibhujawa: A filter can legitimately remove every row from a partition, after which the next `DataDesignerStage` calls `preview(..., num_records=0)`. I reproduced this with Data Designer 0.9.1: a retained one-row batch succeeds, while a fully filtered batch raises `DataDesignerGenerationError`. Could we return the empty batch before calling `preview()`? The existing empty-batch test mocks `preview(0)`, so it does not exercise the real behavior.

eval/llm_judge/llm_judge_workflow.py
```diff
@@ -0,0 +1,441 @@
...
+            return actual > expected
+        if operator == "gte":
+            return actual >= expected
+        if operator == "lt":
+            return actual < expected
+        if operator == "lte":
+            return actual <= expected
+        if operator == "in":
+            return actual in expected
+        return actual not in expected  # noqa: TRY300
```
VibhuJawa: codex: vibhujawa: Could we handle `not_in` explicitly and reject unsupported operators? Currently every unmatched value falls through to `actual not in expected`, so a YAML typo silently changes filtering behavior. I reproduced `operator: gtee` filtering out a row with a scalar value and behaving as `not_in` with a list value. An explicit `operator == "not_in"` branch followed by a `ValueError`, ideally during config validation, would prevent silently producing the wrong dataset.

tutorials/eval/llm_judge/cc_extract_example/prepare_cc_extraction_dataset.py
```diff
@@ -0,0 +1,172 @@
...
+            return None
+
+        language: str | None = None
+        justext_text: str | None = None
+        trafilatura_text: str | None = None
+        try:
+            language = lang_detect(html)
+            stop_words = self.stop_lists.get(language)
+            if stop_words is not None:
+                justext_text = _extract_text(self.justext, html, stop_words, language)
```
VibhuJawa: codex: vibhujawa: Could the JusText and Trafilatura calls have independent exception handling? Currently, if JusText raises, control jumps out of the shared `try` block and Trafilatura is never attempted. I reproduced this by making JusText fail while Trafilatura would succeed: `_extract_text` was called only once and both output fields became `None`. Splitting the calls would preserve whichever extraction succeeds and match the comment that a row is retained when “one extractor fails.”

VibhuJawa (commented):
Mostly looks good to me. One comment around code organization and some issues i found when testing with codex

tutorials/eval/llm_judge/run_llm_judge.py
sarahyurick: Sounds good, I will move it into a new `nemo_curator.eval` folder.

sarahyurick: I created a Linear issue and assigned it to @weijiac0619 - https://linear.app/nvidia/issue/NMCUR-460/move-caption-quality-evaluation-to-nemo-curatoreval-andor .

nemo_curator/eval/llm_judge/workflow.py
praateekmahajan, line 1: nit : Any reason this is called llm_judge_workflow rather than workflow.py? Seems like the import is repetitive? llm_judge.llm_judge_workflow.py

nemo_curator/eval/llm_judge/workflow.py
```diff
@@ -0,0 +1,449 @@
...
+        dd.ModelProvider(
+            name=provider_name,
+            endpoint=endpoint,
+            api_key="unused",  # pragma: allowlist secret
+        )
+    ]
+    return config_builder, model_providers
+
+
+def _start_inference_server(
```
praateekmahajan, line 254: I do think there is need to move this to a common utility, given `ndd_*` benchmarks / pdf benchmark (in future) will also have something like this and then we have  but we can create a future ticket

nemo_curator/eval/llm_judge/workflow.py
```diff
@@ -0,0 +1,449 @@
...
+                    config_builder,
+                    model_providers,
+                    stage.get("runtime_env"),
+                    stage.get("num_workers"),
+                    filters_after_stage,
+                )
+            )
+        return judge_stages
+
+    def run(self) -> WorkflowRunResult:
```
praateekmahajan, line 405: add executor if you want as an optional? Or can keep it.. for the NDD stage do we need to configure similar to how in PDF i had to configure.. i.e. if we have N gpus and N replicas ten the HTTP clients calling that are 4*N and each stage is calling with batch_size=32 which means at a moment the endpoint should have 4*N*32 requests it's predicting

praateekmahajan (approved):

nemo_curator/eval/llm_judge/workflow.py
VibhuJawa: Plus, one , probably add a linear ticket for now or move to a utiluty

nemo_curator/eval/llm_judge/workflow.py
sarahyurick: Not really. I can rename it.

nemo_curator/eval/llm_judge/workflow.py
sarahyurick: Agreed, will be nice to combine in the future.

VibhuJawa (approved):
Approving. Agree with 2 nits  @praateekmahajan  had but  everything else looks good to me

nemo_curator/eval/llm_judge/workflow.py
sarahyurick: I have only ever run with `RayDataExecutor`, IMO adding Xenna as an option isn't super critical. So I will omit for now unless you have any concerns.

Yes I have done the same math. Let me check if that is documented somewhere already, else I can add a quick snippet about it too. Thanks!

sarahyurick: There is some guidance in the tutorial README already, so I will leave it as is for now.

nemo_curator/eval/llm_judge/workflow.py
sarahyurick: Created a Linear ticket for tracking: https://linear.app/nvidia/issue/NMCUR-462/create-start-inference-server-utility

PR #2303, opened 2026-08-11
Bump rapids to 2608

ayushdg:
## Description

This PR bumps rapids to 26.08
1. Adds cuDF streaming and fixes rapidsmpf api changes
2. Fixes cuml KMeans api usage.
3. Fixes cugraph symmetrize error.
4. Removes numpy override and adds a pandas override due to ndd config issues

## Usage

```python
# Add snippet demonstrating usage
```

## Benchmarks

Compared [nightly-2026_08_21__06_28_31_UTC](http://rratzel-ws1:5050/compare?dir_a=login-eos-nightly%3A%2Flustre%2Ffsw%2Fcoreai_dlalgo_ci%2Fcurator_ci%2Fresults%2Fmain&dir_b=login-eos-nightly%3A%2Flustre%2Ffsw%2Fcoreai_dlalgo_ci%2Fcurator_ci%2Fresults%2Fmain-mirror&run_a=nightly-2026_08_21__06_28_31_UTC&run_b=adhoc-adattagupta-6a0f213104e2bc777e7cd0c8f758ccc9aef2c0ed-2026_08_20__15_00_02_UTC-6a0f2131) against the RAPIDS 26.08 adhoc run. Positive changes mean the adhoc run was slower. The table includes benchmarks whose overall runtime changed by more than 5%, plus fuzzy deduplication because its stage-level changes were significant.

| Benchmark | Nightly | Adhoc | Change | Reason |
|---|---:|---:|---:|---|
| `minhash_file_group_task_ray_actors` | 466.1s | 713.7s | +53.1% | Slower `to_parquet`/MinHash write path; adhoc timed out before final metrics were emitted. |
| `minhash_document_batch_xenna` | 573.4s | 821.0s | +43.2% | Slower `to_parquet`/MinHash write path; adhoc timed out before final metrics were emitted. |
| `semdedup_identification_xenna_dataset_size_ratio` | 305.2s | 425.5s | +39.4% | KMeans fit/predict was 61.7% faster, but `to_parquet`/KMeans write was 266.5% slower and dominated. |
| `semdedup_identification_xenna_fit_data_fraction` | 1,454.1s | 1,760.4s | +21.1% | KMeans fit was 52.8% faster, but predict/write (`to_parquet`) was 142.7% slower. |
| `minhash_document_batch_ray_data` | 1,250.6s | 1,400.7s | +12.0% | `to_parquet`/MinHash write was 85.8% slower; the adhoc run also ended with a correctness failure. |
[...]

PR #2325, opened 2026-08-20
perf: use PyArrow fast path for JSONL reads

praateekmahajan:
## What changed

This PR makes direct PyArrow parsing the default for JSONL files and keeps the result as a `pa.Table` until a stage actually needs pandas.

Reader-side `_curator_dedup_id` generation and assignment now work with both Arrow tables and pandas DataFrames. As a result, `LanceReader` now supports `_generate_ids` and `_assign_ids` without converting its Arrow output to pandas first.

When pandas is needed, `DocumentBatch.to_pandas()` handles the conversion and preserves Arrow-backed string columns.

## Choosing an engine

JSONL reads use direct PyArrow parsing by default. Callers can select `engine="pandas"` when they need pandas-specific parsing or type inference.

The two engines do not always infer identical types. For example, pandas may interpret an ISO `created_at` value as a timezone-aware datetime while PyArrow retains it as a string. Mixed-type columns are another situation where the pandas engine may be the better fit.

The PyArrow path continues to support remote files through `fsspec`. It normally reads with an 8 MiB block. If PyArrow reports that a single JSONL record straddles the block boundary, the reader retries with progressively larger blocks so that it can read the complete row. The 256 MiB maximum is a safety ceiling for unusually large individual records, such as rows containing a base64-encoded image or PDF; it is not the default amount read for every row.

## Lance and reader IDs

Because ID handling lives in `BaseReader`, Arrow-backed Lance reads now support `_generate_ids` and `_assign_ids` too.

Lance tasks use a stable identity based on the dataset path, version, and fragment IDs. File readers retain their existing path-based registry keys for backward compatibility; NMCUR-315 tracks moving them to deterministic task IDs as well.

## Compatibility

This work is based directly on `main`, including RAPIDS 26.08 and pandas 3 from #2303. PyArrow remains at 19.0.1 under the current dependency constraints.

## Testing

The tests cover:

- Direct PyArrow and explicit pandas JSONL reads
- ID generation and assignment for Arrow and pandas
- Large JSONL records and remote `fsspec` inputs
- Expected pandas/PyArrow dtype differences
- `DocumentBatch` Arrow-string conversion
- Lance ID generation and assignment
- Related reader, writer, and dedup-removal workflows

## Benchmarks

The table compares the latest completed `main` nightly (`nightly-2026_08_20__06_39_17_UTC`) with this PR (`pr-2325-2026_08_21__00_25_05_UTC-519ad6a9`). Reader time is the mean `jsonl_reader_process_time` per task. Lower is better.

| Benchmark | Mean JSONL reader time, nightly → PR | E2E runtime change |
|---|---:|---:|
| `domain_classification_raydata` | 1.244s → 0.324s (74.0% faster) | 0.8% slower |
| `domain_classification_xenna` | 1.357s → 0.416s (69.4% faster) | 7.5% faster |
| `embedding_generation_raydata` | 0.183s → 0.048s (73.7% faster) | 0.7% faster |
| `embedding_generation_xenna` | 0.212s → 0.054s (74.4% faster) | 0.2% faster |
| `dedup_removal_raydata` | 49.02s → 9.05s (81.5% faster) | 46.9% faster |
| `dedup_removal_xenna` | 33.62s → 9.83s (70.8% faster) | 68.0% faster |
| `score_filter_raydata` | 0.333s → 0.079s (76.3% faster) | 3.8% faster |
| `score_filter_xenna` | 0.303s → 0.069s (77.3% faster) | 2.0% slower |
| `fasttext_filter_raydata` | 0.308s → 0.073s (76.4% faster) | 0.8% faster |
| `fasttext_filter_xenna` | 0.372s → 0.084s (77.3% faster) | 5.8% faster |
| `modifier_raydata` | 0.358s → 0.085s (76.1% faster) | 3.2% faster |
| `modifier_xenna` | 0.341s → 0.082s (75.9% faster) | 4.7% faster |
| `ndd_dynamo_dp8` | 0.175s → 0.046s (73.8% faster) | 7.0% slower |
| `ndd_ray_serve_dp8` | 0.168s → 0.046s (72.8% faster) | <0.1% faster |

Across all 14 entries, the median JSONL reader task is 75.2% faster and median E2E runtime improves by 2.0%. Looking only at the 12 entries where JSONL reading is not the dominant cost, median E2E runtime improves by 0.8%.

### DIFF
--- a/fern/versions/main/pages/curate-text/load-data/read-existing.mdx
+++ b/fern/versions/main/pages/curate-text/load-data/read-existing.mdx
@@ -130,6 +130,9 @@ Both `JsonlReader` and `ParquetReader` support these configuration options:
 | `"pandas"` | `pandas.DataFrame` | You need pandas-specific options or inference, such as `convert_dates` or mixed-type object columns |
 | Another pandas engine, such as `"ujson"` or `"pyarrow"` | `pandas.DataFrame` | You explicitly need an engine supported by `pandas.read_json` |
 
+Use `engine="pandas"` or `engine="ujson"` with `lines=False`. Both `engine="pyarrow_direct"`
+and `engine="pyarrow"` require `lines=True`.
+
 The direct engine does not silently fall back to pandas. If PyArrow cannot infer a consistent
 schema—for example, when one JSON field contains both numbers and strings—select the pandas
 engine explicitly:
--- a/fern/versions/main/pages/curate-text/load-data/read-existing.mdx
+++ b/fern/versions/main/pages/curate-text/load-data/read-existing.mdx
@@ -119,6 +119,51 @@ Both `JsonlReader` and `ParquetReader` support these configuration options:
 | `fields` | list[str] \| None | Column names to read (column selection) | None (all columns) |
 | `read_kwargs` | dict[str, Any] \| None | Extra arguments for the underlying reader | None |
 
+### JSONL Engine Selection
+
+`JsonlReader` parses JSONL directly into a PyArrow table by default. Choose the engine through
+`read_kwargs` when your dataset requires different inference behavior.
+
+| `read_kwargs["engine"]` | Output backing type | Use when |
+| --- | --- | --- |
+| Omitted or `"pyarrow_direct"` | `pyarrow.Table` | Recommended for throughput and Arrow-native processing |
+| `"pandas"` | `pandas.DataFrame` | You need pandas-specific options or inference, such as `convert_dates` or mixed-type object columns |
+| Another pandas engine, such as `"ujson"` or `"pyarrow"` | `pandas.DataFrame` | You explicitly need an engine supported by `pandas.read_json` |
+
+The direct engine does not silently fall back to pandas. If PyArrow cannot infer a consistent
+schema—for example, when one JSON field contains both numbers and strings—select the pandas
+engine explicitly:
+
+```python
+reader = JsonlReader(
+    file_paths="/path/to/data",
+    read_kwargs={
+        "engine": "pandas",
+        "convert_dates": ["created_at"],
+    },
+)
+```
+
+The direct engine accepts these additional options:
+
+| Option | Description | Default |
+| --- | --- | --- |
+| `pyarrow_block_size` | Initial number of bytes PyArrow processes per parser block | 8 MiB |
+| `pyarrow_max_block_size` | Maximum parser block used when retrying an unusually large JSON object | 256 MiB |
+| `storage_options` | Options passed to `fsspec` when opening remote files | `{}` |
+| `compression` | Compression passed to `fsspec`; `"infer"` uses the file extension | `"infer"` |
+| `lines` | JSONL requires one JSON object per line and therefore only supports `True` | `True` |
+
+`pyarrow_block_size` is separate from the top-level `blocksize` parameter. The top-level parameter
+controls how files are grouped into distributed tasks; the PyArrow option controls parsing inside
+each file. When one JSON object is too large for the initial parser window, Curator doubles the
+parser block and retries up to `pyarrow_max_block_size`. The maximum is a safety ceiling for rows
+such as base64-encoded images or PDFs, not the amount read for every row. If the row still does not
+fit, the PyArrow error is returned to the caller.
+
+Remote JSONL paths continue to use `fsspec`, with the resulting file-like stream passed directly to
+PyArrow. This supports the same remote paths and `storage_options` used during file partitioning.
+
 ### Parquet-Specific Features
 
 `ParquetReader` provides these optimizations:
--- a/fern/versions/main/pages/curate-text/load-data/read-existing.mdx
+++ b/fern/versions/main/pages/curate-text/load-data/read-existing.mdx
@@ -130,9 +130,6 @@ Both `JsonlReader` and `ParquetReader` support these configuration options:
 | `"pandas"` | `pandas.DataFrame` | You need pandas-specific options or inference, such as `convert_dates` or mixed-type object columns |
 | Another pandas engine, such as `"ujson"` or `"pyarrow"` | `pandas.DataFrame` | You explicitly need an engine supported by `pandas.read_json` |
 
-Use `engine="pandas"` or `engine="ujson"` with `lines=False`. Both `engine="pyarrow_direct"`
-and `engine="pyarrow"` require `lines=True`.
-
 The direct engine does not silently fall back to pandas. If PyArrow cannot infer a consistent
 schema—for example, when one JSON field contains both numbers and strings—select the pandas
 engine explicitly:
--- a/fern/versions/main/pages/curate-text/load-data/read-existing.mdx
+++ b/fern/versions/main/pages/curate-text/load-data/read-existing.mdx
@@ -150,19 +150,9 @@ The direct engine accepts these additional options:
 | --- | --- | --- |
 | `pyarrow_block_size` | Initial number of bytes PyArrow processes per parser block | 8 MiB |
 | `pyarrow_max_block_size` | Maximum parser block used when retrying an unusually large JSON object | 256 MiB |
-| `storage_options` | Options passed to `fsspec` when opening remote files | `{}` |
-| `compression` | Compression passed to `fsspec`; `"infer"` uses the file extension | `"infer"` |
-| `lines` | JSONL requires one JSON object per line and therefore only supports `True` | `True` |
 
-`pyarrow_block_size` is separate from the top-level `blocksize` parameter. The top-level parameter
-controls how files are grouped into distributed tasks; the PyArrow option controls parsing inside
-each file. When one JSON object is too large for the initial parser window, Curator doubles the
-parser block and retries up to `pyarrow_max_block_size`. The maximum is a safety ceiling for rows
-such as base64-encoded images or PDFs, not the amount read for every row. If the row still does not
-fit, the PyArrow error is returned to the caller.
-
-Remote JSONL paths continue to use `fsspec`, with the resulting file-like stream passed directly to
-PyArrow. This supports the same remote paths and `storage_options` used during file partitioning.
+The direct reader processes 8 MiB chunks and retries with larger chunks, up to 256 MiB, for rows
+containing large payloads such as base64-encoded images or PDFs.
 
 ### Parquet-Specific Features
 
--- a/nemo_curator/stages/text/io/reader/base.py
+++ b/nemo_curator/stages/text/io/reader/base.py
@@ -90,19 +90,19 @@ class BaseReader(ProcessingStage[ReaderTask, DocumentBatch]):
         return self._document_batch(task, output)
 
     def _document_batch(self, task: ReaderTask, output: ReaderOutput) -> DocumentBatch:
-        result = output.data
+        batch = DocumentBatch(
+            dataset_name=task.dataset_name,
+            data=output.data,
+            _metadata=output.metadata if output.metadata is not None else task._metadata,
+        )
         if self._generate_ids or self._assign_ids:
             batch_key = self._id_generator_key(task)
             if self._generate_ids:
-                result = self._generate_ids_func(batch_key, result)
+                self._generate_ids_func(batch_key, batch)
             else:
-                result = self._assign_ids_func(batch_key, result)
+                self._assign_ids_func(batch_key, batch)
 
-        return DocumentBatch(
-            dataset_name=task.dataset_name,
-            data=result,
-            _metadata=output.metadata if output.metadata is not None else task._metadata,
-        )
+        return batch
 
     def _validate_result(self, task: ReaderTask, result: ReaderData) -> None:
         if self.allow_empty:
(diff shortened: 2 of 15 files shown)


nemo_curator/stages/text/io/reader/jsonl.py
```diff
@@ -15,16 +15,156 @@
...
+        table = _pyarrow_select_columns(table, fields, file_path)
+        if table is not None:
+            tables.append(table)
+    if not tables:
+        msg = f"No data read from files in task {paths} with direct PyArrow JSONL reader"
+        logger.error(msg)
+        raise ValueError(msg)
+
+    table = pa.concat_tables(tables, promote_options="permissive")
+    return table.to_pandas(types_mapper=_pyarrow_to_pandas_dtype, use_threads=False)
```
praateekmahajan: Why do we need to return as to_pandas? It can just return as arrow, and if anyone who wants it to do to pandas they can do it.. maybe we should add this to_pandas in documentbatch data class where we call pyarrow_to_pandas_dtype

tests/stages/text/io/reader/test_jsonl.py
```diff
@@ -66,6 +69,94 @@ def test_columns_selection(self, file_group_tasks: list[FileGroupTask]) -> None:
...
+
+    def test_auto_reader_falls_back_to_pandas_for_mixed_column_types(self, tmp_path: Path) -> None:
+        """Auto mode should preserve compatibility when PyArrow rejects mixed JSON types."""
+        file_path = tmp_path / "mixed.jsonl"
+        file_path.write_text('{"value":1}\n{"value":"one"}\n', encoding="utf-8")
+        task = FileGroupTask(dataset_name="ds", data=[str(file_path)], _metadata={})
+
+        result = JsonlReaderStage(read_kwargs={"engine": "auto"}).process(task)
+
+        assert result.data["value"].tolist() == [1, "one"]
```
praateekmahajan, line 140: let's add a test for dtype too?

nemo_curator/stages/text/io/reader/jsonl.py
```diff
@@ -54,30 +194,35 @@ def read_data(
...
-            if fields is not None:
-                df = pandas_select_columns(df, fields, file_path)
-            dfs.append(df)
-        # Concatenate all dataframes
-        if not dfs:
-            msg = f"No data read from files in task {paths} with read_kwargs {read_kwargs} in JSONL reader"
-            logger.error(msg)
-            raise ValueError(msg)
-        return pd.concat(dfs, ignore_index=True)
+        engine = read_kwargs.get("engine", AUTO_ENGINE)
```
praateekmahajan: The issue with the fallback is that, Curator runs the same stage in a distributed fashion, i.e. each concurrent replica of this stage might be working on a different subset of data, therefore one might succeed in reading as pa while other as pd.. but when they're written out their schemas might differ.. does that make sense? if that's right then i think we should read only one way and the user decides that

nemo_curator/stages/text/io/reader/base.py
```diff
@@ -125,26 +125,45 @@ def read_task(
...
+        if isinstance(data, pd.DataFrame):
+            data[CURATOR_DEDUP_ID_STR] = ids
+            return data
+        return data.append_column(CURATOR_DEDUP_ID_STR, pa.array(ids, type=pa.int64()))
+
+    def _assign_ids_func(self, filepath: str | list[str], data: ReaderData) -> ReaderData:
         from nemo_curator.stages.deduplication.id_generator import CURATOR_DEDUP_ID_STR
 
-        if CURATOR_DEDUP_ID_STR not in df.columns:
+        columns = data.columns if isinstance(data, pd.DataFrame) else data.column_names
```
ayushdg: Do you prefer keeping these conditions here or make the method accept DocumentBatch instead of the underlying data and use `get_columns`?

nemo_curator/stages/text/io/reader/base.py
```diff
@@ -125,26 +125,45 @@ def read_task(
...
+        if CURATOR_DEDUP_ID_STR not in columns:
             min_id, max_id = ray.get(self.id_generator.get_batch_range.remote(filepath, None))
-            df[CURATOR_DEDUP_ID_STR] = np.arange(min_id, max_id + 1)
+            data = self._append_ids(data, np.arange(min_id, max_id + 1))
```
ayushdg: Nit: Do you want to pass the min and max to _append_ids and have that generate the np array rather than passing a larger array object?

nemo_curator/stages/text/io/reader/jsonl.py
```diff
@@ -15,16 +15,134 @@
...
 from nemo_curator.tasks import DocumentBatch, EmptyTask
+from nemo_curator.utils.client_utils import is_remote_url
 from nemo_curator.utils.file_utils import FILETYPE_TO_DEFAULT_EXTENSIONS, pandas_select_columns
 
 from .base import BaseFileReader
 
+PANDAS_ENGINE = "pandas"
+PYARROW_DIRECT_ENGINE = "pyarrow_direct"
+DEFAULT_PYARROW_BLOCK_SIZE = 8 * 1024 * 1024
+DEFAULT_PYARROW_MAX_BLOCK_SIZE = 256 * 1024 * 1024
```
ayushdg, line 37: Can you explain what this do and how you chose them?

nemo_curator/stages/text/io/reader/jsonl.py
```diff
@@ -15,16 +15,134 @@
...
+                file_path,
+                mode="rb",
+                compression=compression,
+                **storage_options,
+            ) as stream:
+                return paj.read_json(stream, read_options=read_options)
+        except pa.ArrowInvalid as error:
+            if "straddling object" not in str(error) or block_size >= max_block_size:
+                raise
+            block_size = min(block_size * 2, max_block_size)
```
ayushdg, line 91: In what case would we hit this? Could we end up in a loop where we try reading larger and larger block sizes when we run into this error?

nemo_curator/stages/text/io/reader/jsonl.py
```diff
@@ -15,16 +15,134 @@
...
+    max_block_size: int,
+    storage_options: dict[str, Any],
+    compression: str | None,
+) -> pa.Table:
+    while True:
+        try:
+            read_options = paj.ReadOptions(block_size=block_size, use_threads=False)
+            if not is_remote_url(file_path) and not storage_options and compression == "infer":
+                return paj.read_json(file_path, read_options=read_options)
+            with fsspec.open(
```
ayushdg, line 81: Pyarrow parquet supports remote IO, does jsonl as well?

ayushdg (changes requested):
Minor questions. 

1. Could you also add docstrings/update docs around the engine kwarg and the different options that exist

nemo_curator/stages/text/io/reader/base.py
praateekmahajan: Updated `BaseReader` to construct the `DocumentBatch` before applying IDs. The generate/assign helpers now accept the batch and use `batch.get_columns()`; pandas updates the existing data while Arrow replaces `batch.data` with the table returned by `append_column`. This removes the duplicated representation checks.

nemo_curator/stages/text/io/reader/base.py
praateekmahajan: Moved ID range construction into `_append_ids`. It now accepts `start_id` and `count` rather than an array. The old array argument was passed by reference, so this is mainly a narrower API and allocation scope; `count` also gives generation and assignment the same interface without inclusive-max ambiguity.

nemo_curator/stages/text/io/reader/jsonl.py
praateekmahajan: Added explanations in the constants, reader docstrings, and public reader docs. `pyarrow_block_size` is the initial parser block and is separate from the top-level task-partition `blocksize`. We start at 8 MiB rather than PyArrow's 1 MiB default to reduce retries for moderately large records. The 256 MiB value is only the retry ceiling for unusually large individual rows, such as base64 images or PDFs.

nemo_curator/stages/text/io/reader/jsonl.py
praateekmahajan: The retry is bounded: only a `straddling object` error grows the block, every other Arrow error is re-raised immediately, and an error at `pyarrow_max_block_size` is also re-raised. With the defaults there are at most six attempts (8, 16, 32, 64, 128, and 256 MiB). I added a real-file test with a record larger than the configured maximum to cover termination.

nemo_curator/stages/text/io/reader/jsonl.py
praateekmahajan: `pyarrow.json.read_json` accepts a file-like input but does not expose Parquet's `filesystem=` parameter. I retained `fsspec` to open remote URLs and pass the stream to PyArrow, which also preserves Curator's existing URL schemes, `storage_options`, and compression handling. The reader docstrings/public docs now explain this, and the existing `memory://` test covers the direct remote-stream path.

praateekmahajan:
@ayushdg Addressed all six review items in c5b1c52d: BaseReader ID helpers now operate on DocumentBatch, ID ranges are constructed inside _append_ids, parser block/retry/remote behavior is documented and bounded by a new test, and the public JSONL reader docs now cover engine selection and options. Please take another look.

nemo_curator/stages/text/io/reader/base.py
```diff
@@ -125,26 +125,41 @@ def read_task(
         raise NotImplementedError
 
     # ID helpers ----------------------------------------------------------------
-    def _assign_ids_func(self, filepath: str | list[str], df: pd.DataFrame) -> pd.DataFrame:
+    @staticmethod
+    def _id_generator_key(task: ReaderTask) -> str | list[str]:
+        # TODO(NMCUR-315): Use the deterministic task ID for FileGroupTask as well.
```
sarahyurick, line 130: Nit but can we not reference internal Linear issues as TODOs?

nemo_curator/stages/text/io/reader/jsonl.py
ayushdg: I'm not sure I follow.

nemo_curator/stages/text/io/reader/jsonl.py
praateekmahajan: TLDR is, pyarrow json reads one "block" (default of 1mb) of bytes at time. If a singular row is larger than that then it might error out. To my knowledge, pd.read_json follows a similar code path (with engine=pyarrow) but doesn't expose blocksize.
We do 8mb as default (assuming 8mb is enough for 1 row), however if we have pdf/image bytes in our jsonl row then it can explode to more, which is why we go upto 256 mb (to avoid the runtime error)

nemo_curator/stages/text/io/reader/jsonl.py
praateekmahajan: Sadly not https://arrow.apache.org/docs/python/generated/pyarrow.json.read_json.html

nemo_curator/stages/text/io/reader/jsonl.py
praateekmahajan: Imagine a row, which is <8 mb, then we'll be able to read at first try. If not it'll double that, and goto 16mb and try reading again.. It'll stop at 256mb

nemo_curator/stages/text/io/reader/jsonl.py
```diff
@@ -31,6 +31,10 @@
 
 PANDAS_ENGINE = "pandas"
 PYARROW_DIRECT_ENGINE = "pyarrow_direct"
+# These parser block sizes are independent of JsonlReader.blocksize, which controls
+# how files are grouped into tasks. Start above PyArrow's 1 MiB default to avoid
+# retries for moderately large records, but cap retries for unusually large rows
+# such as records containing base64-encoded images or PDFs.
```
praateekmahajan: Simply change the comment to that we read 8mb of chunk at a time to accomodate for most rows, but go upto 256 for larger rows in case we have payloads of binary objects or something...

nemo_curator/stages/text/io/reader/jsonl.py
```diff
@@ -80,6 +92,7 @@ def _read_jsonl_with_pyarrow(
     read_kwargs: dict[str, Any],
     fields: list[str] | None,
 ) -> pa.Table:
+    """Read JSONL paths directly into Arrow without a pandas conversion."""
```
praateekmahajan: Without a pandas conversion is not necessary to document

nemo_curator/stages/text/io/reader/jsonl.py
```diff
@@ -153,7 +166,14 @@ class JsonlReaderStage(BaseFileReader):
...
         fields (list[str], optional): If specified, only read these fields (columns). Defaults to None.
-        read_kwargs (dict[str, Any], optional): Keyword arguments for the reader. Defaults to {}.
+        read_kwargs (dict[str, Any], optional): Reader options. The default
+            ``engine="pyarrow_direct"`` returns a ``pa.Table``. Select
+            ``engine="pandas"`` for a ``pd.DataFrame`` and pandas-specific
+            parsing or type inference. Other engine values are forwarded to
+            ``pd.read_json`` for backward compatibility. The direct engine
+            accepts ``compression``, ``storage_options``,
+            ``pyarrow_block_size``, and ``pyarrow_max_block_size``;
+            ``lines`` must remain ``True``. Defaults to {}.
```
praateekmahajan: Make it succinct, X is passed to pd._read_Json  while Y is passed to pyarrow_direct.

Also isn't engine="pyarrow" also an option

nemo_curator/stages/text/io/reader/jsonl.py
```diff
@@ -175,8 +195,10 @@ def read_data(
         """Read JSONL files into an Arrow table or pandas DataFrame.
 
         The default ``pyarrow_direct`` engine retains Arrow data without a
-        pandas conversion. Use ``engine="pandas"`` when pandas-specific
-        parsing and inference are required.
+        pandas conversion and does not fall back when Arrow rejects an input.
+        Use ``engine="pandas"`` when pandas-specific parsing and inference are
+        required. Other engine values are delegated to ``pd.read_json`` and
+        also return a pandas DataFrame.
```
praateekmahajan: Mentioning pandas conversion is not required.  Same comment as above keep it succicnt

nemo_curator/stages/text/io/reader/jsonl.py
```diff
@@ -193,6 +215,25 @@ class JsonlReader(CompositeStage[EmptyTask, DocumentBatch]):
...
     1. FilePartitioningStage - partitions files into groups
     2. JsonlReaderStage - reads file groups into DocumentBatches
+
+    Args:
+        file_paths: File paths, directories, or glob patterns to read.
+        files_per_partition: Number of files grouped into each reader task.
+            When set, this takes precedence over ``blocksize``.
+        blocksize: Target storage size for each file-group task. This does not
+            control PyArrow's parser blocks; configure those through
+            ``read_kwargs`` instead.
```
praateekmahajan: Don't mention about `This does not control PyArrow's parser blocks; configure those through ``read_kwargs`` instead.`

nemo_curator/stages/text/io/reader/jsonl.py
```diff
@@ -193,6 +215,25 @@ class JsonlReader(CompositeStage[EmptyTask, DocumentBatch]):
...
+        fields: Optional columns to retain.
+        read_kwargs: Options for JSON parsing and remote storage. Direct
+            PyArrow parsing is the default. Set ``engine="pandas"`` for pandas
+            inference or another pandas-supported engine value to delegate to
+            ``pd.read_json``. The direct engine accepts ``compression``,
+            ``storage_options``, ``pyarrow_block_size``, and
+            ``pyarrow_max_block_size``.
```
praateekmahajan: No need to mention this, just saw something like ptyarrow.json.read_json args are passed to read_kwargs for engine="direct"

nemo_curator/stages/text/io/reader/base.py
praateekmahajan: Good catch, updated it.

nemo_curator/stages/text/io/reader/base.py
praateekmahajan: Because now we also create the `DocumentBatch` before, we just pass that and this method does the generation of np.array and assignment

fern/versions/main/pages/curate-text/load-data/read-existing.mdx
```diff
@@ -119,6 +119,51 @@ Both `JsonlReader` and `ParquetReader` support these configuration options:
...
+| `storage_options` | Options passed to `fsspec` when opening remote files | `{}` |
+| `compression` | Compression passed to `fsspec`; `"infer"` uses the file extension | `"infer"` |
+| `lines` | JSONL requires one JSON object per line and therefore only supports `True` | `True` |
+
+`pyarrow_block_size` is separate from the top-level `blocksize` parameter. The top-level parameter
+controls how files are grouped into distributed tasks; the PyArrow option controls parsing inside
+each file. When one JSON object is too large for the initial parser window, Curator doubles the
+parser block and retries up to `pyarrow_max_block_size`. The maximum is a safety ceiling for rows
+such as base64-encoded images or PDFs, not the amount read for every row. If the row still does not
+fit, the PyArrow error is returned to the caller.
```
praateekmahajan: Keep it succinct.

fern/versions/main/pages/curate-text/load-data/read-existing.mdx
```diff
@@ -119,6 +119,51 @@ Both `JsonlReader` and `ParquetReader` support these configuration options:
...
+
+Remote JSONL paths continue to use `fsspec`, with the resulting file-like stream passed directly to
+PyArrow. This supports the same remote paths and `storage_options` used during file partitioning.
```
praateekmahajan: No need to say this

fern/versions/main/pages/curate-text/load-data/read-existing.mdx
```diff
@@ -119,6 +119,51 @@ Both `JsonlReader` and `ParquetReader` support these configuration options:
...
+
+The direct engine accepts these additional options:
+
+| Option | Description | Default |
+| --- | --- | --- |
+| `pyarrow_block_size` | Initial number of bytes PyArrow processes per parser block | 8 MiB |
+| `pyarrow_max_block_size` | Maximum parser block used when retrying an unusually large JSON object | 256 MiB |
+| `storage_options` | Options passed to `fsspec` when opening remote files | `{}` |
+| `compression` | Compression passed to `fsspec`; `"infer"` uses the file extension | `"infer"` |
+| `lines` | JSONL requires one JSON object per line and therefore only supports `True` | `True` |
```
praateekmahajan: These are supported by pandas as well right? So let's not over-document these three.

ayushdg (approved):
Changes lgtm functionally! 

2 high level questions:

1. With parquet I know we've had some issues around large strings (greater than int32 chars in a table) erroring out where pandas would work, is there something along those lines that we introduce now?
2. Will be caught be QA or other tests but this might change our tutorial behavior in some places? maybe not.

PR #2333, opened 2026-08-25
feat(audio-io): make io stages agent-ready

shubhamNvidia:
**Contracts:** `describe()` + `*_key` params on `convert` and `extract_segments`.

**Two new stages**

- **`ManifestGroupExportStage`** — writes **one file per distinct value of a column** (e.g. a transcript per `speaker_id`) as `txt`, `json`, or `csv`. Optionally also writes a **who-spoke-when `timeline.txt`** ordered by start time across all groups. It's a *tee*, not a sink — rows pass through unchanged, so it can sit anywhere after the data exists. Non-serializable values (like a resident waveform tensor) are dropped from the written columns instead of crashing the export.
  *Why:* "give me a transcript per speaker" was a common ask with no stage behind it — it meant post-processing the manifest by hand.

- **`DocumentBatchJsonlWriterStage`** — appends every row of a `DocumentBatch` to one JSONL manifest (local or cloud via fsspec), truncating once on the driver so batches append within a run, and returning the batch unchanged so downstream metadata survives.
  *Why:* `AudioToDocumentStage` converts `AudioTask` → `DocumentBatch`, but **nothing could write a `DocumentBatch`** — every existing writer takes an `AudioTask`. A pipeline that converted to document form had no terminal sink and simply couldn't be finished. This is that missing sink.

**Also:** `serialize_segments` / `keep_keys` on the conversion path, so the segment list can be carried into document form and the exported columns pinned.


nemo_curator/stages/audio/io/convert.py
```diff
@@ -85,3 +150,91 @@ def process_batch(self, tasks: list[AudioTask]) -> list[DocumentBatch]:
...
+        """Ensure the parent exists on each node without truncating."""
+        self._fs, self._path = url_to_fs(self.output_path)
+        parent_dir = "/".join(self._path.split("/")[:-1])
+        if parent_dir:
+            self._fs.makedirs(parent_dir, exist_ok=True)
+
+    def process(self, task: DocumentBatch) -> DocumentBatch:
+        dataframe = task.to_pandas()
+        if not dataframe.empty:
+            with self._fs.open(self._path, "a", encoding="utf-8") as stream:
```
karpnv, line 215: num_workers() returns 1, which is good, but this is only advisory — nothing enforces it at runtime. The _fs.open(path, "a") pattern with no locking means if the constraint is ever violated (or in a distributed fsspec backend), lines can interleave.

Suggestion: Add a comment/docstring clarifying this is a hard requirement, or add a runtime guard in process().

karpnv (approved):
LGTM in general, small suggestion added

nemo_curator/stages/audio/io/convert.py
```diff
@@ -85,3 +150,91 @@ def process_batch(self, tasks: list[AudioTask]) -> list[DocumentBatch]:
...
+        ),
+        description="Write each DocumentBatch row to one JSONL manifest",
+    )
+
+    def __post_init__(self) -> None:
+        if not self.output_path:
+            msg = "output_path is required for DocumentBatchJsonlWriterStage"
+            raise ValueError(msg)
+
+    def setup(self, _worker_metadata: WorkerMetadata | None = None) -> None:
```
mohammadaaftabv, line 265: P0 (blocker): restart/resume truncates output that was already committed.**

`setup()` is per worker, not driver-once, so a replacement writer actor opens
the complete JSONL with `w` and erases earlier successful batches. 

The same resume-safety problem exists in `group_export.py:132-139,184-188` (first
pending row truncates completed rows for its group) and is explicitly admitted
for segment extraction at `extract_segments.py:342-347`, yet all three stages
inherit `is_resumable=True`. Set them `is_resumable=False` now so
`checkpoint_path` is rejected; a durable version needs source-attributable
shards plus atomic merge/finalization, and truncation cannot live in
restartable worker setup. Please add partial-resume and replacement-worker
tests.

nemo_curator/stages/audio/io/convert.py
```diff
@@ -49,18 +60,72 @@ class AudioToDocumentStage(ProcessingStage[AudioTask, DocumentBatch]):
...
+
+    def _sanitize(self, data: dict) -> dict:
         """Remove non-serializable keys and any remaining tensor values."""
         cleaned = {}
+        keys = self.keep_keys if self.keep_keys is not None else data.keys()
         for k, v in data.items():
-            if k in _NON_SERIALIZABLE_KEYS:
+            if k not in keys or k in self.drop_keys:
+                continue
+            if k in _NON_SERIALIZABLE_KEYS or k == self.segments_key:
```
mohammadaaftabv, line 187: P0 (blocker): the claimed serialization boundary still leaks or corrupts
tensors.** 

Recursive sanitization only runs for `segments_key`. A retained
`{"custom": {"embedding": torch.tensor(...)}}` reaches the new writer and
raises `OverflowError: Maximum recursion level reached`. When `segments`
itself is a tensor, line 128 stores `_DROP_VALUE` and the writer silently emits
`"segments": {}`. Recursively sanitize every retained value, never store the
sentinel itself, and reject/name any remaining non-JSON type. Add both cases as
end-to-end writer regressions.

nemo_curator/stages/audio/io/convert.py
```diff
@@ -49,18 +60,72 @@ class AudioToDocumentStage(ProcessingStage[AudioTask, DocumentBatch]):
...
+        self.serialize_segments = serialize_segments
+        self.segments_key = segments_key
+
     def process(self, task: AudioTask) -> DocumentBatch:
         msg = "AudioToDocumentStage only supports process_batch"
         raise NotImplementedError(msg)
 
-    @staticmethod
-    def _sanitize(data: dict) -> dict:
+    def describe(self) -> StageContract:
```
mohammadaaftabv, line 98: **P0 (blocker): projection is invisible to planning and can silently erase
nonempty batches.** 

The contract preserves all upstream keys regardless of
`keep_keys`/`drop_keys`. With `keep_keys=["audio_filepath"]`, planning approves
a downstream `text` consumer and retains `text` in `produced_keys`, while
runtime removes it and raises `KeyError`. With accepted `keep_keys=[]`, two
tasks become a `(2, 0)` frame; pandas calls it empty and the writer at lines
212-221 emits a zero-byte manifest, losing both rows. Model the configured
allow/deny projection in the contract and reject a nonzero-row/zero-column
batch unless `{}` rows are explicitly supported. Add planner-to-runtime and
empty-projection tests.

mohammadaaftabv (changes requested):

nemo_curator/stages/audio/io/convert.py
```diff
@@ -85,3 +150,91 @@ def process_batch(self, tasks: list[AudioTask]) -> list[DocumentBatch]:
                 _stage_perf=perf,
             )
         ]
+
+
+@dataclass
+class DocumentBatchJsonlWriterStage(AgentReady, ProcessingStage[DocumentBatch, DocumentBatch]):
```
mohammadaaftabv, line 222: DocumentBatchJsonlWriterStage is not needed, and the PR’s “nothing could write a DocumentBatch” rationale is incorrect.
- Existing [BaseWriter](https://github.com/NVIDIA-NeMo/Curator/blob/93b48525eed5eca7449a9b9a1776ac4b913f1218/nemo_curator/stages/text/io/writer/base.py#L30-L104) explicitly consumes DocumentBatch.
- Existing [JsonlWriter](https://github.com/NVIDIA-NeMo/Curator/blob/93b48525eed5eca7449a9b9a1776ac4b913f1218/nemo_curator/stages/text/io/writer/jsonl.py#L23-L48) writes every DocumentBatch row as JSONL.
- The linked benchmark already runs [AudioToDocumentStage → JsonlWriter](https://github.com/NVIDIA-NeMo/Curator/blob/93b48525eed5eca7449a9b9a1776ac4b913f1218/benchmarking/scripts/audio_librispeech_benchmark.py#L199-L206), then merges the shards into one ordered manifest.

mohammadaaftabv (changes requested):

nemo_curator/stages/audio/io/convert.py
shubhamNvidia: You’re right—the original rationale was incorrect because JsonlWriter already accepts DocumentBatch. This stage is retained for different semantics: it is an agent-discoverable sink that writes one configured manifest and preserves DocumentBatch, whereas JsonlWriter creates shards, returns FileGroupTask, and requires a separate merge step. This distinction is now documented.

nemo_curator/stages/audio/io/convert.py
shubhamNvidia: Addressed in b3da05fd. All three stateful IO stages now set is_resumable=False, so checkpointed execution is rejected before any output is touched. Added regression coverage for all three.

nemo_curator/stages/audio/io/convert.py
shubhamNvidia: Addressed in d107c28b. Every retained value is recursively sanitized; tensors and other non-JSON values are dropped with their location reported. Added end-to-end writer regressions.

nemo_curator/stages/audio/io/convert.py
shubhamNvidia: Addressed in d107c28b. The contract now reflects keep_keys/drop_keys, and a nonempty batch producing zero columns raises an error instead of silently losing rows. Added planner/runtime regressions.

nemo_curator/stages/audio/io/convert.py
shubhamNvidia: Addressed in d107c28b. The documentation now explicitly states that num_workers() == 1 is a correctness requirement. A reliable multi-writer implementation would require per-worker shards and atomic merging, not an instance-local guard.

nemo_curator/stages/audio/io/convert.py
shubhamNvidia: Since checkpointing is now rejected before execution, those replacement/resume paths are intentionally unreachable; durable support would require sharded outputs and atomic merging.

nemo_curator/stages/audio/io/extract_segments.py
```diff
@@ -296,9 +304,15 @@ class SegmentExtractionStage(ProcessingStage[AudioTask, AudioTask]):
     """
 
     name: str = "SegmentExtraction"
+    BATCH_ONLY = True  # process() raises; only process_batch is implemented (agent-discovery hint)
     output_dir: str = ""
     output_format: str = DEFAULT_OUTPUT_FORMAT
+    output_key: str = "extracted_path"
```
mohammadaaftabv: P1 — Keep the new field out of the legacy positional layout.

output_key is inserted before the existing batch_size and resources
dataclass fields. 
A previously valid
SegmentExtractionStage("legacy", "/tmp/out", "flac", 17, Resources(cpus=2))
bound batch_size=17 and resources.cpus=2; at this head it silently binds
output_key=17, puts the Resources object in batch_size, and restores the
default resources. 

That can write under an integer task-data key and hand an
object to executor batching instead of the intended size. 

Make output_key
keyword-only (field(default="extracted_path", kw_only=True)) or append it
after every legacy field, validate that it is a nonempty string, and add a
regression asserting all five old positional arguments (and old
__match_args__ positions) retain their bindings.

nemo_curator/stages/audio/io/convert.py
```diff
@@ -49,32 +61,151 @@ class AudioToDocumentStage(ProcessingStage[AudioTask, DocumentBatch]):
...
 
+    def __init__(
+        self,
+        batch_size: int = 64,
+        keep_keys: list[str] | None = None,
+        drop_keys: tuple[str, ...] = (),
+        serialize_segments: bool = False,
+        segments_key: str = "segments",
+    ) -> None:
+        self.batch_size = batch_size
```
mohammadaaftabv: P1 — Preserve legacy serialized instances and subclass defaults.

Before this change AudioToDocumentStage had no initializer: a default old
instance has empty state, subclasses could define an initializer without
calling super(), and a subclass could supply batch_size as a class
default. 
Runtime now requires four initializer-only attributes. Loading an
old-shaped instance at this head makes process_batch() raise
AttributeError: ... keep_keys; a legacy no-super() subclass fails
identically. 
Even class Legacy(AudioToDocumentStage): batch_size = 7
changes from 7 on main to 64 here because the inherited initializer
overwrites it.

nemo_curator/stages/audio/io/convert.py
```diff
@@ -49,32 +61,151 @@ class AudioToDocumentStage(ProcessingStage[AudioTask, DocumentBatch]):
...
+            finally:
+                active.remove(identity)
+
+        if type(value).__module__.startswith("numpy") and callable(item := getattr(value, "item", None)):
+            try:
+                return self._sanitize_nested(item(), path=path, active=active)
+            except ValueError:
+                pass
+
+        try:
```
mohammadaaftabv: P0 — Do not silently discard values supported by the existing conversion
and writer path.

The unchanged AudioToDocumentStage() path now uses stdlib json.dumps as
an admissibility check for every value. 
That is narrower than both the old
DataFrame conversion and the existing pandas-backed JsonlWriter. 
Main
preserves datetime, Decimal, Path, and tuple values; this head silently
drops the first three and changes tuples to lists before returning the
DocumentBatch. 

An end-to-end old pipeline also writes datetime and Decimal
successfully. This is default data/type loss and contradicts line 109's
statement that values are untouched. 

Keep recursive tensor/cycle removal,
but preserve the legacy default for values accepted by the actual downstream
serializer (or make strict normalization explicitly opt-in).

nemo_curator/stages/audio/io/extract_segments.py
```diff
@@ -453,6 +486,16 @@ def _extract_file_segments(
                     try:
                         audio = _read_segment(original_file, start_ms, end_ms, info.samplerate)
                         sf.write(output_path, audio, info.samplerate, subtype=SOUNDFILE_FORMATS[self.output_format])
+                        # collect ALL written paths — a scalar was last-write-wins
+                        # for multi-interval entries (only the final segment's
+                        # path survived, plus a stale path on later failures);
+                        # a pre-existing scalar (re-run on an augmented manifest)
+                        # is folded into the list instead of crashing .append
+                        written = entry.get(self.output_key)
```
mohammadaaftabv: P1 — Keep path bookkeeping from mutating legacy extractor inputs.

Path recording was added inside the shared engine, so it changes both
process_batch() and the public functions under the “Backward-compatible
free functions” heading. 
With a real WAV, main's
extract_segments_by_timestamps(entries, ...) leaves entries unchanged;
this head appends extracted_path to the caller-owned dict. If an input
already has extracted_path="legacy-token", main preserves that scalar and
emits it in metadata; this head converts it to a list containing the old
value plus the generated path, while lines 506-516 remove the column from
metadata/CSV. 

Make bookkeeping explicitly opt-in/pipeline-specific (and keep
legacy wrappers on the non-mutating path), or operate on copies while
preserving pre-existing fields. Add regressions asserting legacy
free-function inputs are unchanged and that a pre-existing same-name field
retains its type and metadata column.

mohammadaaftabv (changes requested):

nemo_curator/stages/audio/io/convert.py
mohammadaaftabv: The non-resumable guard prevents corruption, but it also removes a public
capability that existing checkpointed pipelines had. For complete backward
compatibility, please follow this guard with source-attributable/atomic output
design and restore safe resumability; reverting the guard alone is not an
acceptable fix.

nemo_curator/stages/audio/io/group_export.py
```diff
@@ -0,0 +1,268 @@
...
+
+    def teardown(self) -> None:
+        if self.write_timeline:
+            self._flush_timeline()
+
+    def num_workers(self) -> int | None:
+        return 1  # every group file is appended to from one place; keeps ordering sane
+
+    # ------------------------------------------------------------------ contract
+    def inputs(self) -> tuple[list[str], list[str]]:
```
mohammadaaftabv, line 197: P0: The documented missing-group fallback is rejected before it runs

nemo_curator/stages/audio/io/group_export.py:166-175 declares group_by as a
mechanically required input, while lines 193-194 promise to map a missing value
to missing_group. 

The inherited process_batch() validates first, so a valid
row without speaker_id raises ValueError before process() reaches that
fallback. The tests at tests/stages/audio/io/test_group_export.py:34-38 call
process() directly and therefore miss the framework path.

Do not put an optional fallback key in the legacy required-input list; model it
as optional/advisory in the agent contract (extending the foundation if needed),
and add a process_batch/pipeline regression for a missing group value.

mohammadaaftabv (changes requested):

nemo_curator/stages/audio/io/convert.py
```diff
@@ -49,32 +61,151 @@ class AudioToDocumentStage(ProcessingStage[AudioTask, DocumentBatch]):
...
+        self.serialize_segments = serialize_segments
+        self.segments_key = segments_key
+
     def process(self, task: AudioTask) -> DocumentBatch:
         msg = "AudioToDocumentStage only supports process_batch"
         raise NotImplementedError(msg)
 
-    @staticmethod
-    def _sanitize(data: dict) -> dict:
+    def _removed_keys(self) -> set[str]:
```
mohammadaaftabv, line 110: nemo_curator/stages/audio/io/convert.py:85-104 omits a custom
segments_key from removes_keys when serialize_segments=False, while lines
187-192 always drop it. For segments_key="chunks", validation keeps chunks
in produced_keys with no issue, but the returned frame contains only
audio_filepath.

Add the configured key to _removed_keys() whenever it is not serialized and
regress planner/runtime agreement.

nemo_curator/stages/audio/io/group_export.py
```diff
@@ -0,0 +1,268 @@
...
+        return [], [self.group_by]
+
+    def outputs(self) -> tuple[list[str], list[str]]:
+        return [], []
+
+    def describe(self) -> StageContract:
```
mohammadaaftabv, line 203: P1: Configured transcript reads are absent from the export contract

nemo_curator/stages/audio/io/group_export.py:172-175 declares only
group_by, although TXT and timeline rendering consume text_key at lines
241-257. With text_key="transcript" and only speaker_id available, pipeline
validation reports ok=True with no issues and runtime writes a blank line.

Include text_key in required reads when format == "txt" or
write_timeline=True (and make explicit selected-column requirements equally
clear). Add a planner-to-nonblank-runtime regression.

mohammadaaftabv (changes requested):

nemo_curator/stages/audio/io/extract_segments.py
shubhamNvidia: Thanks for catching this. `output_key` is now keyword-only and validated as a non-empty string, preserving the complete legacy positional layout and `__match_args__`. Added regression coverage for the original five-argument construction.

nemo_curator/stages/audio/io/convert.py
shubhamNvidia: Fixed. `AudioToDocumentStage` now safely handles legacy-shaped instances and subclasses that do not call `super()`. Existing subclass-level `batch_size` defaults are also preserved instead of being overwritten.

nemo_curator/stages/audio/io/convert.py
shubhamNvidia: Agreed. The default conversion path now removes tensors and cycles without applying stricter `json.dumps` compatibility rules. Existing values such as `datetime`, `Decimal`, `Path`, and tuples remain unchanged. Strict JSON normalization is available only through the opt-in `strict_json=True` mode.

nemo_curator/stages/audio/io/extract_segments.py
shubhamNvidia: Fixed. Output-path bookkeeping is now enabled only for the pipeline stage. The backward-compatible free functions remain non-mutating, and pre-existing fields such as `extracted_path` retain their original type and metadata value.

nemo_curator/stages/audio/io/convert.py
shubhamNvidia: The stages remain explicitly non-resumable because their shared aggregate outputs cannot currently be reconstructed safely from partial checkpoints. The previous checkpoint path could truncate or duplicate committed output, so restoring it would require source-attributable shards and atomic finalization rather than removing the guard.

nemo_curator/stages/audio/io/group_export.py
shubhamNvidia: Fixed. `group_by` is now modeled as an optional read, allowing framework validation to reach the documented `missing_group` fallback. Added coverage through `process_batch`, not only direct `process()` calls.

nemo_curator/stages/audio/io/convert.py
shubhamNvidia: Fixed. When segment serialization is disabled, the configured `segments_key` is now included in `removes_keys`, keeping planner output consistent with the returned `DocumentBatch`.

nemo_curator/stages/audio/io/group_export.py
shubhamNvidia: Fixed. `text_key` is now declared as a required read for TXT or timeline output, and explicitly selected JSON/CSV columns are also included. Added planner-to-runtime coverage confirming nonblank transcript output.

karpnv (approved):
Thanks!

nemo_curator/stages/audio/io/extract_segments.py
```diff
@@ -352,13 +474,194 @@ def process_batch(self, tasks: list[AudioTask]) -> list[AudioTask]:
...
+    ) -> tuple[str, bool]:
+        existing = self._segment_reservations.get(resume_key)
+        if existing is not None:
+            return existing["filename"], True
+        filename = make_filename()
+        record: dict[str, Any] = {"filename": filename, "original_file": original_file}
+        if entry.get("speaker_id") is not None:
+            record["speaker_id"] = entry["speaker_id"]
+        self._segment_reservations[resume_key] = record
+        self._persist_output_state()
```
shubhamNvidia, line 626: **[P1] Avoid rewriting the complete reservation state for every segment**

Every new reservation calls `_persist_output_state()`, which serializes the entire accumulated mapping, fsyncs it, and replaces the file. Extracting N segments therefore performs N fsyncs and serializes 1 + 2 + … + N records while holding `_output_lock`—quadratic persistence work before the full metadata CSV rewrite at each batch. This will dominate audio IO at normal curation scale. Preserve crash safety with deterministic filenames from the stable resume identity, or an append/fsync WAL that is compacted periodically, and add a multi-batch regression showing serialized work grows linearly.

nemo_curator/stages/audio/io/group_export.py
```diff
@@ -0,0 +1,322 @@
...
+    identity: str,
+    *,
+    canonical: bool,
+    reserved: set[str] | None = None,
+) -> str:
+    """Return a name determined only by a group's value and typed identity."""
+    raw = str(value)
+    safe = _safe_name(value)
+    if canonical and raw == safe and safe not in (reserved or set()):
+        return safe
```
shubhamNvidia, line 64: **[P1] Bound generated group filenames by bytes**

Canonical strings return the entire sanitized value, and the hashed branch also retains its full prefix. A valid group such as a 260-character source ID/URL therefore reaches `open()` with a component above the common 255-byte `NAME_MAX`, raises `ENAMETOOLONG`, and aborts this `error_policy="fail"` stage. Truncate by encoded bytes while reserving room for the extension and an identity digest so truncation cannot collide; cover long ASCII and multibyte labels and verify order-independent stable names.

shubhamNvidia (commented):
reviwed

shubhamNvidia (commented):
reviewed

PR #1608, opened 2026-03-14
refactor(audio): AudioBatch -> AudioTask single-dict redesign

mohammadaaftabv:
## Summary

Redesigns the audio task model from `AudioBatch` (list-of-dicts) to `AudioTask` (single-dict) and eliminates the `AudioTaskStage` intermediate class — all audio stages now subclass `ProcessingStage[AudioTask, AudioTask]` directly, matching the pattern used by video, text, and image modalities.

### Task model: `AudioBatch` → `AudioTask`

- **Before**: `AudioBatch(Task[list[dict]])` — a batch of manifest entries crammed into one task.  Stages had to loop internally, `num_items` was ambiguous, and `validate_input` couldn't use `hasattr` because `data` was a list.
- **After**: `AudioTask(Task[dict])` — one manifest entry per task.  An `_AttrDict` subclass exposes dict keys as attributes so `base.py`'s `validate_input` works identically to all other modalities.  `num_items` always returns `1`.
- Naming follows the repository convention: `VideoTask`, `FileGroupTask`, `_EmptyTask`, etc.

### Stage hierarchy: direct `ProcessingStage` subclassing

No audio-specific intermediate base class. Each stage subclasses `ProcessingStage[AudioTask, AudioTask]` directly:

| Stage type | Override | Returns |
|---|---|---|
| CPU stages (`GetAudioDurationStage`, `GetPairwiseWerStage`, ALM stages) | `process(task) → AudioTask` | Mutates `task.data` in-place, returns `task` |
| Filtering stages (`PreserveByValueStage`) | `process_batch` only; `process()` raises `NotImplementedError` | `process_batch` filters out entries that don't pass |
| GPU stages (`InferenceAsrNemoStage`) | `process_batch(tasks) → list[AudioTask]`; `process()` raises `NotImplementedError` | Batched GPU inference with inline validation |
| IO stages (`AudioToDocumentStage`) | `process_batch(tasks) → list[DocumentBatch]`; `process()` raises `NotImplementedError` | Aggregates N tasks into one DataFrame |

**Key design decisions:**
- **In-place mutation** — CPU stages mutate `task.data` and return the same `AudioTask` object. ALM stages that replace the entire dict use `task.data.clear()` + `task.data.update(result)` to preserve the `_AttrDict` type.
[...]

PR #2334, opened 2026-08-25
fix(audio-postprocessing): accept diarizer segment shapes; make TimestampMapperStage agent-ready

shubhamNvidia:
**Contracts:** `describe()` + `*_key` params on `TimestampMapperStage` (maps split-clip timestamps back to the original file).

**Bug fix — diarize → timestamp-map pipelines crashed.**
Two producers emit two different segment shapes: VAD and speaker-separation emit `[start, end]` pairs, while the diarizers emit `{start, end, speaker}` dicts. The stage only read pairs, so a real diarizer's output raised **`KeyError: 0`** and killed the pipeline. It now accepts both shapes, and **echoes back the shape it was given** — rewriting a diarizer's dict as a bare pair would have thrown away the speaker label, which is the whole point of running diarization. A single malformed segment is now skipped instead of taking down the batch.

### DIFF
--- a/nemo_curator/stages/audio/postprocessing/timestamp_mapper.py
+++ b/nemo_curator/stages/audio/postprocessing/timestamp_mapper.py
@@ -35,11 +35,13 @@ Output control uses two layers:
   always blocked, even if accidentally added to ``passthrough_keys``.
 """
 
+from collections.abc import Mapping
 from dataclasses import dataclass, field
 from typing import Any
 
 from loguru import logger
 
+from nemo_curator.stages.audio._agent._agent_ready import AgentReady, ConditionalWrite, Gates, IOSpec, StageContract
 from nemo_curator.stages.base import ProcessingStage
 from nemo_curator.stages.resources import Resources
 from nemo_curator.tasks import AudioTask
@@ -53,6 +55,46 @@ _NEVER_PASS_KEYS = frozenset(
         "segments",
     }
 )
+
+
+def _segment_bounds(seg: Any) -> tuple[float, float] | None:  # noqa: ANN401 - shape is the point
+    """``(start_sec, end_sec)`` from either segment shape, or ``None`` if unreadable.
+
+    Two producers, two shapes. VAD and SpeakerSep emit ``[start, end]`` pairs; the diarizers
+    emit ``{start, end, speaker}`` dicts (``InferenceSortformerStage.diarize`` is typed
+    ``list[list[dict[str, Any]]]``). Reading only pairs raised ``KeyError: 0`` on real
+    diarizer output, so a diarize->map pipeline died where it should merely have worked.
+
+    Unreadable segments return ``None`` rather than raising: one malformed entry should be
+    skipped, not take down a whole batch.
+    """
+    if isinstance(seg, Mapping):
+        start, end = seg.get("start"), seg.get("end")
+    else:
+        try:
+            start, end = seg[0], seg[1]
+        except (TypeError, IndexError, KeyError):
+            return None
+    try:
+        return float(start), float(end)
+    except (TypeError, ValueError):
+        return None
+
+
+def _ordered_segments(diar_segments: Any) -> list[tuple[Any, float, float]]:  # noqa: ANN401 - shape is the point
+    """``(original_segment, start, end)`` for each readable segment, earliest first.
+
+    The original is carried alongside its bounds so the output can echo the shape it was
+    given -- rewriting a diarizer's ``{start, end, speaker}`` as a bare pair would throw away
+    the speaker label, which is the one thing a diarization pipeline is run for.
+    """
+    out = []
+    for seg in diar_segments or []:
+        bounds = _segment_bounds(seg)
+        if bounds is not None:
+            out.append((seg, bounds[0], bounds[1]))
+    return sorted(out, key=lambda t: t[1])
+
 
 _DEFAULT_PASSTHROUGH_KEYS: list[str] = [
     "speaker_id",
@@ -102,7 +144,7 @@ def _translate_to_original(
 
 
 @dataclass
-class TimestampMapperStage(ProcessingStage[AudioTask, AudioTask]):
+class TimestampMapperStage(AgentReady, ProcessingStage[AudioTask, AudioTask]):
     """
     Normalize task data at the pipeline output boundary.
 
@@ -124,6 +166,17 @@ class TimestampMapperStage(ProcessingStage[AudioTask, AudioTask]):
     """
 
     passthrough_keys: list[str] | None = field(default=None)
+    audio_filepath_key: str = "audio_filepath"
+    original_file_key: str = "original_file"
+    original_start_ms_key: str = "original_start_ms"
+    original_end_ms_key: str = "original_end_ms"
+    duration_ms_key: str = "duration_ms"
+    duration_key: str = "duration"
+    start_ms_key: str = "start_ms"
+    end_ms_key: str = "end_ms"
+    diar_segments_key: str = "diar_segments"
+    speaking_duration_key: str = "speaking_duration"
+    mappings_key: str = "segment_mappings"
     name: str = "TimestampMapper"
     batch_size: int = 1
     resources: Resources = field(default_factory=lambda: Resources(cpus=1.0))
@@ -144,38 +197,157 @@ class TimestampMapperStage(ProcessingStage[AudioTask, AudioTask]):
         return [], []
 
     def outputs(self) -> tuple[list[str], list[str]]:
-        return [], ["original_file", "original_start_ms", "original_end_ms", "duration_ms", "duration"]
+        return [], [
+            self.original_file_key,
+            self.original_start_ms_key,
+            self.original_end_ms_key,
+            self.duration_ms_key,
+            self.duration_key,
+        ]
+
+    def describe(self) -> StageContract:
+        always_constructed = {
+            self.original_file_key,
+            self.original_start_ms_key,
+            self.original_end_ms_key,
+            self.duration_ms_key,
+            self.duration_key,
+        }
+        passthrough_keys = list(
+            dict.fromkeys(
+                key
+                for key in (self.passthrough_keys or [])
+                if key not in _NEVER_PASS_KEYS and key not in always_constructed
+            )
+        )
+        conditional_writes = [
+            ConditionalWrite(
+                writes=IOSpec(
+                    data_keys=[
+                        self.original_file_key,
+                        self.original_start_ms_key,
+                        self.original_end_ms_key,
+                        self.duration_ms_key,
+                        self.duration_key,
+                    ]
+                ),
+                condition=(
+                    "the mapper resolves a valid mapping/timing branch or its no-mapping fallback "
+                    "and successfully emits an output row"
+                ),
+            ),
+            ConditionalWrite(
+                writes=IOSpec(data_keys=[self.diar_segments_key]),
+                condition=(
+                    "mappings are absent or empty, no valid start/end branch takes priority, "
+                    "at least one readable diarization segment exists, and the mapper emits an output row"
+                ),
+                value_origin="transforms_upstream_same_key",
+            ),
+            ConditionalWrite(
+                writes=IOSpec(data_keys=[self.speaking_duration_key]),
+                condition=(
+                    "mappings are absent or empty, no valid start/end branch takes priority, "
+                    "at least one readable diarization segment exists, and speaking duration is assigned"
+                ),
+            ),
+        ]
+        if passthrough_keys:
+            conditional_writes.append(
+                ConditionalWrite(
+                    writes=IOSpec(data_keys=passthrough_keys),
+                    condition=(
+                        "the same input key is present, non-null, allowed by passthrough_keys, "
+                        "not safety-blocked, not already constructed as a core output, "
+                        "and the mapper successfully emits an output row"
+                    ),
+                    value_origin="upstream_same_key",
+                )
+            )
+        return StageContract(
+            # original_file is optional-with-fallback in every branch (process()
+            # falls back to audio_filepath / 'unknown'), so it must NOT gate
+            # composition — requiring it false-rejected runnable topologies.
+            reads_one_of=[
+                IOSpec(data_keys=[self.start_ms_key, self.end_ms_key]),
+                IOSpec(data_keys=[self.diar_segments_key]),
+                IOSpec(data_keys=[self.duration_key]),
+            ],
+            writes=IOSpec(
+                data_keys=[
+                    self.original_file_key,
+                    self.original_start_ms_key,
+                    self.original_end_ms_key,
+                    self.duration_ms_key,
+                    self.duration_key,
+                    self.diar_segments_key,
+                    self.speaking_duration_key,
+                ]
+            ),
+            # This stage builds its output from an allowlist (``passthrough_keys``) and hard-
+            # blocks _NEVER_PASS_KEYS, so no waveform or audio blob can leave it -- exactly what
+            # ``sanitizes_output`` means. Leaving it unset made the validator report
+            # ``tensor_into_sink`` against a JSON sink placed after this stage, i.e. it refused a
+            # pipeline that was already safe.
+            gates=Gates(
+                sanitizes_output=True,
+                # The concat->original mappings are read from THIS task's ``_metadata``, so every
+                # position it resolves comes from the row it was handed.
+                per_row_independent=True,
+            ),
+            metadata_reads=[self.mappings_key],
+            preserves_upstream_keys=False,
+            conditional_writes=conditional_writes,
+        )
 
     def process(self, task: AudioTask) -> AudioTask | list[AudioTask]:
-        mappings = (task._metadata or {}).get("segment_mappings")
+        mappings = (task._metadata or {}).get(self.mappings_key)
         item = task.data
 
         if mappings:
-            concat_start = item.get("start_ms", 0)
-            concat_end = item.get("end_ms", 0)
-            if concat_end <= concat_start:
-                logger.warning(
-                    f"[TimestampMapper] Skipping task with invalid range: start_ms={concat_start}, end_ms={concat_end}"
-                )
-                return []
-            original_ranges = _translate_to_original(mappings, concat_start, concat_end)
-
-            if len(original_ranges) > 1:
-                logger.debug(
-                    f"[TimestampMapper] Rejecting segment "
-                    f"[{concat_start}-{concat_end}ms] that spans "
-                    f"{len(original_ranges)} concat mappings"
-                )
-                return []
-
-            if len(original_ranges) == 1:
-                result = self._build_output_item(item, original_ranges[0])
+            # start_ms/end_ms is the precise per-segment range, so map it first and fall back
+            # to coarser diar_segments only when absent. Both VAD sources run on a full-length
+            # signal, so these are always concat-time and map directly. Do NOT treat the
+            # presence of diar_segments as a signal to offset instead: that collapsed every
+            # per-speaker VAD sub-segment onto the diar union.
+            start_ms = item.get(self.start_ms_key)
+            end_ms = item.get(self.end_ms_key)
+            diar_segments = item.get(self.diar_segments_key)
+            if start_ms is not None and end_ms is not None:
+                if end_ms <= start_ms:
+                    logger.warning(
+                        f"[TimestampMapper] Skipping task with invalid range: start_ms={start_ms}, end_ms={end_ms}"
+                    )
+                    return []
+                original_ranges = _translate_to_original(mappings, start_ms, end_ms)
+                if len(original_ranges) > 1:
+                    logger.warning(
+                        f"[TimestampMapper] Rejecting segment "
+                        f"[{start_ms}-{end_ms}ms] that spans "
+                        f"{len(original_ranges)} concat mappings"
+                    )
+                    return []
+                if len(original_ranges) == 1:
+                    result = self._build_output_item(item, original_ranges[0])
+                else:
+                    logger.warning(
+                        f"[TimestampMapper] No overlapping mappings for task {task.task_id} "
+                        f"[{start_ms}-{end_ms}ms], dropping"
+                    )
+                    return []
             else:
-                logger.warning(
-                    f"[TimestampMapper] No overlapping mappings for task {task.task_id} "
-                    f"[{concat_start}-{concat_end}ms], dropping"
-                )
-                return []
+                if not diar_segments:
+                    logger.warning(
+                        f"[TimestampMapper] Task {task.task_id} has mappings but no start_ms/end_ms "
+                        f"or diar_segments to resolve against, dropping"
+                    )
+                    return []
+                result = self._build_output_from_diar_and_mappings(item, diar_segments, mappings)
+                if result is None:
+                    logger.warning(
+                        f"[TimestampMapper] No overlapping mappings for diar segments in task {task.task_id}, dropping"
+                    )
+                    return []
         else:
             result = self._build_output_item_no_mapping(item)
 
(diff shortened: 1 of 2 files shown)


nemo_curator/stages/audio/postprocessing/timestamp_mapper.py
```diff
@@ -192,63 +364,96 @@ def _copy_passthrough(self, item: dict[str, Any], result: dict[str, Any]) -> Non
...
+        diar_segments: list,
+        mappings: list[dict[str, Any]],
+    ) -> dict[str, Any] | None:
+        """Compose concat-time diar segments through concat->original mappings.
+
+        Each diar segment is ``[start_sec, end_sec]`` in concatenated time.
+        We translate every segment to original-file time and span the union.
+        Returns ``None`` if no segment overlaps any mapping.
+        """
+        ranges: list[dict[str, Any]] = []
```
mohammadaaftabv, line 388: nemo_curator/stages/audio/postprocessing/timestamp_mapper.py:388-400` collects
all translated diarization ranges, throws away the individual intervals, and
emits only `ranges[0]["original_file"]` plus a global minimum start and maximum
end. 

It does not emit `diar_segments` or compute `speaking_duration` under the
default passthrough list, so it also drops every nested speaker label this PR is
intended to preserve.

A valid same-source, `SegmentConcatenationStage`-shaped case demonstrates silent
corruption: mappings for concat `0-1000 ms -> a.wav 10000-11000 ms` and concat
`1500-2500 ms -> a.wav 30000-31000 ms`, with diar windows `[0.2, 0.8]` and
`[1.6, 2.2]`, produce one `a.wav:10200-30700` row of duration 20.5 seconds. The
canonical extractor classifies that row as combo 4 and reads the whole 20.5
seconds, including 19.3 seconds that neither diar window selected. If
`diar_segments` is explicitly passed through, those intervals remain in concat
coordinates while the scalar fields are in original-file coordinates. With
multi-file mappings, the same reduction additionally combines coordinates from
other files under the first file's name.

Preserve each individually translated interval and its speaker metadata, compute
speaking duration from those translated intervals, and group/fan out by original
file when one row cannot represent the result. Otherwise reject a
non-representable multi-mapping result as the precise start/end branch already
does. Add mapping-plus-dictionary, same-file-gap, and multi-file regressions.
This repeats the mapped-diar findings already recorded in the post-#1608 corpus
on superseded PR #2173.

nemo_curator/stages/audio/postprocessing/timestamp_mapper.py
```diff
@@ -144,38 +197,157 @@ def inputs(self) -> tuple[list[str], list[str]]:
...
+                    self.diar_segments_key,
+                    self.speaking_duration_key,
+                ]
+            ),
+            # This stage builds its output from an allowlist (``passthrough_keys``) and hard-
+            # blocks _NEVER_PASS_KEYS, so no waveform or audio blob can leave it -- exactly what
+            # ``sanitizes_output`` means. Leaving it unset made the validator report
+            # ``tensor_into_sink`` against a JSON sink placed after this stage, i.e. it refused a
+            # pipeline that was already safe.
+            gates=Gates(
```
mohammadaaftabv, line 292: `nemo_curator/stages/audio/postprocessing/timestamp_mapper.py:292-297` declares
an unconditional sanitizer, but `_copy_passthrough()` at lines 358-363 rejects
only five literal key names. 

With
`passthrough_keys=["audio_tensor"]` and a tensor under that key,
`validate_pipeline([mapper, ManifestWriterStage(...)],
initial_tensor_keys={"audio_tensor"})` returns `ok=True`, `keys_ok=True`, and no
issues because the sanitizer clears residency. The mapper then preserves the
tensor and the writer's `json.dumps(task.data)` raises `TypeError: Object of type
Tensor is not JSON serializable`.

The agent-ready framework explicitly supports renamed tensor carriers, so a
name-only blocklist cannot prove sanitization. Either recursively reject/drop
non-JSON audio/tensor values (including nested values) before copying them, or
leave `sanitizes_output` false whenever arbitrary passthrough is allowed. Add a
planner-to-real-writer regression using a custom tensor key.

nemo_curator/stages/audio/postprocessing/timestamp_mapper.py
```diff
@@ -192,63 +364,96 @@ def _copy_passthrough(self, item: dict[str, Any], result: dict[str, Any]) -> Non
...
+            first_start = ordered[0][1]
+            # max, not the last segment's end: diarized speech overlaps, so the segment that
+            # starts last does not necessarily finish last.
+            last_end = max(end for _s, _st, end in ordered)
+            result[self.original_start_ms_key] = int(first_start * 1000)
+            result[self.original_end_ms_key] = int(last_end * 1000)
+            result[self.duration_ms_key] = int((last_end - first_start) * 1000)
+            result[self.duration_key] = last_end - first_start
+            result[self.speaking_duration_key] = round(sum(end - start for _s, start, end in ordered), 3)
+            # Echo the shape we were handed, so a diarizer's speaker labels survive.
```
mohammadaaftabv, line 430: nemo_curator/stages/audio/postprocessing/timestamp_mapper.py:430-435` echoes
Sortformer-style dictionaries on the no-mapping path, but those rows carry the
speaker only inside each dictionary. `detect_combo()` therefore sees no
top-level `speaker_id`, chooses timestamp combo 2, and silently extracts one
bounding clip while ignoring the diarization intervals. Calling the diar path
directly is not a workaround: `_intervals_from_diar_segments()` still sorts and
unpacks pairs and raises `KeyError: 0` on the dictionary shape.

Update the extractor/detector to parse dictionary segments and group their
nested speaker labels, or normalize/fan out the mapper output into the existing
per-speaker row schema. Add an end-to-end Sortformer-shaped
mapper-to-extractor regression; unit-testing the mapper alone does not establish
the claimed pipeline compatibility.

nemo_curator/stages/audio/postprocessing/timestamp_mapper.py
```diff
@@ -124,6 +166,17 @@ class TimestampMapperStage(ProcessingStage[AudioTask, AudioTask]):
     """
 
     passthrough_keys: list[str] | None = field(default=None)
+    audio_filepath_key: str = "audio_filepath"
```
mohammadaaftabv, line 169: nemo_curator/stages/audio/postprocessing/timestamp_mapper.py:168-182` inserts
eleven key fields between the existing `passthrough_keys` and
`name`/`batch_size`/`resources` fields. Before this PR,
`TimestampMapperStage(None, "legacy-name", 7, Resources(cpus=2))` validly set
those three existing options. On this head the same call binds `"legacy-name"`
to `audio_filepath_key`, `7` to `original_file_key`, and the `Resources` object
to `original_start_ms_key`; processing a normal duration row then raises
`TypeError: unhashable type: 'Resources'`.

Append the new key knobs after the old fields, or make only the new knobs
keyword-only, and add a compatibility test for the pre-PR positional binding.
This is required by the audio agent-readiness golden rule that new knobs retain
today's behavior by default.

nemo_curator/stages/audio/postprocessing/timestamp_mapper.py
```diff
@@ -144,38 +197,157 @@ def inputs(self) -> tuple[list[str], list[str]]:
...
+                    writes=IOSpec(data_keys=passthrough_keys),
+                    condition=(
+                        "the same input key is present, non-null, allowed by passthrough_keys, "
+                        "not safety-blocked, not already constructed as a core output, "
+                        "and the mapper successfully emits an output row"
+                    ),
+                    value_origin="upstream_same_key",
+                )
+            )
+        return StageContract(
```
mohammadaaftabv, line 267: nemo_curator/stages/audio/postprocessing/timestamp_mapper.py:267-300` omits
`cardinality`, so both configured and static discovery default to `"1:1"`.
However, lines 317-350 intentionally return `[]` for invalid, cross-mapping,
non-overlapping, or unresolvable mapping rows. The audio `AGENTS.md` explicitly
requires `cardinality="filter"` even when dropping is incidental, and the
required harness currently fails with `cardinality '1:1' != expected 'filter'`.

Declare `cardinality="filter"` and add `assert_agent_ready` coverage for both a
kept row and every supported drop shape.

nemo_curator/stages/audio/postprocessing/timestamp_mapper.py
```diff
@@ -144,38 +197,157 @@ def inputs(self) -> tuple[list[str], list[str]]:
...
+            # original_file is optional-with-fallback in every branch (process()
+            # falls back to audio_filepath / 'unknown'), so it must NOT gate
+            # composition — requiring it false-rejected runnable topologies.
+            reads_one_of=[
+                IOSpec(data_keys=[self.start_ms_key, self.end_ms_key]),
+                IOSpec(data_keys=[self.diar_segments_key]),
+                IOSpec(data_keys=[self.duration_key]),
+            ],
+            writes=IOSpec(
```
mohammadaaftabv, line 276: nemo_curator/stages/audio/postprocessing/timestamp_mapper.py:276-285` puts
`diar_segments` and `speaking_duration` in the mechanical `writes` list even
though the duration, start/end, and mapped branches omit them. 

A duration-only
pipeline therefore validates cleanly when followed by
`PreserveByValueStage(input_value_key="diar_segments", ...)`; at runtime the
mapper omits that key and the consumer raises its missing-input error. The
foundation's mandatory `assert_agent_ready` harness independently fails with
`declared write 'diar_segments' missing from task.data`.

Keep only the five always-produced core fields in `writes`, retain the two
existing `conditional_writes` entries for diarization fields, and add branch
fixtures to the required conformance test. The changed tests currently contain
no `assert_agent_ready` call, which is why this false planner guarantee passes
CI.

nemo_curator/stages/audio/postprocessing/timestamp_mapper.py
```diff
@@ -144,38 +197,157 @@ def inputs(self) -> tuple[list[str], list[str]]:
...
+                        "and the mapper successfully emits an output row"
+                    ),
+                    value_origin="upstream_same_key",
+                )
+            )
+        return StageContract(
+            # original_file is optional-with-fallback in every branch (process()
+            # falls back to audio_filepath / 'unknown'), so it must NOT gate
+            # composition — requiring it false-rejected runnable topologies.
+            reads_one_of=[
```
mohammadaaftabv, line 271: `nemo_curator/stages/audio/postprocessing/timestamp_mapper.py:271-275` requires
start/end, diar segments, or duration through `reads_one_of`. The existing
runtime branch at lines 440-456 deliberately accepts an audio-path-only task and
emits a zero-duration row. Consequently, validation reports
`unsatisfied_reads` and `ok=False` for an input that the same configured stage
successfully processes.

Do not model optional timing probes as mandatory `reads_one_of` alternatives;
represent them as optional reads (extending the contract surface if necessary)
while leaving the zero-duration fallback composable. Add an audio-path-only
validator-to-runtime regression so agent-readiness does not change the existing
default behavior.

mohammadaaftabv (changes requested):

nemo_curator/stages/audio/postprocessing/timestamp_mapper.py
```diff
@@ -54,6 +56,46 @@
...
+    if isinstance(seg, Mapping):
+        start, end = seg.get("start"), seg.get("end")
+    else:
+        try:
+            start, end = seg[0], seg[1]
+        except (TypeError, IndexError, KeyError):
+            return None
+    try:
+        return float(start), float(end)
+    except (TypeError, ValueError):
```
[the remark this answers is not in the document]
shubhamNvidia: Thanks for catching this. Segment bounds now reject non-finite values and non-positive ranges before timestamp conversion. Regression tests cover NaN, infinity, reversed, and zero-length segments.

nemo_curator/stages/audio/postprocessing/timestamp_mapper.py
shubhamNvidia: Fixed across the complete mapper-to-extraction path. Dictionary-shaped diarization segments are recognized, grouped by their nested speaker labels, and extracted per speaker without affecting the existing pair-shaped segment behavior.

nemo_curator/stages/audio/postprocessing/timestamp_mapper.py
shubhamNvidia: Fixed. Timing, path, diarization, and passthrough inputs are now declared as optional reads instead of mutually required alternatives, matching the runtime path-only and zero-duration fallback behavior.

nemo_curator/stages/audio/postprocessing/timestamp_mapper.py
shubhamNvidia: Fixed. The newly introduced key configuration fields are keyword-only, preserving the legacy positional order of `passthrough_keys`, `name`, `batch_size`, and `resources`.

nemo_curator/stages/audio/postprocessing/timestamp_mapper.py
shubhamNvidia: Agreed. The mapper now checks the actual JSON serializability of passthrough values rather than relying only on known key names. Custom tensors and nested non-serializable values are removed before reaching manifest writers.

nemo_curator/stages/audio/postprocessing/timestamp_mapper.py
shubhamNvidia: Fixed. Mapped diarization now preserves individual translated intervals, speaker metadata, `diar_segments`, and `speaking_duration`. Results spanning multiple original files are rejected instead of being collapsed into an invalid single-file range.

nemo_curator/stages/audio/postprocessing/timestamp_mapper.py
shubhamNvidia: Fixed. `diar_segments` and `speaking_duration` are no longer declared as unconditional writes. They remain represented through conditional writes for the runtime branches that produce them.

nemo_curator/stages/audio/postprocessing/timestamp_mapper.py
shubhamNvidia: Fixed. `TimestampMapperStage` now declares filter cardinality because invalid or unresolvable mapped rows can be dropped.

nemo_curator/stages/audio/io/extract_segments.py
```diff
@@ -189,7 +232,11 @@ def detect_combo(entries: list) -> int:
         return 2
 
     first = entries[0]
-    has_speaker = "speaker_id" in first
+    has_nested_speaker = any(
```
mohammadaaftabv: P0 — Sortformer extraction remains batch-order dependent

Anchor: nemo_curator/stages/audio/io/extract_segments.py:234-240 and :382-390.

detect_combo() chooses one extraction mode from entries[0], then process_batch() applies it to all 64 possible rows. A valid path-only/no-speech mapper row followed by a two-speaker Sortformer row selects combo 2 and writes a zero-length generic clip plus one 1.5-second envelope. It does not write the two 0.5-second speaker clips. Reversing the same rows selects combo 3. _expand_nested_speaker_entries() also suppresses nested speaker grouping whenever an inherited top-level speaker_id exists.

This is silent persisted-audio selection and speaker-attribution loss, which is a P0 under the skill rubric. Classify/partition each row independently (or normalize every row to canonical extraction jobs), prefer nested labels for mapping-shaped diarization, and add order-invariant mixed/no-speech and outer-plus-nested-speaker regressions.

nemo_curator/stages/audio/postprocessing/timestamp_mapper.py
```diff
@@ -101,8 +156,27 @@ def _translate_to_original(
     return results
 
 
+def _jsonable_passthrough(value: Any) -> Any:  # noqa: ANN401 - arbitrary passthrough value
```
mohammadaaftabv: P0 — The sanitizer fix is too strict for legacy passthroughs and incomplete for core diarization

Anchors: nemo_curator/stages/audio/postprocessing/timestamp_mapper.py:159-175, :315-319, :381-392, and :432-439.

The legacy default copied configured passthrough values unchanged, and the canonical mapper -> AudioToDocumentStage -> pandas writer successfully serialized pandas.Timestamp, datetime, Decimal, pd.NA, tuples, and NumPy values. The new stdlib json.dumps admissibility test silently drops several of those columns and normalizes tuple/NumPy types before returning the AudioTask. A self-referential container also recurses before the guarded json.dumps call.

At the same time, dictionary diarization records are rebuilt with {**segment, ...} outside the guard. A supported mapping record with embedding=torch.ones(2) therefore retains the tensor while gates.sanitizes_output=True; the raw manifest writer still raises.

Preserve exact legacy values/types by default and advertise sanitizes_output=False in that mode. Put strict normalization behind a new keyword-only opt-in (or at the raw sink), and make the strict path cycle-aware and recursive across both passthrough and core diarization records before claiming the sanitizer gate. Cover both the canonical pandas route and a real raw-writer route.

nemo_curator/stages/audio/postprocessing/timestamp_mapper.py
```diff
@@ -144,38 +233,144 @@ def inputs(self) -> tuple[list[str], list[str]]:
...
-                return []
-
-            if len(original_ranges) == 1:
-                result = self._build_output_item(item, original_ranges[0])
+            # start_ms/end_ms is the precise per-segment range, so map it first and fall back
+            # to coarser diar_segments only when absent. Both VAD sources run on a full-length
+            # signal, so these are always concat-time and map directly. Do NOT treat the
+            # presence of diar_segments as a signal to offset instead: that collapsed every
+            # per-speaker VAD sub-segment onto the diar union.
+            start_ms = item.get(self.start_ms_key)
```
mohammadaaftabv, line 415: P0 — Mapped end_ms-only rows are now silently dropped

Anchor: nemo_curator/stages/audio/postprocessing/timestamp_mapper.py:336-367.

Base deliberately used zero as the missing bound default. With end_ms=500 and a mapping from concat 0..2000 to source.wav:5000..7000, base retained source.wav:5000..5500; head returns []. 

Use the new diarization fallback only when neither scalar key is present. If either scalar is supplied, retain the legacy per-key zero defaults and scalar precedence. Add end-only and partial-scalar-plus-diarization regressions.

nemo_curator/stages/audio/io/extract_segments.py
```diff
@@ -189,7 +232,11 @@ def detect_combo(entries: list) -> int:
         return 2
 
     first = entries[0]
```
mohammadaaftabv, line 191: P0 — Explicit diar_segments: null now aborts the extraction batch

Anchor: nemo_curator/stages/audio/io/extract_segments.py:234-240.

first.get("diar_segments", []) is None, not [], when JSON contains a null. The new generator then raises TypeError. Base accepted the row: with speaker_id it selected combo 3 and the existing interval helper skipped it; without a speaker it selected combo 2. Since only the first row is inspected, one null optional field aborts otherwise extractable later rows. Scan only a non-None value and cover both speaker/no-speaker null rows plus a null-first mixed batch.

nemo_curator/stages/audio/postprocessing/timestamp_mapper.py
```diff
@@ -54,6 +66,49 @@
...
+
+    Two producers, two shapes. VAD and SpeakerSep emit ``[start, end]`` pairs; the diarizers
+    emit ``{start, end, speaker}`` dicts (``InferenceSortformerStage.diarize`` is typed
+    ``list[list[dict[str, Any]]]``). Reading only pairs raised ``KeyError: 0`` on real
+    diarizer output, so a diarize->map pipeline died where it should merely have worked.
+
+    Unreadable segments return ``None`` rather than raising: one malformed entry should be
+    skipped, not take down a whole batch.
+    """
+    if isinstance(seg, Mapping):
```
mohammadaaftabv, line 81: P1 — Early float coercion changes legacy pair types and timestamp boundaries

Anchors: nemo_curator/stages/audio/postprocessing/timestamp_mapper.py:81-94 and nemo_curator/stages/audio/io/extract_segments.py:149-164.

Pair-shaped intervals already accepted Python/NumPy numeric objects. Coercing every bound to Python float changes ordinary integer pair outputs to floats. With Decimal("1.001") or np.longdouble("1.001"), base starts at 1001 ms while head truncates to 1000 ms; head can then expose a 1000..1501 scalar span with duration_ms=500. The extractor also sorts after truncating starts to milliseconds, so reversed 1.0009/1.0001 inputs tie at 1000 and keep the wrong order, changing segment index/filename assignment. Use float views only for finite/range validation, retain original legacy pair values for ordering/arithmetic/output, and sort before millisecond conversion. Test integer, Decimal, long-double, and same-millisecond reversed values in both stages.

nemo_curator/stages/audio/postprocessing/timestamp_mapper.py
```diff
@@ -127,6 +201,21 @@ class TimestampMapperStage(ProcessingStage[AudioTask, AudioTask]):
     name: str = "TimestampMapper"
     batch_size: int = 1
     resources: Resources = field(default_factory=lambda: Resources(cpus=1.0))
+    audio_filepath_key: str = field(default="audio_filepath", kw_only=True)
```
mohammadaaftabv, line 239: P1 — New key parameters permit destructive aliases and empty keys

Anchor: nemo_curator/stages/audio/postprocessing/timestamp_mapper.py:204-230.

start_ms_key=end_ms_key="position" constructs, validates with no error, and silently drops the row as zero-length. original_start_ms_key=original_end_ms_key="boundary" declares a duplicate write and retains only the end, and original_file_key=duration_key silently replaces the path with the numeric duration. Empty names are also accepted. 

Add operation/scope-aware non-empty and collision validation while retaining intentional in-place aliases, then cover a collision matrix and a fully renamed successful configuration. This repeats the configurable-key failure class already fixed on sibling PR 2335.

nemo_curator/stages/audio/postprocessing/timestamp_mapper.py
```diff
@@ -144,38 +233,144 @@ def inputs(self) -> tuple[list[str], list[str]]:
...
+            self.duration_key,
+        }
+        passthrough_keys = list(
+            dict.fromkeys(
+                key
+                for key in (self.passthrough_keys or [])
+                if key not in _NEVER_PASS_KEYS and key not in always_constructed
+            )
+        )
+        conditional_writes = [
```
mohammadaaftabv, line 333: P1 — Conditional writes are still credited when their prerequisite key cannot exist

Anchor: nemo_curator/stages/audio/postprocessing/timestamp_mapper.py:259-307.

Every ConditionalWrite has empty requires_keys, which PR 2332 defines as always reachable. Seeded with only audio_filepath and duration, mapper -> default PreserveByValueStage("diar_segments", ...) reports ok=True, keys_ok=True, and only a conditional_read warning; runtime deterministically emits no diar_segments, and the consumer raises ValueError.

Add requires_keys=[self.diar_segments_key] to the two diarization-derived writes. Split the bundled passthrough declaration into one conditional per key with requires_keys=[key], because requirements are conjunctive. Add absent/present diarization and passthrough planner regressions.

mohammadaaftabv (changes requested):

nemo_curator/stages/audio/postprocessing/timestamp_mapper.py
shubhamNvidia: Agreed. Default passthrough now preserves legacy values and types unchanged and advertises `sanitizes_output=False`. Strict recursive JSON sanitization is available only through the new keyword-only `sanitize_output=True` option; it is cycle-aware and applies to the complete output, including nested diarization records. Tests cover legacy passthrough types and a real raw-writer path with nested tensors.

nemo_curator/stages/audio/io/extract_segments.py
shubhamNvidia: Fixed. An explicit `diar_segments=None` is now treated as an empty optional value during detection rather than being iterated. Because rows are also classified independently, a null row can no longer abort otherwise extractable rows in the same batch. Tests cover null values with and without speaker metadata and a null-first mixed batch.

nemo_curator/stages/audio/postprocessing/timestamp_mapper.py
shubhamNvidia: Fixed. The diarization fallback is now used only when neither scalar boundary is present. If either `start_ms` or `end_ms` is supplied, the missing boundary retains the legacy zero default and the scalar branch keeps precedence. An end-only mapped-row regression was added.

nemo_curator/stages/audio/postprocessing/timestamp_mapper.py
shubhamNvidia: Fixed. The diarization-derived conditional writes now require `diar_segments`, and passthrough writes are declared individually with `requires_keys=[key]` instead of as one conjunctive bundle. A duration-only mapper can therefore no longer satisfy a downstream diarization consumer in planning. Contract and planner regressions were added.

nemo_curator/stages/audio/postprocessing/timestamp_mapper.py
shubhamNvidia: Fixed. Float conversion is now used only for finite/range validation; compatible original numeric values are retained for ordering, arithmetic, and output. Diarization intervals are sorted using the original start values before millisecond conversion, preventing same-millisecond truncation from changing segment order. Precision regressions cover preserved numeric boundaries and reversed sub-millisecond starts.

nemo_curator/stages/audio/postprocessing/timestamp_mapper.py
shubhamNvidia: Fixed. The mapper now validates all configurable key names as non-empty strings and rejects destructive collisions across input and output roles. Intentional in-place aliases for file, start, and end fields remain supported. Regression coverage includes collision cases, legacy in-place aliases, and a fully renamed successful configuration.

nemo_curator/stages/audio/io/extract_segments.py
shubhamNvidia: Thanks for catching this. Extraction routing is now determined independently for each row instead of using the first row for the entire batch. Mixed timestamp and diarization batches now behave consistently regardless of row order, and nested segment speaker labels take precedence over an inherited outer `speaker_id`. Regression coverage includes both row orders, mixed/no-speech input, and outer-plus-nested speaker labels.

nemo_curator/stages/audio/postprocessing/timestamp_mapper.py
```diff
@@ -140,45 +266,199 @@ def __post_init__(self):
...
+                if result is None:
+                    logger.warning(
+                        f"[TimestampMapper] No overlapping mappings for diar segments in task {task.task_id}, dropping"
+                    )
+                    return []
         else:
             result = self._build_output_item_no_mapping(item)
 
+        if self.sanitize_output:
+            sanitized = _jsonable_value(result, set())
```
shubhamNvidia, line 459: **[P1] Preserve required core outputs during strict sanitization**

`sanitize_output=True` runs the complete result through `_jsonable_value()`, which silently drops unrecognized values. A normal `AudioTask(data={"audio_filepath": pathlib.Path(...)})` therefore returns without `original_file`, although `contract().writes` declares that key unconditional and the class promises its core fields are always present. This creates a planner/runtime mismatch and can fail downstream long after the source was discarded. Normalize this required path with `os.fspath()`/`str()` before sanitizing (or reject it clearly), verify all unconditional core keys survive, and cover mapped and unmapped `Path` inputs with a contract-conformance test.

shubhamNvidia (commented):
reviewed

shubhamNvidia (commented):
reviewed

PR #2335, opened 2026-08-25
feat(audio-filtering): annotate-or-filter mode + agent-ready quality stages

shubhamNvidia:
**Contracts:** `describe()` + `*_key` params on the band, SIGMOS and UTMOS filters, with honest `cardinality="filter"` so the planner knows these can drop rows.

**New functionality**
- **`action="annotate"` vs `"filter"`** — the same stage can now either *score every row and keep them all*, or *score and drop* the ones that fail. Previously only dropping was possible, so "show me the quality distribution before I pick a threshold" wasn't expressible.
- **Segment-level scoring** (`mode="task" | "segments" | "auto"`) — score a whole file or each segment of a diarized/VAD'd file.
- **In-memory input** (`input_residency`) — score a waveform handed over from the previous stage instead of re-reading the file.

### DIFF
--- a/nemo_curator/stages/audio/filtering/band.py
+++ b/nemo_curator/stages/audio/filtering/band.py
@@ -39,7 +39,12 @@ from huggingface_hub import hf_hub_download
 from loguru import logger
 
 from nemo_curator.backends.base import NodeInfo, WorkerMetadata
-from nemo_curator.stages.audio.common import resolve_waveform_from_item
+from nemo_curator.stages.audio._agent._agent_ready import AgentReady, Gates, StageContract
+from nemo_curator.stages.audio._agent._residency import (
+    resolve_audio,
+    scoped_audio_conditional_writes,
+    scoped_audio_io_specs,
+)
 from nemo_curator.stages.audio.filtering.band_filter_module.predict import BandPredictor
 from nemo_curator.stages.base import ProcessingStage
 from nemo_curator.stages.resources import Resources
@@ -47,21 +52,36 @@ from nemo_curator.tasks import AudioTask
 
 _HF_REPO_ID = "nvidia/nemocurator-speech-bandwidth-filter"
 _HF_MODEL_FILENAME = "band_classifier_model_band_7000_samples.joblib"
+_VALID_MODES = {"task", "segments", "auto"}
+_VALID_ACTIONS = {"filter", "annotate"}
 
 
 @dataclass
-class BandFilterStage(ProcessingStage[AudioTask, AudioTask]):
+class BandFilterStage(AgentReady, ProcessingStage[AudioTask, AudioTask]):
     """
     Band filter stage for bandwidth classification.
 
-    Classifies audio as "full_band" or "narrow_band" and filters
-    based on the specified band_value to pass.
+    Classifies audio as "full_band" or "narrow_band". With action="filter"
+    (default) only items matching band_value pass; with action="annotate" every
+    item is kept and only the prediction is recorded.
 
     Args:
         model_path: Local path to band classifier model (.joblib). If not provided,
             the model is downloaded from HuggingFace (nvidia/nemocurator-speech-bandwidth-filter).
         cache_dir: Directory to cache downloaded models.
         band_value: Which band type to pass ("full_band" or "narrow_band")
+        mode: Where to classify — "task" (top-level audio), "segments" (nested segments list),
+            or "auto" (segments when segments_key is present, else task; default). Setting
+            "task" or "segments" overrides the auto-detection.
+        action: "filter" drops items that fail the band check; "annotate" keeps every item —
+            including items that fail or cannot be classified — and only writes prediction_key.
+        audio_filepath_key: Key in data dict for the input audio file path.
+        waveform_key: Key in data dict for the in-memory waveform tensor.
+        sample_rate_key: Key in data dict for the waveform sample rate.
+        segments_key: Key in data dict holding the nested segments list (segments/auto mode).
+        prediction_key: Key where the band prediction ("full_band"/"narrow_band") is written.
+        input_residency: Which input to use — "waveform" (in-memory only), "file"
+            (audio_filepath only), or "auto" (waveform first, file fallback; default).
 
     Note:
         GPU is used automatically when resources specify gpus > 0.
@@ -78,6 +98,14 @@ class BandFilterStage(ProcessingStage[AudioTask, AudioTask]):
     model_path: str | None = None
     cache_dir: str | None = None
     band_value: Literal["full_band", "narrow_band"] = "full_band"
+    mode: Literal["task", "segments", "auto"] = "auto"
+    action: Literal["filter", "annotate"] = "filter"
+    audio_filepath_key: str = "audio_filepath"
+    waveform_key: str = "waveform"
+    sample_rate_key: str = "sample_rate"
+    segments_key: str = "segments"
+    prediction_key: str = "band_prediction"
+    input_residency: Literal["file", "waveform", "auto"] = "auto"
 
     name: str = "BandFilter"
     batch_size: int = 1
@@ -92,12 +120,52 @@ class BandFilterStage(ProcessingStage[AudioTask, AudioTask]):
         if self.band_value not in self._VALID_BAND_VALUES:
             msg = f"band_value must be one of {self._VALID_BAND_VALUES!r}, got {self.band_value!r}"
             raise ValueError(msg)
+        if self.mode not in _VALID_MODES:
+            msg = f"mode must be one of {_VALID_MODES!r}, got {self.mode!r}"
+            raise ValueError(msg)
+        if self.action not in _VALID_ACTIONS:
+            msg = f"action must be one of {_VALID_ACTIONS!r}, got {self.action!r}"
+            raise ValueError(msg)
 
     def inputs(self) -> tuple[list[str], list[str]]:
         return [], []
 
     def outputs(self) -> tuple[list[str], list[str]]:
-        return [], ["band_prediction"]
+        return [], [self.prediction_key]
+
+    def describe(self) -> StageContract:
+        reads, reads_one_of, writes = scoped_audio_io_specs(
+            self.input_residency,
+            mode=self.mode,
+            audio_filepath_key=self.audio_filepath_key,
+            waveform_key=self.waveform_key,
+            sample_rate_key=self.sample_rate_key,
+            segments_key=self.segments_key,
+            output_keys=[self.prediction_key],
+        )
+        return StageContract(
+            reads=reads,
+            reads_one_of=reads_one_of,
+            writes=writes,
+            conditional_writes=scoped_audio_conditional_writes(
+                self.mode,
+                segments_key=self.segments_key,
+                output_keys=[self.prediction_key],
+                assignment_condition=(
+                    "audio resolves, the predictor returns 'full_band' or 'narrow_band', "
+                    f"'{self.prediction_key}' is assigned"
+                    + (", and the item matches the configured band and is retained" if self.action == "filter" else "")
+                ),
+            ),
+            cardinality="filter" if self.action == "filter" else "1:1",
+            cardinality_options=["filter", "annotate"],
+            gates=Gates(
+                requires_gpu=self.resources.requires_gpu,
+                requires_internet_first_run=self.model_path is None,
+                # Each clip's features are extracted from that clip alone.
+                per_row_independent=True,
+            ),
+        )
 
     def setup_on_node(
         self, _node_info: NodeInfo | None = None, _worker_metadata: WorkerMetadata | None = None
@@ -136,10 +204,7 @@ class BandFilterStage(ProcessingStage[AudioTask, AudioTask]):
         if self._predictor is None:
             try:
                 model_path = self._resolve_model_path()
-                self._predictor = BandPredictor(
-                    model_path=model_path,
-                    feature_cache_size=100,
-                )
+                self._predictor = BandPredictor(model_path=model_path)
                 logger.info("Band predictor loaded successfully")
             except Exception as e:
                 logger.error(f"Failed to initialize Band predictor: {e}")
@@ -147,24 +212,30 @@ class BandFilterStage(ProcessingStage[AudioTask, AudioTask]):
 
     def process(self, task: AudioTask) -> AudioTask | list[AudioTask]:
         """
-        Filter audio based on bandwidth classification.
+        Filter or annotate audio based on bandwidth classification.
 
-        When ``task.data`` contains a ``"segments"`` key (nested mode from VAD),
-        each segment is evaluated individually and only survivors are kept.
+        Segment handling follows ``mode``: with ``mode="auto"`` (default), nested
+        mode is used when ``task.data`` contains the ``segments_key``; ``mode="task"``
+        or ``mode="segments"`` overrides that auto-detection. In nested mode each
+        segment is evaluated individually. With ``action="filter"`` only survivors
+        are kept; with ``action="annotate"`` every item is kept — including items
+        that fail the band check or cannot be classified — and only the prediction
+        annotation is written.
 
         Returns:
-            AudioTask if passes the band filter, [] if filtered out.
+            AudioTask if it passes (or ``action="annotate"``), [] if filtered out.
         """
-        if "segments" in task.data:
+        use_segments = self.mode == "segments" or (self.mode == "auto" and self.segments_key in task.data)
+        if use_segments:
             survivors = []
-            for seg in task.data["segments"]:
+            for seg in task.data.get(self.segments_key, []):
                 temp = AudioTask(data=seg)
                 result = self._process_single(temp)
-                if result is not None:
+                if result is not None or self.action == "annotate":
                     survivors.append(temp.data)
-            task.data["segments"] = survivors
-            return task if survivors else []
-        return self._process_single(task) or []
+            task.data[self.segments_key] = survivors
+            return task if survivors or self.action == "annotate" else []
+        return self._process_single(task) or (task if self.action == "annotate" else [])
 
     def _process_single(self, task: AudioTask) -> AudioTask | None:
         """Run band classification on a single (non-nested) task."""
@@ -172,7 +243,17 @@ class BandFilterStage(ProcessingStage[AudioTask, AudioTask]):
             logger.error("Band predictor not available")
             return None
 
-        audio = resolve_waveform_from_item(task.data, task.task_id)
+        try:
+            audio = resolve_audio(
+                task.data,
+                residency=self.input_residency,  # type: ignore[arg-type]
+                audio_filepath_key=self.audio_filepath_key,
+                waveform_key=self.waveform_key,
+                sample_rate_key=self.sample_rate_key,
+            )
+        except (OSError, RuntimeError) as e:  # corrupt/unreadable audio -> skip the row, don't crash the batch
+            logger.error(f"Failed to load audio for {task.data.get(self.audio_filepath_key)!r}: {e}")
+            return None
         if audio is None:
             return None
         waveform, sample_rate = audio
@@ -180,16 +261,16 @@ class BandFilterStage(ProcessingStage[AudioTask, AudioTask]):
         try:
             pred = self._predictor.predict_audio(waveform, sample_rate)
             if isinstance(pred, str) and not pred.startswith("Error") and pred in ("full_band", "narrow_band"):
-                task.data["band_prediction"] = pred
+                task.data[self.prediction_key] = pred
             else:
                 logger.warning(f"[{task.task_id}] BandFilter: unexpected prediction value: {pred!r}")
         except Exception as e:  # noqa: BLE001
             logger.exception(f"[BandFilter] Prediction error: {e}")
             return None
 
-        actual = task.data.get("band_prediction", "unknown")
+        actual = task.data.get(self.prediction_key, "unknown")
         if actual != self.band_value:
             logger.info(f"[{task.task_id}] BAND FILTER FAILED: prediction '{actual}' != target '{self.band_value}'")
-            return None
+            return task if self.action == "annotate" else None
 
         return task
--- a/nemo_curator/stages/audio/filtering/band_filter_module/predict.py
+++ b/nemo_curator/stages/audio/filtering/band_filter_module/predict.py
@@ -26,18 +26,15 @@ from .features import AudioFeatureExtractor
 class BandPredictor:
     """Class to predict band label (full_band/narrow_band) for audio waveforms."""
 
-    def __init__(self, model_path: str, feature_cache_size: int = 100):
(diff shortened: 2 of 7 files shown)


nemo_curator/stages/audio/filtering/band.py
```diff
@@ -136,60 +204,73 @@ def _initialize_predictor(self) -> None:
...
+        return self._process_single(task) or (task if self.action == "annotate" else [])
 
     def _process_single(self, task: AudioTask) -> AudioTask | None:
         """Run band classification on a single (non-nested) task."""
         if self._predictor is None:
             logger.error("Band predictor not available")
             return None
 
-        audio = resolve_waveform_from_item(task.data, task.task_id)
+        try:
```
mohammadaaftabv, line 246: nemo_curator/stages/audio/filtering/band.py:246-264` delegates `auto` selection to
`resolve_audio()` but writes only the prediction back to the task. With a resident waveform,
no resident sample rate, and a valid `audio_filepath`, the resolver loads the file: the
predictor scored a `(1, 16000)` waveform with sum `4000`, while the emitted task retained the
different `(1, 8000)` resident waveform with sum `0` and still had no `sample_rate`. The score
therefore describes different audio from the audio handed to downstream stages. It also changes
the prior Band default, which inferred the resident waveform's rate from the file header, in
violation of the audio `AGENTS.md` golden rule that a new knob must preserve current behavior.

Preserve the legacy default by using the file only to obtain the missing header sample rate when
a resident waveform is present. If shared resolution intentionally chooses the file instead,
replace the task's waveform and sample rate atomically (or remove the stale resident waveform)
so the prediction and emitted data agree. Add a regression where transformed resident samples
deliberately differ from the referenced file, plus a file-only downstream-residency regression.

nemo_curator/stages/audio/filtering/sigmos.py
```diff
@@ -74,33 +91,39 @@ def _get_audio_numpy_sr(item: dict[str, Any], task_id: str) -> tuple[np.ndarray,
...
-            return audio, int(sr)
-        except Exception as e:  # noqa: BLE001
-            logger.error(f"[{task_id}] Failed to load audio file: {e}")
-            return None
+    # Thin wrapper over the shared resolver (_residency.resolve_audio): delegate
+    # file/waveform resolution, then return a mono float32 1-D numpy array.
+    # Audio-file load errors are swallowed and reported as None (matches the
+    # original); a waveform without sample_rate falls back to the file path,
+    # exactly like resolve_audio.
+    try:
```
mohammadaaftabv, line 99: nemo_curator/stages/audio/filtering/sigmos.py:99-121` now obtains a tensor from the shared
resolver and calls `ensure_mono()` before converting it to float. A valid
`torch.int16` stereo waveform reaches `torch.mean()` and raises
`RuntimeError: mean(): ... Got: Short`; that reduction is also outside the helper's guarded
resolution block, so it escapes and aborts the pipeline. The pre-PR NumPy path accepted the same
input and returned mono `float32`, and the stage documentation still promises tensor/array input.

Convert resident tensors to `float32` before mono reduction (preferably in the shared resolver so
all consumers get one normalization contract), and keep normalization inside the load-error
boundary. Add a stereo integer-PCM test that exercises `process()`, not only the helper.

nemo_curator/stages/audio/filtering/sigmos.py
```diff
@@ -62,9 +69,19 @@
...
-
-
-def _get_audio_numpy_sr(item: dict[str, Any], task_id: str) -> tuple[np.ndarray, int] | None:
+_VALID_MODES = {"task", "segments", "auto"}
+_VALID_ACTIONS = {"filter", "annotate"}
+
+
+def _get_audio_numpy_sr(  # noqa: PLR0913 (complexity accepted: keyword-only residency/key knobs mirror the stage fields)
+    item: dict[str, Any],
+    task_id: str,
```
mohammadaaftabv, line 78: nemo_curator/stages/audio/filtering/utmos.py:78-95` returns `None` immediately whenever a
waveform exists without `sample_rate`, even in `input_residency="auto"`. 

This contradicts both
the new public docstring and contract, which say `auto` tries the resident pair and then the file.
A row with an incomplete resident value and a valid WAV therefore never reaches
`resolve_audio()`: filter mode silently drops it, while annotate mode keeps an unscored row.

Remove the early return and let the shared resolver perform the documented fallback, or retain
it only for explicit `input_residency="waveform"`. Add an auto-residency regression containing
both an incomplete resident value and a valid file path and assert that file inference runs.

nemo_curator/stages/audio/filtering/band.py
```diff
@@ -92,12 +120,52 @@ def __post_init__(self):
...
+            raise ValueError(msg)
 
     def inputs(self) -> tuple[list[str], list[str]]:
         return [], []
 
     def outputs(self) -> tuple[list[str], list[str]]:
-        return [], ["band_prediction"]
+        return [], [self.prediction_key]
+
+    def describe(self) -> StageContract:
```
mohammadaaftabv, line 136: nemo_curator/stages/audio/filtering/band.py:136-159`,
`nemo_curator/stages/audio/filtering/sigmos.py:246-282`, and
`nemo_curator/stages/audio/filtering/utmos.py:183-210` put every score in both
`writes.data_keys` and `writes.segment_data_keys` for `mode="auto"`, although runtime chooses
exactly one scope. 

In the current #2332 integration, the mechanical planner treats both as
guaranteed. It accepted a nested Band source -> default auto Band annotate -> top-level
`PreserveByValueStage("band_prediction", ...)` pipeline with `ok=True`, `keys_ok=True`, and no
issues; runtime wrote `band_prediction` only into the segment and the selector raised on the
missing top-level key. The required `assert_agent_ready` harness fails the same successful
fixture because the declared top-level write is absent. Unscorable annotate rows create the
second side of the same problem: they are retained without any score despite an unconditional
declared write.

This overlaps the still-open conditional-write limitation in dependency PR #2332, so fix it
once in that foundation rather than patching three stages independently: branch planner state by
scope, never advance a conditional key as guaranteed, and make conformance understand mutually
exclusive and optional writes. Alternatively require an explicit scope until the planner can
represent `auto`. Define the annotate failure value or absence contract explicitly, then add
`assert_agent_ready` coverage for task/segments/auto and success/unscorable execution before
this dependent PR merges.

nemo_curator/stages/audio/filtering/band.py
```diff
@@ -78,6 +98,14 @@ class BandFilterStage(ProcessingStage[AudioTask, AudioTask]):
     model_path: str | None = None
     cache_dir: str | None = None
     band_value: Literal["full_band", "narrow_band"] = "full_band"
+    mode: Literal["task", "segments", "auto"] = "auto"
+    action: Literal["filter", "annotate"] = "filter"
+    audio_filepath_key: str = "audio_filepath"
```
mohammadaaftabv, line 103: The new key fields at `nemo_curator/stages/audio/filtering/band.py:103-108`,
`nemo_curator/stages/audio/filtering/sigmos.py:206-216`, and
`nemo_curator/stages/audio/filtering/utmos.py:155-159` have no collision validation. For example,
`BandFilterStage(prediction_key="waveform", action="annotate")` replaces the audio tensor with
the string `full_band` at `band.py:264`. SIGMOS permits two score fields to share one key, so
later assignments silently replace earlier dimensions and its own contract fails the
well-formedness audit; a score key that equals `waveform_key` is popped at `sigmos.py:435-444`
before resolution. UTMOS has the equivalent destructive pop at `utmos.py:309-313`.

Validate all configurable keys as non-empty and require output keys to be distinct from one
another and disjoint from the audio/file/rate/segments input keys. Add constructor and runtime
collision tests for each stage.

nemo_curator/stages/audio/filtering/band.py
```diff
@@ -92,12 +120,52 @@ def __post_init__(self):
...
+                output_keys=[self.prediction_key],
+                assignment_condition=(
+                    "audio resolves, the predictor returns 'full_band' or 'narrow_band', "
+                    f"'{self.prediction_key}' is assigned"
+                    + (", and the item matches the configured band and is retained" if self.action == "filter" else "")
+                ),
+            ),
+            cardinality="filter" if self.action == "filter" else "1:1",
+            cardinality_options=["filter", "annotate"],
+            gates=Gates(
```
mohammadaaftabv, line 162: `nemo_curator/stages/audio/filtering/band.py:162-165` and
`nemo_curator/stages/audio/filtering/sigmos.py:285-287` declare first-run internet only when
`model_path is None`, but their resolvers download when a non-`None` path is missing
(`band.py:194-201`; `sigmos.py:345-349`). `setup_on_node()` skips prefetch for that same missing
explicit path, so an offline worker sees an undeclared download or failure. Static class
discovery compounds this: none of the three classes supplies `AGENT_STATIC`, so it reports
`requires_internet_first_run=False` and unknown row independence even for UTMOS, whose configured
contract correctly declares internet.

Reject a nonexistent explicit path eagerly, or derive both the gate and prefetch condition from
the same existence check. Add conservative static hints for invariant first-run internet and
per-row independence (plus GPU where invariant), and test static, cached, valid-path, and
missing-path discovery.

nemo_curator/stages/audio/filtering/band.py
```diff
@@ -78,6 +98,14 @@ class BandFilterStage(ProcessingStage[AudioTask, AudioTask]):
     model_path: str | None = None
     cache_dir: str | None = None
     band_value: Literal["full_band", "narrow_band"] = "full_band"
+    mode: Literal["task", "segments", "auto"] = "auto"
```
mohammadaaftabv, line 101: The new Band fields at `nemo_curator/stages/audio/filtering/band.py:101-108`, SIGMOS fields at
`nemo_curator/stages/audio/filtering/sigmos.py:203-216`, and UTMOS fields at
`nemo_curator/stages/audio/filtering/utmos.py:152-159` were inserted before the existing
`name`, `batch_size`, and `resources` fields. Previously valid positional calls now bind the old
name/batch/resource arguments to `mode` or residency fields and raise during construction; all
three historical signatures reproduced this failure.

Keep the legacy fields in their original positional order and make only the newly introduced
knobs keyword-only with `dataclasses.KW_ONLY`. Add one signature/constructor regression per
stage using its pre-PR positional call shape.

nemo_curator/stages/audio/filtering/band.py
mohammadaaftabv, line 1: The audio authoring contract at `nemo_curator/stages/audio/AGENTS.md:30-43` requires every
agent-ready conversion to add an `assert_agent_ready(...)` test and a conforming capability card.
None of the three changed test modules calls `assert_agent_ready` or
`assert_residency_consumption`, and neither this PR nor current dependency #2332 contains the
required `nemo_curator/audio_agent/knowledge/cards/` tree. This is not just documentary coverage:
the mandatory harness immediately exposes the auto-scope failure above.

Add configuration-matrix conformance and residency-consumption tests for all three stages and
ship/audit the three capability cards in this PR or in an explicit merge-blocking dependency.

mohammadaaftabv, line 116: The `__post_init__` methods validate `mode` and `action` but not `input_residency`
(`band.py:116-128`, `sigmos.py:222-230`, `utmos.py:165-172`). Type annotations do not validate at
runtime, and the shared residency helper treats every unknown string as `auto`; a typo such as
`wavefrom` therefore advertises and performs file fallback instead of failing closed.

Validate against `{file, waveform, auto}` in one shared helper and add the same invalid-value
test for all three stages.

mohammadaaftabv (changes requested):

nemo_curator/stages/audio/filtering/band.py
shubhamNvidia: Agreed. Band now preserves its legacy behavior: a resident waveform missing only its rate uses the file header without replacing samples. A selected file load hydrates waveform and sample rate together, so the score and downstream residency describe the same audio. Added deliberately different resident/file and downstream-auto regressions.

nemo_curator/stages/audio/filtering/sigmos.py
shubhamNvidia: Agreed. Resident audio now converts supported signed PCM16/PCM32 to normalized float32 before mono reduction, inside the guarded resolution path. Added a process-level stereo PCM16 regression covering scale, dtype, and shape.

nemo_curator/stages/audio/filtering/sigmos.py
shubhamNvidia: Agreed. Removed the premature rejection. Auto mode now falls back to the file when the resident pair is incomplete and atomically replaces both resident fields with the selected file pair. Explicit waveform mode still fails closed. Both orphan-pair directions are covered.

nemo_curator/stages/audio/filtering/band.py
shubhamNvidia: Agreed. Score and file-hydration outputs are now conditional rather than guaranteed planner writes, with task/segment scope represented separately. Added auto task/segment × successful/unscorable runtime and conformance coverage. Possible tensor hydration is tracked conservatively for JSON-sink safety without advancing its keys as guaranteed.

nemo_curator/stages/audio/filtering/band.py
shubhamNvidia: Agreed. Constructors now reject empty keys, duplicate output keys, and output collisions with filepath, waveform, sample-rate, or segment inputs before model setup or inference. Added coverage for all three stages.

nemo_curator/stages/audio/filtering/band.py
shubhamNvidia: Agreed. Missing or empty explicit Band/SIGMOS model paths now fail without silently downloading, while valid explicit paths bypass downloads and default/cache paths follow the expected download or reuse flow. Added conservative static internet and row-independence hints with configured/static regressions.

nemo_curator/stages/audio/filtering/band.py
shubhamNvidia: Agreed. Restored every legacy positional field to its original slot and made only the newly introduced controls keyword-only. Exact pre-PR constructor binding is covered for Band, SIGMOS, and UTMOS.

nemo_curator/stages/audio/filtering/band.py
shubhamNvidia: Added `assert_agent_ready` coverage across task, segment, and auto scopes, including successful and unscorable annotation paths, plus file/waveform residency-consumption tests for all three stages. The audited Band, SIGMOS, and UTMOS cards are maintained in draft PR #2343.

shubhamNvidia: Agreed. All three constructors now use the shared residency validator and reject values outside `file`, `waveform`, and `auto`, including misspellings and incorrectly cased values.

nemo_curator/stages/audio/filtering/band.py
mohammadaaftabv: P1 — auto read validation approves a branch runtime will not use

Current anchors:

nemo_curator/stages/audio/filtering/band.py:156-167
nemo_curator/stages/audio/filtering/band.py:268-277

The conditional-write portion of the existing thread is fixed, but
scoped_audio_io_specs(mode="auto") still puts parent-audio and child-audio
forms into one unconditional reads_one_of. With parent waveform and
sample_rate plus segments=[{"segment_num": 0}], validate_pipeline()
returns ok=True, keys_ok=True, and no issues. Runtime sees the segment key,
chooses the nested branch, invokes the predictor zero times, empties segments,
and drops the entire row. SIGMOS and UTMOS use the same helper and branch rule.

Represent the segments_key present/absent predicate in the planner and check
only the runtime-selected scope. If conditional reads cannot yet be expressed,
planned pipelines should require explicit mode="task" or mode="segments".
Add the mixed parent-audio/unhydrated-child regression. This is a follow-up to
open Thread 4 (discussion_r3987637625).

nemo_curator/stages/audio/filtering/band.py
mohammadaaftabv: P1 — audio input roles may still alias and crash normal rows

Current anchors:

nemo_curator/stages/audio/filtering/band.py:139-148
nemo_curator/stages/audio/filtering/sigmos.py:245-262
nemo_curator/stages/audio/filtering/utmos.py:179-188

The shared validator rejects output collisions, but not collisions among
audio_filepath_key, waveform_key, sample_rate_key, and segments_key.
BandFilterStage(segments_key="audio_filepath", action="annotate") therefore
constructs; an ordinary filepath row is treated as nested data, the path string
is iterated character by character, and AudioTask(data=seg) raises
ValueError. Aliasing waveform and rate keys instead produces duplicate,
unsatisfiable read specs. Require these input-role values to be pairwise
distinct and cover both collision shapes through all three callers. This is a
follow-up to open Thread 3 (discussion_r3987640021).

nemo_curator/stages/audio/filtering/band_filter_module/predict.py
```diff
@@ -26,18 +26,15 @@
 class BandPredictor:
     """Class to predict band label (full_band/narrow_band) for audio waveforms."""
 
-    def __init__(self, model_path: str, feature_cache_size: int = 100):
+    def __init__(self, model_path: str):
```
mohammadaaftabv: P1 / backward compatibility — preserve BandPredictor's constructor and cache contract

Anchor: nemo_curator/stages/audio/filtering/band_filter_module/predict.py:29

Before this PR, BandPredictor(model_path, feature_cache_size=100) was the
documented signature and instances exposed feature_cache_size plus the
bounded feature_cache. The PR removes the second argument and the cache. Both
historical positional and keyword calls now raise TypeError, and repeated
inference loses configured caching behavior.

Retain the optional argument and compatible state. If the old sampled-hash
cache was removed because it could collide or omitted sample rate, replace it
with a safe bounded key that includes sample rate, shape, dtype, and robust
audio identity; alternatively accept and deprecate the argument while
intentionally disabling caching. Add direct predictor constructor/cache tests,
because the current stage tests mock the predictor and cannot catch this break.

nemo_curator/stages/audio/filtering/sigmos.py
```diff
@@ -74,33 +94,44 @@ def _get_audio_numpy_sr(item: dict[str, Any], task_id: str) -> tuple[np.ndarray,
...
+            logger.warning(f"[{task_id}] No {waveform_key}+{sample_rate_key} found")
+        else:
+            logger.warning(f"[{task_id}] No {waveform_key}+{sample_rate_key} or valid {audio_filepath_key} found")
+        return None
 
-    logger.warning(f"[{task_id}] No waveform+sample_rate or valid audio_filepath found")
-    return None
+    try:
+        waveform, sample_rate = resolved
+        mono = normalize_audio_waveform(waveform, stage_name="SIGMOSFilterStage", mono=True)
```
mohammadaaftabv, line 127: P1 / backward compatibility — SIGMOS changes legacy Torch PCM amplitudes

Anchor: nemo_curator/stages/audio/filtering/sigmos.py:125

The integer-stereo crash is gone, but the fix changes a previously supported
default input. The pre-PR SIGMOS path cast a torch.int16 waveform to float32
without scaling; the current shared normalizer divides by 32768. For the
deterministic stereo samples [[32767, -32768], [16384, -16384]], the base
passes [24575.5, -24576.0] to the model while this head passes
[0.7499847, -0.75]. That silently changes MOS values and keep/drop decisions.
Equivalent NumPy integer arrays remain unscaled because the resolver converts
them to float32 before normalization, so results now also depend on whether the
same samples arrived as Torch or NumPy.

Preserve the legacy cast-only SIGMOS conversion under existing defaults and
make PCM amplitude normalization opt-in if desired. Add base-parity tests for
the exact model input and threshold decision for both Torch and NumPy integer
containers. This belongs as a follow-up to the existing PCM Thread 8
(discussion_r3987624523).

mohammadaaftabv (changes requested):

nemo_curator/stages/audio/filtering/band.py
shubhamNvidia: Fixed in the Foundation contract layer and consumed by all three filtering stages. `mode="auto"` now declares task and segment reads as conditional branches selected by `segments_key` presence, so parent audio can no longer satisfy an unhydrated nested branch. Mixed-scope regressions were added for Band, SIGMOS, and UTMOS.

nemo_curator/stages/audio/filtering/band.py
shubhamNvidia: Fixed. The shared validator now requires filepath, waveform, sample-rate, and segments input keys to be pairwise distinct. Both reported collision shapes are covered across all three filtering stages.

nemo_curator/stages/audio/filtering/band_filter_module/predict.py
shubhamNvidia: Agreed on the constructor compatibility issue. Caching was intentionally removed, so it remains disabled. However, `feature_cache_size` is accepted again and the legacy `feature_cache_size` and empty `feature_cache` attributes remain available. A direct regression confirms repeated inference recomputes features rather than using a cache.

nemo_curator/stages/audio/filtering/sigmos.py
shubhamNvidia: Fixed. SIGMOS now preserves its legacy cast-only conversion for integer Torch inputs, matching NumPy behavior and avoiding implicit amplitude normalization. Base-parity tests verify the exact model input for both Torch and NumPy containers.

nemo_curator/stages/audio/filtering/band.py
```diff
@@ -136,60 +245,86 @@ def _initialize_predictor(self) -> None:
...
         Returns:
-            AudioTask if passes the band filter, [] if filtered out.
+            AudioTask if it passes (or ``action="annotate"``), [] if filtered out.
         """
-        if "segments" in task.data:
+        use_segments = self.mode == "segments" or (self.mode == "auto" and self.segments_key in task.data)
+        if use_segments:
             survivors = []
-            for seg in task.data["segments"]:
+            for seg in task.data.get(self.segments_key, []):
```
shubhamNvidia, line 272: **[P1] Handle nullable segment containers before iterating**

The default `mode="auto"` selects segment mode from key presence, but this loop assumes the value is iterable. A normal nullable manifest row with `segments=None` raises `TypeError` here before `action="annotate"` can retain the unscorable row (and before filter mode can drop it); the same unchecked loop is present in SIGMOS and UTMOS. One null row can therefore abort the whole pipeline. Normalize null consistently as empty/absent, handle other non-list containers through the stage's per-item policy, and validate elements are mappings before constructing `AudioTask`. Please cover null containers and malformed entries for all three filters in annotate and filter modes.

shubhamNvidia (commented):
reviewed

shubhamNvidia (commented):
reviewed

mohammadaaftabv (approved):

PR #2336, opened 2026-08-25
feat(audio-metrics): in-memory input + agent-ready contracts

shubhamNvidia:
**Contracts:** `describe()` + `*_key` params on the bandwidth, SQUIM and WER metric stages, including `conditional_writes` (a metric is only written when the audio resolves and estimation succeeds).

**New functionality**
- **In-memory input** (`input_residency`) — read the waveform already in the task rather than re-opening the file for every metric stage. On a chain of several metric stages this removes one file read per stage per row.

Per-segment scoring and `metric_skip_reason` already existed upstream; the contracts here just declare them, so the planner knows a metric is written only when the audio resolves and estimation succeeds.

Missing configured WER text-key handling is intentionally out of scope here and remains in draft #2353. This PR can merge independently; #2353 can rebase onto main afterward.

### DIFF
--- a/nemo_curator/stages/audio/metrics/bandwidth.py
+++ b/nemo_curator/stages/audio/metrics/bandwidth.py
@@ -21,12 +21,15 @@ import librosa
 import numpy as np
 from loguru import logger
 
+from nemo_curator.stages.audio._agent._agent_ready import AgentReady, ConditionalWrite, Gates, IOSpec, StageContract
+from nemo_curator.stages.audio._agent._residency import InputResidency, residency_read_specs
+from nemo_curator.stages.audio.common import ensure_mono, ensure_waveform_2d
 from nemo_curator.stages.base import ProcessingStage
 from nemo_curator.tasks import AudioTask
 
 
 @dataclass
-class BandwidthEstimationStage(ProcessingStage[AudioTask, AudioTask]):
+class BandwidthEstimationStage(AgentReady, ProcessingStage[AudioTask, AudioTask]):
     """
     Stage that estimates audio bandwidth by analyzing power spectra.
 
@@ -41,6 +44,10 @@ class BandwidthEstimationStage(ProcessingStage[AudioTask, AudioTask]):
         frequency_threshold: Threshold in dB below peak for bandwidth estimation. Defaults to -50.0.
         audio_filepath_key: Key for the audio file path in the manifest. Defaults to "audio_filepath".
         segments_key: Key for the segments in the manifest. Defaults to "segments".
+        waveform_key: Key for an in-memory waveform tensor. Defaults to "waveform".
+        sample_rate_key: Key for the in-memory waveform sample rate. Defaults to "sample_rate".
+        input_residency: Which input to use — "file" (audio_filepath only; default, unchanged),
+            "waveform" (in-memory only), or "auto" (waveform first, file fallback).
 
     Returns:
         The same data as in the input data, but with bandwidth estimates added to each segment.
@@ -52,6 +59,11 @@ class BandwidthEstimationStage(ProcessingStage[AudioTask, AudioTask]):
     frequency_threshold: float = -50.0
     audio_filepath_key: str = "audio_filepath"
     segments_key: str = "segments"
+    duration_key: str = "duration"
+    metrics_key: str = "metrics"
+    waveform_key: str = "waveform"
+    sample_rate_key: str = "sample_rate"
+    input_residency: InputResidency = "file"
 
     # Stage metadata
     name: str = "BandwidthEstimation"
@@ -60,17 +72,80 @@ class BandwidthEstimationStage(ProcessingStage[AudioTask, AudioTask]):
         return [], [self.audio_filepath_key]
 
     def outputs(self) -> tuple[list[str], list[str]]:
-        return [], [self.audio_filepath_key, "metrics"]
+        return [], [self.audio_filepath_key, self.metrics_key]
+
+    def describe(self) -> StageContract:
+        # An audio source (file or in-memory waveform, per input_residency) AND
+        # (segments OR duration). Each audio-source shape is paired with both refinements.
+        reads_one_of = []
+        for spec in residency_read_specs(
+            self.input_residency,
+            audio_filepath_key=self.audio_filepath_key,
+            waveform_key=self.waveform_key,
+            sample_rate_key=self.sample_rate_key,
+        ):
+            reads_one_of.append(IOSpec(data_keys=[*spec.data_keys, self.segments_key], accepts=list(spec.accepts)))
+            reads_one_of.append(IOSpec(data_keys=[*spec.data_keys, self.duration_key], accepts=list(spec.accepts)))
+        return StageContract(
+            reads_one_of=reads_one_of,
+            writes=IOSpec(data_keys=[self.metrics_key], segment_data_keys=[self.metrics_key]),
+            conditional_writes=[
+                ConditionalWrite(
+                    writes=IOSpec(data_keys=[self.metrics_key]),
+                    condition=(
+                        f"audio resolves, '{self.segments_key}' is absent, the top-level item is not skipped "
+                        "for speaker/text, its time range is valid, and bandwidth estimation completes"
+                    ),
+                    value_origin="augments_upstream_same_key",
+                ),
+                ConditionalWrite(
+                    writes=IOSpec(segment_data_keys=[self.metrics_key]),
+                    condition=(
+                        f"audio resolves, '{self.segments_key}' is present, and an individual segment is not "
+                        "skipped for speaker/text, has a valid range, and bandwidth estimation completes"
+                    ),
+                    value_origin="augments_upstream_same_key",
+                ),
+                ConditionalWrite(
+                    writes=IOSpec(segment_data_keys=[self.metrics_key]),
+                    condition=(
+                        f"'{self.segments_key}' is present, an individual segment raises a caught ValueError, "
+                        f"and '{self.metrics_key}.metric_skip_reason' is assigned"
+                    ),
+                    value_origin="augments_upstream_same_key",
+                ),
+            ],
+            # The threshold is measured against the peak of this clip's own power spectrum, not
+            # against a level taken over the corpus.
+            gates=Gates(per_row_independent=True),
+        )
 
     def validate_input(self, task: AudioTask) -> bool:
-        """OR-shaped: needs audio_filepath AND (segments OR duration)."""
+        """Needs an audio source AND (segments OR duration).
+
+        The audio source is ``audio_filepath_key`` (default) or, when ``input_residency``
+        allows it, an in-memory ``waveform_key``+``sample_rate_key``.
+        """
         data = task.data
-        if not hasattr(data, self.audio_filepath_key):
-            logger.error(f"Task {task.task_id} missing '{self.audio_filepath_key}'")
+        has_waveform = data.get(self.waveform_key) is not None and data.get(self.sample_rate_key) is not None
+        has_file = self.audio_filepath_key in data
+        if self.input_residency == "waveform":
+            has_audio = has_waveform
+        elif self.input_residency == "file":
+            has_audio = has_file
+        else:  # auto
+            has_audio = has_waveform or has_file
+        if not has_audio:
+            logger.error(
+                f"Task {task.task_id} missing audio input for input_residency={self.input_residency!r}: "
+                f"need '{self.audio_filepath_key}' or '{self.waveform_key}'+'{self.sample_rate_key}'"
+            )
             return False
-        if hasattr(data, self.segments_key) or hasattr(data, "duration"):
+        if self.segments_key in data or self.duration_key in data:
             return True
-        logger.error(f"Task {task.task_id} missing required attributes: need '{self.segments_key}' OR 'duration'")
+        logger.error(
+            f"Task {task.task_id} missing required attributes: need '{self.segments_key}' OR '{self.duration_key}'"
+        )
         return False
 
     def _estimate_bandwidth(self, audio: "np.ndarray", sample_rate: int) -> int:
@@ -112,14 +187,29 @@ class BandwidthEstimationStage(ProcessingStage[AudioTask, AudioTask]):
         segment_audio_array = audio[int(start * sample_rate) : int(end * sample_rate)]
         bandwidth = self._estimate_bandwidth(segment_audio_array, sample_rate)
 
-        if "metrics" not in audio_segment:
-            audio_segment["metrics"] = {}
+        if self.metrics_key not in audio_segment:
+            audio_segment[self.metrics_key] = {}
 
-        audio_segment["metrics"]["bandwidth"] = int(bandwidth)
+        audio_segment[self.metrics_key]["bandwidth"] = int(bandwidth)
 
-    def process(self, task: AudioTask) -> AudioTask:
-        """Estimate bandwidth for audio entry."""
-        data_entry = task.data
+    def _resolve_entry_audio(self, data_entry: dict[str, Any]) -> tuple["np.ndarray", int]:
+        """Return ``(mono_1d_audio, sample_rate)`` from a waveform or the file.
+
+        When ``input_residency`` allows it and an in-memory waveform is present, it is used
+        directly; otherwise the audio file is loaded at its native rate (default, unchanged).
+        """
+        if self.input_residency != "file":
+            waveform = data_entry.get(self.waveform_key)
+            sr = data_entry.get(self.sample_rate_key)
+            if waveform is not None and sr is not None:
+                audio = ensure_mono(ensure_waveform_2d(waveform)).squeeze(0)
+                return audio.detach().cpu().numpy(), int(sr)
+            if self.input_residency == "waveform":
+                msg = (
+                    f"[{self.name}] Missing '{self.waveform_key}'+'{self.sample_rate_key}' for entry: "
+                    f"{data_entry.get('audio_item_id', 'unknown')} (input_residency='waveform')"
+                )
+                raise ValueError(msg)
         audio_path = data_entry.get(self.audio_filepath_key)
         if not audio_path:
             msg = (
@@ -132,6 +222,12 @@ class BandwidthEstimationStage(ProcessingStage[AudioTask, AudioTask]):
         except Exception as ex:
             msg = f"[{self.name}] Failed to load audio: {audio_path}"
             raise RuntimeError(msg) from ex
+        return audio, sample_rate
+
+    def process(self, task: AudioTask) -> AudioTask:
+        """Estimate bandwidth for audio entry."""
+        data_entry = task.data
+        audio, sample_rate = self._resolve_entry_audio(data_entry)
 
         if self.segments_key in data_entry:
             for segment in data_entry[self.segments_key]:
@@ -139,7 +235,7 @@ class BandwidthEstimationStage(ProcessingStage[AudioTask, AudioTask]):
                     self.get_bandwidth(segment, audio, sample_rate)
                 except ValueError as ex:
                     logger.warning(f"[{self.name}] skipping segment in {task.task_id}: {ex}")
-                    segment.setdefault("metrics", {})["metric_skip_reason"] = str(ex)
+                    segment.setdefault(self.metrics_key, {})["metric_skip_reason"] = str(ex)
         else:
             self.get_bandwidth(data_entry, audio, sample_rate)
 
--- a/nemo_curator/stages/audio/metrics/squim.py
+++ b/nemo_curator/stages/audio/metrics/squim.py
@@ -26,13 +26,16 @@ from loguru import logger
 from torchaudio.pipelines import SQUIM_OBJECTIVE
 
 from nemo_curator.backends.base import NodeInfo, WorkerMetadata
+from nemo_curator.stages.audio._agent._agent_ready import AgentReady, ConditionalWrite, Gates, IOSpec, StageContract
+from nemo_curator.stages.audio._agent._residency import InputResidency, residency_read_specs
+from nemo_curator.stages.audio.common import ensure_mono, ensure_waveform_2d
 from nemo_curator.stages.base import ProcessingStage
 from nemo_curator.stages.resources import Resources
 from nemo_curator.tasks import AudioTask
 
 
 @dataclass
-class TorchSquimQualityMetricsStage(ProcessingStage[AudioTask, AudioTask]):
+class TorchSquimQualityMetricsStage(AgentReady, ProcessingStage[AudioTask, AudioTask]):
     """
     Stage that calculates Squim quality metrics for audio files.
 
@@ -45,6 +48,10 @@ class TorchSquimQualityMetricsStage(ProcessingStage[AudioTask, AudioTask]):
         batch_size: Number of audio tasks to be processed at once. Defaults to 32.
         compute_batch_size: Number of waveforms to process per GPU inference call. Defaults to 32.
         segments_key: Key for the segments in the manifest. Defaults to "segments".
+        waveform_key: Key for an in-memory waveform tensor. Defaults to "waveform".
+        sample_rate_key: Key for the in-memory waveform sample rate. Defaults to "sample_rate".
+        input_residency: Which input to use — "file" (audio_filepath only; default, unchanged),
+            "waveform" (in-memory only), or "auto" (waveform first, file fallback).
 
     Returns:
         The same data as in the input data, but with Squim quality metrics added to each segment.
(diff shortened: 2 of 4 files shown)


nemo_curator/stages/audio/metrics/bandwidth.py
```diff
@@ -52,6 +59,11 @@ class BandwidthEstimationStage(ProcessingStage[AudioTask, AudioTask]):
     frequency_threshold: float = -50.0
     audio_filepath_key: str = "audio_filepath"
     segments_key: str = "segments"
+    duration_key: str = "duration"
```
mohammadaaftabv, line 62: nemo_curator/stages/audio/metrics/bandwidth.py:62,144,182` adds
`duration_key` to the constructor, contract, and input validation, but `get_bandwidth()` still
falls back to the literal `"duration"`. With
`BandwidthEstimationStage(duration_key="clip_len", input_residency="waveform")`, a task carrying
`waveform`, `sample_rate`, and `clip_len=1.0` returns `validate_input=True` and then raises
`ValueError: Invalid segment time range ... end=0.0`. The same accepted configuration on a
nested segment is caught and silently records `metric_skip_reason` instead of bandwidth.

Use `audio_segment.get(self.duration_key, 0.0)` at line 182 (and document the new key), then add
top-level and nested regressions proving a non-default duration key reaches the computation.

nemo_curator/stages/audio/metrics/bandwidth.py
```diff
@@ -112,14 +187,29 @@ def get_bandwidth(self, audio_segment: dict[str, Any], audio: "np.ndarray", samp
...
+
+        audio_segment[self.metrics_key]["bandwidth"] = int(bandwidth)
+
+    def _resolve_entry_audio(self, data_entry: dict[str, Any]) -> tuple["np.ndarray", int]:
+        """Return ``(mono_1d_audio, sample_rate)`` from a waveform or the file.
+
+        When ``input_residency`` allows it and an in-memory waveform is present, it is used
+        directly; otherwise the audio file is loaded at its native rate (default, unchanged).
+        """
+        if self.input_residency != "file":
```
mohammadaaftabv, line 201: nemo_curator/stages/audio/metrics/bandwidth.py:201-206` and
`nemo_curator/stages/audio/metrics/squim.py:201-206` feed resident tensors through
`ensure_waveform_2d()` and `ensure_mono()` without converting their dtype or PCM scale. A valid
stereo `torch.int16` waveform crashes in `torch.mean()` with `Got: Short`; mono integer input
survives with raw PCM magnitudes. For identical PCM16 content, the file path produces float32
samples peaking at `0.5`, while resident SQUIM input peaks at `16384`, so even the non-crashing
case scores different audio numerically.

Normalize supported integer PCM to the same floating-point scale used by file loading before
mono reduction/resampling (and reject unsupported shapes/dtypes clearly). Add mono and stereo
NumPy/Torch PCM16 file-versus-resident parity tests for both stages; merely calling `.float()`
is insufficient because it preserves the wrong amplitude.

nemo_curator/stages/audio/metrics/squim.py
```diff
@@ -126,31 +190,58 @@ def _compute_metrics_batched(self, waveforms: list[torch.Tensor]) -> list[tuple[
...
+    def _resolve_entry_audio(self, data_entry: dict) -> tuple[Any, int]:
+        """Return ``(mono_1d_audio_ndarray, sample_rate)`` from a waveform or the file.
 
-        Returns a list of (task_idx, segment_idx, waveform) tuples.
+        When ``input_residency`` allows it and an in-memory ``waveform_key``+
+        ``sample_rate_key`` is present, that waveform is used directly; otherwise the
+        audio file is read at its native sample rate (the default, unchanged behavior).
+        Per-segment slicing downstream is identical for either source.
         """
+        if self.input_residency != "file":
```
mohammadaaftabv, line 201: nemo_curator/stages/audio/metrics/squim.py:201-232` falls back to disk in `auto` mode whenever
either half of the resident `(waveform, sample_rate)` pair is absent, but it leaves the partial
resident fields untouched. In a deterministic reproduction with a 16,000-sample `0.25` file
and an unrelated 8,000-sample zero waveform with no sample rate, the model received the file
(`shape=(16000,)`, `sum=4000`) while the returned task retained the zero waveform
(`shape=(1,8000)`, `sum=0`), still lacked `sample_rate`, and carried the file-derived SQUIM
metrics. Bandwidth has the same resolution branch at `bandwidth.py:201-225`.

Give partial resident pairs one explicit shared policy: reject them, infer the missing rate and
score the resident waveform, or replace/drop both resident keys when falling back to the file.
Add a deliberately different waveform/file regression; using two representations of the same
audio would not detect this split-brain result.

nemo_curator/stages/audio/metrics/squim.py
```diff
@@ -66,18 +78,70 @@ def inputs(self) -> tuple[list[str], list[str]]:
...
+
+    def describe(self) -> StageContract:
+        # An audio source (file or in-memory waveform, per input_residency) is required;
+        # segments only refine WHERE metrics are attached (optional, read at runtime).
+        return StageContract(
+            reads_one_of=residency_read_specs(
+                self.input_residency,
+                audio_filepath_key=self.audio_filepath_key,
+                waveform_key=self.waveform_key,
+                sample_rate_key=self.sample_rate_key,
```
mohammadaaftabv, line 91: nemo_curator/stages/audio/metrics/bandwidth.py:91-117`,
`nemo_curator/stages/audio/metrics/squim.py:93-112`, and
`nemo_curator/stages/audio/metrics/wer.py:92-124` put the metrics key in both task and segment
`writes`, although runtime selects exactly one scope and may write neither. 

The current #2332
foundation treats `conditional_writes` as advisory and advances every key in `writes` as
guaranteed. Consequently, for each of Bandwidth, SQUIM, and ComputeWER, a segmented input
followed by `PreserveByValueStage("metrics", ...)` validates with `ok=True`, `keys_ok=True`, no
issues, and `metrics` in `produced_keys`; runtime writes only `segments[0].metrics`, and the
approved downstream stage raises for missing top-level `metrics`. The required
`assert_agent_ready` check also fails because the declared top-level write is absent.

The same contradiction appears at `wer.py:388-407`: `GetPairwiseWerStage` guarantees
`wer_pct`, but accepted rows with a null prediction return unchanged and make an approved
consumer fail. 

Resolve the existing #2332 foundation limitation before presenting these as
agent-ready: model guaranteed writes separately from possible conditional writes and do not
advance a possible key mechanically. Alternatively add an explicit task/segments mode whose
configured contract has one exact scope. Add planner-to-runtime and conformance regressions for
task, segment, empty/skipped-segment, and nullable-text branches.

nemo_curator/stages/audio/metrics/wer.py
```diff
@@ -81,7 +83,48 @@ def inputs(self) -> tuple[list[str], list[str]]:
...
 
     def outputs(self) -> tuple[list[str], list[str]]:
-        return [], ["metrics"]
+        return [], [self.metrics_key]
+
+    def describe(self) -> StageContract:
+        # Mirrors validate_input's OR shape: per-segment WER over ``segments``,
+        # OR top-level WER over hypothesis+reference keys on the row.
+        return StageContract(
+            reads_one_of=[
```
mohammadaaftabv, line 92: nemo_curator/stages/audio/metrics/wer.py:92-95` considers the literal presence of `segments`
an entire input alternative, but `get_wer()` requires the configured hypothesis/reference keys
inside each segment and also reads hard-coded `start`, `end`, and `duration` at
`wer.py:230-237`. The planner therefore approves `segments=[{}]`, `validate_input()` returns
true, and the stage emits no metric. A pipeline that remaps timing to `begin`/`finish` is also
approved but silently reports zero character/word rates because those reads are invisible.

Declare the nested hypothesis/reference keys with `segment_data_keys`; add configurable
`start_key`, `end_key`, and `duration_key` fields and use them in runtime. Cover missing nested
text plus remapped timing in planner-to-runtime tests. If PR #2353 is intended to supply the
text-key portion, make it an explicit dependency rather than calling this head agent-ready.

nemo_curator/stages/audio/metrics/bandwidth.py
```diff
@@ -52,6 +59,11 @@ class BandwidthEstimationStage(ProcessingStage[AudioTask, AudioTask]):
     frequency_threshold: float = -50.0
     audio_filepath_key: str = "audio_filepath"
     segments_key: str = "segments"
+    duration_key: str = "duration"
+    metrics_key: str = "metrics"
```
mohammadaaftabv, line 63: The new fields precede existing initializer slots at
`nemo_curator/stages/audio/metrics/bandwidth.py:62-69`,
`nemo_curator/stages/audio/metrics/squim.py:65-75`, and
`nemo_curator/stages/audio/metrics/wer.py:65-72`. A formerly valid positional Bandwidth call
that supplied `name="legacy-name"` now binds that value to `duration_key` and keeps the default
name. The old SQUIM tail `(name, Resources(...), model)` now binds those values to
`metrics_key`, `waveform_key`, and `sample_rate_key`, silently retaining the default GPU
resources. ComputeWER similarly binds the old `name` argument to `metrics_key`.

Keep every pre-PR field in its original positional slot and make the new knobs keyword-only
(or append them after all legacy init fields). Add regressions that construct each class with
the exact pre-PR positional signature and assert every bound field.

nemo_curator/stages/audio/metrics/bandwidth.py
```diff
@@ -112,14 +187,29 @@ def get_bandwidth(self, audio_segment: dict[str, Any], audio: "np.ndarray", samp
...
 
-        if "metrics" not in audio_segment:
-            audio_segment["metrics"] = {}
-
-        audio_segment["metrics"]["bandwidth"] = int(bandwidth)
-
-    def process(self, task: AudioTask) -> AudioTask:
-        """Estimate bandwidth for audio entry."""
-        data_entry = task.data
+        if self.metrics_key not in audio_segment:
```
mohammadaaftabv, line 190: There is no non-empty/distinct-key validation for the new output containers. At
`nemo_curator/stages/audio/metrics/bandwidth.py:190-193`, `metrics_key="waveform"` passes input
validation and then indexes the waveform tensor as a mapping, raising `IndexError`. At
`nemo_curator/stages/audio/metrics/wer.py:239-345`, `metrics_key="text"` reads the transcript
string as the metrics mapping and raises `TypeError: 'str' object does not support item
assignment`; SQUIM reaches the same class of failure after inference at `squim.py:274-282`.
These collisions also overwrite the semantic role associated with the shared literal key.

Reject empty keys and output/container aliases against every input key in `__post_init__`, and
fail clearly before expensive work when a pre-existing metrics value is not a mutable mapping.
Add unique-remap success tests and collision tests for all three stages.

nemo_curator/stages/audio/metrics/squim.py
```diff
@@ -66,18 +78,70 @@ def inputs(self) -> tuple[list[str], list[str]]:
...
+                    writes=IOSpec(segment_data_keys=[self.metrics_key]),
+                    condition=(
+                        f"'{self.segments_key}' is present; an individual segment is not marked no-speaker, "
+                        "has non-blank text and a positive requested frame span, its waveform is collected, "
+                        "and whole-batch model inference completes"
+                    ),
+                    value_origin="augments_upstream_same_key",
+                ),
+            ],
+            gates=Gates(
```
mohammadaaftabv, line 113: The configured SQUIM contract declares `requires_internet_first_run=True` and
`per_row_independent=False` at `nemo_curator/stages/audio/metrics/squim.py:113-119`, while
`setup_on_node()` downloads the model at `squim.py:158-162`. Instance-free public discovery,
however, currently reports `requires_internet_first_run=False` and
`per_row_independent=None`, so an agent can approve an offline environment before it constructs
the stage. It also advertises the runtime-only `model` field at `squim.py:75` as an `Any`
constructor parameter.

Add `AGENT_STATIC = StaticHints(...)` with the invariant internet and per-row facts, and remove
`model` from agent parameters (`init=False` or the registry's explicit internal-field
mechanism). Add a static-catalog regression comparing these invariant gates with the configured
contract.

tests/stages/audio/metrics/test_metrics.py
```diff
@@ -51,6 +51,85 @@ def test_no_segments_computes_on_entry(self, audio_task: Callable[..., AudioTask
         assert "bandwidth" in result.data["metrics"]
         assert result.data["metrics"]["bandwidth"] > 0
 
+    def test_waveform_residency_matches_file(self, audio_task: Callable[..., AudioTask], tmp_path: Path) -> None:
```
mohammadaaftabv, line 54: The four classes now claim `AgentReady`, but the additions in
`tests/stages/audio/metrics/test_metrics.py:54-131` never call `assert_agent_ready` or
`assert_residency_consumption`. The current #2332 + #2336 dependency integration also has no
`nemo_curator/audio_agent/knowledge/cards/` directory. This violates the checklist in
`nemo_curator/stages/audio/AGENTS.md:30-43`, and is why the impossible writes above survive the
otherwise-green tests.

Add one executable `assert_agent_ready` fixture per converted stage, file/waveform consumption
coverage for Bandwidth and SQUIM, and cards that distinguish Bandwidth's Hz output, SQUIM's
score semantics, ComputeWER's ratio-valued nested metrics, and GetPairwiseWer's 0-100
percentage. Run the card audit and public `describe` smoke as part of this PR's tests.

nemo_curator/stages/audio/metrics/bandwidth.py
```diff
@@ -60,17 +72,80 @@ def inputs(self) -> tuple[list[str], list[str]]:
...
+
+        The audio source is ``audio_filepath_key`` (default) or, when ``input_residency``
+        allows it, an in-memory ``waveform_key``+``sample_rate_key``.
+        """
         data = task.data
-        if not hasattr(data, self.audio_filepath_key):
-            logger.error(f"Task {task.task_id} missing '{self.audio_filepath_key}'")
+        has_waveform = data.get(self.waveform_key) is not None and data.get(self.sample_rate_key) is not None
+        has_file = self.audio_filepath_key in data
+        if self.input_residency == "waveform":
```
mohammadaaftabv, line 132: The catch-all branches at `nemo_curator/stages/audio/metrics/bandwidth.py:132-137,201` and
`nemo_curator/stages/audio/metrics/squim.py:133-138,201` treat every string other than `file`
or `waveform` as `auto`; the shared `residency_read_specs()` does likewise. For example,
`input_residency="wavefrom"` constructs successfully, advertises both forms, and validates a
file input. A misspelled YAML value therefore changes source precedence instead of failing at
configuration time.

Validate the runtime value against `{"file", "waveform", "auto"}` centrally or in
`__post_init__`, and add typo regressions for both stages.

mohammadaaftabv (changes requested):

nemo_curator/stages/audio/metrics/bandwidth.py
shubhamNvidia: Agreed. `BandwidthEstimationStage` now uses `self.duration_key` at runtime. Added top-level and nested regressions using a remapped duration key.

nemo_curator/stages/audio/metrics/bandwidth.py
shubhamNvidia: Agreed. Added shared metrics-only normalization for supported NumPy/Torch floating-point and signed PCM inputs. PCM scaling occurs before mono reduction, with mono/stereo PCM16 file-versus-resident parity and invalid shape/dtype regressions.

nemo_curator/stages/audio/metrics/squim.py
shubhamNvidia: Agreed. `waveform` and `auto` modes now reject incomplete waveform/sample-rate pairs instead of silently using a different file. File-only behavior remains unchanged, and both incomplete-pair cases are covered.

nemo_curator/stages/audio/metrics/squim.py
shubhamNvidia: Agreed. Metric outputs are now represented only as conditional writes and are not advanced as guaranteed planner outputs. Conditional producers remain discoverable, nested input keys can be seeded explicitly, and planner/runtime regressions cover skipped and valid paths.

nemo_curator/stages/audio/metrics/wer.py
shubhamNvidia: Added keyword-only `start_key`, `end_key`, and `duration_key` fields and now use them at runtime, including the historical nonzero-start fallback behavior. Missing configured nested-text handling remains intentionally scoped to draft PR #2353. This PR can merge first; #2353 can rebase afterward.

nemo_curator/stages/audio/metrics/bandwidth.py
shubhamNvidia: Agreed. All new parameters are keyword-only, preserving every pre-PR positional slot. Added exact-signature and value-binding regressions for Bandwidth, SQUIM, and ComputeWER.

nemo_curator/stages/audio/metrics/bandwidth.py
shubhamNvidia: Agreed. Configurable keys are now checked for empty values and input/output collisions, including hard-coded runtime segment keys. Existing metrics containers must be mutable mappings and are validated before expensive processing.

nemo_curator/stages/audio/metrics/squim.py
shubhamNvidia: Agreed. Added invariant static hints for first-run internet access and per-row independence. The runtime-only model remains positionally compatible but is excluded from public agent parameters, with registry and static-contract regressions.

tests/stages/audio/metrics/test_metrics.py
shubhamNvidia: Added executable `assert_agent_ready` coverage for all four metric stages and file/waveform residency-consumption coverage for Bandwidth and SQUIM using model-free fixtures. The four capability cards remain in draft PR #2343, where their card audit and public discovery coverage run.

nemo_curator/stages/audio/metrics/bandwidth.py
shubhamNvidia: Agreed. Added centralized validation for `file`, `waveform`, and `auto`, invoked during stage initialization. Invalid or misspelled residency modes now fail immediately, with regressions for Bandwidth and SQUIM.

nemo_curator/stages/audio/metrics/_common.py
```diff
@@ -0,0 +1,154 @@
...
+    protected_fields: tuple[str, ...],
+    reserved_input_keys: tuple[str, ...] = (),
+) -> None:
+    """Reject empty key values and output/input-container collisions."""
+    for field_name, key in keys.items():
+        if not isinstance(key, str) or not key.strip():
+            msg = f"[{stage_name}] '{field_name}' must be a non-empty string"
+            raise ValueError(msg)
+
+    output_key = keys[output_field]
```
mohammadaaftabv: P1 — Collision hardening removes valid legacy in-place/scoped key configurations

nemo_curator/stages/audio/metrics/_common.py:50-54 rejects every output/input
literal alias without considering where the keys live or whether the stage
intentionally replaces a value. These aliases were already part of the public
constructors and some ran successfully at the base SHA.

GetPairwiseWerStage(text_key="text", pred_text_key="pred", wer_key="text")
reads both strings before assignment and validly replaces text with the WER
percentage; this head now fails during construction. Likewise,
ComputeWERStage(segments_key="metrics") validly consumes a top-level
metrics list and writes a separate metrics mapping inside each segment.
Bandwidth and SQUIM also successfully handled segments_key="metrics", and
their segmented paths allowed audio_filepath_key="metrics" because the path
is top-level while outputs are nested. The current check turns all of those
working pipelines into constructor errors.

Make collision validation scope- and operation-aware: retain Pairwise's legacy
in-place write, allow top-level segment/path keys to share a literal with nested
outputs, and reject an alias only in a runtime branch where both values occupy
the same mapping incompatibly. Add execution regressions for the exact pre-PR
calls above; constructor/signature tests alone do not cover this compatibility
surface.

nemo_curator/stages/audio/metrics/_common.py
```diff
@@ -0,0 +1,154 @@
...
+def validate_metric_keys(
+    stage_name: str,
+    *,
+    keys: dict[str, Any],
+    output_field: str,
+    protected_fields: tuple[str, ...],
+    reserved_input_keys: tuple[str, ...] = (),
+) -> None:
+    """Reject empty key values and output/input-container collisions."""
+    for field_name, key in keys.items():
```
mohammadaaftabv: nemo_curator/stages/audio/metrics/_common.py:45-48 applies a new str.strip()
requirement not just to new knobs, but also to every pre-existing configurable
key. At the base SHA, ordinary dict/JSON rows with "" as a literal key ran
successfully through BandwidthEstimationStage(audio_filepath_key=""),
TorchSquimQualityMetricsStage(segments_key=""),
ComputeWERStage(hypothesis_text_key=""), and
GetPairwiseWerStage(wer_key=""); this head rejects all of them at
construction. Empty keys are unusual, but they were valid Python/JSON keys and
no previous contract prohibited them, so this is a functional break under a
complete-backward-compatibility requirement.

Grandfather the pre-existing key fields, or introduce an opt-in strict-
validation/deprecation cycle, and enforce the new invariant only for newly
added parameters until a breaking release. Add at least one execute-through
empty-key regression per affected stage family.

mohammadaaftabv (changes requested):

nemo_curator/stages/audio/metrics/_common.py
```diff
@@ -0,0 +1,154 @@
...
+
+    File mode deliberately ignores resident values to preserve its historical
+    file-only behavior. Waveform and auto modes reject a waveform without its
+    sample rate (or vice versa) instead of falling back to a potentially
+    different file while stale resident state remains on the row.
+    """
+    if residency == "file":
+        return False
+
+    has_waveform = item.get(waveform_key) is not None
```
mohammadaaftabv, line 111: P2 — Resident sample rates are accepted and silently coerced without validation

nemo_curator/stages/audio/metrics/_common.py:81-91 treats any non-None
sample-rate value as a complete resident pair. The two resolvers then call
int(...), so 0, -1, True, and 16000.5 all pass validate_input() and
resolve to 0, -1, 1, and a silently truncated 16000. The resulting
failure or wrong timing occurs later in slicing, STFT, or resampling.

Validate resident rates centrally as positive, losslessly integral, and
non-boolean before declaring the pair usable; explicitly decide whether numeric
strings remain compatible. Add Bandwidth and SQUIM tests in waveform and
auto modes.

mohammadaaftabv (changes requested):

nemo_curator/stages/audio/metrics/_common.py
shubhamNvidia: Thanks for catching this. The validation was too broad and rejected legacy in-place and cross-scope aliases. I’ve limited strict validation to newly introduced fields while preserving the existing key behavior. Regression tests were added for these legacy configurations.

nemo_curator/stages/audio/metrics/_common.py
shubhamNvidia: Agreed, rejecting empty values for pre-existing key parameters was a backward-incompatible change. Empty legacy keys remain supported, while non-empty validation is applied only to the newly added configuration fields.

nemo_curator/stages/audio/metrics/_common.py
shubhamNvidia: Fixed. Resident sample rates are now validated as positive, non-boolean, and losslessly integral values. Compatible representations such as `16000`, `16000.0`, NumPy integers, scalar tensors, and `"16000"` remain supported, while invalid or fractional values are rejected.

nemo_curator/stages/audio/metrics/bandwidth.py
```diff
@@ -52,25 +72,119 @@ class BandwidthEstimationStage(ProcessingStage[AudioTask, AudioTask]):
...
+        # An audio source (file or in-memory waveform, per input_residency) AND
+        # (segments OR duration). Each audio-source shape is paired with both refinements.
+        reads_one_of = []
+        for spec in residency_read_specs(
+            self.input_residency,
+            audio_filepath_key=self.audio_filepath_key,
+            waveform_key=self.waveform_key,
+            sample_rate_key=self.sample_rate_key,
+        ):
+            reads_one_of.append(IOSpec(data_keys=[*spec.data_keys, self.segments_key], accepts=list(spec.accepts)))
```
shubhamNvidia, line 120: **[P1] Make the auto file branch conditional on resident keys being absent**

These unconditional alternatives let the file spec satisfy planning even when only half of the resident pair is present. For `input_residency="auto"` with initial keys `{audio_filepath, duration, waveform}`, `validate_pipeline()` returns `ok=True`, `keys_ok=True`, and no issues via the file alternative, but `validate_input()` raises in `resident_pair_is_complete()` because `sample_rate` is missing and runtime never falls back to the file. SQUIM has the same contract/runtime mismatch. Model auto selection with conditional reads so the resident branch requires both keys and the file branch forbids both (or align runtime with the advertised fallback), and cover both orphan-key directions for both stages.

shubhamNvidia (commented):
reviewed

shubhamNvidia (commented):
reviewed

PR #2337, opened 2026-08-25
feat(audio-inference): optional diarization fan-out + agent-ready contracts

shubhamNvidia:
**Contracts:** `describe()` + `*_key` params on the ASR, speaker-diarization (pyannote / sortformer) and VAD (whisperx) stages, with skip/notes control columns declared.

**New functionality**
- **`fanout=True` on the diarizers and VAD** — instead of attaching a nested list of segments to one task, emit **one task per speaker segment** (carrying `start`/`end`/`speaker`/`segment_num`/`original_file`), and declare it to Ray via `ray_stage_spec` so the executor schedules it as a fan-out. That makes "transcribe each speaker turn separately" a normal pipeline instead of a manual explode step. **Defaults to `False`** — existing nested-list behavior is unchanged.
- **`write_rttm` (pyannote)** — RTTM files were written unconditionally on every run; this makes it a toggle, so a pipeline that only needs segments in memory stops emitting files it never reads. **Defaults to `True`** — unchanged behavior.
- **`vad_onset` / `vad_offset` (pyannote)** — the VAD sensitivity thresholds were hard-coded; they are now parameters, so speech detection can be tuned per corpus without editing the stage.
- **In-memory input** (`input_residency`) on the inference stages.

**Reconciled onto the current base class.** Upstream refactored these onto a shared `AdapterInferenceStage`; `ASRStage` now inherits `(AgentReady, AdapterInferenceStage[ASRAdapter])`, keeping the base's adapter lifecycle (`setup`/`teardown`/`_create_adapter`) and adding the agent-ready parts on top. Transcription behavior is unchanged.

### DIFF
--- a/nemo_curator/stages/audio/inference/asr/stage.py
+++ b/nemo_curator/stages/audio/inference/asr/stage.py
@@ -30,6 +30,7 @@ import torchaudio
 from loguru import logger
 
 from nemo_curator.models.asr.base import ASRAdapter, ASRResult
+from nemo_curator.stages.audio._agent._agent_ready import AgentReady, ConditionalWrite, Gates, IOSpec, StageContract
 from nemo_curator.stages.audio.inference.base import AdapterInferenceStage
 from nemo_curator.stages.resources import Resources
 
@@ -97,16 +98,16 @@ _MONO_DIMENSIONS = 1
 _CHANNEL_FIRST_DIMENSIONS = 2
 
 
-def _set_note(task_data: dict[str, Any], stage_name: str, value: str) -> None:
-    notes = task_data.get(_NOTES_KEY)
+def _set_note(task_data: dict[str, Any], notes_key: str, stage_name: str, value: str) -> None:
+    notes = task_data.get(notes_key)
     if not isinstance(notes, dict):
         notes = {}
-        task_data[_NOTES_KEY] = notes
+        task_data[notes_key] = notes
     notes[stage_name] = value
 
 
 @dataclass
-class ASRStage(AdapterInferenceStage[ASRAdapter]):
+class ASRStage(AgentReady, AdapterInferenceStage[ASRAdapter]):
     """Audio speech-recognition stage with a pluggable adapter.
 
     The stage writes ``pred_text_key`` and optional control columns ``_skipme``
@@ -140,13 +141,19 @@ class ASRStage(AdapterInferenceStage[ASRAdapter]):
 
     resources: Resources = field(default_factory=lambda: Resources(gpus=1.0))
     batch_size: int = 32
+    # Additive agent-facing key knobs live after the existing constructor fields
+    # so legacy positional arguments retain their meaning.
+    skip_me_key: str = _SKIP_ME_KEY
+    notes_key: str = _NOTES_KEY
+    BATCH_ONLY = True
+    INTERNAL_KEY_FIELDS = frozenset({"source_lang_key", "extras_key", "skip_me_key", "notes_key"})
 
     def __post_init__(self) -> None:
         super().__post_init__()
         if not self.pred_text_key:
             msg = "ASRStage.pred_text_key must be non-empty"
             raise ValueError(msg)
-        if self.pred_text_key in {_SKIP_ME_KEY, _NOTES_KEY}:
+        if self.pred_text_key in {self.skip_me_key, self.notes_key}:
             msg = f"ASRStage.pred_text_key cannot use reserved control column {self.pred_text_key!r}"
             raise ValueError(msg)
         if self.extras_key is not None:
@@ -154,7 +161,7 @@ class ASRStage(AdapterInferenceStage[ASRAdapter]):
             if not self.extras_key:
                 msg = "ASRStage.extras_key must be non-empty or None"
                 raise ValueError(msg)
-            if self.extras_key in {self.pred_text_key, _SKIP_ME_KEY, _NOTES_KEY}:
+            if self.extras_key in {self.pred_text_key, self.skip_me_key, self.notes_key}:
                 msg = f"ASRStage.extras_key cannot collide with another output column: {self.extras_key!r}"
                 raise ValueError(msg)
         if int(self.batch_size) <= 0:
@@ -188,10 +195,45 @@ class ASRStage(AdapterInferenceStage[ASRAdapter]):
         )
 
     def outputs(self) -> tuple[list[str], list[str]]:
-        optional_outputs = [self.pred_text_key, _SKIP_ME_KEY, _NOTES_KEY]
+        optional_outputs = [self.pred_text_key, self.skip_me_key, self.notes_key]
         if self.extras_key is not None:
             optional_outputs.append(self.extras_key)
         return [], optional_outputs
+
+    def describe(self) -> StageContract:
+        reads = (
+            IOSpec(data_keys=[self.waveform_key, self.sample_rate_key], accepts=["waveform"])
+            if self.waveform_key
+            else IOSpec(data_keys=[self.audio_filepath_key], accepts=["file"])
+        )
+        conditional_writes = [
+            ConditionalWrite(
+                writes=IOSpec(data_keys=[self.skip_me_key]),
+                condition="the adapter skips an item or audio preparation fails",
+            ),
+            ConditionalWrite(
+                writes=IOSpec(data_keys=[self.notes_key]),
+                condition="a configured language allowlist rejects or cannot identify the item language",
+            ),
+        ]
+        if self.extras_key is not None:
+            conditional_writes.append(
+                ConditionalWrite(
+                    writes=IOSpec(data_keys=[self.extras_key]),
+                    condition="the adapter returns non-empty metadata for the item",
+                )
+            )
+        return StageContract(
+            reads=reads,
+            writes=IOSpec(data_keys=[self.pred_text_key]),
+            conditional_writes=conditional_writes,
+            cardinality="1:1",
+            gates=Gates(
+                requires_gpu=self.resources.requires_gpu,
+                requires_internet_first_run=True,
+                per_row_independent=True,
+            ),
+        )
 
     def _resolve_language(self, task: AudioTask) -> str | None:
         code = self._resolve_language_code(task)
@@ -397,26 +439,28 @@ class ASRStage(AdapterInferenceStage[ASRAdapter]):
             unsupported_language = result.unsupported_language
             missing_language = self._supported_language_codes is not None and not item["language_code"]
             if missing_language:
-                _set_note(task.data, self.name, "skipped (missing language)")
-                _set_note(task.data, self.pred_text_key, "language_missing")
+                _set_note(task.data, self.notes_key, self.name, "skipped (missing language)")
+                _set_note(task.data, self.notes_key, self.pred_text_key, "language_missing")
             elif unsupported_language:
                 _set_note(
                     task.data,
+                    self.notes_key,
                     self.name,
                     f"skipped (unsupported language: {unsupported_language})",
                 )
                 _set_note(
                     task.data,
+                    self.notes_key,
                     self.pred_text_key,
                     f"lang_not_supported:{unsupported_language}",
                 )
             if result.skipped:
-                task.data[_SKIP_ME_KEY] = result.skip_reason or "empty_audio"
+                task.data[self.skip_me_key] = result.skip_reason or "empty_audio"
                 skipped_count += 1
 
         if skipped_count:
             logger.info(
-                f"ASRStage ({self.adapter_target}): marked {skipped_count}/{len(tasks)} tasks with {_SKIP_ME_KEY}",
+                f"ASRStage ({self.adapter_target}): marked {skipped_count}/{len(tasks)} tasks with {self.skip_me_key}",
             )
         logger.debug(
             f"ASRStage ({self.adapter_target}): generated {len(results)} predictions",
--- a/nemo_curator/stages/audio/inference/speaker_diarization/pyannote.py
+++ b/nemo_curator/stages/audio/inference/speaker_diarization/pyannote.py
@@ -34,6 +34,14 @@ from pyannote.audio.pipelines.utils.hook import ProgressHook
 from pyannote.core import Segment
 
 from nemo_curator.backends.base import NodeInfo, WorkerMetadata
+from nemo_curator.backends.utils import RayStageSpecKeys
+from nemo_curator.stages.audio._agent._agent_ready import AgentReady, Gates, IOSpec, StageContract
+from nemo_curator.stages.audio._agent._residency import (
+    InputResidency,
+    cleanup_temp_files,
+    residency_read_specs,
+    resolve_audio_path,
+)
 from nemo_curator.stages.audio.common import get_audio_duration
 from nemo_curator.stages.audio.inference.vad.whisperx_vad import WhisperXVADModel
 from nemo_curator.stages.audio.tagging.utils import add_non_speaker_segments
@@ -73,7 +81,7 @@ def has_overlap(turn: Segment, overlaps: list) -> bool:
 
 
 @dataclass
-class PyAnnoteDiarizationStage(ProcessingStage[AudioTask, AudioTask]):
+class PyAnnoteDiarizationStage(AgentReady, ProcessingStage[AudioTask, AudioTask]):
     """
     Stage that performs speaker diarization and overlap detection using PyAnnote.
 
@@ -85,6 +93,8 @@ class PyAnnoteDiarizationStage(ProcessingStage[AudioTask, AudioTask]):
         embedding_batch_size: Batch size for speaker embeddings
         min_length: Minimum segment length in seconds
         max_length: Maximum segment length in seconds
+        num_speakers_key: Key in output data for the distinct-speaker count derived
+            from the diarization result (passthrough mode only). Defaults to "num_speakers".
         xenna_num_workers: If set, caps workers cluster-wide. Prefer ``with_(num_workers=...)`` for new code.
     """
 
@@ -102,8 +112,24 @@ class PyAnnoteDiarizationStage(ProcessingStage[AudioTask, AudioTask]):
     max_length: float = 40.0
 
     audio_filepath_key: str = "resampled_audio_filepath"
+    waveform_key: str = "waveform"
+    sample_rate_key: str = "sample_rate"
     segments_key: str = "segments"
     overlap_segments_key: str = "overlap_segments"
+    num_speakers_key: str = "num_speakers"
+    input_residency: InputResidency = "file"
+    write_rttm: bool = True
+    vad_onset: float = 0.5
+    vad_offset: float = 0.363
+    fanout: bool = False
+    start_key: str = "start"
+    end_key: str = "end"
+    start_ms_key: str = "start_ms"
+    end_ms_key: str = "end_ms"
+    duration_key: str = "duration"
+    segment_num_key: str = "segment_num"
+    speaker_key: str = "speaker"
+    original_file_key: str = "original_file"
 
     # Stage metadata
     name: str = "PyAnnoteDiarization"
@@ -121,7 +147,107 @@ class PyAnnoteDiarizationStage(ProcessingStage[AudioTask, AudioTask]):
         return [], [self.audio_filepath_key]
 
     def outputs(self) -> tuple[list[str], list[str]]:
-        return [], [self.audio_filepath_key, self.segments_key, self.overlap_segments_key]
+        if self.fanout:
+            return [], [
+                self.audio_filepath_key,
+                self.start_key,
+                self.end_key,
+                self.start_ms_key,
+                self.end_ms_key,
+                self.duration_key,
+                self.segment_num_key,
+                self.speaker_key,
+                self.original_file_key,
+            ]
+        return [], [self.audio_filepath_key, self.segments_key, self.overlap_segments_key, self.num_speakers_key]
+
+    def describe(self) -> StageContract:
+        if self.fanout:
+            writes = [
+                self.audio_filepath_key,
+                self.start_key,
+                self.end_key,
+                self.start_ms_key,
+                self.end_ms_key,
+                self.duration_key,
+                self.segment_num_key,
+                self.speaker_key,
+                self.original_file_key,
+            ]
+            cardinality = "1:N fan-out"
+        else:
+            writes = [self.segments_key, self.overlap_segments_key, self.num_speakers_key]
+            cardinality = "1:1"
+        return StageContract(
+            reads_one_of=residency_read_specs(
+                self.input_residency,
+                audio_filepath_key=self.audio_filepath_key,
+                waveform_key=self.waveform_key,
+                sample_rate_key=self.sample_rate_key,
+            ),
+            writes=IOSpec(data_keys=writes),
+            cardinality=cardinality,
+            cardinality_options=["passthrough", "fan_out"],
+            iteration_key=self.segments_key if self.fanout else None,
+            gates=Gates(
+                requires_gpu=self.resources.requires_gpu,
+                writes_to_disk=self.write_rttm,
+                runtime_secrets=["HF_TOKEN"],
+                # ``add_vad_segments`` draws from one unseeded ``random.Random()`` built in
+                # ``setup()``, so the durations a file gets depend on how many files that worker
+                # segmented before it.
+                per_row_independent=False,
+            ),
+        )
+
+    def ray_stage_spec(self) -> dict[str, Any]:
+        if self.fanout:
+            return {RayStageSpecKeys.IS_FANOUT_STAGE: True}
+        return {}
+
+    def _segment_child_data(
+        self,
+        item: dict[str, Any],
+        segment: dict[str, Any],
+        segment_num: int,
+    ) -> dict[str, Any]:
+        child = {k: v for k, v in item.items() if k not in {self.segments_key, self.overlap_segments_key}}
+        child.update({k: v for k, v in segment.items() if k not in {"start", "end", "speaker"}})
+        start = float(segment.get("start", 0.0))
+        end = float(segment.get("end", start))
+        child[self.start_key] = start
+        child[self.end_key] = end
+        child[self.start_ms_key] = round(start * 1000)
+        child[self.end_ms_key] = round(end * 1000)
+        child[self.duration_key] = max(0.0, end - start)
+        child[self.segment_num_key] = segment_num
+        if "speaker" in segment:
+            child[self.speaker_key] = segment["speaker"]
+        # provenance chain mirrors whisperx: configured key, then the CANONICAL
+        # audio_filepath, then the legacy resampled key — never skip canonical
+        # (with the default audio_filepath_key='resampled_audio_filepath' the
+        # old chain probed the same key twice and fell to "unknown").
+        original_file = item.get(
+            self.original_file_key,
+            item.get(
+                self.audio_filepath_key,
+                item.get("audio_filepath", item.get("resampled_audio_filepath", "unknown")),
+            ),
+        )
+        child.setdefault(self.original_file_key, original_file)
+        return child
+
+    def _fanout_segments(self, task: AudioTask, segments: list[dict[str, Any]]) -> list[AudioTask]:
+        return [
+            AudioTask(
+                dataset_name=task.dataset_name,
+                filepath_key=task.filepath_key or self.audio_filepath_key,
+                data=self._segment_child_data(task.data, segment, index),
+                _metadata=dict(task._metadata or {}),
+                _stage_perf=list(task._stage_perf),
+            )
+            for index, segment in enumerate(segments)
+        ]
 
     def num_workers(self) -> int | None:
         return self.xenna_num_workers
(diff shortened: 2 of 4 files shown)


nemo_curator/stages/audio/inference/speaker_diarization/pyannote.py
mohammadaaftabv, line 146: P0 (blocker): the normal executor rejects every new waveform-residency configuration before the
residency code runs. PyAnnoteDiarizationStage.inputs() remains file-only at
nemo_curator/stages/audio/inference/speaker_diarization/pyannote.py:146-147; the same is true at
speaker_diarization/sortformer.py:229-230 and vad/whisperx_vad.py:145-146. 

None of these stages overrides process_batch, so the ProcessingStage implementation performs this legacy validation before calling the new process() path. On the current PR #2332 stack, each configured contract says it reads waveform + sample_rate, but validate_input() returns false and process_batch([task]) raises for the missing file. 

The PyAnnote/WhisperX fallback from resampled_audio_filepath to canonical
audio_filepath is unreachable for the same reason. This contradicts the runtime-validation contract in
.cursor/rules/processing-stage-patterns.mdc. 

Make input_spec_for_task()/validate_input() genuinely residency-aware (including the two alternatives in auto and the canonical path fallback), or override
process_batch with equivalent validation. Add file, waveform, auto, and canonical-fallback regressions
that call process_batch or a real backend; direct process() tests do not exercise this failure.

nemo_curator/stages/audio/inference/speaker_diarization/pyannote.py
```diff
@@ -121,7 +147,107 @@ def inputs(self) -> tuple[list[str], list[str]]:
...
+                per_row_independent=False,
+            ),
+        )
+
+    def ray_stage_spec(self) -> dict[str, Any]:
+        if self.fanout:
+            return {RayStageSpecKeys.IS_FANOUT_STAGE: True}
+        return {}
+
+    def _segment_child_data(
```
mohammadaaftabv, line 208: P0 (blocker): fanout=True emits timestamp annotations, not audio segments. 

The child builders at
nemo_curator/stages/audio/inference/speaker_diarization/pyannote.py:208-238,
speaker_diarization/sortformer.py:295-329, and vad/whisperx_vad.py:199-232 shallow-copy the parent
row and only add boundary metadata. 

A file-backed child retains the same full-recording path; a
waveform-backed child retains the exact same full waveform object. ASRStage._build_items() at
inference/asr/stage.py:258-272 then reads that full path/waveform and never consumes start/end.
The model-free reproduction used two 250 ms intervals over a 16,000-sample parent and the ASR adapter
received two identical 16,000-sample arrays, from sample 0 through 15,999. 

The advertised “transcribe
each speaker turn separately” flow therefore transcribes the complete recording once per interval and
silently returns duplicated/wrong transcripts while multiplying GPU work and resident memory. 

Follow
the existing VADSegmentationStage._build_segment_item() pattern: slice and clone the resident waveform
using sample-rate-aware bounds, or materialize stable per-segment files and point each child to its own
file. 

Add an end-to-end fan-out → spy-ASR test using nonconstant audio so both child lengths and content
are asserted for file and waveform residency.

nemo_curator/stages/audio/inference/speaker_diarization/pyannote.py
```diff
@@ -121,7 +147,107 @@ def inputs(self) -> tuple[list[str], list[str]]:
...
+                self.end_ms_key,
+                self.duration_key,
+                self.segment_num_key,
+                self.speaker_key,
+                self.original_file_key,
+            ]
+        return [], [self.audio_filepath_key, self.segments_key, self.overlap_segments_key, self.num_speakers_key]
+
+    def describe(self) -> StageContract:
+        if self.fanout:
```
mohammadaaftabv, line 165: P1: the new contracts are not conformant with the new waveform fan-out results. 

PyAnnote declares audio_filepath_key in every fan-out result at
nemo_curator/stages/audio/inference/speaker_diarization/pyannote.py:165-188; Sortformer does so at
speaker_diarization/sortformer.py:248-271; WhisperX does so at
vad/whisperx_vad.py:162-185. 

Waveform-only children have no such path, and Sortformer intentionally
refuses to preserve its soon-deleted materialization path at sortformer.py:315-320. 

Exact
assert_agent_ready checks fail with declared write 'resampled_audio_filepath' missing for PyAnnote
and WhisperX and declared write 'audio_filepath' missing for Sortformer. PyAnnote also installs
num_speakers before fan-out at pyannote.py:453-460, then carries it into every child without declaring
it. 

These lies let the planner select a file-only downstream stage for rows that contain only a
waveform. Declare the representation actually emitted for each residency (including waveform/sample
rate), use conditional writes for preserved source paths, and either declare or remove num_speakers.
Per nemo_curator/stages/audio/AGENT_READY.md, add at least one assert_agent_ready(...) test per stage;
this PR currently changes no tests at all which should be avoided when adding functionality to stages.

nemo_curator/stages/audio/inference/asr/stage.py
```diff
@@ -188,11 +195,46 @@ def _create_adapter(self) -> ASRAdapter:
...
+            ),
+        ]
+        if self.extras_key is not None:
+            conditional_writes.append(
+                ConditionalWrite(
+                    writes=IOSpec(data_keys=[self.extras_key]),
+                    condition="the adapter returns non-empty metadata for the item",
+                )
+            )
+        return StageContract(
```
mohammadaaftabv, line 226: P1: the ASR contract allows a pipeline that its default runtime behavior breaks. 

The contract at
nemo_curator/stages/audio/inference/asr/stage.py:226-235 has no removes_keys, but process_batch()
removes the configured waveform whenever keep_waveform=False at stage.py:334-336. With ASR followed
by a waveform-resident GetAudioDurationStage, validate_pipeline() reports ok=True, keys_ok=True,
and no issues; runtime ASR leaves only pred_text and sample_rate, and the next stage raises. Declare
the mutation in the returned contract:

removes_keys=[self.waveform_key] if self.waveform_key and not self.keep_waveform else [],

Add contract/planner/runtime regressions for both keep_waveform values.

nemo_curator/stages/audio/inference/speaker_diarization/pyannote.py
```diff
@@ -102,8 +112,24 @@ class PyAnnoteDiarizationStage(ProcessingStage[AudioTask, AudioTask]):
     max_length: float = 40.0
 
     audio_filepath_key: str = "resampled_audio_filepath"
+    waveform_key: str = "waveform"
```
mohammadaaftabv, line 115: P1: the new public fields were inserted into the middle of three existing dataclass signatures. 

In PyAnnote they precede the old segment keys at
nemo_curator/stages/audio/inference/speaker_diarization/pyannote.py:114-132; in Sortformer they precede
the old diar_segments_key/RTTM/model settings at sortformer.py:130-159; and in WhisperX they precede
the old name/resources fields at vad/whisperx_vad.py:122-141. 

Replaying valid pre-PR positional
calls made PyAnnote bind "legacy_segments" as waveform_key, Sortformer bind it as waveform_key
instead of diar_segments_key, and WhisperX bind a former stage name as waveform_key while leaving
name="WhisperXVAD". 

The ASR change in this PR explicitly appends its new fields to preserve exactly
this compatibility. Do the same for all three stages (or make only the new fields keyword-only without
disturbing old positional slots), and add signature-binding regressions using the pre-PR constructors.

nemo_curator/stages/audio/inference/speaker_diarization/pyannote.py
```diff
@@ -239,11 +396,12 @@ def process(self, task: AudioTask) -> AudioTask:
         diarization = diarization.crop(Segment(0, len(s[0]) / fs))
 
         # Write RTTM file (cloud-aware via fsspec)
-        logger.info(f"Writing {len(diarization._tracks)} turns to RTTM file")
-        rttm_filepath = os.path.splitext(file_path)[0] + ".rttm"
-        rttm_fs, rttm_path = url_to_fs(rttm_filepath)
-        with rttm_fs.open(rttm_path, "w") as rttm_file:
-            diarization.write_rttm(rttm_file)
+        if self.write_rttm:
```
mohammadaaftabv, line 399: P1: waveform input plus the default write_rttm=True leaks one RTTM per task. 

The residency helper
materializes a temporary WAV, pyannote.py:398-404 writes a sibling .rttm, and the finally block at
pyannote.py:369-372 cleans only the registered WAV. The reproduction observed the WAV gone, the RTTM
still present, and no RTTM path in the returned task. 

Long-running waveform pipelines therefore fill
/tmp with unreachable output. 

This is the same defect recorded in the post-#1608 corpus for superseded
PR #2176. Either register the sibling RTTM for cleanup when the input path is temporary, or introduce an
explicit durable RTTM destination/output key and return the artifact. Add a waveform-mode regression
that asserts both the intended artifact and the absence of leaked siblings after success and failure.

nemo_curator/stages/audio/inference/speaker_diarization/pyannote.py
```diff
@@ -121,7 +147,107 @@ def inputs(self) -> tuple[list[str], list[str]]:
...
+                self.duration_key,
+                self.segment_num_key,
+                self.speaker_key,
+                self.original_file_key,
+            ]
+            cardinality = "1:N fan-out"
+        else:
+            writes = [self.segments_key, self.overlap_segments_key, self.num_speakers_key]
+            cardinality = "1:1"
+        return StageContract(
```
mohammadaaftabv, line 181: P1: the agent-ready surface understates the execution conditions it was added to expose. 

PyAnnote's
configured gates at nemo_curator/stages/audio/inference/speaker_diarization/pyannote.py:181-200 report
no first-run internet requirement, even though setup downloads both the PyAnnote pipeline and WhisperX
VAD at pyannote.py:260-285; WhisperX has the same omission at
vad/whisperx_vad.py:178-191. 

The classes also provide no conservative AGENT_STATIC hints. Exact
instance-free discovery of default PyAnnote reported writes_to_disk=False, no internet, no runtime
secret, unknown row independence, and no cardinality options, whereas its configured default reports a
disk write, HF_TOKEN, row dependence, and both fan-out modes. 

An agent can therefore approve an
offline or unsandboxed catalog smoke before construction reveals the real requirements. Set the dynamic
internet gates correctly and add StaticHints for mandatory/default GPU, download, secret, disk/output,
row-independence, dispatch, and cardinality facts. Test both static_contract(Class) and configured
contracts for all four changed stages.

nemo_curator/stages/audio/inference/speaker_diarization/pyannote.py
mohammadaaftabv, line 411: P1: valid waveform-only inputs can produce different durable identities on every run. 

The PyAnnote
contract accepts only waveform + sample_rate, but when no optional manifest ID/path exists the branch
at nemo_curator/stages/audio/inference/speaker_diarization/pyannote.py:411-424 derives speaker IDs from
the random materialized WAV stem, while pyannote.py:226-237 records original_file="unknown".
Sortformer similarly derives an RTTM session name from its random temp WAV at
speaker_diarization/sortformer.py:381-390; identical reruns create different tmp*.rttm files. 

Require
or derive a stable logical input identity (for example audio_item_id, explicit session_name, or a
stable source/content ID), never use the temporary materialization path as provenance, declare that
required identity in the contract, and add identical-input rerun tests.

mohammadaaftabv (changes requested):

nemo_curator/stages/audio/inference/speaker_diarization/pyannote.py
shubhamNvidia: Agreed. PyAnnote, Sortformer, and WhisperX now perform residency-aware validation through normal `process_batch()` execution. File, waveform, auto, canonical filepath fallback, resident preference, and incomplete-pair behavior are covered by model-free regressions.

nemo_curator/stages/audio/inference/speaker_diarization/pyannote.py
shubhamNvidia: Agreed. Fan-out children now contain cloned channel-first waveform slices bounded to each segment, plus their sample rate. Full-recording paths are removed. End-to-end spy-ASR tests verify exact child lengths and nonconstant content for file and waveform inputs.

nemo_curator/stages/audio/inference/speaker_diarization/pyannote.py
shubhamNvidia: Agreed. Fan-out contracts now declare waveform/sample-rate outputs, tensor production, timing/provenance metadata, and removed full-path and parent-container keys. PyAnnote retains and declares the parent `num_speakers` count. Added `assert_agent_ready` and planner regressions for all three stages.

nemo_curator/stages/audio/inference/asr/stage.py
shubhamNvidia: Agreed. `ASRStage.describe()` now declares `removes_keys=[waveform_key]` when a waveform key is configured and `keep_waveform=False`. Planner and runtime tests cover retention, removal, and `skip_if_output_exists` rows.

nemo_curator/stages/audio/inference/speaker_diarization/pyannote.py
shubhamNvidia: Agreed. Only PR-added fields are now keyword-only. The exact pre-PR positional slots for PyAnnote, Sortformer, and WhisperX are preserved and covered by signature-binding regressions.

nemo_curator/stages/audio/inference/speaker_diarization/pyannote.py
shubhamNvidia: Agreed. RTTMs written beside temporary waveform materializations are registered for cleanup on both success and failure. RTTMs written beside real file inputs remain durable, with regressions for all three cases.

nemo_curator/stages/audio/inference/speaker_diarization/pyannote.py
shubhamNvidia: Agreed. Added narrowed static hints and corrected configured gates for GPU, first-run internet, runtime secrets, disk output, row independence, and fan-out cardinality. Local/injected Sortformer models correctly avoid the internet gate. Corresponding capability cards were updated in draft PR #2343.

nemo_curator/stages/audio/inference/speaker_diarization/pyannote.py
shubhamNvidia: Agreed. Temporary materialization paths are no longer used as speaker/session identities. Explicit IDs and stable source paths retain precedence; waveform-only inputs fall back to a deterministic waveform-and-sample-rate hash. Identical reruns and explicit-ID precedence are covered.

nemo_curator/stages/audio/inference/speaker_diarization/pyannote.py
```diff
@@ -212,18 +442,97 @@ def add_vad_segments(  # noqa: PLR0913
...
+        temp_paths: list[str] = []
+        file_path = resolve_audio_path(
+            audio_input,
+            residency=self.input_residency,  # type: ignore[arg-type]
+            audio_filepath_key=self.audio_filepath_key,
+            waveform_key=self.waveform_key,
+            sample_rate_key=self.sample_rate_key,
+            register_temp=temp_paths,
+        )
+        if file_path is None and self.audio_filepath_key != "audio_filepath":
```
mohammadaaftabv: P0 (backward compatibility): keep configured source selection strict by
default. 

These branches silently substitute canonical audio_filepath when
the configured key is missing. 
On main,
audio_filepath_key="clean_audio" fails if clean_audio is absent; this head
instead processes the raw recording. 
The default
resampled_audio_filepath can likewise bypass a missing resampling stage,
silently producing output from the wrong audio. 

The differential reproduction
raised KeyError('clean_audio') on main but ran WhisperX on the raw fallback
here. 

Put alias fallback behind an explicit opt-in whose default is false, and
preserve exact configured-key enforcement in file mode. When opt-in fallback
is used, derive source_path, stable identity, and original_file from the
source actually selected, not the invalid preferred path. Add regressions for
missing configured keys and explicit fallback in both stages.

nemo_curator/stages/audio/inference/speaker_diarization/pyannote.py
```diff
@@ -281,14 +583,21 @@ def process(self, task: AudioTask) -> AudioTask:
...
 
         # Update entry
         data_entry[self.segments_key] = segments
         data_entry[self.overlap_segments_key] = overlap_segments
 
-        speakers = {seg["speaker"] for seg in segments if seg.get("speaker") != "no-speaker"}
+        # Distinct speakers across turns AND overlap-only turns; "no-speaker" is a
+        # silence/VAD placeholder, not a real speaker.
+        speakers = {seg["speaker"] for seg in (*segments, *overlap_segments) if seg.get("speaker") != "no-speaker"}
+        data_entry[self.num_speakers_key] = len(speakers)
```
mohammadaaftabv: P0 (backward compatibility): do not add or overwrite num_speakers under legacy defaults. 

Main's non-fan-out stages wrote only their segment outputs; these unconditional assignments now add a serialized column and destroy an upstream manifest's existing num_speakers. 

They also break old-valid key aliases: InferenceSortformerStage(diar_segments_key="num_speakers") returned the segment list on main but returns integer 1 here. 

Make derived-count output disabled/opt-in by default (for example, num_speakers_key: str | None = None) and condition the output declaration, runtime write, and fan-out carry-through on it. Validate enabled output keys as distinct. Add default non-addition, upstream-preservation, and legacy-alias
regressions. 

Merely preserving an existing value is insufficient for complete
compatibility because an unchanged default invocation must not add the new
field either.

nemo_curator/stages/audio/inference/speaker_diarization/pyannote.py
```diff
@@ -281,14 +583,21 @@ def process(self, task: AudioTask) -> AudioTask:
                     )
 
         # Add non-speaker segments
```
mohammadaaftabv, line 626: speaker_diarization/pyannote.py:585-617 appends synthesized
"no-speaker" gaps to segments, then fans out only segments while removing
and ignoring overlap_segments. A five-second reproduction with speech at
[1,2] and an overlap turn at [2,3] emitted three children: silence [0,1],
speech [1,2], and silence [2,5]; the overlap speaker disappeared. 

That
contradicts the advertised one-child-per-speaker-turn ASR pipeline and loses
real speech.

Define the fan-out population explicitly: exclude synthetic no-speaker
entries and include overlap turns with explicit overlap metadata (or expose a
documented policy knob). 

Add a regression using the real
add_non_speaker_segments; the present tests mock away the behavior that
reveals this failure. This is distinct from the outdated earlier comment about
children carrying the full recording.

nemo_curator/stages/audio/inference/speaker_diarization/pyannote.py
```diff
@@ -212,18 +442,97 @@ def add_vad_segments(  # noqa: PLR0913
...
+            _channel_first_waveform(data_entry[self.waveform_key])
+            if self.input_residency != "file"
+            and data_entry.get(self.waveform_key) is not None
+            and data_entry.get(self.sample_rate_key) is not None
+            else None
+        )
+        resident_sample_rate = int(data_entry[self.sample_rate_key]) if resident_waveform is not None else None
+        audio_input = data_entry if resident_waveform is None else {**data_entry, self.waveform_key: resident_waveform}
+        temp_paths: list[str] = []
+        file_path = resolve_audio_path(
```
mohammadaaftabv, line 509: The new calls at speaker_diarization/pyannote.py:468-475,
speaker_diarization/sortformer.py:452-459, and
vad/whisperx_vad.py:376-383 route resident float audio through PR 2332's
resolve_audio_path. That helper writes a temporary WAV without a subtype, so
SoundFile uses PCM16. The model therefore does not see the resident signal it
was given. The direct reproduction reported subtype PCM_16 and changed
[1e-5, -1e-5, 1.25, -1.25] to
[0, -3.0517578e-5, 0.9999695, -1].

Write temporary WAVs with subtype="FLOAT" or pass resident arrays directly
to backends that support them. 

Assert the exact low-amplitude/model-observed
samples for all three stages; current child-slice tests use the original array
and cannot detect what the model consumed. Also make the contract honest about
temporary disk materialization (the currently open gate thread is incomplete
for input_residency="waveform"/"auto"). The helper fix belongs in the PR
2332 prerequisite, and PR 2337 should test the integrated behavior.

nemo_curator/stages/audio/inference/vad/whisperx_vad.py
```diff
@@ -155,34 +351,103 @@ def setup(self, _: WorkerMetadata | None = None) -> None:
...
+        if file_path is None:
+            msg = f"[{self.name}] Missing audio input for key '{self.audio_filepath_key}'"
+            raise ValueError(msg)
+        try:
+            duration = (
+                _resident_audio_duration(resident_waveform, resident_sample_rate)
+                if resident_waveform is not None and resident_sample_rate is not None
+                else data_entry.get("duration", get_audio_duration(file_path))
+            )
+            if duration < self.min_length:
```
mohammadaaftabv, line 422: vad/whisperx_vad.py:404-415 returns [] for a short recording with
fanout=True; all three fan-out helpers can return an empty list when no
segments are found. 

These stages inherit is_resumable=True. On the intended
stack, BaseStageAdapter receives no output or sentinel, takes its empty-output
early return, emits no parent -1 delta, and leaves the source pending forever.
The reproduction observed outputs=[], deltas=[], and
stage_resumable=True.

Represent 1-to-0 as an attributable None/NoneTask, or teach the adapter to
account for an explicit zero-child result; mark these modes non-resumable until
that is safe. 

Add no-speech and short-recording checkpoint/restart tests for
PyAnnote, Sortformer, and WhisperX fan-out.

nemo_curator/stages/audio/inference/base.py
```diff
@@ -16,24 +16,275 @@
...
+    if np.issubdtype(array.dtype, np.integer):
+        msg = f"unsupported resident waveform integer dtype {array.dtype}; supported PCM dtypes are int16 and int32"
+        raise ValueError(msg)
+    if not np.issubdtype(array.dtype, np.floating):
+        msg = f"unsupported resident waveform dtype {array.dtype}; expected floating-point or signed PCM int16/int32"
+        raise ValueError(msg)
+    return np.ascontiguousarray(array, dtype=np.float32)
+
+
+def _fanout_audio_slice(
```
mohammadaaftabv, line 164: inference/base.py:164-180 clamps slice indices to the waveform, but the three
child builders publish the raw model timestamps. 

For a one-second waveform,
[-0.2,0.4] emits 0.4 seconds of samples while claiming duration 0.6;
[0.8,1.5] emits 0.2 seconds while claiming 0.7; reversed bounds emit an empty
array. 

Normalize or reject boundaries once and derive the slice, seconds,
milliseconds, and duration from the same normalized sample indices. 
Add negative, overrun, and reversed-boundary tests.

nemo_curator/stages/audio/inference/speaker_diarization/sortformer.py
mohammadaaftabv, line 232: speaker_diarization/sortformer.py:224-230 uses an injected diar_model
directly and does not move it to CUDA, while describe() hardcodes
requires_gpu=True at sortformer.py:341-343. A CPU-injected stage with CPU
resources therefore has a false planning gate. Derive this gate from the
configured resources/model path and make runtime behavior match, or reject CPU
configuration explicitly. Add the missing injected-model GPU-gate assertion.

tests/stages/audio/inference/test_review_fixes.py
mohammadaaftabv, line 1: All tests for five source files live in the new 1,008-line
tests/stages/audio/inference/test_review_fixes.py. AGENTS.md:30-33 requires
unit-test files to map one-to-one to source files and cautions against mocking
away the boundary under test. Split these regressions into the matching ASR,
base, PyAnnote, Sortformer, and WhisperX test modules. In particular, preserve a
real non-speaker/overlap fan-out case and model-observed residency samples.

After the PR 2332 package is present, Ruff also reports I001 on this test
file because its _agent imports are separated from the other first-party
nemo_curator imports. Reorder them before testing the rebased stack.

mohammadaaftabv (changes requested):

nemo_curator/stages/audio/inference/speaker_diarization/pyannote.py
shubhamNvidia: Thanks for catching this. The configured audio path is now strict by default. Falling back to  requires explicitly enabling , preserving existing configured-path behavior.

nemo_curator/stages/audio/inference/speaker_diarization/pyannote.py
shubhamNvidia: Fixed.  is now disabled by default for both diarization stages. Users can explicitly set  when the derived count is required.

nemo_curator/stages/audio/inference/speaker_diarization/pyannote.py
shubhamNvidia: Fixed. PyAnnote fanout now excludes synthetic  gap entries, includes overlap turns, and marks emitted overlap children with .

nemo_curator/stages/audio/inference/speaker_diarization/pyannote.py
shubhamNvidia: Fixed in the shared residency helper. Temporary WAV files created from resident waveforms now use the  subtype, preserving the exact model-observed samples. Disk-write contracts were also updated accordingly.

nemo_curator/stages/audio/inference/vad/whisperx_vad.py
shubhamNvidia: Agreed. Fanout configurations are now marked non-resumable until the backend can safely account for inputs that legitimately produce zero children. Passthrough behavior remains resumable.

nemo_curator/stages/audio/inference/base.py
shubhamNvidia: Fixed. Waveform slicing and published start/end/duration metadata now use the same clamped, sample-aligned boundaries, preventing metadata from describing samples outside the emitted waveform.

nemo_curator/stages/audio/inference/speaker_diarization/sortformer.py
shubhamNvidia: Fixed. A Sortformer stage using an injected CPU model no longer reports a GPU requirement. Default model loading continues to declare GPU requirements.

tests/stages/audio/inference/test_review_fixes.py
shubhamNvidia: Fixed. The catch-all regression file was removed. Tests are now organized under their corresponding inference source modules, with shared model-free fixtures placed in a dedicated helper. Ruff, DCO, pre-commit, compatibility, tutorial, agent-contract, and non-GPU inference checks pass.

nemo_curator/stages/audio/inference/asr/stage.py
```diff
@@ -140,21 +158,35 @@ class ASRStage(AdapterInferenceStage[ASRAdapter]):
...
+        gates=Gates(
+            requires_gpu=True,
+            requires_internet_first_run=True,
+            per_row_independent=True,
+        )
+    )
 
     def __post_init__(self) -> None:
         super().__post_init__()
+        _validate_control_keys(self.skip_me_key, self.notes_key)
```
shubhamNvidia: **[P1] Reject control-key collisions with audio inputs**

The new controls are checked against each other and the other outputs, but not against configured input keys. `ASRStage(..., waveform_key="waveform", skip_me_key="waveform", keep_waveform=True)` constructs successfully; when an adapter returns a skipped result, `assemble()` replaces the waveform array with the skip-reason string. With `keep_waveform=False`, the later waveform pop instead deletes that control marker. Either outcome is destructive while planning can still treat the literal key as resident audio. Require the new control/output keys to be disjoint from all configured input-role keys and cover skipped rows under both waveform-retention modes.

nemo_curator/stages/audio/inference/speaker_diarization/sortformer.py
```diff
@@ -194,10 +279,161 @@ def _configure_streaming(self) -> None:
...
+                self.start_ms_key,
+                self.end_ms_key,
+                self.duration_key,
+                self.segment_num_key,
+                self.speaker_key,
+                self.original_file_key,
+            ]
+            cardinality = "1:N fan-out"
+        else:
+            writes = [self.filepath_key, self.diar_segments_key]
```
shubhamNvidia: **[P1] Do not guarantee a filepath for waveform-only output**

In non-fan-out waveform mode this stage never materializes a durable path into the returned task: a valid `{waveform, sample_rate}` row comes back with those keys plus `diar_segments`, while `audio_filepath` remains absent. Declaring `filepath_key` as an unconditional write lets the planner advance a key runtime does not produce and approve a file-only downstream consumer. Remove the input filepath from produced writes (it is merely preserved when already present); keep only `diar_segments` and the opt-in speaker count, and add exact conformance/planner-to-runtime tests for file and waveform non-fan-out configurations.

shubhamNvidia (commented):
reviewed

shubhamNvidia (commented):
reviewed

PR #2338, opened 2026-08-25
feat(audio-segmentation): in-memory segment handoff + agent-ready contracts

shubhamNvidia:
**Contracts:** `describe()` + `*_key` params on speaker-separation and VAD segmentation, with honest cardinality (these produce a nested per-segment list / fan out).

**New functionality**
- **In-memory handoff** (`keep_waveform_in_task`, `keep_segment_waveform_in_task`) — hand the cut waveforms straight to the next stage instead of writing every segment to disk and re-reading it. On a diarize→ASR chain this removes a full write+read of the corpus.
- **Explicit disk control** (`write_to_disk`, `separated_audio_dir`) — writing separated audio is now opt-in and its destination is configurable, rather than implied.

### DIFF
--- a/nemo_curator/stages/audio/segmentation/speaker_separation.py
+++ b/nemo_curator/stages/audio/segmentation/speaker_separation.py
@@ -30,6 +30,7 @@ Example:
     )
 """
 
+import os
 from dataclasses import dataclass, field
 from typing import Any
 
@@ -44,7 +45,13 @@ except ImportError:
     SortformerEncLabelModel = None
 
 from nemo_curator.backends.base import WorkerMetadata
-from nemo_curator.stages.audio.common import resolve_waveform_from_item
+from nemo_curator.stages.audio._agent._agent_ready import AgentReady, Gates, IOSpec, StageContract
+from nemo_curator.stages.audio._agent._residency import (
+    InputResidency,
+    accepts_for_residency,
+    resolve_audio,
+    write_audio_stable,
+)
 from nemo_curator.stages.audio.segmentation.speaker_separation_module.speaker_sep import SpeakerSeparator
 from nemo_curator.stages.base import ProcessingStage
 from nemo_curator.stages.resources import Resources
@@ -61,7 +68,7 @@ def _pydub_to_waveform_sr(seg: AudioSegment) -> tuple[torch.Tensor, int]:
 
 
 @dataclass
-class SpeakerSeparationStage(ProcessingStage[AudioTask, AudioTask]):
+class SpeakerSeparationStage(AgentReady, ProcessingStage[AudioTask, AudioTask]):
     """
     Speaker separation stage using NeMo SortFormer diarization model.
 
@@ -75,8 +82,28 @@ class SpeakerSeparationStage(ProcessingStage[AudioTask, AudioTask]):
         min_duration: Minimum segment duration in seconds
         gap_threshold: Gap threshold for merging speaker segments
         buffer_time: Buffer time around speaker segments
+        audio_filepath_key: Key in data dict for the input audio file path.
+        waveform_key: Key in data dict for the in-memory waveform tensor.
+        sample_rate_key: Key in data dict for the waveform sample rate.
+        speaker_id_key: Key where each child task's speaker id is written.
+        num_speakers_key: Key where the detected speaker count is written.
+        duration_key: Key where each child's speech duration in seconds is written.
+        diar_segments_key: Key where each child's diarization segments are written.
+        input_residency: Which input to use — "waveform" (in-memory only), "file"
+            (audio_filepath only), or "auto" (waveform first, file fallback; default).
+        keep_waveform_in_task: Keep each per-speaker waveform in the task (default True,
+            today's behavior). Set False to emit only on-disk paths (requires write_to_disk).
+        write_to_disk: Also write each per-speaker track to ``separated_audio_dir`` and set
+            ``audio_filepath_key`` to it, so file-based downstream stages can consume it.
+            Defaults to False (in-memory only, unchanged).
+        separated_audio_dir: Directory for per-speaker WAVs (required when write_to_disk=True).
 
     Note:
+        By default (write_to_disk=False) per-speaker child tasks DROP the parent's
+        audio_filepath (it points at the full multi-speaker file) and carry
+        ``original_file`` for provenance; downstream consumes the per-speaker waveform.
+        With write_to_disk=True, each child instead gets its own audio_filepath.
+
         GPU assignment is handled by the executor via _resources.
         Use .with_(resources=Resources(gpus=X)) to configure GPU allocation.
     """
@@ -86,6 +113,18 @@ class SpeakerSeparationStage(ProcessingStage[AudioTask, AudioTask]):
     min_duration: float = 0.8
     gap_threshold: float = 0.1
     buffer_time: float = 0.5
+    audio_filepath_key: str = "audio_filepath"
+    waveform_key: str = "waveform"
+    sample_rate_key: str = "sample_rate"
+    speaker_id_key: str = "speaker_id"
+    num_speakers_key: str = "num_speakers"
+    duration_key: str = "duration"
+    diar_segments_key: str = "diar_segments"
+    input_residency: InputResidency = "auto"
+    # Output residency (both default to today's behavior: in-memory waveform only, no disk).
+    keep_waveform_in_task: bool = True
+    write_to_disk: bool = False
+    separated_audio_dir: str | None = None
 
     name: str = "SpeakerSeparation"
     batch_size: int = 1
@@ -94,12 +133,64 @@ class SpeakerSeparationStage(ProcessingStage[AudioTask, AudioTask]):
     def __post_init__(self):
         super().__init__()
         self._separator = None
+        if not (self.keep_waveform_in_task or self.write_to_disk):
+            msg = "At least one of keep_waveform_in_task or write_to_disk must be True"
+            raise ValueError(msg)
+        if self.write_to_disk and not self.separated_audio_dir:
+            msg = "separated_audio_dir is required when write_to_disk=True"
+            raise ValueError(msg)
 
     def inputs(self) -> tuple[list[str], list[str]]:
         return [], []
 
     def outputs(self) -> tuple[list[str], list[str]]:
-        return [], ["waveform", "sample_rate", "speaker_id", "num_speakers", "duration"]
+        outs: list[str] = []
+        if self.keep_waveform_in_task:
+            outs.extend([self.waveform_key, self.sample_rate_key])
+        outs.extend([self.speaker_id_key, self.num_speakers_key, self.duration_key, self.diar_segments_key])
+        if self.write_to_disk:
+            outs.append(self.audio_filepath_key)
+        return [], outs
+
+    def describe(self) -> StageContract:
+        forms = accepts_for_residency(self.input_residency)
+        reads_one_of = []
+        if "waveform" in forms:
+            reads_one_of.append(IOSpec(data_keys=[self.waveform_key, self.sample_rate_key], accepts=["waveform"]))
+        if "file" in forms:
+            reads_one_of.append(IOSpec(data_keys=[self.audio_filepath_key], accepts=["file"]))
+        writes: list[str] = []
+        produces: list[str] = []
+        if self.keep_waveform_in_task:
+            writes.extend([self.waveform_key, self.sample_rate_key])
+            produces.append("tensor")
+        writes.extend(
+            [self.speaker_id_key, self.num_speakers_key, self.duration_key, self.diar_segments_key, "original_file"]
+        )
+        if self.write_to_disk:
+            writes.append(self.audio_filepath_key)
+            produces.append("disk")
+        return StageContract(
+            reads_one_of=reads_one_of,
+            writes=IOSpec(data_keys=writes, produces=produces),
+            # children drop the parent's audio_filepath (and blob keys)
+            preserves_upstream_keys=False,
+            cardinality="1:N fan-out",
+            # One child per detected speaker; speaker_id is the per-child key that
+            # identifies which slice of the iteration a child is (role-resolvable).
+            iteration_key=self.speaker_id_key,
+            gates=Gates(
+                requires_gpu=self.resources.requires_gpu,
+                requires_internet_first_run=True,
+                writes_to_disk=self.write_to_disk,
+                output_path_params=["separated_audio_dir"],
+                # Diarization runs on one file's audio, and ``num_speakers`` counts the speakers
+                # found in THAT file. Unlike SplitLongAudioStage, a shared output directory is
+                # still safe here: ``write_audio_stable`` names each per-speaker WAV after a
+                # digest of its own samples rather than after the source basename.
+                per_row_independent=True,
+            ),
+        )
 
     def setup_on_node(self, _node_info: Any = None, _worker_metadata: Any = None) -> None:  # noqa: ANN401
         try:
@@ -160,6 +251,17 @@ class SpeakerSeparationStage(ProcessingStage[AudioTask, AudioTask]):
     # each child gets its own duration from the diarization result.
     _INHERITED_DROP_KEYS = frozenset({"audio", "waveform", "duration", "num_samples"})
 
+    def _write_speaker_wav(self, waveform: torch.Tensor, sr: int, original_file: str, speaker_id: str) -> str:
+        """Write one per-speaker waveform to ``separated_audio_dir`` and return the path."""
+        stem = os.path.splitext(os.path.basename(str(original_file)))[0] or "audio"
+        return write_audio_stable(
+            waveform,
+            sr,
+            output_dir=self.separated_audio_dir,
+            stem=stem,
+            tag=str(speaker_id),
+        )
+
     def _build_speaker_tasks(
         self,
         speaker_audio_data: dict,
@@ -174,21 +276,46 @@ class SpeakerSeparationStage(ProcessingStage[AudioTask, AudioTask]):
                 logger.debug(f"Skipping {speaker_id}: duration {result.duration:.2f}s < {self.min_duration}s")
                 continue
             spk_waveform, spk_sr = _pydub_to_waveform_sr(result.audio)
-            speaker_data = {
-                **{k: v for k, v in item.items() if k not in self._INHERITED_DROP_KEYS},
-                "waveform": spk_waveform,
-                "sample_rate": spk_sr,
-                "speaker_id": speaker_id,
-                "num_speakers": num_speakers,
-                "duration": result.duration,
-                "diar_segments": result.diar_segments,
+            # Drop the parent's file path(s) too: they point at the FULL
+            # multi-speaker file, so a file-preferring downstream stage would
+            # process the whole file per speaker instead of this speaker's
+            # extracted waveform. With the path gone, downstream resolves the
+            # per-speaker waveform (input_residency="auto") instead.
+            drop_keys = {
+                *self._INHERITED_DROP_KEYS,
+                self.waveform_key,
+                self.duration_key,
+                self.audio_filepath_key,
+                "audio_filepath",
             }
+            speaker_data = {
+                **{k: v for k, v in item.items() if k not in drop_keys},
+                self.speaker_id_key: speaker_id,
+                self.num_speakers_key: num_speakers,
+                self.duration_key: result.duration,
+                self.diar_segments_key: result.diar_segments,
+                # Source identity must survive the audio_filepath drop above —
+                # TimestampMapper (and any provenance consumer) reads original_file.
+                "original_file": item.get("original_file")
+                or item.get(self.audio_filepath_key)
+                or item.get("audio_filepath")
+                or "unknown",
+            }
+            # Output residency: keep the in-memory waveform (default) and/or persist a
+            # per-speaker WAV and point audio_filepath at it (opt-in write_to_disk).
+            if self.keep_waveform_in_task:
+                speaker_data[self.waveform_key] = spk_waveform
+                speaker_data[self.sample_rate_key] = spk_sr
+            if self.write_to_disk:
+                speaker_data[self.audio_filepath_key] = self._write_speaker_wav(
+                    spk_waveform, spk_sr, speaker_data["original_file"], speaker_id
+                )
             spk_task = AudioTask(
                 data=speaker_data,
                 dataset_name=task.dataset_name,
+                _metadata=dict(task._metadata or {}),
+                _stage_perf=list(task._stage_perf),
             )
-            if task._metadata:
-                spk_task._metadata = dict(task._metadata)
             results.append(spk_task)
         return results
 
@@ -208,7 +335,13 @@ class SpeakerSeparationStage(ProcessingStage[AudioTask, AudioTask]):
         results: list[AudioTask] = []
 
         try:
-            audio_result = resolve_waveform_from_item(item, task.task_id)
+            audio_result = resolve_audio(
+                item,
+                residency=self.input_residency,  # type: ignore[arg-type]
+                audio_filepath_key=self.audio_filepath_key,
+                waveform_key=self.waveform_key,
+                sample_rate_key=self.sample_rate_key,
+            )
             if audio_result is None:
                 return []
             waveform, sample_rate = audio_result
--- a/nemo_curator/stages/audio/segmentation/vad_segmentation.py
+++ b/nemo_curator/stages/audio/segmentation/vad_segmentation.py
@@ -47,7 +47,8 @@ from loguru import logger
 from silero_vad import get_speech_timestamps, load_silero_vad
 
 from nemo_curator.backends.base import WorkerMetadata
-from nemo_curator.stages.audio.common import ensure_waveform_2d, load_audio_file
+from nemo_curator.stages.audio._agent._agent_ready import AgentReady, Gates, IOSpec, StageContract
+from nemo_curator.stages.audio._agent._residency import InputResidency, accepts_for_residency, resolve_audio
 from nemo_curator.stages.base import ProcessingStage
 from nemo_curator.stages.resources import Resources
 from nemo_curator.tasks import AudioTask
@@ -57,7 +58,7 @@ SILERO_TARGET_RATE = 16000
 
 
 @dataclass
-class VADSegmentationStage(ProcessingStage[AudioTask, AudioTask]):
+class VADSegmentationStage(AgentReady, ProcessingStage[AudioTask, AudioTask]):
     """
     Stage to segment audio using Voice Activity Detection (VAD).
 
@@ -75,6 +76,20 @@ class VADSegmentationStage(ProcessingStage[AudioTask, AudioTask]):
         speech_pad_ms: Padding in ms to add before/after speech segments.
         waveform_key: Key to get waveform data.
         sample_rate_key: Key to get sample rate.
+        audio_filepath_key: Key in data dict for the input audio file path.
+        segments_key: Key where the nested segments list is written (nested=True).
+        start_ms_key: Key where each segment's start time in milliseconds is written.
+        end_ms_key: Key where each segment's end time in milliseconds is written.
+        segment_num_key: Key where each segment's index is written.
+        duration_key: Key where each segment's duration in seconds is written.
+        original_file_key: Key carrying the source file path for provenance.
+        nested: If True, return one task with all segment dicts under segments_key
+            instead of fanning out one task per segment (default False).
+        input_residency: Which input to use — "waveform" (in-memory only), "file"
+            (audio_filepath only), or "auto" (waveform first, file fallback; default).
+        keep_segment_waveform_in_task: If True (default), store each segment's waveform
+            in the segment item. If False, nested segments are metadata-only — waveform
+            consumers such as SegmentConcatenation will skip them.
 
     Note:
         Default resources: cpus=1.0, gpus=0.0 (CPU). Silero VAD is lightweight.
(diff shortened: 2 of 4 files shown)


nemo_curator/stages/audio/segmentation/speaker_separation.py
```diff
@@ -208,7 +335,13 @@ def process(self, task: AudioTask) -> list[AudioTask]:
         results: list[AudioTask] = []
 
         try:
-            audio_result = resolve_waveform_from_item(item, task.task_id)
+            audio_result = resolve_audio(
```
mohammadaaftabv, line 338: nemo_curator/stages/audio/segmentation/speaker_separation.py:338-347
replaces the main-branch resolve_waveform_from_item(..., mono=True) path with
resolve_audio(). 

The shared resolver applies mono=True only when loading a
file; it returns a resident tensor without downmixing. I passed a valid
two-channel (2, 3200) waveform: main's resolver produced (1, 3200), while
this branch passed (2, 3200) into SpeakerSeparator. Its packaging path then
hands the channel-first array to SoundFile as frames-first, raises
LibsndfileError: Format not recognised, and the broad catch turns that into a
successful empty fan-out. Every stereo waveform-resident row is therefore
silently lost.

The same substitution changes another supported default case: when a task has
a waveform and file path but no sample-rate scalar, main reads only the file
header and keeps the supplied waveform; this PR falls back to loading and
processing the file's different samples. Preserve both legacy behaviors by
making resolve_audio(..., mono=True) downmix resident tensors and retaining a
header-only sample-rate fallback (or normalize and perform that fallback in
this stage). Add model-free stereo and waveform-plus-path-without-rate
regressions which assert the exact tensor handed to the separator.

nemo_curator/stages/audio/segmentation/speaker_separation.py
mohammadaaftabv, line 362: nemo_curator/stages/audio/segmentation/speaker_separation.py:362-375 puts the
new _write_speaker_wav() operation under the existing catch-all processing
handler. Making that write raise OSError("simulated disk full") on a valid
detected speaker logged the exception but returned []. 

A full/unavailable
output volume can thus erase every source row while the executor sees no
failure; if a later speaker fails, earlier files for the same parent can also
remain as an incomplete result set.

Keep an explicit skip policy for model/inference failures if that is intended,
but let persistence errors escape. For multi-speaker writes, stage all outputs
and either publish the complete set or clean only files created by the failed
attempt. Add single- and multi-speaker write-failure tests that require the
stage/executor to fail rather than return an empty list.

nemo_curator/stages/audio/segmentation/vad_segmentation.py
mohammadaaftabv, line 241: nemo_curator/stages/audio/segmentation/vad_segmentation.py:241-273 slices a
segment waveform but copies every unrelated parent field, including
audio_filepath, into the child. 

A 0.25-second child in the focused repro held
a (2, 4000) segment tensor and the original one-second path; a second stage
configured with input_residency="file" reopened (1, 16000) samples and
processed the complete recording once for that child. The planner returned
ok=True, keys_ok=True, with no issue. 

This is precisely the stale-path case
that the same PR explicitly prevents for speaker-separation children at
speaker_separation.py:279-290.

The other output setting is also unsafe: with
keep_segment_waveform_in_task=False, the builder drops the waveform (and
always drops num_samples) while describe() still claims upstream keys are
preserved. The planner approved a waveform-resident consumer; at runtime every
child lacked the waveform and the consumer returned [].

Remove the parent audio_filepath_key from every segment child and retain it
only as original_file_key. Declare the configured waveform/path/num-sample
removals so file- or waveform-only consumers are rejected when no corresponding
segment carrier exists; alternatively add a segment-file output mode and point
the child at that actual file. Recompute num_samples if it is meant to remain.
Add planner-to-runtime tests for file and waveform consumers with both keep
settings.

nemo_curator/stages/audio/segmentation/vad_segmentation.py
```diff
@@ -247,23 +314,23 @@ def process(self, task: AudioTask) -> AudioTask | list[AudioTask]:
             if not segments:
                 logger.warning("No speech segments detected by VAD")
                 if self.nested:
-                    task.data["segments"] = []
+                    task.data[self.segments_key] = []
                     return task
```
mohammadaaftabv, line 318: nemo_curator/stages/audio/segmentation/vad_segmentation.py:314-319 returns the
nested task immediately when VAD finds no speech, before the waveform removal
at line 333. 

With nested=True and
keep_segment_waveform_in_task=False, the contract declares the waveform
removed and produces no tensor, so a manifest sink validates. Runtime output is
{"segments": [], "sample_rate": ..., "waveform": Tensor(...)} and JSON
serialization raises TypeError: Object of type Tensor is not JSON serializable.

Apply the nested-mode cleanup before both the empty and populated returns (or
through one shared finalizer), then add the required assert_agent_ready
fixture for the no-speech branch and an end-to-end serializable-sink regression.

nemo_curator/stages/audio/segmentation/speaker_separation.py
```diff
@@ -174,21 +276,46 @@ def _build_speaker_tasks(
...
+            }
             speaker_data = {
-                **{k: v for k, v in item.items() if k not in self._INHERITED_DROP_KEYS},
-                "waveform": spk_waveform,
-                "sample_rate": spk_sr,
-                "speaker_id": speaker_id,
-                "num_speakers": num_speakers,
-                "duration": result.duration,
-                "diar_segments": result.diar_segments,
+                **{k: v for k, v in item.items() if k not in drop_keys},
```
mohammadaaftabv, line 292: nemo_curator/stages/audio/segmentation/speaker_separation.py:291-311 copies
the parent's sample-rate key and replaces it with spk_sr only when the
waveform is retained. 

In disk-only mode, a real 8-kHz input produced a written
speaker WAV with an 11,025-Hz header (the current PyDub packaging rate), while
the child and final-serialization path still reported sample_rate=8000.
Downstream timing derived from that scalar is wrong even though the file itself
is readable.

Always set self.sample_rate_key to spk_sr and declare it for disk output, or
remove it together with the resident waveform and require file consumers to
read the header. Add a non-default-rate test comparing child metadata with
soundfile.info(child[audio_filepath_key]).samplerate.

nemo_curator/stages/audio/segmentation/speaker_separation.py
```diff
@@ -86,6 +113,18 @@ class SpeakerSeparationStage(ProcessingStage[AudioTask, AudioTask]):
     min_duration: float = 0.8
     gap_threshold: float = 0.1
     buffer_time: float = 0.5
+    audio_filepath_key: str = "audio_filepath"
```
mohammadaaftabv, line 116: nemo_curator/stages/audio/segmentation/speaker_separation.py:116-131 inserts
all new fields before the existing name, batch_size, and resources
parameters. Likewise,
nemo_curator/stages/audio/segmentation/vad_segmentation.py:104-119 inserts
audio_filepath_key and the new output keys before the existing
waveform_key, sample_rate_key, nested, name, batch_size, and
resources fields. 

Existing positional calls still construct, but bind values
to unrelated fields. For example, the old VAD tail
("samples", "hz", True) now yields
audio_filepath_key="samples", waveform_key="hz", sample_rate_key=True, nested=False; an old speaker name becomes its path key.

Preserve the old field order and append the additions after every legacy field,
or mark only the newly added fields keyword-only. Pin both old signatures in
compatibility tests so this cannot silently recur.

nemo_curator/stages/audio/segmentation/vad_segmentation.py
```diff
@@ -98,12 +122,59 @@ def __post_init__(self):
...
+            self.end_ms_key,
+            self.segment_num_key,
+            self.duration_key,
+            self.original_file_key,  # _build_segment_item always writes it
+        ]
+        produces = []
+        if self.keep_segment_waveform_in_task:
+            writes.append(self.waveform_key)
+            produces.append("tensor")
+        if self.nested:
```
mohammadaaftabv, line 157: nemo_curator/stages/audio/segmentation/vad_segmentation.py:157-173 reduces
nested writes to only segments_key and leaves
writes.segment_data_keys=[], even though each segments[] item contains
sample rate, bounds, index, duration, provenance, and optionally waveform. 

It also always sets iteration_key=self.segments_key; fan-out children do not
contain that key and are distinguished by segment_num_key. The rendered
corpus shows that this exact iteration-key defect was raised on the closed
predecessor PR #2177 and remains unchanged here.

This is not only missing metadata: the planner rejects a safe
nested-VAD -> segment-concatenation -> disk-only conversion -> manifest chain
as tensor_into_sink because the tensor has no segment-level carrier it can
trace and remove. Build the nested IOSpec with
data_keys=[segments_key] plus the actual segment_data_keys, and choose
segments_key only for nested mode versus segment_num_key for fan-out. Add
configured assert_agent_ready(..., segments_key=...) and full-chain planner
regressions for both waveform settings.

nemo_curator/stages/audio/segmentation/speaker_separation.py
```diff
@@ -94,12 +133,64 @@ class SpeakerSeparationStage(ProcessingStage[AudioTask, AudioTask]):
...
+            reads_one_of.append(IOSpec(data_keys=[self.waveform_key, self.sample_rate_key], accepts=["waveform"]))
+        if "file" in forms:
+            reads_one_of.append(IOSpec(data_keys=[self.audio_filepath_key], accepts=["file"]))
+        writes: list[str] = []
+        produces: list[str] = []
+        if self.keep_waveform_in_task:
+            writes.extend([self.waveform_key, self.sample_rate_key])
+            produces.append("tensor")
+        writes.extend(
+            [self.speaker_id_key, self.num_speakers_key, self.duration_key, self.diar_segments_key, "original_file"]
```
mohammadaaftabv, line 168: nemo_curator/stages/audio/segmentation/speaker_separation.py:168-177
hard-codes "original_file" and declares
preserves_upstream_keys=False, although _build_speaker_tasks() at lines
284-303 copies almost all upstream metadata. 

The result is broken in both
directions: the planner rejects a downstream text consumer even though every
runtime child retains text, while remapping upstream provenance (for example
to source_path) yields the advertised original_file="unknown". The
hard-coded provenance defect was also reported on closed predecessor PR #2177.

Add original_file_key and use it consistently in the contract, lookup, child
assignment, and disk filename call. Model the real behavior as preserving
upstream keys plus exact conditional removals/replacements for parent audio
carriers instead of discarding the whole upstream schema. Add key-remapping and
speaker -> scalar-metadata-consumer planner/runtime tests.

tests/stages/audio/segmentation/test_speaker_separation.py
```diff
@@ -318,3 +371,57 @@ def test_empty_input(self) -> None:
...
+    channels = 1
+    frame_rate = 16000
+
+    def get_array_of_samples(self) -> list[int]:
+        return [0, 500, -500, 0] * 100
+
+
+# Lifted from tests/stages/audio/test_agent_simulation_pipelines.py: it drives only
+# SpeakerSeparationStage, and was the sole coverage of fan-out metadata isolation.
+def test_agent_fanout_children_have_isolated_metadata() -> None:
```
mohammadaaftabv, line 389: tests/stages/audio/segmentation/test_speaker_separation.py:389-427 labels its
metadata-isolation check as an agent test, but neither changed test file calls
the foundation's mandatory assert_agent_ready(...) harness. PR #2332's
AGENT_READY.md explicitly requires one per stage and requires segment-level
writes and externally consumed output semantics to be documented. The missing
empty-speech fixture would have failed immediately on the undeleted waveform,
and the current public static contract also reports
SpeakerSeparationStage.requires_internet_first_run=False even though its
configured contract and setup download SortFormer.

Add model-free conformance fixtures covering file/waveform/auto, nested/fan-out,
empty/nonempty, disk/memory, and remapped keys; add an AGENT_STATIC internet
gate test; and land the capability-card entry that explains, in particular,
that each child's copied num_speakers describes the parent recording rather
than re-evaluating the separated child.

mohammadaaftabv (changes requested):

nemo_curator/stages/audio/segmentation/speaker_separation.py
shubhamNvidia: Thanks for catching this. Resident stereo tensors are now downmixed before separation, and when the sample rate is missing we read only the file header while preserving the supplied waveform. Model-free regressions verify the exact tensor passed to the separator.

nemo_curator/stages/audio/segmentation/speaker_separation.py
shubhamNvidia: Fixed. Model/inference failures retain the existing skip behavior, while persistence errors now propagate. Multi-speaker outputs are staged and published atomically, with failure cleanup limited to files created by the current attempt. Single- and multi-speaker failure tests were added.

nemo_curator/stages/audio/segmentation/vad_segmentation.py
shubhamNvidia: Fixed. Fan-out children no longer retain the parent recording path or stale num_samples. The contract now declares the exact carrier removals for each output mode, and planner/runtime tests cover file and waveform consumers with both waveform-retention settings.

nemo_curator/stages/audio/segmentation/vad_segmentation.py
shubhamNvidia: Fixed. Nested output now uses a shared finalization path, so the top-level waveform is removed for both empty and populated results when segment waveforms are not retained. Empty-speech conformance and serialization regressions were added.

nemo_curator/stages/audio/segmentation/speaker_separation.py
shubhamNvidia: Fixed. Speaker children now always receive the separated audio’s actual sample rate, including disk-only mode. A regression verifies that child metadata matches the written WAV header.

nemo_curator/stages/audio/segmentation/speaker_separation.py
shubhamNvidia: Fixed. All newly introduced configuration fields are keyword-only, preserving the legacy positional order for both segmentation stages. Compatibility tests pin the previous constructor behavior.

nemo_curator/stages/audio/segmentation/vad_segmentation.py
shubhamNvidia: Fixed. Nested mode now declares the actual segment_data_keys, while the iteration key is segments_key for nested mode and segment_num_key for fan-out. Conformance and full-chain planner tests cover configured keys and both waveform modes.

nemo_curator/stages/audio/segmentation/speaker_separation.py
shubhamNvidia: Fixed. original_file_key is configurable and used consistently for lookup, child metadata, contracts, and persisted filenames. The contract now preserves upstream metadata while declaring the exact audio-carrier replacements/removals. Remapped-key and downstream metadata-consumer tests were added.

tests/stages/audio/segmentation/test_speaker_separation.py
shubhamNvidia: Fixed. Model-free assert_agent_ready coverage now exercises residency modes, nested/fan-out behavior, empty/non-empty results, disk/memory outputs, and remapped keys. Speaker separation also advertises its first-run network requirement. The capability card documents that num_speakers is the parent-recording aggregate copied to each child.

nemo_curator/stages/audio/segmentation/speaker_separation.py
```diff
@@ -160,38 +288,160 @@ def _initialize_separator(self) -> None:
...
+                    waveform,
+                    sample_rate,
+                    speaker_data[self.original_file_key],
+                    speaker_id,
+                    output_dir=staging_dir,
+                )
+                final_path = os.path.join(output_dir, os.path.basename(staged_path))
+                staged.append((speaker_data, staged_path, final_path))
+
+            for speaker_data, staged_path, final_path in staged:
```
mohammadaaftabv, line 359: P0 — concurrent rollback deletes another successful task's stable WAV.

speaker_separation.py:332-345. A deterministic two-thread probe made task B
 return two paths successfully, after which task A's rollback deleted B's
 first path. Retain immutable published files or use an ownership/atomic
 directory protocol, and add a concurrent duplicate-row regression.

nemo_curator/stages/audio/segmentation/speaker_separation.py
```diff
@@ -91,15 +124,110 @@ class SpeakerSeparationStage(ProcessingStage[AudioTask, AudioTask]):
...
+    diar_segments_key: str = "diar_segments"
+    original_file_key: str = "original_file"
+    input_residency: InputResidency = "auto"
+    # Output residency (both default to today's behavior: in-memory waveform only, no disk).
+    keep_waveform_in_task: bool = True
+    write_to_disk: bool = False
+    separated_audio_dir: str | None = None
+
+    AGENT_STATIC: ClassVar[StaticHints] = StaticHints(gates=Gates(requires_internet_first_run=True))
+
```
mohammadaaftabv, line 145: P1 — configurable key collisions and residency typos are accepted.

speaker_separation.py:144-152; vad_segmentation.py:123-132. 

Reproductions
 erase VAD's nested segment list/start boundary and speaker identity. Invalid
 residency silently becomes auto. 
Call PR2332's validation helpers with
 stage-specific distinct/nonempty key sets and test the rejected cases.

nemo_curator/stages/audio/segmentation/vad_segmentation.py
```diff
@@ -176,54 +263,63 @@ def _build_segment_item(
...
+                self.end_ms_key,
+                self.segment_num_key,
+                self.duration_key,
                 "num_samples",
-            )
+            }
         }
+        if not self.keep_segment_waveform_in_task:
+            segment_waveform = None
         segment_data.update(
```
mohammadaaftabv, line 333: nemo_curator/stages/audio/segmentation/vad_segmentation.py:282-299

P1 — Preserve the established output schema for the existing VAD key parameters. waveform_key and sample_rate_key already existed and selected input lookup keys; main always emitted children under canonical waveform and sample_rate. At this head, lines 282-299 reuse those input settings as output names. The same old VADSegmentationStage(waveform_key="samples", sample_rate_key="hz") call changes from a child containing waveform/sample_rate to one containing only samples/hz, so existing downstream stages lose their audio input. An old input name such as waveform_key="start_ms" now also lets the tensor overwrite the numeric boundary. Add separate output-key settings defaulted to the legacy canonical names (or put output remapping behind an explicit opt-in), retain the legacy outputs() surface/order for old configurations, and add base-versus-head custom-key and collision regressions.

nemo_curator/stages/audio/segmentation/vad_segmentation.py
```diff
@@ -176,54 +263,63 @@ def _build_segment_item(
             k: v
             for k, v in item.items()
             if k
```
mohammadaaftabv, line 316: P1 — Default VAD fan-out no longer preserves the legacy path/provenance schema. 

Main copies audio_filepath into each fan-out child and nested segment, but lines 262-278 now delete it unconditionally and only write original_file. 

A default legacy manifest/exporter therefore loses an established field. The stale full-recording path is unsafe as a segment carrier, but strict compatibility requires retaining it by default as legacy provenance and making removal or segment-file replacement an explicit new mode; the contract should distinguish provenance from a consumable segment path. Also avoid the truthiness chain at lines 289-294: main preserves an explicitly present falsey original_file, whereas this code replaces it. Add differential default fan-out, nested, and falsey-provenance tests.

nemo_curator/stages/audio/segmentation/speaker_separation.py
```diff
@@ -160,38 +288,160 @@ def _initialize_separator(self) -> None:
...
             if result.duration < self.min_duration:
                 logger.debug(f"Skipping {speaker_id}: duration {result.duration:.2f}s < {self.min_duration}s")
                 continue
             spk_waveform, spk_sr = _pydub_to_waveform_sr(result.audio)
+            # Drop the parent's file path(s) too: they point at the FULL
+            # multi-speaker file, so a file-preferring downstream stage would
+            # process the whole file per speaker instead of this speaker's
+            # extracted waveform. With the path gone, downstream resolves the
+            # per-speaker waveform (input_residency="auto") instead.
+            drop_keys = {
```
mohammadaaftabv, line 382: P1 — Default speaker fan-out deletes a previously preserved field. Main's _build_speaker_tasks() carries the parent audio_filepath into each child; lines 369-385 now remove it whenever the default write_to_disk=False is used and expose only a renamed original_file. 

That breaks existing manifests and provenance consumers even though the child waveform remains available. Keep legacy path preservation as the default compatibility behavior and make stale-path removal an explicit new mode (with the contract treating the retained value as provenance, not separated audio); disk mode may replace it with the actual child path. Preserve explicitly present falsey original_file values instead of replacing them through the or chain, and add a base-versus-head default-row regression.

nemo_curator/stages/audio/segmentation/vad_segmentation.py
```diff
@@ -176,54 +263,63 @@ def _build_segment_item(
...
+                    or item.get("audio_filepath")
+                    or "unknown"
+                ),
             }
         )
+        if segment_waveform is not None:
+            segment_data[self.waveform_key] = segment_waveform
         return segment_data
 
+    def _finalize_nested(self, task: AudioTask, segments: list[dict[str, Any]]) -> AudioTask:
```
mohammadaaftabv, line 353: P1 — The default resolver/finalizer changes three established VAD branches. Main skips a row whose resident waveform lacks sample_rate, even when a path is present; this head's auto path loads and processes that different file instead. Main also retains the original waveform for nested=True with no speech, whereas _finalize_nested() now deletes it even with keep_segment_waveform_in_task=True, and main's file loader leaves the discovered sample rate on a nested task while the shared resolver does not. In a model-free differential run, base never invoked inference for a (1, 3200) resident tensor without a rate, while this head inferred on an unrelated (1, 1600) file. 

Preserve the legacy branches for the default configuration and put the new fallback/cleanup semantics behind explicit controls; add divergent resident-versus-file, empty nested, and file-only nested regressions.

nemo_curator/stages/audio/segmentation/speaker_separation.py
```diff
@@ -239,10 +488,13 @@ def process(self, task: AudioTask) -> list[AudioTask]:
...
             logger.exception(f"[SpeakerSeparation] Failed to process task {task.task_id}: {e}")
+            return []
         finally:
             if waveform is not None:
                 del waveform
                 if torch.cuda.is_available():
                     torch.cuda.empty_cache()
 
-        return results
+        # Packaging and persistence are intentionally outside the inference catch:
```
mohammadaaftabv: P2 — Moving all packaging outside the catch changes the established error policy. Main invokes _build_speaker_tasks() inside the broad handler, so malformed separator audio or conversion failures log and return []; lines 498-500 now let those legacy failures escape together with newly introduced disk-write failures. A deterministic malformed-audio result returned zero children on base and raises AttributeError here. Keep persistence failures fail-fast, but split persistence from legacy in-memory conversion/packaging so the default path retains its skip behavior. Add separate regressions for a default packaging exception and a disk-persistence exception.

mohammadaaftabv (changes requested):

nemo_curator/stages/audio/segmentation/speaker_separation.py
shubhamNvidia: Thanks for catching this. Published speaker WAVs are now treated as immutable and are never removed during rollback, preventing one concurrent task from deleting another task’s successful output. Cleanup is limited to temporary files owned by the failing attempt. A deterministic two-thread duplicate-row regression was added.

nemo_curator/stages/audio/segmentation/speaker_separation.py
shubhamNvidia: Fixed. Both stages now use the shared validation helpers for residency values and stage-specific key constraints. Invalid residency values and unsafe empty or colliding keys are rejected early, with regression coverage for each case.

nemo_curator/stages/audio/segmentation/vad_segmentation.py
shubhamNvidia: Agreed. Input lookup keys are now separate from output keys. The new output settings default to the legacy canonical `waveform` and `sample_rate` names, so existing custom-input configurations retain their original downstream schema. Compatibility and collision regressions were added.

nemo_curator/stages/audio/segmentation/vad_segmentation.py
shubhamNvidia: Fixed. VAD children continue to retain `audio_filepath` as legacy provenance, while the contract marks it invalid as a consumable segment-audio path. Explicitly present falsey provenance values are also preserved. Differential tests cover default fan-out, nested output, and falsey provenance.

nemo_curator/stages/audio/segmentation/speaker_separation.py
shubhamNvidia: Fixed. In-memory speaker children retain the parent `audio_filepath` for backward-compatible provenance, while the contract prevents downstream stages from treating it as separated audio. Disk mode still replaces it with the actual speaker file path, and falsey provenance values are preserved. A base-versus-head regression was added.

nemo_curator/stages/audio/segmentation/vad_segmentation.py
shubhamNvidia: Fixed. The default behavior now matches the established implementation: resident audio without a sample rate is skipped rather than replaced from disk, empty nested output retains the waveform when requested, and file-only nested input keeps the discovered sample rate. Regressions cover all three branches.

nemo_curator/stages/audio/segmentation/speaker_separation.py
shubhamNvidia: Fixed. Legacy inference, conversion, and in-memory packaging failures continue to log and return an empty result, while disk-persistence failures propagate to the executor. Separate regressions verify both behaviors.

nemo_curator/stages/audio/segmentation/speaker_separation.py
```diff
@@ -160,38 +316,150 @@ def _initialize_separator(self) -> None:
...
+        self,
+        waveform: torch.Tensor,
+        sr: int,
+        original_file: str,
+        speaker_id: str,
+        *,
+        output_dir: str | None = None,
+    ) -> str:
+        """Write one per-speaker waveform to ``separated_audio_dir`` and return the path."""
+        stem = os.path.splitext(os.path.basename(str(original_file)))[0] or "audio"
```
shubhamNvidia: **[P1] Bound the source stem before building the stable filename**

This newly persisted filename keeps the full source stem and then appends the speaker tag, digest, and extension. A source named with a valid 240-character stem reproduces `OSError: [Errno 36] File name too long` when `_write_speaker_wav()` tries to publish `<stem>_spk_0_<digest>.wav`; the input component is valid, but the derived component is 267 bytes. This makes `write_to_disk=True` fail for otherwise valid recordings. Sanitize and truncate the human-readable stem to a budget that includes the suffix (the digest already supplies uniqueness), and add a near-`NAME_MAX` regression.

nemo_curator/stages/audio/segmentation/vad_segmentation.py
```diff
@@ -94,16 +117,131 @@ class VADSegmentationStage(ProcessingStage[AudioTask, AudioTask]):
...
+            input_keys={
+                "audio_filepath_key": self.audio_filepath_key,
+                "waveform_key": self.waveform_key,
+                "sample_rate_key": self.sample_rate_key,
+            },
+            output_keys=metadata_output_keys,
+        )
+        validate_audio_key_configuration(
+            self.name,
+            input_keys={},
```
shubhamNvidia: **[P1] Reject cross-role VAD output/input key collisions**

Leaving `input_keys` empty here misses destructive collisions introduced by the separate output-key controls. All of `output_waveform_key="audio_filepath"`, `output_sample_rate_key="audio_filepath"`, and the pair `output_waveform_key="sample_rate", output_sample_rate_key="waveform"` construct successfully. A valid fan-out then returns a tensor or integer under `audio_filepath`, or swaps the waveform and rate carriers, so downstream stages receive values of the wrong semantic type. Allow the role-aligned defaults/reuse, but reject the path collision and cross-role swaps; cover the emitted child as well as constructor validation.

nemo_curator/stages/audio/segmentation/vad_segmentation.py
```diff
@@ -94,16 +117,131 @@ class VADSegmentationStage(ProcessingStage[AudioTask, AudioTask]):
...
+                "segments will carry no audio — SegmentConcatenation (and any waveform "
+                "consumer) will silently drop every segment. Metadata-only use intended?"
+            )
 
     def inputs(self) -> tuple[list[str], list[str]]:
         return [], []
 
     def outputs(self) -> tuple[list[str], list[str]]:
-        return [], ["waveform", "sample_rate", "start_ms", "end_ms", "segment_num", "duration"]
+        return [], [
```
shubhamNvidia: **[P2] Keep `outputs()` aligned with metadata-only mode**

`VADSegmentationStage(keep_segment_waveform_in_task=False).outputs()` still lists `output_waveform_key`, while `_build_segment_item()` deliberately omits it and `describe().writes` correctly excludes it. `Pipeline.describe()` consumes this legacy surface, so it tells users and tooling that metadata-only children contain a waveform that runtime never emits. Append the waveform key only when `keep_segment_waveform_in_task` is true and add a small parity assertion against the configured contract.

shubhamNvidia (commented):
reviewed

shubhamNvidia (commented):
reviewed

PR #2339, opened 2026-08-25
feat(audio-tagging): output control + in-memory handoff + agent-ready contracts

shubhamNvidia:
**Contracts:** `describe()` + `*_key` params across the tagging stages (ASR-align, merge-alignment-diarization, prepare-module-segments, resample, split, and the Chinese-conversion / inverse-text-normalization text stages). Includes the **`SplitASRAlignJoinStage` composite contract**, so the planner can see through the composite to the sub-stages it expands into rather than treating it as an opaque box.

**New functionality**
- **In-memory handoff on resample** (`keep_waveform_in_task`, `write_to_disk`) — a resample feeding an ASR stage no longer has to land on disk first.
- **Stable output naming on resample (fixes duplicate and stale files).** The output was named after the input path. That broke in three ways, all now covered by tests:
  - A row fed from an **in-memory waveform** has only a scratch path, which differs every run — so every run wrote a *fresh* set of files. Measured on one corpus: **129 files on disk for 65 manifest rows.**
  - **Two recordings sharing a basename** (same filename in different folders) wrote to the same output and silently overwrote each other.
  - **Re-running at a different target rate** produced the same filename, so the pipeline served the *old* conversion.

  The output is now named from a digest of the audio content plus the settings being applied (`target_sample_rate`, `target_nchannels`, `target_format`). Rows that already carry a real input path keep the name they have always had — the digest only stands in where the path cannot.
- **`update_audio_filepath` on resample** — choose whether downstream stages see the resampled file or the original path, instead of that being decided for you.
- **Configurable outputs** — `output_dir` on split (where the chunks go) and `output_suffix` on the text stages (`_ITN`, `_simplified`), so a pipeline can direct these without editing code.


nemo_curator/stages/audio/tagging/resample_audio.py
```diff
@@ -57,12 +67,25 @@ class ResampleAudioStage(ProcessingStage[AudioTask, AudioTask]):
     # Key names
     audio_filepath_key: str = "audio_filepath"
     resampled_audio_filepath_key: str = "resampled_audio_filepath"
+    waveform_key: str = "waveform"
+    sample_rate_key: str = "sample_rate"
     duration_key: str = "duration"
     audio_item_id_key: str = "audio_item_id"
+    original_audio_filepath_key: str = "original_audio_filepath"
+
+    input_residency: InputResidency = "file"
```
mohammadaaftabv, line 96: P0: Waveform-only input is rejected before the new in-memory path runs

nemo_curator/stages/audio/tagging/resample_audio.py:76-79 exposes
input_residency="waveform", and describe() advertises the corresponding
waveform/sample-rate alternative at lines 123-128, but inputs() still
unconditionally requires audio_filepath at lines 98-100. 

The normal backend
entry point is ProcessingStage.process_batch(), which validates inputs()
before calling process(). A waveform-only AudioTask therefore raises
ValueError: ... failed validation and never reaches resolve_audio_path; the
new tests at tests/stages/audio/tagging/test_resample_audio.py:59-71 avoid the
failure by calling process() directly.

Make runtime validation residency-aware: file mode must require the configured
path, waveform mode must require both resident keys, and auto mode must accept
either complete representation. Add a regression through process_batch()
(or the real backend adapter), not only a direct process() call, for every
advertised residency.

nemo_curator/stages/audio/tagging/resample_audio.py
```diff
@@ -96,29 +190,63 @@ def process(self, task: AudioTask) -> AudioTask:
         t0 = time.perf_counter()
         data_entry = task.data
 
-        if self.audio_filepath_key not in data_entry:
-            msg = "Absolute audio filepath is required"
+        temp_paths: list[str] = []
```
mohammadaaftabv, line 243: P0: auto consumes the file even when a resident waveform is present

nemo_curator/stages/audio/tagging/resample_audio.py:193-201 delegates the
new auto mode to the foundation's resolve_audio_path. At the current PR
#2332 head that helper checks and returns audio_filepath before inspecting the
resident waveform, although the foundation contract says auto residency
prefers resident audio. With a one-second file and a different two-second
resident waveform in the same valid task, this stage reported a one-second
result: it silently converted the file.

Fix resolve_audio_path in the dependency so auto first materializes a
non-null waveform/sample-rate pair and only falls back to the path; alternatively
select that order in this stage before calling the helper. Add a deliberately
divergent file-plus-waveform regression so merely checking that both routes are
accepted cannot miss the precedence error.

nemo_curator/stages/audio/tagging/resample_audio.py
```diff
@@ -129,20 +257,43 @@ def process(self, task: AudioTask) -> AudioTask:
...
                 msg = f"Error converting {input_audio_path}: {e}"
                 raise RuntimeError(msg) from e
 
-        # Update metadata — preserve original URL for cloud paths
-        data_entry[self.audio_filepath_key] = original_audio_filepath
-        data_entry[self.resampled_audio_filepath_key] = output_audio_path
+        # Input temp WAV (materialized from a waveform) is no longer needed after conversion.
+        cleanup_temp_files(temp_paths)
+
+        # Update metadata — preserve original URL for cloud paths.
```
mohammadaaftabv, line 348: P0: Disk-only conversion leaves stale resident audio in the task

nemo_curator/stages/audio/tagging/resample_audio.py:274-291 writes the
converted file but, when keep_waveform_in_task=False, never removes a
waveform/sample-rate pair that arrived on the task. A reproduced stereo/8 kHz
resident input was written as mono/16 kHz while the returned task still held
the old stereo tensor and sample_rate=8000. The next audio stage in auto
mode is then entitled to consume that stale tensor while the manifest points
at and describes the converted file.

On a disk-only output, call the foundation's drop_resident_audio helper for
the configured keys and declare those removals in the contract. Cover a task
that starts with both resident keys and verify the written file, returned keys,
and a following auto-residency consumer all agree.

nemo_curator/stages/audio/tagging/split.py
```diff
@@ -90,32 +153,39 @@ def process(self, task: AudioTask) -> AudioTask:
...
-        audio_path = data_entry["resampled_audio_filepath"]
+        audio_path = data_entry[self.audio_filepath_key]
         _fs, resolved_path = url_to_fs(audio_path)
 
         # parent_url preserves protocol prefix (e.g. "s3://bucket/dir") for stored paths;
         # resolved_parent is the fsspec-resolved counterpart for torchaudio I/O.
         parent_url, filename = audio_path.rsplit("/", 1) if "/" in audio_path else ("", audio_path)
         resolved_parent = resolved_path.rsplit("/", 1)[0] if "/" in resolved_path else ""
         stem = filename.rsplit(".", 1)[0] if "." in filename else filename
+        if self.output_dir is not None:
```
mohammadaaftabv: P0: Split names can overwrite different chunks in the shared output directory

nemo_curator/stages/audio/tagging/split.py:182-186 derives the flattened
stem only from the source path, while lines 195-223 add only the split ordinal
and count. Two rows naming the same source but using valid split plans
[0,4,8] and [0,3,8] therefore produced the same two paths; the second row
overwrote 40/40-frame chunks with 30/50-frame chunks, leaving the first row's
metadata pointing at the wrong audio. per_row_independent=False is advisory
and does not give the rows distinct output identities. The eight-hex-character
path digest is also too short for corpus-scale flattening; for example,
/dataset/spk25433/utt.wav and /dataset/spk158142/utt.wav both map to
aa600732.

Derive output identity from the source plus the effective split plan and
settings, and retain at least a 128-bit digest (or the full digest). Add a
two-row regression that processes the same source with different segment
boundaries and asserts both path uniqueness and preserved sample ranges, plus
a many-path collision test for the flattened namespace.

nemo_curator/stages/audio/tagging/inference/nemo_asr_align.py
```diff
@@ -268,13 +277,36 @@ def setup(self, _: WorkerMetadata | None = None) -> None:
...
-        return ["data"], ["duration", self.segments_key, "split_filepaths", "split_metadata"]
+            return ["data"], [self.resampled_audio_filepath_key, self.segments_key]
+        return ["data"], ["duration", self.segments_key, self.split_filepaths_key, self.split_metadata_key]
+
+    def describe(self) -> StageContract:
+        if self.infer_segment_only:
+            reads = IOSpec(data_keys=[self.resampled_audio_filepath_key, self.segments_key])
+            writes = IOSpec(segment_data_keys=[self.text_key, self.words_key])
+        else:
+            reads = IOSpec(
```
mohammadaaftabv, line 301: P0: The full-audio ASR contract writes at the wrong metadata scope

nemo_curator/stages/audio/tagging/inference/nemo_asr_align.py:293-299
declares top-level text/alignment and segments[] text/words writes in full
mode. The normal split path actually writes text and alignment inside
split_metadata[] at lines 439-445. 

In the focused reproduction,
validate_pipeline([aligner, MergeAlignmentDiarizationStage(...)]) returned
ok=True, keys_ok=True, the aligner produced only nested values, and the merge
stage then failed backend validation because top-level alignment was absent.
**This is a planner-approved pipeline that crashes on a valid input.**

Describe full mode as a nested-list transform with
iteration_key=split_metadata_key and nested text/alignment writes; keep
the separate segment-only contract tied to segments_key, and represent the
top-level fallback branch conditionally. Add assert_agent_ready fixtures for
split metadata, the no-splits branch, and segment-only mode, followed by a
planner-plus-runtime chain regression.

nemo_curator/stages/audio/tagging/resample_audio.py
```diff
@@ -76,14 +99,85 @@ def inputs(self) -> tuple[list[str], list[str]]:
...
+    def describe(self) -> StageContract:
+        writes = [self.audio_item_id_key, self.duration_key]
+        produces = []
+        if self.write_to_disk:
+            writes.append(self.resampled_audio_filepath_key)
+            produces.append("disk")
+        if self.keep_waveform_in_task:
+            writes.extend([self.waveform_key, self.sample_rate_key])
+            produces.append("tensor")
+        if self.update_audio_filepath:
```
mohammadaaftabv, line 155: P1: Resample's declared outputs disagree with two accepted branches

nemo_curator/stages/audio/tagging/resample_audio.py:120-121 declares an
audio_filepath write whenever update_audio_filepath=True, but runtime makes
that write only inside if self.write_to_disk at lines 277-285. Thus
write_to_disk=False, keep_waveform_in_task=True, update_audio_filepath=True is accepted and advertised even though the knob is
silently ignored. 

Conversely, when disk output and path replacement are
enabled, produce_audio_filepath may add the configured
original_audio_filepath at lines 279-285, but neither outputs() nor
describe() declares that new key; the reproduction observed it as an
undeclared write.

Use reject_sinkless_conversion (including its update-without-disk check) in
__post_init__, and declare the provenance key on the path-replacement branch.
Add assert_agent_ready coverage for every supported sink combination and an
explicit constructor-failure test for the unsupported combination.

nemo_curator/stages/audio/tagging/split.py
```diff
@@ -41,36 +44,74 @@ class SplitLongAudioStage(ProcessingStage[AudioTask, AudioTask]):
...
         suggested_max_len: Target maximum length for audio segments in seconds
         min_len: Minimum length for any split segment
+        output_dir: Optional directory for written split audio. When unset,
+            split files remain beside the source audio for backward compatibility.
     """
 
     # Split parameters
     suggested_max_len: float = 3600.0
     min_len: float = 1.0
+    duration_key: str = "duration"
```
mohammadaaftabv, line 73: P1: The new key fields break legacy positional construction

nemo_curator/stages/audio/tagging/split.py:54-64 inserts nine key fields
before the pre-existing name argument. Before this PR,
SplitLongAudioStage(120.0, 2.0, "custom") set name="custom"; it now sets
duration_key="custom" and a normal task crashes with KeyError('custom').
The analogous new fields at lines 296-302 shift the second positional argument
of JoinSplitAudioMetadataStage. **The test at
tests/stages/audio/tagging/test_split.py:52-59 calls this preservation while
actually asserting the new, incompatible binding.**

Keep every old field, including name, in its previous positional slot and
append the new fields afterward (or adopt an explicit, project-wide keyword-only
migration rather than changing selected constructors silently). **Correct the
regression to compare the pre-PR signatures and assert that legacy calls still
bind name.**

nemo_curator/stages/audio/tagging/split.py
```diff
@@ -213,13 +294,38 @@ class JoinSplitAudioMetadataStage(ProcessingStage[AudioTask, AudioTask]):
...
+            self.split_metadata_key,
+            self.split_offsets_key,
+            self.split_timestamps_key,
+        ]
 
     def outputs(self) -> tuple[list[str], list[str]]:
-        return [], [self.text_key, "alignment"]
+        return [], [self.text_key, self.alignment_key]
+
+    def describe(self) -> StageContract:
```
mohammadaaftabv, line 382: P1: Join's contract neither matches its writes nor its removals

nemo_curator/stages/audio/tagging/split.py:315-328 declares unconditional
top-level text/alignment writes and no removals. With all four declared inputs
present and split_filepaths=None, lines 342-345 remove only that key and emit
neither declared output; the conformance harness fails with declared write 'alignment' missing from task.data. On the normal branch, lines 391-393 remove
both split_filepaths and split_metadata, but the planner keeps both in
produced_keys because removes_keys is empty.

Normalize the no-split, empty-split, and populated-split exits so their output
and cleanup postconditions are stable, then declare the guaranteed removals
and any genuinely conditional writes. Add assert_agent_ready and planner
tests for all three branches, including a downstream consumer of a removed key.

nemo_curator/stages/audio/tagging/resample_audio.py
```diff
@@ -76,14 +99,85 @@ def inputs(self) -> tuple[list[str], list[str]]:
...
+            writes.append(self.audio_filepath_key)
+        return StageContract(
+            reads_one_of=residency_read_specs(
+                self.input_residency,
+                audio_filepath_key=self.audio_filepath_key,
+                waveform_key=self.waveform_key,
+                sample_rate_key=self.sample_rate_key,
+            ),
+            writes=IOSpec(data_keys=writes, produces=produces),
+            gates=Gates(
```
mohammadaaftabv, line 180: P1: Static discovery hides the new stages' mandatory side effects

The configured contracts at
nemo_curator/stages/audio/tagging/resample_audio.py:130-135 and
nemo_curator/stages/audio/tagging/split.py:101-105 correctly report FFmpeg
and/or disk-output gates, but neither class supplies conservative
AGENT_STATIC hints near its declaration (resample_audio.py:49-50 and
split.py:36-37). 

Instance-free discovery consequently reports the resampler
as requiring neither FFmpeg nor disk and the splitter as a pure, non-writing
stage with no redirectable output parameter. **An agent can therefore approve a
catalog/smoke run before it knows it must provide FFmpeg and sandbox filesystem
writes.**

Add StaticHints that conservatively expose writes_to_disk,
requires_ffmpeg, output_path_params, and row-independence before
construction, then assert both static_contract(Class) and configured
contracts in tests.

nemo_curator/stages/audio/tagging/text/itn.py
```diff
@@ -45,17 +46,27 @@ class InverseTextNormalizationStage(ProcessingStage[AudioTask, AudioTask]):
...
 
     def inputs(self) -> tuple[list[str], list[str]]:
-        return [], ["segments"]
+        return [], [self.segments_key]
 
     def outputs(self) -> tuple[list[str], list[str]]:
-        return [], ["segments"]
+        return [], [self.segments_key]
+
+    def describe(self) -> StageContract:
```
mohammadaaftabv, line 69: P1: Three new contracts claim writes that valid no-op inputs do not make

nemo_curator/stages/audio/tagging/text/itn.py:63-68 and
nemo_curator/stages/audio/tagging/text/chinese_conversion.py:59-64 declare an
unconditional output on every segment, but their runtime branches at
itn.py:80-87 and chinese_conversion.py:76-82 skip segments whose input text
key is absent (and ITN also skips empty text). Similarly,
merge_alignment_diarization.py:67-73 promises text/words on segments while
lines 197-201 legitimately no-op for empty alignment or segments. These are
valid according to the declared reads. Focused assert_agent_ready runs report
the declared writes missing for ITN and merge; Chinese conversion has the same
guarded-write shape at the cited lines.

Either materialize stable empty/fallback outputs or move the writes to
ConditionalWrite entries with accurate branch descriptions. Add the
foundation-mandated assert_agent_ready(...) fixtures for missing, empty, and
populated inputs; none of the three submitted test files currently invokes the
conformance harness.

mohammadaaftabv (changes requested):

nemo_curator/stages/audio/tagging/split.py
```diff
@@ -82,6 +123,28 @@ def get_split_points(self, metadata: dict) -> list[float]:
...
+        split_name: str,
+        parent_url: str,
+        resolved_parent: str,
+        resolved_output_dir: str,
+    ) -> tuple[str, str]:
+        """Build the stored and resolved paths for one split."""
+        if self.output_dir is None:
+            split_filepath = f"{parent_url}/{split_name}" if parent_url else split_name
+            split_resolved = f"{resolved_parent}/{split_name}" if resolved_parent else split_name
+            return split_filepath, split_resolved
```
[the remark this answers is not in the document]
shubhamNvidia: Fixed in `209bd1b2`. Remote chunks are encoded locally and uploaded through fsspec, so the advertised remote paths now exist. Added remote-write and failure-cleanup tests.

nemo_curator/stages/audio/tagging/resample_audio.py
```diff
@@ -129,20 +257,43 @@ def process(self, task: AudioTask) -> AudioTask:
...
+        if self.keep_waveform_in_task:
+            waveform, sample_rate = load_audio_file(output_audio_path, mono=False)
+            data_entry[self.waveform_key] = waveform
+            data_entry[self.sample_rate_key] = sample_rate
         duration = get_audio_duration(output_audio_path)
         data_entry[self.duration_key] = duration
+        if not self.write_to_disk:
+            try:  # noqa: SIM105
+                os.remove(output_audio_path)
+            except OSError:
```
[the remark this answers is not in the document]
shubhamNvidia: Fixed in `209bd1b2`. Temporary converted output is cleaned in `finally` when an in-memory handoff fails. Added a regression for the failure path.

nemo_curator/stages/audio/tagging/resample_audio.py
shubhamNvidia: Fixed in `209bd1b2`. Runtime validation now handles `file`, `waveform`, and `auto` residency correctly, with coverage through `process_batch()`.

nemo_curator/stages/audio/tagging/resample_audio.py
shubhamNvidia: Fixed in foundation dependency commit `89dde016`. `auto` now prefers a complete resident waveform/sample-rate pair and falls back to the file path. Added a deliberately divergent input regression.

nemo_curator/stages/audio/tagging/resample_audio.py
shubhamNvidia: Fixed in `209bd1b2`. Disk-only conversion removes stale waveform/sample-rate values, declares those removals, and verifies a downstream `auto` consumer reads the converted file.

nemo_curator/stages/audio/tagging/split.py
shubhamNvidia: Fixed in `209bd1b2`. Shared-output identity now includes the source, effective split plan, settings, and a 128-bit digest. Added different-plan and known short-hash collision regressions.

nemo_curator/stages/audio/tagging/inference/nemo_asr_align.py
shubhamNvidia: Fixed in `f3802046`, with shared planner support in foundation commit `89dde016`. Nested writes are tracked under `split_metadata[]`, while top-level fallback writes are conditional. Added conformance and planner/runtime-chain tests.

nemo_curator/stages/audio/tagging/resample_audio.py
shubhamNvidia: Fixed in `209bd1b2`. Invalid sink combinations are rejected, path replacement requires disk output, and the provenance write is declared conditionally. Supported combinations are covered by conformance tests.

nemo_curator/stages/audio/tagging/split.py
shubhamNvidia: Fixed in `209bd1b2`. Existing positional fields retain their original positions and new fields are appended. Added regressions for legacy construction.

nemo_curator/stages/audio/tagging/split.py
shubhamNvidia: Fixed in `acc9a007`. All valid branches now provide stable text/alignment outputs and remove both temporary split keys. Added conformance and downstream planner tests.

nemo_curator/stages/audio/tagging/resample_audio.py
shubhamNvidia: Fixed in `209bd1b2`. Both stages now expose conservative static hints for disk writes, FFmpeg, output paths, and row independence, with static/configured contract tests.

nemo_curator/stages/audio/tagging/text/itn.py
shubhamNvidia: Fixed in `acc9a007`. ITN, Chinese conversion, and alignment/diarization now use conditional nested writes without changing runtime behavior. Added missing, empty, and populated-input conformance tests.

nemo_curator/stages/audio/tagging/resample_audio.py
```diff
@@ -58,12 +87,28 @@ class ResampleAudioStage(ProcessingStage[AudioTask, AudioTask]):
     # Key names
     audio_filepath_key: str = "audio_filepath"
     resampled_audio_filepath_key: str = "resampled_audio_filepath"
+    waveform_key: str = "waveform"
```
mohammadaaftabv: Compatibility comment 1 — positional constructor API

Anchor: nemo_curator/stages/audio/tagging/resample_audio.py:90-99

Related anchors:
nemo_curator/stages/audio/tagging/inference/nemo_asr_align.py:82-89
nemo_curator/stages/audio/tagging/merge_alignment_diarization.py:61-62
nemo_curator/stages/audio/tagging/prepare_module_segments.py:69-71
nemo_curator/stages/audio/tagging/text/chinese_conversion.py:50-51
nemo_curator/stages/audio/tagging/text/itn.py:55-56

P1 — New key fields still silently shift six legacy constructors. 

The earlier positional-compatibility fix covers Split/Join/Composite, but the same
break remains in every other modified dataclass. Exact pre-PR calls still
construct, so this silently changes their meaning instead of raising a useful
error. For example,
NeMoASRAlignerStage(1, 40, 100, 10, 5000, False, "txt", "words", False)
set compute_timestamps=False at the base SHA; this head assigns False to
alignment_key and leaves timestamps enabled. Likewise,
ResampleAudioStage("/tmp", "wav", 16000, "wav", 1, "src", "dst", "secs", "item", "custom")
used to bind duration_key, audio_item_id_key, and name; this head binds
those values to waveform_key, sample_rate_key, and duration_key, while
restoring the default item key and name. The old positional name is also
consumed as alignment_key for Merge or segments_key for Prepare, Chinese,
and ITN. 

Make all additive fields keyword-only (especially the fields added
to BaseASRProcessorStage, where merely appending them still shifts the
subclass's inherited positional slots), or otherwise preserve every pre-PR
positional slot. Add one regression per affected class using its complete
pre-PR positional signature and asserting every legacy field.

nemo_curator/stages/audio/tagging/resample_audio.py
```diff
@@ -142,18 +343,36 @@ def process(self, task: AudioTask) -> AudioTask:
...
+                    data_entry,
+                    output_audio_path,
+                    key=self.audio_filepath_key,
+                    original_key=self.original_audio_filepath_key,
+                )
+        if self.keep_waveform_in_task:
+            waveform, sample_rate = load_audio_file(output_audio_path, mono=False)
+            data_entry[self.waveform_key] = waveform
+            data_entry[self.sample_rate_key] = sample_rate
+        elif self.write_to_disk:
```
mohammadaaftabv: Compatibility comment 2 — default Resample row preservation

Anchor: nemo_curator/stages/audio/tagging/resample_audio.py:364-369

P1 — The stale-resident fix deletes unrelated data in legacy file mode.
drop_resident_audio() now runs whenever disk output is enabled and the
converted waveform is not retained, including the default
input_residency="file" route. 

In a differential reproduction, a default stage received audio_filepath together with upstream waveform and sample_rate values. The base implementation preserved that resident pair;
this head deleted it even though the file, not the resident representation,
was selected for conversion. That changes the stage's historical
preserve-unmentioned-keys behavior and breaks a following waveform-only
consumer. The existing stale-data issue requires removal when resident audio
was actually selected by waveform/auto; it does not require deleting
unrelated keys on the legacy file route. Track which representation
resolve_audio_path() selected and drop the pair only after a resident-source,
disk-only conversion. 

Keep removes_keys empty for the legacy file configuration and retain it for resident-source configurations. 

Add both a base-vs-head default-file preservation regression and the existing resident-source stale-data regression.

nemo_curator/stages/audio/tagging/split.py
```diff
@@ -234,13 +408,16 @@ def process(self, task: AudioTask) -> AudioTask:
         words_aligned = 0
 
         # Check if this is a meta-entry with split information
-        if "split_filepaths" in data_entry:
-            if data_entry["split_filepaths"] is None:
-                del data_entry["split_filepaths"]
+        if self.split_filepaths_key in data_entry:
+            if data_entry[self.split_filepaths_key] is None:
```
mohammadaaftabv, line 450: Compatibility comment 3 — Join no/empty-split behavior

Anchors: nemo_curator/stages/audio/tagging/split.py:412-416 and
nemo_curator/stages/audio/tagging/split.py:436-440

P1 — The Join contract fix deletes legacy no/empty-split state. For
{"split_filepaths": None, "split_metadata": [{"text": "latent"}], "split_offsets": [1.0]},
the base stage removed only the sentinel and retained split_metadata; it did
not fabricate transcript/alignment output. 

This head deletes
split_metadata and adds text="" plus alignment=[]. With empty lists, the
base likewise retained split_metadata=[] and emitted neither output, while
this head removes the list and injects both keys. 

A downstream stage that
still consumes the retained metadata therefore loses valid state. 

Restore the
two pre-PR runtime branches and describe their writes/removals conditionally
instead of changing row contents to satisfy an unconditional contract. Add
exact-dictionary regressions for None, empty, and populated split inputs;
the populated branch already matches the base behavior.

nemo_curator/stages/audio/tagging/resample_audio.py
```diff
@@ -76,13 +121,111 @@ def setup_on_node(
...
+        return has_file or has_waveform
+
     def outputs(self) -> tuple[list[str], list[str]]:
-        return [], [
-            self.audio_filepath_key,
-            self.audio_item_id_key,
-            self.resampled_audio_filepath_key,
-            self.duration_key,
-        ]
+        outputs = [self.audio_item_id_key, self.duration_key]
```
mohammadaaftabv: Compatibility comment 4 — legacy outputs() API

Anchor: nemo_curator/stages/audio/tagging/resample_audio.py:135-143

P2 — Default ResampleAudioStage.outputs() drops audio_filepath. At the
base SHA the public default returned
['audio_filepath', 'audio_item_id', 'resampled_audio_filepath', 'duration'].
This head returns only ['audio_item_id', 'duration', 'resampled_audio_filepath'], even though the default runtime still
preserves/reassigns audio_filepath at lines 348-350. 

Existing pipeline
introspection therefore sees a column disappear without any configuration
change. 

Preserve the exact legacy return for the default/file-compatible
configuration while extending it for the new sink modes, and add an exact
default-return regression against the pre-PR API.

nemo_curator/stages/audio/tagging/resample_audio.py
```diff
@@ -76,13 +121,111 @@ def setup_on_node(
...
+
+        A real input path is stable across runs, so it can name the output. A scratch path
+        materialised from a waveform is not: naming a persistent output after it made every run
+        write a fresh set of files (measured: 129 on disk against 65 manifest rows). So hash the
+        audio instead, which is the identity the path was standing in for -- and folding in the
+        settings makes the "already converted, skip it" check below correct, not merely fast.
+
+        The path branch is unchanged, so pipelines reading real files keep their output names.
+        """
+        if not from_scratch_file:
```
mohammadaaftabv, line 260: P0 — a 32-bit filename digest can silently substitute another recording

Anchor: nemo_curator/stages/audio/tagging/resample_audio.py:208-210

P0 — The claimed basename-collision fix still has deterministic collisions.
The real-file route retains only eight hex characters (32 bits) of the path
digest. The concrete valid paths /dataset/spk25433/utt.wav and
/dataset/spk158142/utt.wav both map to utt_aa600732.wav. Once the first row
creates that target, line 309 accepts it for the second row whenever sample
rate and channel count match, so the second manifest row silently points to
the first recording. This is corpus corruption, not merely a duplicate name.
Preserve the legacy name where it is unambiguous, but associate each cached
output with the full source identity/content and disambiguate a mismatch with
at least a 128-bit digest (or the full digest). Add a regression using these
exact two paths and different waveforms, verifying both output identity and
sample content.

This does not duplicate the outdated Split naming thread: that fix lengthened
the Split digest to 128 bits, while the resampler still uses the known 32-bit
collision.

nemo_curator/stages/audio/tagging/resample_audio.py
```diff
@@ -76,13 +121,111 @@ def setup_on_node(
...
+            stem = os.path.splitext(os.path.basename(local_audio_path))[0]
+            return f"{stem}_{hashlib.sha256(local_audio_path.encode()).hexdigest()[:8]}"
+        # Keep the source name on the front so a clip stays traceable by eye.
+        stem = os.path.splitext(os.path.basename(str(source)))[0] if source else "clip"
+        return f"{stem}_{self._audio_digest(local_audio_path)}"
+
+    def _matches_target(self, path: str) -> bool:
```
mohammadaaftabv: P0 — a truncated target with a valid header is accepted as complete

Anchor: nemo_curator/stages/audio/tagging/resample_audio.py:215-228

P0 — Header-only cache validation serves truncated audio as a completed
conversion. _matches_target() checks only rate and channel count. A
one-second 16-kHz PCM WAV truncated to 1,000 bytes still has a readable
16-kHz/mono header and 478 frames; _matches_target() returns True, line
309 skips FFmpeg, and the manifest receives roughly 0.03 seconds of audio.
The new atomic rename prevents a future in-process interruption from creating
this stump, but it cannot repair an output left by the previous implementation
or damaged externally because the skip occurs first. Validate source identity
and expected frame/duration/content, or require an atomically written
completion sidecar and refuse targets without a matching marker. Add a
pre-seeded truncated-target regression that proves conversion runs and repairs
the file.

nemo_curator/stages/audio/tagging/prepare_module_segments.py
```diff
@@ -94,8 +112,8 @@ def get_words_list_from_all_segments(self, metadata: dict[str, Any]) -> list[dic
                 - sisdr_squim: The SI-SDR score of the word if available
                 - bandwidth: The bandwidth of the word if available
         """
-        segments = metadata["segments"]
-        audio_duration = metadata.get("duration", 0.0)
+        segments = metadata[self.segments_key]
```
mohammadaaftabv, line 136: P0 — Prepare's hidden literal reads defeat its configurable contract

Anchor: nemo_curator/stages/audio/tagging/prepare_module_segments.py:115-121

P0 — A planner-approved renamed-key chain can silently delete all segments.
describe() advertises configurable segment/duration/text/word keys, but
runtime still branches on literal overlap_segments and reads literal
alignment. 

In a focused ASR-align → Prepare reproduction using
segments_key="utterances", duration_key="length",
text_key="transcript", words_key="tokens", and an upstream alignment at
token_alignment, validation returned ok=True, keys_ok=True; Prepare then
read the absent literal alignment and replaced utterances with [].
Add and use alignment_key and overlap_segments_key, model the
alignment-versus-pre-enriched-segment alternatives honestly, and make the
row-identity keys read at lines 429-434 configurable as well. Add a fully
renamed planner-plus-runtime chain regression. This is also required by
stages/audio/AGENTS.md: every task.data key read or written must be a
constructor *_key field.

nemo_curator/stages/audio/tagging/inference/nemo_asr_align.py
```diff
@@ -268,13 +283,48 @@ def setup(self, _: WorkerMetadata | None = None) -> None:
...
+        return ["data"], ["duration", self.segments_key, self.split_filepaths_key, self.split_metadata_key]
 
     def outputs(self) -> tuple[list[str], list[str]]:
         if self.infer_segment_only:
-            return ["data"], ["resampled_audio_filepath", self.segments_key]
-        return ["data"], ["duration", self.segments_key, "split_filepaths", "split_metadata"]
+            return ["data"], [self.resampled_audio_filepath_key, self.segments_key]
+        return ["data"], ["duration", self.segments_key, self.split_filepaths_key, self.split_metadata_key]
+
+    def describe(self) -> StageContract:
```
mohammadaaftabv, line 309: P1 — the ASR contract still rejects runtime-valid inputs and promises absent writes

Anchor: nemo_curator/stages/audio/tagging/inference/nemo_asr_align.py:294-327

P1 — The ASR contract remains unsound on valid branches. In segment mode,
a sub-min_len segment causes no text/word write, and
compute_timestamps=False writes text but never words, although both are
unconditional contract writes. In full mode, an empty split list writes only
top-level fallback values while missing/None split paths can return without
either the claimed nested or top-level writes. The read side is also stricter
than runtime: full mode requires duration and segments it never consumes, and
segment mode requires resampled_audio_filepath even though line 125 accepts
audio_filepath as a fallback. Move every data-dependent write into an
accurate ConditionalWrite, make words conditional on timestamps, remove
unused full-mode reads, and express segment audio as a configured path
alternative. Add conformance plus planner/runtime tests for short segments,
timestamps off, empty/missing split paths, mismatched path/metadata counts,
and the original-file fallback.

The normal populated split scope from existing thread
discussion_r3960935689 is fixed; this comment covers the remaining branches
rather than repeating that thread.

nemo_curator/stages/audio/tagging/resample_audio.py
```diff
@@ -58,12 +87,28 @@ class ResampleAudioStage(ProcessingStage[AudioTask, AudioTask]):
...
+
+    input_residency: InputResidency = "file"
+    keep_waveform_in_task: bool = False
+    write_to_disk: bool = True
+    update_audio_filepath: bool = False
 
     # Stage metadata
     name: str = "ResampleAudio"
 
+    def __post_init__(self) -> None:
```
mohammadaaftabv, line 128: P1 — unsupported residency strings silently acquire auto behavior

Anchor: nemo_curator/stages/audio/tagging/resample_audio.py:104-110

P1 — Invalid input_residency values silently behave like auto.
__post_init__() validates sink combinations but never calls the foundation's
validate_input_residency(). 
A stage configured with input_residency="disk" constructs successfully, validates a file row, its
contract advertises both file and waveform alternatives, and resolution falls
through as if auto were selected. 

A YAML typo can therefore change which
audio representation is processed instead of failing at construction. Call
validate_input_residency(self.input_residency, stage_name=...) before sink
validation and add constructor rejection tests for unsupported values.

nemo_curator/stages/audio/tagging/resample_audio.py
```diff
@@ -76,13 +121,111 @@ def setup_on_node(
     def inputs(self) -> tuple[list[str], list[str]]:
         return [], [self.audio_filepath_key]
 
+    def validate_input(self, task: AudioTask) -> bool:
```
mohammadaaftabv, line 161: P2 — resident sample rates are accepted solely for being non-null
Anchor: nemo_curator/stages/audio/tagging/resample_audio.py:124-133

P2 — Malformed resident sample rates pass validation and are silently
coerced. The waveform alternative checks only sample_rate is not None.
The shared resolver then converts True to 1 and 16000.5 to 16000,
while zero and negative rates pass this stage's validation and fail later in
SoundFile/FFmpeg. 

Require a positive, losslessly integral, non-boolean sample
rate before declaring the resident alternative usable, and cover both
waveform and auto through process_batch().

nemo_curator/stages/audio/tagging/inference/nemo_asr_align.py
mohammadaaftabv, line 165: P1 — static discovery hides the default model download

Anchor: nemo_curator/stages/audio/tagging/inference/nemo_asr_align.py:158-159

P1 — Static discovery says the default aligner needs no first-run internet.
setup_on_node() downloads model_name when model_path is absent and the
configured contract correctly reports requires_internet_first_run=True, but
static_contract(NeMoASRAlignerStage) reports False because the class has no
conservative AGENT_STATIC hint. Catalog planning can therefore approve an
offline environment before instantiation and fail only during setup. Add a
static internet gate (plus the existing GPU fact) and a class-level
regression comparing static and default-configured requirements.

nemo_curator/stages/audio/tagging/prepare_module_segments.py
```diff
@@ -32,7 +33,7 @@
 
 
 @dataclass
-class PrepareModuleSegmentsStage(ProcessingStage[AudioTask, AudioTask]):
+class PrepareModuleSegmentsStage(AgentReady, ProcessingStage[AudioTask, AudioTask]):
```
mohammadaaftabv, line 36: P1 — the AgentReady conversion omits its required cards and one conformance test

Anchor: nemo_curator/stages/audio/tagging/prepare_module_segments.py:35-36

P1 — This AgentReady conversion is incomplete under its own dependency's
authoring contract. The PR converts the tagging stages to AgentReady, but
the intended PR2332+PR2339 stack has no
nemo_curator/audio_agent/knowledge/cards/ directory or capability cards for
these stages, and test_prepare_module_segments.py never calls
assert_agent_ready. nemo_curator/stages/audio/AGENTS.md and
AGENT_READY.md require both for every converted stage; cards carry semantic
meaning that key-role matching cannot prove. Add schema-conformant cards for
each public converted stage and run card_conformance.audit() with zero
violations, plus add Prepare's missing conformance fixture. If cards are
intentionally supplied by another stacked PR, declare and pin that dependency
before this one merges.

mohammadaaftabv (changes requested):

nemo_curator/stages/audio/tagging/resample_audio.py
shubhamNvidia: Agreed. All newly added agent knobs across ResampleAudioStage, NeMoASRAlignerStage, MergeAlignmentDiarizationStage, PrepareModuleSegmentsStage, ChineseConversionStage, and ITNStage are now keyword-only via `KW_ONLY`, preserving every legacy positional slot (including the base ASR processor class). Added exact pre-PR positional-signature regressions per class.

nemo_curator/stages/audio/tagging/resample_audio.py
shubhamNvidia: Agreed. `drop_resident_audio` now runs only when a resident/materialized source was actually converted; the default `input_residency="file"` route no longer deletes a pre-existing resident pair, and `removes_keys` is empty for that configuration. Added base-vs-head file-mode preservation plus the resident-source stale-data regressions.

nemo_curator/stages/audio/tagging/split.py
shubhamNvidia: Agreed. Restored the two legacy runtime branches: the None and empty-split cases only strip the sentinel and never delete `split_metadata` or fabricate `text`/`alignment`. The contract now declares `text`/`alignment` as conditional writes for the populated branch. Added exact-dict regressions for None, empty, and populated split inputs.

nemo_curator/stages/audio/tagging/resample_audio.py
shubhamNvidia: Agreed. The default/file-compatible `outputs()` again returns `audio_filepath`, `audio_item_id`, `resampled_audio_filepath`, `duration`, extended for the new sink modes. Added an exact default-return regression.

nemo_curator/stages/audio/tagging/resample_audio.py
shubhamNvidia: Agreed. The file-route id now uses a 64-bit digest (`[:8]` -> `[:16]`), matching `_audio_digest`, so two distinct paths sharing a basename no longer collide onto one output and a second row cannot cache-hit the first recording. Added a distinct-paths/same-stem regression asserting distinct output stems.

nemo_curator/stages/audio/tagging/resample_audio.py
shubhamNvidia: Addressed. Every write goes through a temp file plus atomic `os.replace` with cleanup, so this stage never leaves a truncated target; we also hardened `_matches_target` and added a truncated-target regression. We scoped the residual exposure to externally/previously truncated files rather than a stage-produced result.

nemo_curator/stages/audio/tagging/prepare_module_segments.py
shubhamNvidia: Agreed. Added configurable `alignment_key`/`overlap_segments_key` and configurable row-identity keys, used them in `process()` and `describe()`, and guarded segment assembly so an empty/word-less result never erases existing input segments. Added a fully renamed-key planner-plus-runtime regression proving segments are not wiped and per-row seeds stay distinct.

nemo_curator/stages/audio/tagging/inference/nemo_asr_align.py
shubhamNvidia: Agreed. Data-dependent writes are now conditional: short segments write nothing, `compute_timestamps=False` writes text but not words (words conditional on timestamps), an empty split writes only the top-level fallback, and missing/None split writes neither. Removed the unused full-mode `duration`/`segments` reads and expressed segment audio as a `resampled_audio_filepath`-or-`audio_filepath` alternative. Added branch tests.

nemo_curator/stages/audio/tagging/resample_audio.py
shubhamNvidia: Agreed. `__post_init__` now calls `validate_input_residency(...)` before sink validation, so an unsupported value such as `"disk"` fails at construction instead of silently behaving like `auto`. Added a constructor rejection test.

nemo_curator/stages/audio/tagging/resample_audio.py
shubhamNvidia: Agreed. The waveform alternative now requires a positive, non-boolean, losslessly-integer sample rate before it is considered usable; `True`, `0`, negative, and `16000.5` are rejected while `16000` passes. Covered through `validate_input`/`process_batch`.

nemo_curator/stages/audio/tagging/inference/nemo_asr_align.py
shubhamNvidia: Agreed. Added a conservative `AGENT_STATIC = StaticHints(gates=Gates(requires_internet_first_run=True, requires_gpu=True))` so instance-free discovery matches the default-configured contract. Added a static-vs-configured regression.

nemo_curator/stages/audio/tagging/prepare_module_segments.py
shubhamNvidia: The capability cards for these stages live in the full agent branch (draft PR #2343), where the card audit runs with zero violations; I updated the Prepare card for the new configurable keys and added the Prepare conformance coverage there. This module PR stays stacked on that dependency for the card tree.

nemo_curator/stages/audio/tagging/resample_audio.py
mohammadaaftabv: P0 — auto still consumes an invalid resident pair instead of its valid
file fallback. 

validate_input() correctly rejects malformed resident
sample rates, but in auto it returns true when the file alternative is
valid. process() then calls the shared resolver, which checks only that the
resident values are non-None, selects them first, and coerces the rate. I
reproduced a valid file plus waveform and sample_rate=0: validation
returned True, then process_batch() failed with LibsndfileError ... SF_INFO struct incomplete instead of converting the file. 
Select resident audio only with the same usability predicate used by validation; otherwise
fall back to the file. Centralize that predicate in the resolver (or select
the branch before calling it), and cover a valid file plus resident rates
None, 0, negative, boolean, and fractional values.

nemo_curator/stages/audio/tagging/split.py
mohammadaaftabv: P0 — Split identity hardening is applied only to output_dir, leaving the
default route corruptible. The source/plan digest at lines 258-261 runs only
when output_dir is set. The default source-adjacent route still writes
<stem>.<ordinal>_of_<count>.wav, so two rows for the same source with
different boundary plans and the same chunk count overwrite one another.
Distinct siblings such as utt.wav and utt.flac do the same; the reproduced
first chunk changed from mean +0.19998 to -0.20001 after the second row.
Yet the default contract reports per_row_independent=True. Apply durable
source + effective-plan ownership to both routes. To retain compatibility,
keep the historical name for the ordinary unambiguous owner and atomically
disambiguate an actual mismatch; test same-source/different-plan and
same-stem/different-extension rows, including sample ranges and metadata.

nemo_curator/stages/audio/tagging/resample_audio.py
mohammadaaftabv: P0 — disk-only auto leaves a partial resident representation behind while
the planner removes it. For input_residency="auto", disk output, and no
resident output, the contract unconditionally removes waveform and
sample_rate. Runtime drops them only when a complete resident pair was
selected. A valid row with a file plus waveform but no sample_rate falls
back to the file and returns with the tensor intact. The planner then approves
a manifest writer with no issue because it believes the tensor is gone, but
serialization fails. On disk-only non-file modes, remove stale/partial
resident keys after file fallback too, or represent the branch through honest
conditional-removal semantics. Add the exact auto file + waveform-only ->
serializing-sink regression.

nemo_curator/stages/audio/tagging/split.py
mohammadaaftabv: P0 — Join's populated branch deletes a key its contract preserves. The
contract lists only split_filepaths in removes_keys, but the normal join
path deletes both split_filepaths and split_metadata. A downstream
consumer configured to read split_metadata was approved by
validate_pipeline() with ok=True, keys_ok=True, and no issues; after
Join deleted the key, the consumer failed input validation. The None and
empty branches must retain split_metadata for compatibility, so do not make
the removal unconditional. Add conditional-removal support/metadata (or
otherwise stop guaranteeing the key after this stage) and test the
planner-plus-runtime populated branch while retaining the exact legacy
dictionaries for None and empty inputs.

nemo_curator/stages/audio/tagging/merge_alignment_diarization.py
```diff
@@ -48,18 +55,37 @@ class MergeAlignmentDiarizationStage(ProcessingStage[AudioTask, AudioTask]):
...
     words_key: str = "words"
 
-    # Stage metadata
+    # Stage metadata (legacy positional slot)
     name: str = "MergeAlignmentDiarization"
 
+    # Agent-added knobs are keyword-only (KW_ONLY sentinel) so the legacy positional slots
+    # above keep their historical order and meaning.
+    _: KW_ONLY
+    alignment_key: str = "alignment"
```
mohammadaaftabv, line 69: nemo_curator/stages/audio/tagging/merge_alignment_diarization.py:68-69
nemo_curator/stages/audio/tagging/split.py:84-85
nemo_curator/stages/audio/tagging/resample_audio.py:116-117

P0 — New configurable key aliases can silently drop data or create cyclic
rows. 

Three configurations made possible only by this PR reproduce the
problem: (1) MergeAlignmentDiarizationStage(alignment_key="items", segments_key="items") creates item["words"][0] is item, so the row is
cyclic and json.dumps fails; (2)
SplitLongAudioStage(split_filepaths_key="bundle", split_metadata_key="bundle") overwrites the path list with metadata while
its contract advertises duplicate writes; and (3)
ResampleAudioStage(write_to_disk=False, keep_waveform_in_task=True, waveform_key="resident", sample_rate_key="resident") validates and runs
but replaces the waveform with integer 16000, while its contract still
labels resident as waveform and claims tensor output. Reject aliases between
incompatible roles during __post_init__, using the foundation helper where
applicable and explicitly allowing only intentional in-place/cross-scope
transforms. Add these exact constructor/runtime regressions. Do not impose a
blanket rule on pre-existing aliases that were valid at the merge base.

mohammadaaftabv (changes requested):

nemo_curator/stages/audio/tagging/merge_alignment_diarization.py
shubhamNvidia: Thanks for catching this. I added targeted validation for the newly configurable incompatible key aliases:

- `alignment_key` and `segments_key` must be distinct.
- Generated split output keys must be distinct.
- `waveform_key` and `sample_rate_key` must be distinct.

The validation intentionally does not impose new restrictions on pre-existing legacy aliases or valid cross-scope configurations. Exact regression tests were added for all three reported cases, and the focused tagging suite passes.

nemo_curator/stages/audio/tagging/resample_audio.py
```diff
@@ -99,33 +310,134 @@ def process(self, task: AudioTask) -> AudioTask:
...
+            # which downstream stages read as the shared ``item_id`` role.
+            output_stem = f"{output_stem}_{self._audio_digest(local_audio_path)}"
+        elif inherited_id and self.write_to_disk:
+            # The legacy file route names the output after the inherited id, which keeps every
+            # tutorial's output names stable -- but two DIFFERENT recordings carrying the same id
+            # would then share one output path, and the second would either overwrite the first
+            # or (if the lengths agree) be served the first recording as a finished conversion.
+            # Keep the legacy name for the first source seen under an id in this worker; a later
+            # source with the same id gets a path digest so it can never alias the first.
+            owner = self._stem_owners.setdefault(str(output_stem), local_audio_path)
```
shubhamNvidia: **[P0] Do not keep inherited-id ownership only in one worker**

`_stem_owners` is instance-local, so this collision defense disappears across Ray workers or restarts. I reproduced it sequentially with two independent stage instances sharing an output directory: two different same-length WAVs both carried `audio_item_id="dup"`; both returned `dup.wav`, and the second instance accepted the first file through `_matches_target()` because rate, channels, and frame count matched. Reading the second row's advertised output yielded the first row's +0.2 samples instead of its -0.2 source. Encode the source/content identity in every ambiguous filename, or atomically publish and verify a durable ownership sidecar; a per-worker dictionary cannot protect a distributed corpus. Add a two-instance (and ideally concurrent) regression with equal-duration, different-content sources.

nemo_curator/stages/audio/tagging/resample_audio.py
```diff
@@ -80,13 +158,146 @@ def setup_on_node(
...
+        write a fresh set of files (measured: 129 on disk against 65 manifest rows). So hash the
+        audio instead, which is the identity the path was standing in for -- and folding in the
+        settings makes the "already converted, skip it" check below correct, not merely fast.
+
+        The path branch uses a 128-bit suffix. The former eight-character suffix had known
+        collisions, allowing two same-length recordings to share and reuse one output.
+        """
+        if not from_scratch_file:
+            stem = os.path.splitext(os.path.basename(local_audio_path))[0]
+            return f"{stem}_{hashlib.sha256(local_audio_path.encode()).hexdigest()[:32]}"
```
shubhamNvidia: **[P1] Bound the readable stem when lengthening the digest**

The 128-bit suffix fixes the short-hash collision, but retaining the entire input stem can now make a valid source impossible to resample. A source component containing a 240-character stem plus `.wav` is valid (244 bytes); this line derives a 277-byte `<stem>_<32 hex>.wav` output component, and FFmpeg fails on `ENAMETOOLONG`. The prior eight-hex form fit this exact case, so this is a regression from the hardening change. Sanitize/truncate the readable prefix to a budget that includes the digest and extension, and add a near-`NAME_MAX` input regression.

nemo_curator/stages/audio/tagging/split.py
```diff
@@ -213,13 +384,54 @@ class JoinSplitAudioMetadataStage(ProcessingStage[AudioTask, AudioTask]):
     """
 
     text_key: str = "text"
+
+    # Stage metadata. Keep the legacy positional fields before additive keys.
     name: str = "JoinSplitAudioMetadata"
 
+    split_filepaths_key: str = "split_filepaths"
+    split_metadata_key: str = "split_metadata"
```
shubhamNvidia: **[P1] Reject Join aliases between temporary and generated keys**

The newly configurable Join keys have no collision validation. `JoinSplitAudioMetadataStage(split_metadata_key="alignment")` accepts a populated row, computes the joined alignment into `meta_entry[alignment_key]`, and then the cleanup loop removes that same key as temporary `split_metadata`; the result contains text but no alignment even though `describe()` conditionally promises both. The analogous aliases with `split_filepaths_key` or `text_key` can delete the other generated output. Require the temporary split keys to be distinct from each other and from `text_key`/`alignment_key`, and cover the populated branch plus contract parity.

shubhamNvidia (commented):
reviewed

shubhamNvidia (commented):
reviewed

PR #2340, opened 2026-08-25
feat(audio-alm): make ALM stages agent-ready

shubhamNvidia:
**Contracts only — no behavior change.** Adds `describe()` + `*_key` params to the ALM data-builder / overlap stages and the pretrain extraction / io / planning stages.

Because window-building **replaces** the row rather than adding to it, the contract also declares `preserves_upstream_keys=False` and lists the fields it drops (`removes_keys`) — so the planner knows a key consumed upstream of ALM will not survive past it, instead of discovering that at runtime.

Source only — the `tests/stages/audio/alm/` folder is untouched and identical to `main`.


nemo_curator/stages/audio/alm/pretrain/io.py
```diff
@@ -239,6 +251,15 @@ class SnippetManifestWriterStage(ProcessingStage[AudioTask, AudioTask]):
     name: str = "SnippetManifestWriter"
     batch_size: int = 1
     resources: Resources = field(default_factory=lambda: Resources(cpus=1.0))
+    AGENT_STATIC: ClassVar[StaticHints] = StaticHints(
```
mohammadaaftabv, line 276: P0 (blocker).** Both the static and configured writer contracts set
`per_row_independent=True`, but this stage appends rows to shared shards which
`_merge_manifest_shards()` later merges by opening the public manifest in `"w"` mode.
The dependency's `AGENT_READY.md` explicitly lists appending to a shared file as a
`False` case and warns that an incorrect `True` can republish a partial corpus. In a
deterministic reproduction, a full run over `X,Y` produced `['X', 'Y']`; a narrowed
rerun over only changed row `X` then replaced the same manifest with `['X']`, silently
deleting unchanged row `Y`. `SnippetExtractionStage` makes the equivalent claim for
the shared tar at `extraction.py:139-144`, and final tar merging has the same replacement
semantics. 

Set the manifest writer to `per_row_independent=False`, set extraction to
`False` in normal mode (it can be `True` only for a genuinely side-effect-free dry
run), and use conservative static hints. These may become `True` only after delta
finalization reconciles new rows/members with untouched prior outputs. Please add a
full-`X,Y` then delta-`X` regression covering both manifest and tar preservation.

nemo_curator/stages/audio/alm/alm_data_builder.py
```diff
@@ -159,17 +167,38 @@ def __post_init__(self) -> None:
         self._drop_fields_top_level_set = {f.strip() for f in self.drop_fields_top_level.split(",") if f.strip()}
 
     def inputs(self) -> tuple[list[str], list[str]]:
-        return [], ["audio_filepath", "segments", "audio_sample_rate"]
+        return [], [self.audio_filepath_key, self.segments_key, self.audio_sample_rate_key]
+
+    def outputs(self) -> tuple[list[str], list[str]]:
+        return [], [self.windows_key, self.stats_key, self.truncation_events_key]
+
+    def describe(self) -> StageContract:
```
mohammadaaftabv, line 178: P1.** `preserves_upstream_keys=False` makes the planner clear every inherited key
not named in `writes`, but `writes` omits `audio_filepath_key`. Runtime retains that
path explicitly in the low-sample-rate branch (`:230-235`) and through the normal copy
(`:315-318`). Consequently, a valid `ALMDataBuilderStage() ->
GetAudioDurationStage()` chain is rejected with `key_removed_upstream` even though the
processed row still contains `kept.wav`. 

Make the configured audio path a guaranteed
write/pass-through (and reject it in `drop_fields_top_level`, or re-add it after that
filter), record its upstream-value lineage, and add a composition regression proving a
downstream file consumer remains valid.

nemo_curator/stages/audio/alm/alm_data_overlap.py
```diff
@@ -172,17 +175,28 @@ def __post_init__(self) -> None:
             raise ValueError(msg)
 
     def inputs(self) -> tuple[list[str], list[str]]:
-        return [], ["windows"]
+        return [], [self.windows_key]
+
+    def outputs(self) -> tuple[list[str], list[str]]:
```
mohammadaaftabv, line 202: P1.** The new contract exposes only `filtered_windows_key`, while `_filter_overlaps`
writes eight additional documented task fields at `:246-255`:
`total_dur_window`, `total_dur_list_window`,
`total_dur_list_window_timestamps`, `filtered`, `filtered_dur`,
`filtered_dur_list`, `manifest_filepath`, and `swift_filepath`.
`assert_agent_ready()` fails verbatim with those eight names as “undeclared new
top-level keys”, and an agent cannot discover or safely configure them. There is also
a broken custom-key chain: the builder exposes `stats_key`, but overlap still reads the
literal `"stats"` at `:242`, so renaming builder stats makes both provenance outputs
`None`. Add `*_key` fields for every task-data read/write required by the agent-ready
golden rule, use them in runtime code, list all guaranteed outputs in `outputs()` and
`writes`, and add default plus renamed-key conformance tests.

nemo_curator/stages/audio/alm/pretrain/extraction.py
```diff
@@ -115,6 +116,35 @@ def inputs(self) -> tuple[list[str], list[str]]:
     def outputs(self) -> tuple[list[str], list[str]]:
         return [], [self.audio_filepath_key, "snippet_id", "duration", "segments"]
 
+    def describe(self) -> StageContract:
```
mohammadaaftabv, line 128: P1.** This contract inherits `preserves_upstream_keys=True`. Normal snippet output
unconditionally removes `alignment`, `_snippet_plan`, `audio_size`, and
`resampled_audio_filepath` at `:333-338`; the valid zero-plan path and every
all-snippets-failed path call `_make_stub_task()`, which rebuilds the row from only five
keys at `:365-378`. The planner therefore approves extraction followed by a consumer of
upstream `text` (`ok=True`, `keys_ok=True`, no issues), while an empty-plan input drops
`text` and that consumer fails at runtime. 

Either preserve the safe upstream fields in
stubs and model the four guaranteed removals, or conservatively set
`preserves_upstream_keys=False` and describe conditional pass-through/transformed
fields with `ConditionalWrite`. Add conformance tests for a normal snippet, an empty
plan, an unreadable source, and an all-write-failed source.

nemo_curator/stages/audio/alm/pretrain/extraction.py
```diff
@@ -115,6 +116,35 @@ def inputs(self) -> tuple[list[str], list[str]]:
...
+        return StageContract(
+            reads=IOSpec(data_keys=[self.audio_filepath_key, _PLAN_DATA_KEY], accepts=["file"]),
```
mohammadaaftabv, line 130: P1.** The contract publishes `audio_filepath_key` while declaring a disk product and
remaining wrappable, so key/role validation accepts a normal file-consuming stage
after extraction. Runtime instead assigns only `<snippet_id>.<format>` at `:293-321`;
the canonical audio-pretrain page explicitly says this is a tar-internal member name,
not a standalone filesystem path. 
A downstream file consumer therefore receives a
nonexistent path even though validation approved the chain (and dry-run intentionally
has no tar at all). 

Represent tar-member residency/reference explicitly and require a
resolver-aware consumer, or mark this stage non-wrappable/withhold the ordinary file
output until the contract vocabulary can express it. Add a negative composition test
against an ordinary `accepts=['file']` consumer.

nemo_curator/stages/audio/alm/pretrain/io.py
```diff
@@ -154,6 +156,16 @@ def inputs(self) -> tuple[list[str], list[str]]:
     def outputs(self) -> tuple[list[str], list[str]]:
         return [], [self.audio_filepath_key, "id", "segments"]
 
+    def describe(self) -> StageContract:
```
mohammadaaftabv, line 172: **P1.** `describe()` unconditionally promises `segments`, but `process()` validates
only `id` and the configured audio path before emitting the original row. A manifest
row containing only those two fields is emitted with keys `['audio_filepath', 'id']`;
the immediately following `OverlapFilterStage` then raises its missing-`segments`
validation error. The canonical audio-pretrain schema calls `segments` required, and
the post-1608 corpus already records reader-side validation as the outstanding fix.

Reject or skip a row whose `segments` key is absent or not a list, with its manifest
line number, and add the missing-`segments` `assert_agent_ready()` regression. Making
the write conditional would still leave the standard pipeline invalid.

nemo_curator/stages/audio/alm/pretrain/io.py
```diff
@@ -249,6 +270,17 @@ def inputs(self) -> tuple[list[str], list[str]]:
     def outputs(self) -> tuple[list[str], list[str]]:
         return [], []
 
+    def describe(self) -> StageContract:
```
mohammadaaftabv, line 295: P1.** The manifest writer declares no reads even though `_is_origin_stub()` uses
`snippet_id` to decide whether to write; a task with that discriminator missing is
silently treated as a stub. The metrics contract at `:362-373` declares only metadata,
while runtime `:394-417` depends on `id`, `snippet_id`, `segments`, and `duration`.
With `initial_keys=set()`, validation reports `ok=True`/`keys_ok=True` for the metrics
stage, but processing a row without `id` creates no shard at all; missing remaining
fields can silently turn real output into zero/stub metrics. Add configurable key
fields and declare the extraction output shape each sink mechanically requires (at
minimum `snippet_id` for the writer and all four fields for metrics), then add validator
regressions proving a non-extraction row is rejected before execution.

nemo_curator/stages/audio/alm/alm_data_builder.py
```diff
@@ -124,7 +125,7 @@ def _record_window_loss(  # noqa: PLR0913
 
 
 @dataclass
-class ALMDataBuilderStage(ProcessingStage[AudioTask, AudioTask]):
+class ALMDataBuilderStage(AgentReady, ProcessingStage[AudioTask, AudioTask]):
```
mohammadaaftabv, line 128: P1.** This PR converts nine source stages but deliberately changes no tests. That
violates both `.cursor/rules/coding-standards.mdc` (“all source changes” need relevant
tests) and the dependency's agent-ready checklist, which requires one
`assert_agent_ready(...)` test per converted stage. The 67 focused legacy ALM tests
pass, but none exercises contract/runtime parity; the first direct conformance call on
`ALMDataOverlapStage` fails on eight undeclared outputs, and the other deterministic
probes expose false key-flow and gate claims. Add fixture-based conformance coverage for
all nine stages, including branch/cardinality cases, custom key names, static-versus-
configured gates for required-argument stages, and the durable-output delta case before
merging.

nemo_curator/stages/audio/alm/pretrain/extraction.py
```diff
@@ -115,6 +116,35 @@ def inputs(self) -> tuple[list[str], list[str]]:
     def outputs(self) -> tuple[list[str], list[str]]:
         return [], [self.audio_filepath_key, "snippet_id", "duration", "segments"]
 
+    def describe(self) -> StageContract:
+        return StageContract(
```
mohammadaaftabv, line 129: P1.** `SnippetExtractionStage` and `PretrainMetricsAggregatorStage` require path
constructor arguments but define no `AGENT_STATIC`, so the sanctioned instance-free
catalog cannot recover their configured gates. Exact static discovery reports
`writes_to_disk=False`, `output_path_params=None`, `lifecycle_side_effects=False`, and
no stable-ID/row-independence constraint; configured default extraction reports
`True`, both output paths, `True`, and stable-ID required, while configured metrics
reports disk/lifecycle `True` and row-independence `False`. 

This is the same regression
class PR 2332 already fixed and tests for `ManifestWriterStage`. The required-argument
`SnippetRepetitionFilterStage` has the same issue at `planning.py:549-562`: static
discovery says no first-run internet while a Hub identifier says `True`, and its
`hf_token` is not named in `runtime_secrets`. 


Add conservative `StaticHints` for every
invariant/possible safety gate and static-versus-configured regressions. Also either
move the dry-run return before the directory creation at `extraction.py:148-164` or
keep its configured lifecycle gate true; today a supposedly side-effect-free dry run
creates both output directories.

mohammadaaftabv (changes requested):

nemo_curator/stages/audio/alm/pretrain/io.py
shubhamNvidia: Addressed in 2ae98808. Manifest writing and normal extraction now declare per_row_independent=False; extraction declares independence only for its side-effect-free dry-run mode. Added configured/static contract regressions for these declarations.
For context, the standard ALM reader has no include_files narrowing parameter, so the full agent already refuses changed-files-only execution for that source. The reported manual narrowed rerun demonstrates why the previous contract was inaccurate, but it is not the standard agent execution path.
This change corrects the declarations. Durable delta merging and preservation of untouched tar members remain unsupported.

nemo_curator/stages/audio/alm/alm_data_builder.py
shubhamNvidia: Fixed in `2ae98808`. The default contract now guarantees `audio_filepath`, so downstream file consumers validate correctly. When explicitly configured in `drop_fields_top_level`, runtime and contract consistently remove it. Added planner and branch regressions.

nemo_curator/stages/audio/alm/alm_data_overlap.py
shubhamNvidia: Fixed in `2ae98808`. All nine outputs now have configurable key fields and are consistently used by runtime, `outputs()`, and `describe()`. Added default and renamed-key conformance tests.

nemo_curator/stages/audio/alm/pretrain/extraction.py
shubhamNvidia: Fixed conservatively in `2ae98808`. Extraction now declares `preserves_upstream_keys=False`, reports its guaranteed removals, and exposes only the stable snippet output shape. Added normal, empty-plan, unreadable-source, and failure-path coverage.

nemo_curator/stages/audio/alm/pretrain/extraction.py
shubhamNvidia: Fixed in `2ae98808`. The tar-internal member name is no longer advertised as an ordinary filesystem output. Added a planner regression ensuring extraction cannot satisfy a normal file consumer.

nemo_curator/stages/audio/alm/pretrain/io.py
shubhamNvidia: Fixed in `2ae98808`. Manifest rows are now skipped with a warning when the row is malformed or `segments` is not a list. Added missing/malformed-segment conformance coverage.

nemo_curator/stages/audio/alm/pretrain/io.py
shubhamNvidia: Fixed in `2ae98808`. The manifest writer now declares its configurable snippet-ID discriminator, and metrics declares configurable `id`, `snippet_id`, `segments`, and `duration` reads plus required metadata. Added validation and positional-compatibility regressions.

nemo_curator/stages/audio/alm/alm_data_builder.py
shubhamNvidia: Fixed in `2ae98808`. Added conformance coverage for all nine converted ALM stages, including branch behavior, custom keys, planner composition, static gates, and delta safety.

nemo_curator/stages/audio/alm/pretrain/extraction.py
shubhamNvidia: Fixed in `2ae98808`. Extraction, metrics, and repetition filtering now provide conservative static hints. Repetition filtering reports its token/internet requirements, and extraction dry-run returns before creating output directories. Added static-versus-configured and dry-run regressions.

nemo_curator/stages/audio/alm/alm_data_builder.py
mohammadaaftabv: P1 — Preserve the legacy low-rate audio path

Location: nemo_curator/stages/audio/alm/alm_data_builder.py:239-240

P1 — Preserve the legacy low-rate audio path. This contracts-only patch changes
_process_single_entry() for an existing configuration. At base 7961210f, the
low-sample-rate return always included "audio_filepath": audio_file regardless of
drop_fields_top_level; at head 92eb6c78, this guard omits it when that list contains
the configured path. 
A deterministic ALMDataBuilderStage(drop_fields_top_level="words,segments,audio_filepath")
reproduction returns audio_filepath="legacy.wav" on base and no path on head, so
rejected/low-rate rows now lose source identity and downstream consumers can break.
Restore the unconditional legacy branch and represent its branch-specific pass-through
separately in the agent contract. Add a base-equivalence regression for both normal and
low-rate rows.

nemo_curator/stages/audio/alm/pretrain/extraction.py
```diff
@@ -314,6 +347,7 @@ def _make_snippet_task(
         # Drop source-file-specific fields that don't apply to the snippet.
         new_data.pop("audio_size", None)
         new_data.pop("resampled_audio_filepath", None)
+        new_data["id"] = str(new_data.get("id") or task.task_id)
```
mohammadaaftabv: P1 — Preserve the source ID's exact value and type

nemo_curator/stages/audio/alm/pretrain/extraction.py:350
nemo_curator/stages/audio/alm/pretrain/extraction.py:378-380

These new assignments normalize or replace the existing row ID in both normal and stub outputs. Base preserves the caller's value; head converts nonzero numeric IDs to strings and treats a valid
falsey ID such as 0 as missing, replacing it with task.task_id. 

In a deterministic dry-run, id=0, task_id="FRAMEWORK" changes from 0 on base to "FRAMEWORK" on
head. That changes durable manifest identity and can break joins and metrics keyed by the
original value; the tutorial also promises that the original ID is preserved unchanged.
Keep the task-data ID verbatim and compute a separate string only for snippet/tar-member
naming. 

Any normalization or synthesis must be an explicit opt-in. Add normal and stub
regressions for string, numeric, empty, and zero-valued IDs.

mohammadaaftabv (changes requested):

nemo_curator/stages/audio/alm/pretrain/io.py
mohammadaaftabv: P1 — Do not silently drop legacy reader rows by default

nemo_curator/stages/audio/alm/pretrain/io.py:103-112
nemo_curator/stages/audio/alm/pretrain/io.py:215-217

P1 — The contract fix introduced a data-plane compatibility regression. A manifest
row containing id and audio_filepath but no segments was previously emitted by
this reusable reader; current code silently drops it. 

The exact base/head reproduction changes from one output row to zero. Because this PR is scoped as contracts-only/no behavior change, stricter schema enforcement cannot become the default. Preserve the
permissive default, or add an explicit strict-schema option whose default retains the old
behavior and make the configured contract reflect the selected mode. Add a base-versus-
head regression proving default row preservation.

nemo_curator/stages/audio/alm/pretrain/io.py
mohammadaaftabv: nemo_curator/stages/audio/alm/pretrain/io.py:289-290
nemo_curator/stages/audio/alm/pretrain/io.py:392-393

ProcessingStage.process_batch() now rejects rows the sinks intentionally tolerated
before this PR: the manifest writer treated a missing snippet_id as a stub and passed
it through, while metrics skipped a missing ID and defaulted absent
snippet/segment/duration values. 

Base validates and returns those rows; head fails
validation and raises ValueError. Keep conservative describe().reads for agent
planning, but restore the legacy inputs() behavior, or put strict runtime validation
behind a new opt-in that defaults off. Add comparative process-batch regressions rather
than tests that enshrine the new rejection.

nemo_curator/stages/audio/alm/pretrain/extraction.py
```diff
@@ -113,25 +123,56 @@ def inputs(self) -> tuple[list[str], list[str]]:
...
+                per_row_independent=self.dry_run,
+            ),
+        )
 
     def setup_on_node(
         self,
         _node_info: NodeInfo | None = None,
         _worker_metadata: WorkerMetadata | None = None,
     ) -> None:
+        if self.dry_run:
```
mohammadaaftabv: nemo_curator/stages/audio/alm/pretrain/extraction.py:162-175

P2 — Dry-run setup no longer performs its established directory creation. Base
creates output_dir and the tar parent before returning; head returns first and creates
neither. For complete compatibility, retain those directory creations and report
lifecycle_side_effects=True in dry-run mode. If a side-effect-free dry run is wanted,
make that a separately justified opt-in rather than changing the existing default under
a contract-only PR.

mohammadaaftabv (changes requested):

nemo_curator/stages/audio/alm/pretrain/io.py
```diff
@@ -154,6 +169,16 @@ def inputs(self) -> tuple[list[str], list[str]]:
     def outputs(self) -> tuple[list[str], list[str]]:
         return [], [self.audio_filepath_key, "id", "segments"]
 
+    def describe(self) -> StageContract:
+        return StageContract(
+            writes=IOSpec(data_keys=[self.audio_filepath_key, "id", "segments"]),
+            cardinality="1:N fan-out",
+            gates=Gates(per_row_independent=True),
```
mohammadaaftabv: P1 — per_row_independent=True is contradicted by this reader's
corpus-wide duplicate checks. 
seen_ids and seen_basenames make a row's
result depend on which other rows are present: a full manifest containing
A: x/foo.wav and B: y/foo.wav raises on B, while a narrowed manifest
containing only B emits it. 

This fails AGENT_READY's “X alone versus X plus
other rows” test, and an optimistic True can let a delta publish rows that a
full run would never produce. Set this gate to False, including the static
view, and add a full-versus-narrowed regression.

nemo_curator/stages/audio/alm/alm_data_builder.py
```diff
@@ -159,17 +167,42 @@ def __post_init__(self) -> None:
...
+            # process() rebuilds task.data selectively (words/segments stripped;
+            # the min_sample_rate branch keeps only a four-key subset).
+            preserves_upstream_keys=False,
+            # Named, not merely implied by preserves_upstream_keys: that flag says the entry is
+            # rebuilt, while this says WHICH keys do not survive. Without the names, a downstream
+            # sink check still counts a waveform this stage has already dropped, and refuses a
+            # pipeline that would have run -- the kind of wrong hard gate that pushes a caller
+            # into faking a value to get past it. Derived from the configured parameter so the
+            # declaration tracks whatever the caller actually set.
+            removes_keys=sorted(self._drop_fields_top_level_set),
```
mohammadaaftabv: P1 — Do not declare generated outputs removed. drop_fields_top_level is
an unrestricted legacy knob, but describe() copies every configured value
into removes_keys even when runtime re-adds that key after filtering. With
drop_fields_top_level="words,segments,windows", the contract both writes and
removes windows; assert_agent_ready() fails with “declared removes_keys
'windows' but it is still present”, and Builder→Overlap validation loses
windows although runtime produces it. Derive removals from the net output
(drop set minus unconditional/conditional rewrites) and add generated/custom
output-key collision regressions. Rejecting this previously accepted
configuration would conflict with the requested compatibility guarantee.

nemo_curator/stages/audio/alm/pretrain/extraction.py
```diff
@@ -113,25 +123,56 @@ def inputs(self) -> tuple[list[str], list[str]]:
...
+            writes=IOSpec(
+                data_keys=["id", "snippet_id", "duration", "segments"],
+                # Disk output only happens when not dry-running; keep this consistent
+                # with the writes_to_disk gate below.
+                produces=[] if self.dry_run else ["disk"],
+            ),
+            cardinality="1:N fan-out",
+            iteration_key=_PLAN_DATA_KEY,
+            preserves_upstream_keys=False,
+            removes_keys=["alignment", _PLAN_DATA_KEY, "audio_size", "resampled_audio_filepath"],
```
mohammadaaftabv: P1 — removes_keys is false for supported path aliases.
audio_filepath_key="alignment" worked before this PR. 

Runtime pops that field and then re-adds it as the emitted tar-member name, while the configured
contract says it is absent; assert_agent_ready() fails with “declared
removes_keys 'alignment' still present in result 0”. 

Derive removals after
excluding self.audio_filepath_key and test every removal alias. 

Do not reject alignment, audio_size, or resampled_audio_filepath merely to satisfy the
contract because that would itself remove pre-PR functionality; reject only
genuinely impossible collisions such as using _snippet_plan as both path and
plan.

mohammadaaftabv (changes requested):

nemo_curator/stages/audio/alm/alm_data_builder.py
mohammadaaftabv: P1 — The claimed custom-key coverage is incomplete. The reader still
hardcodes id/segments; planning hardcodes segments and
_snippet_plan; extraction hardcodes id, snippet_id, duration,
segments, removal keys, and several conditional fields.
build_contract(OverlapFilterStage()).params exposes no key knob. If the
segments semantic role is available under turns, validation reports
ok=True but keys_ok=False (dangling_key), and runtime
process_batch() raises because it can read only literal segments. 

Append
key fields after all legacy dataclass slots, use them consistently in runtime,
contracts, and the pipeline factory, and add a renamed end-to-end chain.

The required capability cards are also absent. The nested AGENTS.md requires
cards under nemo_curator/audio_agent/knowledge/cards/ and a clean card audit,
but that package/path is absent even from the tested PR2332 stack. Treat this as
an unresolved dependency/stack deliverable before claiming the AgentReady
checklist complete.

nemo_curator/stages/audio/alm/alm_data_builder.py
shubhamNvidia: Thanks for catching this. The low-sample-rate branch now always preserves the configured audio path, matching the pre-PR behavior even when it appears in drop_fields_top_level. The branch-specific pass-through is represented as a conditional write, with regressions for both normal and low-rate rows.

nemo_curator/stages/audio/alm/pretrain/extraction.py
shubhamNvidia: Fixed. Extraction no longer normalizes or replaces the source ID. The original value and type are preserved for normal and stub outputs, while a separate string fallback is used only for snippet and tar-member naming. Tests cover string, numeric, empty, and zero-valued IDs.

nemo_curator/stages/audio/alm/pretrain/io.py
shubhamNvidia: Agreed. The standalone reader is permissive by default again, so legacy rows without segments continue to be emitted. Strict schema validation is now explicit through strict_schema=True, which the standard pretraining pipeline enables before the planning stages. The configured contract reflects both modes.

nemo_curator/stages/audio/alm/pretrain/io.py
shubhamNvidia: Fixed. The writer and metrics stages retain their conservative describe().reads declarations for agent planning, while inputs() again preserves the legacy permissive process_batch() behavior. Regression tests cover missing discriminator and metrics fields.

nemo_curator/stages/audio/alm/pretrain/extraction.py
shubhamNvidia: Fixed. Dry-run setup once again creates output_dir and the tar parent directory without creating a tar file. The configured lifecycle contract now reports lifecycle_side_effects=True in dry-run mode, matching runtime behavior.

nemo_curator/stages/audio/alm/pretrain/io.py
shubhamNvidia: Agreed. The reader is now marked per_row_independent=False in both static and configured contracts because duplicate ID and basename checks depend on the complete manifest. Coverage verifies the conservative gate.

nemo_curator/stages/audio/alm/alm_data_builder.py
shubhamNvidia: Fixed. removes_keys is now derived from the effective drop set after excluding generated outputs and branch-specific rewrites. Existing drop_fields_top_level configurations remain supported, and tests cover generated-field drops and invalid new output-key collisions.

nemo_curator/stages/audio/alm/pretrain/extraction.py
shubhamNvidia: Fixed. Extraction now calculates net removals after excluding the configured audio path, so supported aliases such as alignment, audio_size, and resampled_audio_filepath are re-added without being declared removed. Only genuinely impossible collisions, such as sharing the path and plan key, are rejected.

nemo_curator/stages/audio/alm/alm_data_builder.py
shubhamNvidia: The module-side issue is fixed. Configurable keys are now propagated consistently through the reader, overlap filter, planner, repetition filter, extraction, writer, metrics aggregator, and pipeline factory. A renamed end-to-end chain verifies runtime and contract alignment.\n\nThe nine ALM capability cards are present in the synchronized agent/adv integration branch. They are not duplicated in this module PR because the audio_agent package is delivered through the foundation/integration stack.

nemo_curator/stages/audio/alm/pretrain/pipeline.py
```diff
@@ -95,6 +95,11 @@ def build_audio_pretrain_pipeline(  # noqa: PLR0913
     audio_path_resolution: str = AUDIO_PATH_RESOLUTION_BASENAME,
     dataset_name: str = "long_form_audio",
     dry_run: bool = False,
+    id_key: str = "id",
```
mohammadaaftabv, line 98: P0 (blocker) — Thread renamed fields through reconciliation. 

The new factory knobs make id_key, segments_key, and duration_key supported configurations, but finalization still rebuilds post-reconcile metrics from literal id, segments, and duration. This path runs whenever even one tar member is missing or corrupt. With the exact renamed schema used by the new chain test (source_id, turns, clip_duration), one valid surviving manifest row is recomputed as zero snippets, zero segments, and zero duration, and every per_original[*].out_* value is reset to zero. 

This silently corrupts metrics for an advertised valid audio configuration. Pass the three configured keys through finalize_audio_pretrain_outputs() into the collector, update the driver call, and add a renamed-schema regression where one of two tar members is reconciled away while the survivor's metrics remain intact.

nemo_curator/stages/audio/alm/alm_data_builder.py
```diff
@@ -148,6 +149,13 @@ class ALMDataBuilderStage(ProcessingStage[AudioTask, AudioTask]):
 
     # Top-level fields to drop from output entry (comma-separated)
     drop_fields_top_level: str = "words,segments"
```
mohammadaaftabv, line 151: P1 — Apply the default segment drop to segments_key. 

The new segments_key remaps the read but not drop_fields_top_level, whose default remains the literal segments. 

Consequently, ALMDataBuilderStage(segments_key="turns") keeps the complete source turns list on normal-rate rows even though the default-key stage drops its source segments. The contract simultaneously says upstream keys are not preserved, so a Builder -> consumer-of-turns chain runs successfully but validation rejects it with key_removed_upstream; the low-rate branch drops turns again. Resolve the default semantic drop through the configured key (or declare the pass-through if retention is intentional) and cover both sample-rate branches with a renamed-key runtime/contract regression.

nemo_curator/stages/audio/alm/pretrain/extraction.py
```diff
@@ -106,14 +153,77 @@ def __post_init__(self) -> None:
...
+        ]
+        return StageContract(
+            reads=IOSpec(data_keys=[self.audio_filepath_key, self.snippet_plan_key, self.id_key], accepts=["file"]),
+            writes=IOSpec(
+                data_keys=output_keys,
+                # Disk output only happens when not dry-running; keep this consistent
+                # with the writes_to_disk gate below.
+                produces=[] if self.dry_run else ["disk"],
+            ),
+            conditional_writes=[
```
mohammadaaftabv: P1 — Make conditional extraction writes depend on their literal source keys. 

The six transformed pass-through fields are currently declared as conditional writes with requires_keys=[] and are absent from optional_reads, although runtime emits each one only when that exact input key exists. As a result, an input with only id, audio_filepath, and _snippet_plan validates successfully before a downstream consumer of text, yet extraction does not emit text and that consumer raises. Declare these keys as optional reads and set each conditional write's requires_keys=[that_key]. Add absent/present regressions proving the planner does not advertise text (or the other five fields) without its literal source and still propagates it when present.

nemo_curator/stages/audio/alm/alm_data_builder.py
```diff
@@ -157,19 +165,69 @@ def __post_init__(self) -> None:
...
+                ConditionalWrite(
+                    writes=IOSpec(data_keys=[self.audio_filepath_key]),
+                    condition=(
+                        f"'{self.audio_sample_rate_key}' is below min_sample_rate, so the legacy low-rate branch "
+                        f"preserves '{self.audio_filepath_key}'"
+                    ),
+                    value_origin="upstream_same_key",
+                )
+            )
+        return StageContract(
```
mohammadaaftabv, line 211: P1 — Declare the optional Swift-path read. process() reads task.data[self.swift_audio_filepath_key] when present and places that value into generated stats, but describe() exposes neither that key nor any optional read. This hides a real data dependency from discovery and planning. Add swift_audio_filepath_key to optional_reads and add present/absent conformance tests showing the dependency is optional while its provenance is preserved when supplied.

nemo_curator/stages/audio/alm/alm_data_overlap.py
```diff
@@ -172,18 +197,40 @@ def __post_init__(self) -> None:
...
+            self.total_dur_list_window_timestamps_key,
+            self.filtered_key,
+            self.filtered_windows_key,
+            self.filtered_dur_key,
+            self.filtered_dur_list_key,
+            self.manifest_filepath_key,
+            self.swift_filepath_key,
+        ]
+
+    def describe(self) -> StageContract:
```
mohammadaaftabv, line 215: P1 follow-up — the stats input is still missing from the contract. 

The overlap stage now declares its configurable outputs, but process() also reads task.data[self.stats_key] when available to derive manifest and Swift provenance. describe() lists only windows_key, so discovery cannot see that optional dependency, and the new contract test does not check it. Add stats_key to optional_reads and cover both present and absent stats inputs.

nemo_curator/stages/audio/alm/pretrain/extraction.py
```diff
@@ -106,14 +153,77 @@ def __post_init__(self) -> None:
...
         self._tar_shard_path: str | None = None
         self._tar: Any = None  # tarfile.TarFile, opened lazily in setup()
 
     def inputs(self) -> tuple[list[str], list[str]]:
-        return [], [self.audio_filepath_key, _PLAN_DATA_KEY]
+        return [], [self.audio_filepath_key, self.snippet_plan_key]
 
     def outputs(self) -> tuple[list[str], list[str]]:
-        return [], [self.audio_filepath_key, "snippet_id", "duration", "segments"]
+        return [], [self.id_key, self.snippet_id_key, self.duration_key, self.segments_key]
```
mohammadaaftabv: P1 — Preserve the legacy outputs() schema and make ID propagation branch-aware. 

At merge-base 7961210f, SnippetExtractionStage.outputs() returns [audio_filepath_key, "snippet_id", "duration", "segments"]; current head returns [id_key, snippet_id_key, duration_key, segments_key]. This removes the audio key that every normal result contains (and each stub contains with value None) and adds id_key as guaranteed even though a supported missing-ID normal result omits it; the stub branch instead includes id_key=None. The exact default tuple changes from ['audio_filepath', 'snippet_id', 'duration', 'segments'] to ['id', 'snippet_id', 'duration', 'segments'], breaking Pipeline.describe() and callers that introspect the existing stage API. 

Keep the exact legacy outputs() tuple, build the agent-only describe() shape independently so a tar member is not advertised as an ordinary file, and model source ID according to those real branches rather than as a guaranteed write. Add exact base-tuple and missing-ID normal/stub regressions.

nemo_curator/stages/audio/alm/pretrain/io.py
```diff
@@ -147,12 +194,41 @@ class ReadLongFormManifestStage(ProcessingStage[EmptyTask, AudioTask]):
...
+    id_key: str = "id"
+    segments_key: str = "segments"
+    strict_schema: bool = False
+    INTERNAL_KEY_FIELDS: ClassVar[frozenset[str]] = frozenset({"id_key"})
+    AGENT_STATIC: ClassVar[StaticHints] = StaticHints(gates=Gates(per_row_independent=False))
 
     def inputs(self) -> tuple[list[str], list[str]]:
         return [], []
 
     def outputs(self) -> tuple[list[str], list[str]]:
```
mohammadaaftabv, line 206: P1 compatibility follow-up — keep the reader's legacy outputs() stable. 

Making the standalone reader permissive by default restores its row behavior, but current default outputs() now returns only [audio_filepath_key, id_key]; base always returns [audio_filepath_key, "id", "segments"]. 

This removes segments from Pipeline.describe() and any caller using the legacy tuple even when the manifest contains segments. Preserve the exact legacy outputs() result and express permissive-mode conditionality only in describe().conditional_writes; those two surfaces serve different compatibility and agent-planning purposes. Add an exact default-tuple regression against 7961210f.

nemo_curator/stages/audio/alm/pretrain/pipeline.py
```diff
@@ -195,19 +210,25 @@ def build_audio_pretrain_pipeline(  # noqa: PLR0913
                 audio_filepath_key=audio_filepath_key,
                 audio_path_resolution=audio_path_resolution,
                 dataset_name=dataset_name,
+                id_key=id_key,
+                segments_key=segments_key,
+                strict_schema=True,
```
mohammadaaftabv: P1 compatibility follow-up — do not turn the legacy failure into silent partial output. 

The standalone default is permissive again, but the standard factory hard-codes strict_schema=True, whose implementation skips malformed rows. 
With a two-row manifest where A has id/audio_filepath but no segments and B is valid, base emits both and the next stage raises on A, so the run fails; current head silently drops A and successfully advances B. That fail-fast-to-partial-success change can publish an incomplete corpus under a PR advertised as contracts-only/no behavior change. 

Make strict factory validation raise on an invalid row, or retain the legacy reader flow and let the next stage reject it. If skip-and-continue is desired, expose it as a separately named opt-in policy. Add a mixed valid/invalid end-to-end regression proving the default never converts rejected input into partial success.

nemo_curator/stages/audio/alm/pretrain/extraction.py
```diff
@@ -106,14 +153,77 @@ def __post_init__(self) -> None:
...
             msg = "target_sample_rate must be > 0"
             raise ValueError(msg)
+        stable_output_keys = [self.id_key, self.snippet_id_key, self.duration_key, self.segments_key]
+        if len(set(stable_output_keys)) != len(stable_output_keys):
+            msg = "id, snippet, duration, and segments output keys must be distinct"
+            raise ValueError(msg)
+        if self.snippet_plan_key in {self.audio_filepath_key, *stable_output_keys}:
+            msg = "snippet_plan_key must be distinct from input and output keys"
+            raise ValueError(msg)
+        if self.audio_filepath_key in set(stable_output_keys):
```
mohammadaaftabv: P0 (blocker) — Do not crash a pre-PR audio_filepath_key configuration to satisfy the new contract. audio_filepath_key was already public and had no exclusion against "id". 

At the merge-base, SnippetExtractionStage(..., audio_filepath_key="id", dry_run=True) constructs, processes, and emits a task with filepath_key="id"; the standard pipeline factory also constructs with that value. Current head raises ValueError("audio_filepath_key must be distinct from stable output keys") before any work runs, so a previously accepted audio configuration now crashes during construction. 

Preserve legacy runtime construction and behavior, handling semantic ambiguity only in the new agent surface (for example by marking that configuration non-wrappable), rather than narrowing an existing parameter. Add a base-equivalence stage and factory regression for the "id" path alias.

mohammadaaftabv (changes requested):

nemo_curator/stages/audio/alm/pretrain/pipeline.py
shubhamNvidia: Thanks for catching this. Finalization now uses the configured ID, segments, and duration keys when rebuilding post-reconciliation metrics. I also added regression coverage for renamed manifest schemas.

nemo_curator/stages/audio/alm/alm_data_builder.py
shubhamNvidia: Fixed. The legacy default segments drop now resolves to the configured segments_key, so renamed segment fields are consistently removed from normal outputs.

nemo_curator/stages/audio/alm/pretrain/extraction.py
shubhamNvidia: Fixed. Optional source fields are now declared through optional_reads, and each conditional pass-through write requires its corresponding upstream key.

nemo_curator/stages/audio/alm/alm_data_builder.py
shubhamNvidia: Fixed. swift_audio_filepath_key is now declared as an optional read because the builder only consults it when present.

nemo_curator/stages/audio/alm/alm_data_overlap.py
shubhamNvidia: Fixed. stats_key is now declared as an optional read while remaining non-blocking when absent.

nemo_curator/stages/audio/alm/pretrain/extraction.py
shubhamNvidia: Fixed. The legacy outputs() tuple is restored. Configurable agent behavior is represented independently in describe(), including conditional ID preservation and invalidation of tar-member paths for filesystem consumers.

nemo_curator/stages/audio/alm/pretrain/io.py
shubhamNvidia: Fixed. The reader’s historical outputs() tuple is restored, while strict and permissive segment behavior remains accurately represented in describe().

nemo_curator/stages/audio/alm/pretrain/pipeline.py
shubhamNvidia: Fixed. The standard pipeline no longer forces strict row skipping and preserves the legacy processing behavior. Strict schema validation remains available as an explicit stage option.

nemo_curator/stages/audio/alm/pretrain/extraction.py
shubhamNvidia: Fixed. The previously supported audio_filepath_key="id" configuration is constructible again. Because that alias is ambiguous for agent planning, it is conservatively marked non-wrappable without changing legacy runtime behavior. The focused ALM and agent-foundation regression suites pass.

nemo_curator/stages/audio/alm/alm_data_overlap.py
```diff
@@ -161,6 +162,30 @@ class ALMDataOverlapStage(ProcessingStage[AudioTask, AudioTask]):
     # Processing parameters (EXACT match to SDP)
     overlap_percentage: int = 0
     target_duration: float = 120.0
+    windows_key: str = "windows"
+    filtered_windows_key: str = "filtered_windows"
+    stats_key: str = "stats"
```
shubhamNvidia: **[P1] Reject aliases among generated output keys.** These fields are independent public knobs, but `__post_init__` does not check that the generated outputs are distinct. For example, `ALMDataOverlapStage(filtered_windows_key="collision", filtered_dur_key="collision")` constructs successfully; on a populated row `_filter_overlaps()` writes the windows list and then overwrites it with the duration float, after which `process()` calls `len()` on that float and raises `TypeError`. Other pairings can silently replace one advertised result. Please reject output-key collisions at construction and add a populated-row regression for this case.

shubhamNvidia (commented):
reviewed

shubhamNvidia (commented):
reviewed

PR #2180, opened 2026-07-07
feat(audio-datasets): agent-ready initial-manifest stages (FLEURS, ReadSpeech)

shubhamNvidia:
## Summary
Makes the dataset initial-manifest stages agent-ready. Additive only.

- `datasets/fleurs/create_initial_manifest.py`, `datasets/readspeech/create_initial_manifest.py` — subclass `AgentReady` and add `describe()` / `StageContract`. These are source stages: they read a dataset layout and emit `audio_filepath` + metadata records, which the contract now declares. Manifest-creation logic unchanged.

## Notes for reviewers
- Additive, backward compatible.

## Test plan
- [ ] `pytest tests/stages/audio/datasets -q`

PR #2341, opened 2026-08-25
feat(audio-datasets): make dataset manifest stages agent-ready

shubhamNvidia:
**Contracts only — no behavior change.** Adds `describe()` + `*_key` params to the FLEURS and ReadSpeech initial-manifest source stages.

These are *sources* (they produce audio paths + text from nothing), and their contracts declare the two facts a planner has to know up front: they **write to disk** and they **need internet on the first run** (when `auto_download` is on) — so an offline or read-only environment is caught before the run starts rather than partway through a download.

### DIFF
--- a/nemo_curator/stages/audio/datasets/fleurs/create_initial_manifest.py
+++ b/nemo_curator/stages/audio/datasets/fleurs/create_initial_manifest.py
@@ -19,6 +19,7 @@ from dataclasses import dataclass
 from huggingface_hub import hf_hub_download
 from loguru import logger
 
+from nemo_curator.stages.audio._agent._agent_ready import AgentReady, Gates, IOSpec, StageContract
 from nemo_curator.stages.audio.datasets.file_utils import extract_archive
 from nemo_curator.stages.base import ProcessingStage
 from nemo_curator.tasks import AudioTask, EmptyTask
@@ -40,7 +41,7 @@ def get_fleurs_filenames(lang: str, split: str) -> tuple[str, str]:
 
 
 @dataclass
-class CreateInitialManifestFleursStage(ProcessingStage[EmptyTask, AudioTask]):
+class CreateInitialManifestFleursStage(AgentReady, ProcessingStage[EmptyTask, AudioTask]):
     """Create initial manifest for the FLEURS dataset.
 
     Dataset link: https://huggingface.co/datasets/google/fleurs
@@ -91,6 +92,19 @@ class CreateInitialManifestFleursStage(ProcessingStage[EmptyTask, AudioTask]):
 
     def outputs(self) -> tuple[list[str], list[str]]:
         return [], [self.filepath_key, self.text_key]
+
+    def describe(self) -> StageContract:
+        return StageContract(
+            writes=IOSpec(data_keys=[self.filepath_key, self.text_key], produces=["disk"]),
+            cardinality="1:N fan-out",
+            # The download stages the split once; after that every emitted row's path and text
+            # come from its own transcript line, and no line is dropped or reordered.
+            gates=Gates(
+                writes_to_disk=True,
+                requires_internet_first_run=self.auto_download,
+                per_row_independent=True,
+            ),
+        )
 
     def language_data_dir(self) -> str:
         """Return the per-language download directory under ``raw_data_dir``.
--- a/nemo_curator/stages/audio/datasets/readspeech/create_initial_manifest.py
+++ b/nemo_curator/stages/audio/datasets/readspeech/create_initial_manifest.py
@@ -19,6 +19,7 @@ from dataclasses import dataclass
 
 from loguru import logger
 
+from nemo_curator.stages.audio._agent._agent_ready import AgentReady, Gates, IOSpec, StageContract
 from nemo_curator.stages.audio.datasets.file_utils import download_file
 from nemo_curator.stages.base import ProcessingStage
 from nemo_curator.tasks import AudioTask, EmptyTask
@@ -33,7 +34,7 @@ DNS_READSPEECH_URL = (
 
 
 @dataclass
-class CreateInitialManifestReadSpeechStage(ProcessingStage[EmptyTask, AudioTask]):
+class CreateInitialManifestReadSpeechStage(AgentReady, ProcessingStage[EmptyTask, AudioTask]):
     """
     Stage to create initial manifest for the DNS Challenge Read Speech dataset.
 
@@ -65,6 +66,23 @@ class CreateInitialManifestReadSpeechStage(ProcessingStage[EmptyTask, AudioTask]
 
     def outputs(self) -> tuple[list[str], list[str]]:
         return [], [self.filepath_key, self.text_key]
+
+    def describe(self) -> StageContract:
+        return StageContract(
+            writes=IOSpec(data_keys=[self.filepath_key, self.text_key], produces=["disk"]),
+            cardinality="1:N fan-out",
+            gates=Gates(
+                writes_to_disk=True,
+                requires_internet_first_run=self.auto_download,
+                output_path_params=[],
+                # Each row is parsed from its own filename. Declared True unconditionally BY
+                # DECISION: ``max_samples`` (default 5000 of ~14k) truncates the SORTED listing,
+                # so a delta enumerating only changed files can admit files a full run would not
+                # have. Accepted rather than deny reuse to the default configuration -- not an
+                # oversight to "fix" back. Same call as CreateInitialManifestAudioFolderStage.
+                per_row_independent=True,
+            ),
+        )
 
     def download_and_extract(self) -> str:
         """Download and extract DNS Challenge Read Speech dataset (~4.88 GB)."""


nemo_curator/stages/audio/datasets/fleurs/create_initial_manifest.py
```diff
@@ -92,6 +93,19 @@ def inputs(self) -> tuple[list[str], list[str]]:
     def outputs(self) -> tuple[list[str], list[str]]:
         return [], [self.filepath_key, self.text_key]
```
mohammadaaftabv, line 94: declares only filepath_key and text_key, but
nemo_curator/stages/audio/datasets/readspeech/create_initial_manifest.py:267-273
unconditionally emits sample_rate, book_id, and reader_id as well. Those
three outputs also remain hard-coded literals, contrary to the dependency's
rule that every task.data read/write key be a configurable *_key field.

A local run with filepath_key="path_custom" and text_key="txt_custom"
produced this exact mismatch:

declared_writes = ['path_custom', 'txt_custom']
actual_keys     = ['book_id', 'path_custom', 'reader_id', 'sample_rate', 'txt_custom']
undeclared      = ['book_id', 'reader_id', 'sample_rate']

This is not merely incomplete documentation. The planner rejects the valid
pipeline ReadSpeech -> PreserveByValueStage("sample_rate", 48000) with
unsatisfied_reads, even though every row produced at runtime contains that
value. The same defect was called out on superseded PR #2180 and remains in
this replacement; no current #2341 thread duplicates it.

Add configurable fields for all three outputs, use them in outputs(),
collect_audio_files(), verify_dataset_structure(), and describe(), and
declare their shared roles or explicit stage-internal status. Add both default-
key and renamed-key regressions that compare the complete runtime output schema
with the contract.

nemo_curator/stages/audio/datasets/fleurs/create_initial_manifest.py
mohammadaaftabv, line 74: Making the class agent-ready at
nemo_curator/stages/audio/datasets/fleurs/create_initial_manifest.py:44
exposes the existing fields at lines 74-88 through the public generated schema.
lang, split, and raw_data_dir all have "" defaults, so stage_params
reports required=False for each and the JSON schema has no required list.
Yet the advertised no-argument configuration immediately raises
ValueError: lang is required for CreateInitialManifestFleursStage.

The mismatch also prevents producer discovery: the catalog's safe default
contract probe returns None, so this newly agent-ready source is placed in
unresolved_stages instead of being indexed as an audio-path/text producer.
Make the three fields genuinely required dataclass arguments (reordering them
before defaulted fields), or add an equally authoritative supported required-
parameter mechanism, and regress the public parameter schema plus default
contract/role-index discovery.

nemo_curator/stages/audio/datasets/fleurs/create_initial_manifest.py
```diff
@@ -92,6 +93,19 @@ def inputs(self) -> tuple[list[str], list[str]]:
...
 
+    def describe(self) -> StageContract:
+        return StageContract(
```
mohammadaaftabv, line 97: The new gates at
nemo_curator/stages/audio/datasets/fleurs/create_initial_manifest.py:98-105
and
nemo_curator/stages/audio/datasets/readspeech/create_initial_manifest.py:72-83
are unsafe in opposite directions:

Neither class supplies conservative AGENT_STATIC hints. Sanctioned
instance-free discovery reports writes_to_disk=False,
requires_internet_first_run=False, output_path_params=None, and unknown
row independence for both default auto_download=True downloaders. An
offline/read-only plan can therefore be accepted before construction.

Configured auto_download=False instances still claim a disk product and
writes_to_disk=True, although FLEURS lines 221-233 and ReadSpeech lines
332-345 only locate/scan pre-staged files. That can reject valid offline,
read-only dataset sources.

With downloads enabled, FLEURS leaves output_path_params=None and
ReadSpeech explicitly declares [] at line 77. The dependency defines these
values as no usable redirectable destination, but both stages download and
extract beneath raw_data_dir; FLEURS can also write its Hugging Face cache.
A safety-conscious smoke therefore has to refuse the stages instead of
redirecting their writes.

Make the configured disk product/write claims conditional on auto_download,
name at least raw_data_dir as the download staging output (and ensure FLEURS'
HF cache is sandboxed), and add conservative StaticHints for the default
disk/network/path/row-independence facts. Test the static contract and both
configured download modes.

nemo_curator/stages/audio/datasets/readspeech/create_initial_manifest.py
mohammadaaftabv, line 1: PR #2341 changes no tests. Current .cursor/rules/coding-standards.mdc:80
requires relevant tests for every source change, while the declared
dependency's nemo_curator/stages/audio/AGENT_READY.md:278-310 specifically
requires one assert_agent_ready(...) test per converted stage. The 24 legacy
dataset tests contain no assertions for describe, build_contract, static
discovery, key remapping, or gate/path behavior; that is why they all pass
despite the three reproducible contract failures above.

Add model/network-free conformance fixtures for both stages, covering default
and custom keys, complete emitted schemas, static contracts, auto_download
true/false gates, redirectable output paths, and FLEURS constructor requiredness.

mohammadaaftabv (changes requested):

nemo_curator/stages/audio/datasets/fleurs/create_initial_manifest.py
```diff
@@ -92,6 +93,19 @@ def inputs(self) -> tuple[list[str], list[str]]:
...
+            writes=IOSpec(data_keys=[self.filepath_key, self.text_key], produces=["disk"]),
+            cardinality="1:N fan-out",
+            # The download stages the split once; after that every emitted row's path and text
+            # come from its own transcript line, and no line is dropped or reordered.
```
melllinia: The comment says "no line is dropped or reordered" but process_transcript at lines 131-132 does exactly that, we should mark it in the comments.

nemo_curator/stages/audio/datasets/fleurs/create_initial_manifest.py
```diff
@@ -92,6 +93,19 @@ def inputs(self) -> tuple[list[str], list[str]]:
     def outputs(self) -> tuple[list[str], list[str]]:
         return [], [self.filepath_key, self.text_key]
 
+    def describe(self) -> StageContract:
+        return StageContract(
+            writes=IOSpec(data_keys=[self.filepath_key, self.text_key], produces=["disk"]),
+            cardinality="1:N fan-out",
```
melllinia, line 112: Both stages declare cardinality="1:N fan-out" but neither defines ray_stage_spec() returning {"is_fanout_stage": True} nor pins num_workers() -> 1. I think the backend here will schedule them as ordinary stages.

sarahyurick: It is fanout because `process` does `EmptyTask` -> `list[AudioTask]`. A community contributor helped add automatic fanout stage detection in https://github.com/NVIDIA-NeMo/Curator/pull/2157.

shubhamNvidia: Thanks, Sarah. Confirmed that ProcessingStage.ray_stage_spec() automatically detects the list[AudioTask] return annotation and sets is_fanout_stage=True. Since the source receives one EmptyTask, a num_workers() -> 1 override is also unnecessary.

nemo_curator/stages/audio/datasets/fleurs/create_initial_manifest.py
shubhamNvidia: Good catch. process_transcript() skips malformed rows containing fewer than three TSV columns, although valid rows retain their original order. I’ll update the comment to state this accurately.

nemo_curator/stages/audio/datasets/fleurs/create_initial_manifest.py
shubhamNvidia: Thanks for catching this. sample_rate, book_id, and reader_id now have configurable key fields and are used consistently in runtime output, validation, and the declared contract. Default- and renamed-key regressions verify that the complete runtime schema matches the contract.

nemo_curator/stages/audio/datasets/fleurs/create_initial_manifest.py
shubhamNvidia: Fixed. lang, split, and raw_data_dir are now marked as agent-required while preserving the existing Python constructor defaults and compatibility. Generated schemas and role-index discovery now correctly recognize the required configuration and source outputs.

nemo_curator/stages/audio/datasets/fleurs/create_initial_manifest.py
shubhamNvidia: Fixed. Configured contracts now report disk and network requirements only when auto_download=True. Conservative static hints describe the default download behavior, and redirectable paths include raw_data_dir plus the FLEURS Hugging Face cache_dir. Tests cover static discovery and both download modes.

nemo_curator/stages/audio/datasets/readspeech/create_initial_manifest.py
shubhamNvidia: Added model- and network-free conformance tests for both dataset stages. Coverage includes complete emitted schemas, default and remapped keys, required-parameter discovery, static contracts, conditional download gates, redirectable paths, fan-out behavior, and compatibility preservation.

nemo_curator/stages/audio/datasets/readspeech/create_initial_manifest.py
```diff
@@ -47,15 +49,33 @@ class CreateInitialManifestReadSpeechStage(ProcessingStage[EmptyTask, AudioTask]
...
+    INTERNAL_KEY_FIELDS: ClassVar[frozenset[str]] = frozenset({"book_id_key", "reader_id_key"})
+
     raw_data_dir: str
     max_samples: int = 5000
     auto_download: bool = True
     filepath_key: str = "audio_filepath"
     text_key: str = "text"
     name: str = "CreateInitialManifestReadSpeech"
     batch_size: int = 1
+    sample_rate_key: str = "sample_rate"
```
mohammadaaftabv, line 76: The fields at
nemo_curator/stages/audio/datasets/readspeech/create_initial_manifest.py:76-78
are not checked for emptiness or collisions. This accepted configuration:

CreateInitialManifestReadSpeechStage(
    raw_data_dir=path,
    auto_download=False,
    sample_rate_key="audio_filepath",
)

emits {'audio_filepath': 48000, ...} because the later sample-rate assignment
at lines 301-307 overwrites the real path. Its contract also contains the same
write twice. Validate each newly added key against every legacy/new output key
and add new-vs-legacy and new-vs-new regressions. This can preserve all old
parameters, defaults, and positional order.

nemo_curator/stages/audio/datasets/readspeech/create_initial_manifest.py
```diff
@@ -203,9 +255,9 @@ def _extract_archive(self, archive_path: str, extract_path: str) -> None:
 
     def parse_filename(self, filename: str) -> dict:
```
mohammadaaftabv, line 277: At nemo_curator/stages/audio/datasets/readspeech/create_initial_manifest.py:256-281,
parse_filename() combines configurable book/reader keys with a literal
temporary chapter slot. 
For book_00025_chp_0019_reader_04069_...wav, setting
book_id_key="chapter" returns/emits chapter="0019", losing book 00025.
The contract remains unique and well-formed, so conformance passes. Parse into
fixed local/canonical fields and remap only while constructing the output row;
add this exact regression.

mohammadaaftabv (changes requested):

nemo_curator/stages/audio/datasets/readspeech/create_initial_manifest.py
shubhamNvidia: Thanks for catching this. The newly added ReadSpeech output-key parameters now reject empty names and collisions with both legacy and newly configurable output keys. This prevents runtime overwrites and duplicate contract writes while preserving existing defaults and positional compatibility. Regression tests cover new-to-legacy and new-to-new collisions.

nemo_curator/stages/audio/datasets/readspeech/create_initial_manifest.py
shubhamNvidia: Fixed. Filename metadata is now parsed into fixed canonical fields, and configurable key remapping happens only when constructing the output row. This prevents `book_id_key="chapter"` from overwriting the parsed book ID. The exact filename and remapped-key case from the comment is covered by a regression test.

nemo_curator/stages/audio/datasets/readspeech/create_initial_manifest.py
```diff
@@ -47,24 +49,95 @@ class CreateInitialManifestReadSpeechStage(ProcessingStage[EmptyTask, AudioTask]
...
+            gates=Gates(
+                writes_to_disk=self.auto_download,
+                requires_internet_first_run=self.auto_download,
+                output_path_params=["raw_data_dir"] if self.auto_download else [],
+                # Each row is parsed from its own filename. Declared True unconditionally BY
+                # DECISION: ``max_samples`` (default 5000 of ~14k) truncates the SORTED listing,
+                # so a delta enumerating only changed files can admit files a full run would not
+                # have. Accepted rather than deny reuse to the default configuration -- not an
+                # oversight to "fix" back. Same call as CreateInitialManifestAudioFolderStage.
+                per_row_independent=True,
```
shubhamNvidia: **[P1] Mark bounded sources as corpus-dependent.** The comment immediately above describes why this cannot be `True`: `max_samples` truncates the globally sorted listing, so with `max_samples=1` an `X` row is emitted when processed alone but disappears when an earlier-sorting `A` is present. A delta can therefore publish rows that the full run excludes. This is the exact conditional case documented by `AGENT_READY.md` and implemented by `CreateInitialManifestAudioFolderStage`; the latter does not make the same call claimed here. Please use `per_row_independent=self.max_samples <= 0` (matching this class's unbounded behavior), make the static hint conservative `False` for the default `5000`, and cover bounded and unbounded contracts.

shubhamNvidia (commented):
reviewed

shubhamNvidia (commented):
reviewed

PR #2342, opened 2026-08-25
feat(audio-advanced-pipelines): make AudioDataFilterStage agent-ready

shubhamNvidia:
**Contracts only — no behavior change.** Adds a `describe()` contract to the composite `AudioDataFilterStage` so the planner can see what the whole bundled pipeline reads and writes, and that it writes to disk.

This composite expands into filtering, postprocessing, preprocessing, and segmentation leaf stages. The following agent-ready migrations are hard prerequisites and must merge before this PR is rebased and merged:

- #2332 — shared agent-ready foundation
- #2334 — timestamp mapper/postprocessing contracts
- #2335 — filtering contracts
- #2338 — segmentation contracts

Until those prerequisites land, this PR's standalone tree does not contain the complete contract stack and its CI is expected to remain blocked. The complete integrated stack validates all four supported VAD/speaker topologies (5/6/9/12 leaf stages) cleanly.

### DIFF
--- a/nemo_curator/stages/audio/advanced_pipelines/audio_data_filter/audio_data_filter.py
+++ b/nemo_curator/stages/audio/advanced_pipelines/audio_data_filter/audio_data_filter.py
@@ -46,17 +46,19 @@ if TYPE_CHECKING:
 
 from loguru import logger
 
+from nemo_curator.stages.audio._agent._agent_ready import AgentReady, Gates, StageContract
 from nemo_curator.stages.audio.filtering import BandFilterStage, SIGMOSFilterStage, UTMOSFilterStage
 from nemo_curator.stages.audio.postprocessing import TimestampMapperStage
 from nemo_curator.stages.audio.preprocessing import MonoConversionStage, SegmentConcatenationStage
 from nemo_curator.stages.audio.segmentation import SpeakerSeparationStage, VADSegmentationStage
 from nemo_curator.stages.base import CompositeStage, ProcessingStage
 from nemo_curator.stages.resources import Resources
+from nemo_curator.tasks import AudioTask
 
 from .config import _deep_merge, get_enabled_stages, load_config
 
 
-class AudioDataFilterStage(CompositeStage):
+class AudioDataFilterStage(AgentReady, CompositeStage[AudioTask, AudioTask]):
     """Complete audio data filtering and curation pipeline (CompositeStage).
 
     Decomposes into independent stages that the executor can schedule with
@@ -89,6 +91,16 @@ class AudioDataFilterStage(CompositeStage):
         self._cfg = load_config(config_path)
         if config:
             self._cfg = _deep_merge(self._cfg, config)
+
+    def describe(self) -> StageContract:
+        return StageContract(
+            wrappable=False,
+            # True for every topology because it is true of each delegate: mono conversion, VAD,
+            # the three quality filters, concatenation, speaker separation and the timestamp
+            # mapper all work from the row they are handed. The factories below leave
+            # ``write_to_disk`` unset throughout, so none of them even opens a shared directory.
+            gates=Gates(per_row_independent=True),
+        )
 
     def decompose(self) -> list[ProcessingStage]:
         """Build a self-consistent pipeline topology based on enabled features."""


nemo_curator/stages/audio/advanced_pipelines/audio_data_filter/audio_data_filter.py
```diff
@@ -46,17 +46,19 @@
 
 from loguru import logger
 
+from nemo_curator.stages.audio._agent._agent_ready import AgentReady, Gates, StageContract
```
mohammadaaftabv: P0 — Preserve the existing public AudioDataFilterStage import. 

This eager import makes the established public stage import fail on this PR's
exact tree: _agent is absent from both its recorded base and current
upstream main, so importing AudioDataFilterStage and collecting its
existing tests raise ModuleNotFoundError. Please land #2332 and rebase this
branch.

nemo_curator/stages/audio/advanced_pipelines/audio_data_filter/audio_data_filter.py
```diff
@@ -90,6 +92,16 @@ def __init__(
         if config:
             self._cfg = _deep_merge(self._cfg, config)
 
+    def describe(self) -> StageContract:
```
mohammadaaftabv, line 97: P1 — The leaf-stage migrations are hard prerequisites, not optional merge
ordering. 

The planner does not use this empty composite contract as a
substitute for its children; it expands the configured topology and calls
describe() on every leaf. 
With only #2332 plus this head, all four
VAD/speaker combinations are partly opaque, and the full topology additionally
reports unsatisfied_reads_in_composite. 

Please make #2334 (timestamp
mapper), #2335 (filters), and #2338 (segmentation) hard prerequisites too,
rebase after their current heads, and run the four-topology matrix in CI. On
the complete stack the 5/6/9/12-leaf plans have no opaque/unrunnable entries
or validation issues.

nemo_curator/stages/audio/advanced_pipelines/audio_data_filter/audio_data_filter.py
```diff
@@ -90,6 +92,16 @@ def __init__(
...
+        return StageContract(
+            wrappable=False,
+            # True for every topology because it is true of each delegate: mono conversion, VAD,
+            # the three quality filters, concatenation, speaker separation and the timestamp
+            # mapper all work from the row they are handed. The factories below leave
+            # ``write_to_disk`` unset throughout, so none of them even opens a shared directory.
+            gates=Gates(per_row_independent=True),
```
mohammadaaftabv, line 104: P1 — Publish the invariant gate through static discovery and add the
required contract regression. 
describe() reports
per_row_independent=True, but the public static catalog never calls this
instance method. 
On the complete dependency stack,
static_contract(AudioDataFilterStage).gates.per_row_independent is None
while build_contract(AudioDataFilterStage()).gates.per_row_independent is
True, so static discovery drops the only semantic gate added here. 

This source change also adds no test. Please declare the invariant through
AGENT_STATIC = StaticHints(gates=Gates(per_row_independent=True)), and add
tests covering static/configured parity, wrappable=False, discovery, and
clean configured expansion/validation for all four topologies.

mohammadaaftabv (changes requested):

nemo_curator/stages/audio/advanced_pipelines/audio_data_filter/audio_data_filter.py
shubhamNvidia: Thanks for catching this. This PR depends on the shared `_agent` support introduced in Foundation. I’ll rebase this branch after Foundation merges so the public `AudioDataFilterStage` import remains valid on the PR base.

nemo_curator/stages/audio/advanced_pipelines/audio_data_filter/audio_data_filter.py
shubhamNvidia: Agreed. The composite expands into IO, Filtering, and Segmentation leaf stages, so their agent-contract migrations are prerequisites. The combined `agent/adv` stack validates all four topologies cleanly; I’ll rebase this PR after those prerequisite PRs merge.

nemo_curator/stages/audio/advanced_pipelines/audio_data_filter/audio_data_filter.py
shubhamNvidia: Thanks for catching this. `describe()` correctly declared the gate, but static discovery does not instantiate the stage or call `describe()`. I added `AGENT_STATIC` with `per_row_independent=True` and a regression test confirming static and configured contracts remain aligned, including non-wrappability.

shubhamNvidia (commented):
reviewed

shubhamNvidia (commented):
reviewed

PR #2345, opened 2026-08-26
[benchmarking] Add separate nemo-curator-benchmarking package

rlratzel:
## Motivation

Curator benchmarks need dependencies and tooling that cannot live in the released Curator image, partly because some benchmark dependencies cannot be redistributed. Today that creates scattered setup logic across Curator, the old benchmarking Dockerfile, shell wrappers, and CI orchestration tooling.

That makes benchmark dependency changes easy to miss: benchmark authors can update a benchmark script or config in Curator, but the environment setup path used by nightly/ad-hoc automation may still need a separate update elsewhere.

This PR makes the benchmark suite its own separate, manually installed package: `nemo-curator-benchmarking`.

## Summary

This adds a package under `benchmarking/pyproject.toml` with a CLI named `curator-benchmark`. The package owns benchmark scripts, configs, dependency declarations, setup/check helpers, and container/image orchestration.

It enables three main workflows:

- Run benchmarks on a local or bare-metal Curator environment.
- Run benchmarks inside a Curator container/image while using benchmark scripts and configs from the current checkout.
- Run the latest benchmark suite against an older released Curator image, so release-to-release comparisons can use a consistent benchmark harness.

It also starts making `nemo-curator-benchmarking` a reusable library for CI orchestration tooling, so downstream automation can stop duplicating config parsing, YAML merging, entry listing, timeout planning, and dependency setup logic.

## User-Facing Commands

- `curator-benchmark run`
- `curator-benchmark check`
- `curator-benchmark setup`
- `curator-benchmark start`
- `curator-benchmark shell`

The package also introduces benchmark dependency groups. Python dependencies are declared in `benchmarking/pyproject.toml`; system dependencies live under `benchmarking/system_deps/` as check/install scripts.

## Replaced Flow

This removes the old dedicated benchmarking Dockerfile and shell-wrapper flow. The intent is that `curator-benchmark` becomes the primary entrypoint instead of requiring users to dig into `benchmarking/tools/run.sh`.

## Design Notes

`benchmarking/runner` remains a peer of `benchmarking/curator_benchmarking` for now. That is transitional. Moving `runner` under `curator_benchmarking` would be a mechanical cleanup better handled in a separate PR, so this PR stays focused on packaging and behavior.

Some tutorial files are vendored into the benchmark package at build time. That is intended as temporary tech debt until the tutorial-backed benchmark utilities can be refactored into a shared module.

## Detailed Changes

- Add `benchmarking/pyproject.toml` for the separate `nemo-curator-benchmarking` package.
- Add the `curator-benchmark` CLI with `run`, `list`, `check`, `setup`, `start`, and `shell` commands.
- Add Docker/image support for one-shot runs, named reusable containers, and externally prepared running containers.
- Add benchmark dependency groups backed by Python extras and modular `benchmarking/system_deps/*/{check,install}.sh` scripts.
- Add `--setup-benchmark-env auto|yes|no` for container/image workflows that should install or skip benchmark environment setup.
- Add reusable `curator_benchmarking` APIs for downstream automation to load and merge configs, list entries, resolve paths, calculate entry timeouts, and map dependency groups.
- Keep direct benchmark script execution supported via `python benchmarking/scripts/<script>.py ...`.
- Keep `python benchmarking/run.py ...` as a compatibility shim for the runner.
- Update benchmark script and data-prep path resolution so installed package layouts work.
- Vendor the small set of tutorial files required by tutorial-backed benchmarks during package build.
- Remove the dedicated benchmarking Dockerfile, Docker build helper, bash variable generator, and `benchmarking/tools/run.sh`.

## Validation

- GitHub checks currently pass: DCO, pre-commit, ruff, and secrets detector.


tests/benchmarking/test_run_sink_hooks.py
```diff
@@ -23,6 +23,8 @@
```
praateekmahajan, line 24: Is this even run in ci?

tests/benchmarking/test_run_sink_hooks.py
```diff
@@ -23,6 +23,8 @@
 
 
 class _Sink:
```
praateekmahajan, line 25: One qualm that I have with AI is that it keeps adding unit tests for new benchmarking/scripts, can we add some instruction somewhere to not add unittests for benchmarking/scripts

benchmarking/curator_benchmarking/_tutorials/__init__.py
```diff
@@ -0,0 +1,26 @@
+# Copyright (c) 2026, NVIDIA CORPORATION.  All rights reserved.
```
praateekmahajan, line 1: Let's see if we can remove somehow?

benchmarking/curator_benchmarking/commands/check.py
```diff
@@ -0,0 +1,299 @@
+# Copyright (c) 2026, NVIDIA CORPORATION.  All rights reserved.
```
praateekmahajan, line 1: See if click (or newer alternatives) can help here

benchmarking/system_deps/audio/check.sh
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env bash
```
praateekmahajan, line 1: I think all this is redundant and too bloated

benchmarking/benchmarks.yaml
```diff
@@ -1587,6 +1613,8 @@ entries:
   - name: interleaved_filter_raydata_materialize
     enabled: true
     script: interleaved_filter_benchmark.py
+    dependencies:
+      - visual
```
praateekmahajan, line 1617: ??

benchmarking/pyproject.toml
```diff
@@ -0,0 +1,95 @@
...
+video = [
+    "opencv-python-headless",
+]
+all = [
+    "albumentations",
+    "opencv-python-headless",
+    "pyloudnorm",
+    "slack-sdk",
+    "wget",
+]
```
praateekmahajan, line 63: Let's see how we don't repeat ourselves across tutorials.. the reason this is not in Curator/pyproject.toml OR Curator/Dockerfile...

benchmarking/setup.py
```diff
@@ -0,0 +1,70 @@
...
+
+from __future__ import annotations
+
+import shutil
+from pathlib import Path
+
+from setuptools import setup
+from setuptools.command.build_py import build_py
+
+VENDORED_TUTORIAL_FILES = (
```
praateekmahajan, line 23: Can we avoid this somehow?

PR #2346, opened 2026-08-26
Audio: add local duration-aware ASR batching

mohammadaaftabv:
## Summary

This PR adds stage-local, duration-aware batching to `ASRStage` with a deliberately small configuration surface:

```yaml
max_audio_sec_per_actor: 240
max_inference_duration_s: 120
local_bucketing: true
batch_size: 32
```

- `max_audio_sec_per_actor` is required and bounds every adapter invocation using `item_count × longest_audio_seconds`.
- `max_inference_duration_s` is the model-specific single-audio limit. Segmentation always occurs, regardless of bucketing mode.
- `local_bucketing` only controls whether segments are reordered for more efficient packing.
- `batch_size` remains the backend candidate-row window; it is not an adapter item-count limit.

This intentionally has no compatibility layer for the earlier experimental APIs: no `BatchPolicy`, `batch_policy`, `adapter_batch_size`, bucket boundaries, per-bucket item limits, `max_items_per_batch_by_bucket`, `estimate_item_cost`, timers, queues, or flush lifecycle.

## Behavior

For each finite `process_batch()` call, the stage:

1. Prepares every eligible input row.
2. Splits every waveform into contiguous model-safe segments using `max_inference_duration_s`.
3. Flattens segments from all eligible rows in that call into one planning window.
4. Plans capacity-bounded adapter calls.
5. Executes the calls and scatters results back to their original segment positions.
6. Stitches segments in temporal order and restores the original parent-row order.

No audio or planner state survives into another `process_batch()` call.

With `local_bucketing: false`, segments retain their input order and are greedily packed under the same padded-audio budget.

With `local_bucketing: true`, segments are stable-sorted by duration. Dynamic programming then chooses contiguous call boundaries that lexicographically minimize:

1. adapter-call count;
2. total padded audio seconds.

Equal-duration inputs retain their original relative order. Output rows are restored to their original order regardless of model execution order.

For sorted durations `[1, 2, 2]` under a 4-second budget, the previous greedy boundary selection produced:

```text
[1, 2], [2] -> 2 calls, 6 padded seconds
```

The final planner produces:

```text
[1], [2, 2] -> 2 calls, 5 padded seconds
```

The planner therefore fixes the known efficiency counterexample without increasing call count.

## Segmentation and result semantics

- `max_inference_duration_s` must be less than or equal to `max_audio_sec_per_actor`.
- The segment limit must represent at least one sample at the configured sample rate.
- Exact boundaries do not create empty tails; every nonempty remainder becomes a segment.
- Segment duration is derived from the prepared waveform rather than manifest metadata.
- A skipped segment marks the stitched parent result as skipped.
- Segment extras remain one flat dictionary, merged in temporal order.
- Adapter result counts are checked strictly before scatter and stitching.

The planner directly consumes internally generated `audio_seconds` values and uses normal tuple ordering for its calls-first, padded-seconds-second score. The earlier `_item_audio_seconds` and `_duration_plan_is_better` helpers were removed; floating-point tolerance is retained only at the budget feasibility boundary.

## Validation

The final source was rebased onto `main` at `7961210f9e622c7fd47e5b0218bafabb883027b8` and published as `2403b83b212d36bf7a627179b997dea650aed52a`.

Audio inference, ASR adapter, segmentation, and configuration suites:

```text
255 passed, 5 deselected
```

Pre-commit, Python compilation, Bash syntax, JSON validation, preparation checks, and the planner audit also passed.

### Local GPU parity

Hardware/model:

- NVIDIA GeForce RTX 3080 Ti
- `nvidia/stt_en_fastconformer_ctc_large`
- one warm-up plus seven measured iterations per performance variant

The fixed cohort contained eight rows in one `process_batch()` call:

- four × 25.3135-second inputs
- four × 115.0935-second inputs
- 561.628 useful audio seconds total
- 230.2-second adapter budget

Compared with the previous implementation (`d1c93b3c2859bd3a2b7cba7a8bf81d1db1468337`):

| Metric | This PR | Previous |
|---|---:|---:|
| Adapter calls | 3 | 3 |
| Padded audio seconds | 561.628 | 561.628 |
| Padding efficiency | 100% | 100% |
| Median adapter time | 0.3724 s | 0.3956 s |
| Median `process_batch` time | 0.4771 s | 0.4941 s |
| Adapter throughput | 1508.12 audio s/s | 1419.67 audio s/s |
| Process throughput | 1177.18 audio s/s | 1136.66 audio s/s |
| Peak VRAM | 1,529,335,808 B | 1,529,335,808 B |

The target and previous implementations produced exact raw and normalized transcript parity. Row order, row count, skip state, exact-once coverage, source identity, call budgets, and GPU identity also matched.

A correctness comparison against `nithinraok/Curator:nkoluguri/integration-test` at `1e2d639946f28dfd289a0a9456480385d4480c28` produced the same transcripts with no missing, empty, skipped, or duplicate rows.

The final comparator passed 38/38 checks. The local launcher completed on its first invocation, so no Kratos run or inference retry was needed.

The performance numbers are a controlled local observation, not a guarantee for every model or duration distribution. The padded-seconds budget is a tuning proxy and does not replace model- and hardware-specific memory validation.

## Documentation

The PR adds a comprehensive local-bucketing guide covering:

- padding-cost theory;
- the exact planning scope and algorithm;
- configuration and tuning;
- segmentation and stitching invariants;
- application to other audio GPU inference stages;
- model-specific caveats;
- testing and parity methodology;
- complexity and explicit non-goals.

Tutorial configurations and documentation for the affected ASR examples are updated to use the final contract.

This PR intentionally remains a draft.


virajkarandikar (approved):

tests/config/test_run.py
```diff
@@ -789,6 +801,26 @@ def test_nemo_fastconformer_tutorial_yaml_uses_shared_adapter_contract():
     assert executor.config == {}
 
 
+def test_nemo_fastconformer_tutorial_accepts_local_bucketing_config() -> None:
+    config_dir = Path(__file__).parents[2] / "tutorials" / "audio" / "nemo_fastconformer"
```
VibhuJawa, line 805: Dont think we should add absolute paths, honstestly

VibhuJawa (approved):
Mostly looks good, minor nit.

mohammadaaftabv:
/ok to test https://github.com/NVIDIA-NeMo/Curator/pull/2346/commits/88476f78f6f100d1f3c2cf8d48b9c45910d1117f

mohammadaaftabv:
/ok to test https://github.com/NVIDIA-NeMo/Curator/pull/2346/commits/77bed5429e19d0e08ceba1011787508f4be28e24

PR #2348, opened 2026-08-27
Update GB200 text benchmarks

sarahyurick:
# Entries to fix
- [x] embedding_generation_raydata
- [x] embedding_generation_xenna
- [x] fasttext_filter_raydata
- [x] fasttext_filter_xenna
- [x] modifier_raydata
- [x] ndd_dynamo
- [x] ndd_ray_serve
- [x] score_filter_raydata
- [x] score_filter_xenna

# NDD Context (generated by Codex)
Why `ndd_ray_serve` needs `--tiktoken-cache-dir` and `ndd_dynamo` doesn't:

- `ndd_ray_serve` (`nemo_curator/core/serve/ray_serve/backend.py`) deploys via `ray.serve.llm.build_openai_app`, which stands up Ray Serve LLM's `LLMServer` deployment. That deployment builds vLLM's own OpenAI-compatible `/v1/responses` handler, `OpenAIServingResponses`. Because gpt-oss uses OpenAI's "Harmony" response format for its tool-calling/reasoning tokens, `OpenAIServingResponses.__init__` eagerly calls `get_stop_tokens_for_assistant_actions` → `load_harmony_encoding(HARMONY_GPT_OSS)`, which tries to download the `o200k_base.tiktoken` vocab file from Azure at every replica startup — that's the slow/failing network call your notes traced.
- Dynamo doesn't use OpenAI's `openai_harmony` Python package at all. It reimplements Harmony's channel/tool-call parsing natively in Rust — `lib/llm/src/protocols/openai/chat_completions/aggregator.rs` (`is_harmony_parser`, `contains_harmony_protocol`) and `lib/llm/src/preprocessor.rs` parse the raw `<|channel|>analysis<|message|>...<|end|>` text markers directly from the model's output. This never touches `openai_harmony`, `load_harmony_encoding`, or any `.tiktoken` vocab file. So `--tiktoken-cache-dir` correctly doesn't apply to `ndd_dynamo`; there's structurally nothing to cache.


benchmarking/4xGB200-64CPU.yaml
```diff
@@ -159,6 +180,7 @@ entries:
       --executor=ray_data
       --engine-kwargs='{"tensor_parallel_size": 1}'
       --autoscaling-config='{"min_replicas": 4, "max_replicas": 4}'
+      --health-check-timeout-s=1200
```
rlratzel, line 183: Does this and the other new CLI option also need to be in `benchmarks.yaml` (the "base" config used by the nightly for H100)?

sarahyurick: Nope, in the base config the default timeout is good. Same with the other new arg, the base config already works without it.

rlratzel (approved):

PR #2352, opened 2026-08-28
fix(readers): fail fast on unresolved input paths

mohammadaaftabv:
## Summary

- Fail fast with actionable diagnostics when no supported input files resolve, with `allow_empty=True` for intentional empty sources.
- Normalize strings, path-like values, directories, globs, local URIs, and chained fsspec paths across worker working directories.
- Prefer construction-anchored inputs, safely fall back to executor working-directory paths, and deduplicate results.

## Validation

Focused partitioning, file-utility, and downstream audio/text reader suites pass. Ruff, pre-commit, and DCO checks pass.


nemo_curator/stages/file_partitioning.py
```diff
@@ -54,15 +62,27 @@ class FilePartitioningStage(ProcessingStage[EmptyTask, FileGroupTask]):
...
 
-    file_paths: str | list[str]
+    file_paths: str | os.PathLike[str] | list[str | os.PathLike[str]]
     files_per_partition: int | None = None
     blocksize: int | str | None = None
     file_extensions: list[str] | None = None
     storage_options: dict[str, Any] | None = None
     limit: int | None = None
     name: str = "file_partitioning"
+    allow_empty: bool = False
```
sarahyurick, line 83: Can we only keep the `allow_empty` bool and not add all the extra `cwd` etc. logic? And then add the absolute path issue in the https://github.com/NVIDIA-NeMo/Curator/tree/main/tutorials/audio/tagging tutorial instead of here?

sarahyurick (changes requested):

PR #2354, opened 2026-08-28
feat: add per-node stage worker sizing

### LINKED ISSUE #2197: Ray Data backend should cap download-stage concurrency by node count and spread workers

praateekmahajan:
## Summary

- Add backend-neutral `ProcessingStage.num_workers_per_node()` and `with_(num_workers_per_node=...)` configuration.
- Translate per-node sizing across Ray Data, Ray Actor Pool, and Xenna while keeping the legacy Xenna stage-spec key compatible.
- Size Ray actor pools and task concurrency as `ceil(num_workers_per_node * alive_node_count)` and use `scheduling_strategy="SPREAD"` for best-effort distribution.
- Route document download stages through the common policy and default Arxiv, Common Crawl, and Wikipedia downloads to two workers per node.
- Centralize conflicting worker-sizing validation and update current Fern/backend guidance.

## Semantics and constraints

- `num_workers_per_node` is a sizing policy, not a hard placement guarantee. Ray actor stages create fixed cluster-wide pools; Ray Data task stages use cluster-wide concurrency caps. Both Ray backends use `SPREAD`, which may still place workers unevenly when resources or other constraints differ.
- `STRICT_SPREAD` is not a valid string scheduling strategy for Ray tasks or actors. Enforcing exact placement would require placement-group creation, gang resource reservation, lifecycle management, and separate handling for heterogeneous or autoscaling clusters; that is intentionally outside this API.
- Per-node values must be positive, finite integers or fractions. The common hook and legacy Xenna stage-spec key use the same validation. Ray Data and Ray Actor Pool round the cluster-wide request up with `ceil`; Xenna receives the per-node value directly.

## API changes

- `num_workers`, `num_workers_per_node`, and Ray Data actor-pool min/max/initial sizing are mutually exclusive.
- `ImageDuplicatesRemovalStage(num_workers_per_node=...)` is replaced by `ImageDuplicatesRemovalStage(...).with_(num_workers_per_node=...)`.
- Existing `xenna_stage_spec()["num_workers_per_node"]` configurations remain supported when the common worker hooks are unset.

## Validation

- 133 changed-area tests and all repository pre-commit hooks pass in `nvcr.io/nvidian/nemo-curator:nightly-2026-08-31` with Ray 2.57.0.
- In the same image, a Ray multi-node `Cluster` probe used a zero-CPU head, two 3-CPU worker nodes, and `ActorPoolStrategy(size=4)`. `map_batches(..., scheduling_strategy="SPREAD")` created two actors on each worker and none on the head.

Replaces #2198.
Fixes #2197.

### DIFF
--- a/fern/versions/main/pages/reference/infrastructure/execution-backends.mdx
+++ b/fern/versions/main/pages/reference/infrastructure/execution-backends.mdx
@@ -67,19 +67,19 @@ executor = XennaExecutor(
         # Execution mode: 'streaming' (default) or 'batch'
         # Batch processes all data for a stage before moving to the next; streaming runs stages concurrently.
         "execution_mode": "streaming",
-        
+
         # Logging interval: seconds between status logs (default: 60)
         # Controls how frequently progress updates are printed
         "logging_interval": 60,
-        
+
         # Ignore failures: whether to continue on failures (default: False)
         # When True, the pipeline continues execution instead of failing fast when stages raise errors.
         "ignore_failures": False,
-        
+
         # CPU allocation percentage: ratio of CPU to allocate (0-1, default: 0.95)
         # Fraction of available CPU resources to use for pipeline execution
         "cpu_allocation_percentage": 0.95,
-        
+
         # Autoscale interval: seconds between auto-scaling checks (default: 180)
         # How often to run the stage auto-scaler.
         "autoscale_interval_s": 180,
@@ -99,7 +99,7 @@ results = pipeline.run(executor)
 | `cpu_allocation_percentage` | `float` | `0.95` | Fraction (0-1) of available CPU resources to allocate |
 | `autoscale_interval_s` | `int` | `180` | Seconds between auto-scaling evaluations |
 
-Set a cluster-wide stage count with `stage.with_(num_workers=N)`, or set `xenna_stage_spec={"num_workers_per_node": N}` for a per-node count. These settings are mutually exclusive. See [Stage Worker Sizing](/reference/infra/stage-worker-sizing#xenna-worker-counts) and the [`XennaExecutor` API](/api/reference/api-reference/executors/xenna-executor).
+Set a cluster-wide stage count with `stage.with_(num_workers=N)`, or a per-node count with `stage.with_(num_workers_per_node=N)`. These settings are mutually exclusive. The legacy Xenna-only `xenna_stage_spec={"num_workers_per_node": N}` form remains supported. See [Stage Worker Sizing](/reference/infra/stage-worker-sizing#xenna-worker-counts) and the [`XennaExecutor` API](/api/reference/api-reference/executors/xenna-executor).
 
 For more details about the underlying engine, refer to the official [NVIDIA Cosmos-Xenna project](https://github.com/nvidia-cosmos/cosmos-xenna/tree/main).
 
@@ -135,7 +135,7 @@ results = pipeline.run(executor)
 | `show_progress` | `bool` | `True` | Display tqdm progress bars during stage execution and shuffle inserts |
 | `progress_interval` | `float` | `10.0` | Minimum interval in seconds between progress bar updates |
 
-A positive stage `num_workers()` requests an exact actor count when resources fit. If fewer actors fit, this executor warns and caps the request to available CPU/GPU capacity. See [Stage Worker Sizing](/reference/infra/stage-worker-sizing#ray-actor-pool-resource-capping).
+A positive stage `num_workers()` requests an exact actor count. `num_workers_per_node()` requests that count per alive node. If fewer actors fit, this executor warns and caps the request to available CPU/GPU capacity. See [Stage Worker Sizing](/reference/infra/stage-worker-sizing#ray-actor-pool-resource-capping).
 
 <Accordion title="Example: Fuzzy Deduplication">
 
@@ -190,7 +190,7 @@ results = pipeline.run(executor)
 |-----|------|---------|-------------|
 | `ignore_failures` | `bool` | `False` | If `True`, continue pipeline execution even when tasks fail |
 
-Ray Data uses `ActorPoolStrategy` for actor stages and `TaskPoolStrategy` for task stages. Configure a fixed pool with `stage.with_(num_workers=N)`, or configure actor autoscaling with `min_workers`, `max_workers`, and `initial_workers` in `ray_stage_spec`. See [Stage Worker Sizing](/reference/infra/stage-worker-sizing#ray-data-worker-pools) for copy-pasteable examples and precedence rules.
+Ray Data uses `ActorPoolStrategy` for actor stages and `TaskPoolStrategy` for task stages. Configure a fixed cluster-wide pool with `stage.with_(num_workers=N)`, a fixed per-node pool with `stage.with_(num_workers_per_node=N)`, or actor autoscaling with `min_workers`, `max_workers`, and `initial_workers` in `ray_stage_spec`. Per-node sizing uses a cluster-wide pool plus best-effort `SPREAD` placement; it is not a hard node-affinity constraint. See [Stage Worker Sizing](/reference/infra/stage-worker-sizing#ray-data-worker-pools) for copy-pasteable examples and conflict rules.
 
 Ray Data may fuse adjacent operators when their compute and resource requirements are compatible. Matching resources enables the optimization but does not guarantee it; Ray Data makes the final optimizer decision. The [resource matching and fusion reference](/reference/infra/stage-worker-sizing#resource-matching-and-ray-data-fusion) explains the video-stage CPU reservation behavior.
 
@@ -207,7 +207,7 @@ For worker counts, resources, and backend-specific `with_()` overrides, see [Sta
 Ray-based executors provide enhanced scalability and performance for large-scale data processing tasks. These executors are beneficial for:
 
 - **Large-scale classification tasks**: Distributed inference across multi-GPU setups
-- **Deduplication workflows**: Parallel processing of document similarity computations  
+- **Deduplication workflows**: Parallel processing of document similarity computations
 - **Resource-intensive stages**: Automatic scaling based on computational demands
 
 ## Choosing a Backend
--- a/fern/versions/main/pages/reference/infrastructure/execution-backends.mdx
+++ b/fern/versions/main/pages/reference/infrastructure/execution-backends.mdx
@@ -172,7 +172,7 @@ from nemo_curator.backends.ray_data import RayDataExecutor
 
 executor = RayDataExecutor(
     config={"ignore_failures": False},
-    ignore_head_node=True,  # Exclude head node from computation
+    ignore_head_node=True,  # Exclude head from setup and per-node sizing
 )
 results = pipeline.run(executor)
 ```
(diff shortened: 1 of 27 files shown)


nemo_curator/backends/utils.py
```diff
@@ -137,6 +137,20 @@ class RayStageSpecKeys(str, Enum):
...
+    """Return a stage's validated per-node worker request."""
+    value = stage.num_workers_per_node()
+    if value is None:
+        return None
+    if isinstance(value, bool) or not isinstance(value, (int, float)):
+        msg = f"num_workers_per_node() for stage {stage.name} must be a positive number."
+        raise TypeError(msg)
+    if value <= 0:
+        msg = f"num_workers_per_node() for stage {stage.name} must be > 0."
+        raise ValueError(msg)
```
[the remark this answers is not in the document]
praateekmahajan: Fixed in 8cc664a9: the shared validator now rejects non-finite floats before either Ray backend reaches math.ceil, with regression coverage for NaN and positive infinity.

nemo_curator/backends/ray_data/adapter.py
```diff
@@ -141,6 +167,9 @@ def process_dataset(self, dataset: Dataset) -> Dataset:
             )
             raise ValueError(msg)
 
+        if total_pool_size is not None:
+            map_batches_kwargs["scheduling_strategy"] = "SPREAD"
```
ayushdg, line 171: Thoughts on `SPREAD` vs `STRICT_SPREAD` for this given that `num_workers_per_node` seems to be a fixed/opinionated config?

nemo_curator/backends/ray_data/adapter.py
```diff
@@ -90,6 +97,18 @@ def _build_resource_kwargs(self, ray_stage_spec: dict) -> dict[str, float]:
             kwargs["num_gpus"] = self.stage.resources.gpus  # type: ignore[reportArgumentType]
         return kwargs
 
+    def _total_pool_size(self) -> int | None:
+        """Return the cluster-wide pool size implied by num_workers_per_node()."""
+        num_workers_per_node = get_stage_num_workers_per_node(self.stage)
+        if num_workers_per_node is None:
+            return None
+
+        node_count = get_alive_ray_node_count(ignore_head_node=self.ignore_head_node)
```
ayushdg, line 106: I think the docs mention this but we're not guaranteed to avoid scheduling on the head node in some cases. 
Do you forsee this causing issues for setups?

praateekmahajan: Yup, I think the whole `ignore_head_node` thing is spilling to more parts of the code. It is needed for anyscale compatibility but I wish there was a better way to do it.

nemo_curator/backends/ray_data/adapter.py
praateekmahajan: Good question, I tried it out. It'll require a Placement Group, which also means resource reservation outside ray data scheduler

nemo_curator/backends/ray_actor_pool/utils.py
```diff
@@ -164,6 +174,19 @@ def calculate_optimal_actors_for_resources(
...
+        requested_actors = max(1, math.ceil(num_workers_per_node * num_nodes))
+        if requested_actors > max_actors_resources:
+            msg = (
+                f"Stage {stage.name} requires {requested_actors} actors from num_workers_per_node(), "
+                f"but only {max_actors_resources} fit with available resources. "
+                f"Capping actor count to {max_actors_resources}."
+            )
+            logger.warning(msg)
+            return max_actors_resources
+        return requested_actors
```
ayushdg, line 198: Don't we also need to specify some strategy like `spread` while creating the actors in scheduling strategy?

Side note: Unrelated to this PR but `num_workers` isn't respected in RayActorlPoolExecutor. But we can also make the assumption that its value comes from a streaming pipeline. Now when every stage does all the compute anyway.

ayushdg (commented):
Minor comment. Things look good at a high level otherwise.

nemo_curator/backends/ray_actor_pool/utils.py
praateekmahajan: Good catch on first.

Re the side note, not sure i follow. num_workers is respected IIRC as of this PR https://github.com/NVIDIA-NeMo/Curator/pull/2077/changes

fern/versions/main/pages/reference/infrastructure/execution-backends.mdx
```diff
@@ -172,7 +172,7 @@ from nemo_curator.backends.ray_data import RayDataExecutor
 
 executor = RayDataExecutor(
     config={"ignore_failures": False},
-    ignore_head_node=True,  # Exclude head node from computation
+    ignore_head_node=True,  # Exclude head from setup and per-node sizing
```
VibhuJawa: I have a tangential question,  for single node how does this work out. 

Like, with Slurm Arrays , I am using 1 node all the time

VibhuJawa (commented):
Mostly looks good from inital review. Just had a tangential question

fern/versions/main/pages/reference/infrastructure/execution-backends.mdx
praateekmahajan: We should never have ignore_head_node=True.. it was added only so that Curator can work on Anyscale where head node is not-schedulable on. Ideally we get rid of this and improve code s.t. we can just say "exclude this resource" from everything rather than carrying ignore_head_node everywhere

VibhuJawa: Should we then just switch this to False even in docs ? 

Like, have a small section (if needed) to turn=True for anyscale deployment ?

praateekmahajan: Added that comment

nemo_curator/stages/base.py
```diff
@@ -189,6 +205,33 @@ def num_workers(self) -> int | None:
...
+        configured = [
+            name
+            for name, value in (
+                ("num_workers()", self.num_workers()),
+                ("num_workers_per_node()", self.num_workers_per_node()),
+            )
+            if value is not None
+        ]
+
+        from nemo_curator.backends.ray_data.utils import get_configured_actor_pool_sizing_keys
```
weijiac0619: should we move get_configured_actor_pool_sizing_keys from nemo_curator.backends.ray_data.utils to backends.utils?

tests/backends/ray_data/test_adapter.py
```diff
@@ -91,22 +100,87 @@ def test_process_dataset_uses_compute_for_actor_stages_and_ray_default_for_task_
...
+            (float("nan"), ValueError),
+            (float("inf"), ValueError),
+            (True, TypeError),
+            ("2", TypeError),
+        ],
+    )
+    def test_num_workers_per_node_rejects_invalid_values(self, value: object, error: type[Exception]):
+        stage = ConfigurableTaskStage(num_workers_per_node=value)  # type: ignore[arg-type]
+
+        with pytest.raises(error, match="num_workers_per_node"):
```
weijiac0619: Is this intentional? it seems inconsistent with the conflict validation in _validate_worker_sizing() which fires at constuction

praateekmahajan: Resolved

nemo_curator/stages/base.py
praateekmahajan: Good call

weijiac0619 (approved):
Thanks for your effort on this! Great work

PR #2360, opened 2026-08-31
fix(io): preserve Arrow schemas through Parquet writes

praateekmahajan:
## Summary

- write Arrow-backed `DocumentBatch` inputs directly with `pyarrow.parquet.write_table`
- preserve field selection and supported compression, encoding, and row-group options on the direct path
- use the writer's configured fsspec filesystem so `storage_options` works without converting Arrow input to pandas
- retain the existing pandas writer for pandas inputs and pandas-only/unsupported direct-writer options

Why: in a 50,000×768 FP32 probe, direct Arrow writing avoided ~40% higher peak RSS and ~73% longer writes in the current Arrow → pandas-object → Arrow path, without requiring a global switch to Arrow-backed pandas dtypes.

Linear: [NMCUR-316](https://linear.app/nvidia/issue/NMCUR-316/preserve-arrow-native-schemas-through-embedding-reads-and-parquet)

## Scope

This intentionally does not change `DocumentBatch.to_pandas()`, JSONL reader dtypes, or ID assignment. The direct JSONL reader already returns a `pyarrow.Table`; preserving that representation through the Parquet writer avoids the unnecessary pandas conversion without changing pandas behavior elsewhere.

`partition_cols` keeps its existing pandas behavior for now. Deterministic partitioned output for both pandas and Arrow inputs is tracked in [NMCUR-432](https://linear.app/nvidia/issue/NMCUR-432/make-parquetwriter-partitioned-output-explicit).

Both supported round trips were exercised in the nightly image:

1. JSONL → Arrow table → Parquet → `ParquetReaderStage`
2. JSONL → Arrow table → existing pandas conversion → Parquet → `ParquetReaderStage`

Both reread successfully and produced FP32 embeddings accepted by the existing KMeans normalization. The direct path preserved the logical schema. The pandas path retained `large_string`, signed non-null IDs, and `list<float32>` embeddings, while nullable integers and nullable integers inside object-backed structs were inferred as doubles. That pre-existing pandas behavior is left unchanged.

### DIFF
--- a/nemo_curator/stages/text/io/writer/parquet.py
+++ b/nemo_curator/stages/text/io/writer/parquet.py
@@ -13,7 +13,6 @@
 # limitations under the License.
 
 from dataclasses import dataclass, field
-from inspect import signature
 from typing import Any
 
 import pyarrow as pa
@@ -22,6 +21,8 @@ import pyarrow.parquet as pq
 from nemo_curator.tasks import DocumentBatch
 
 from .base import BaseWriter
+
+_PANDAS_ONLY_WRITE_OPTIONS = {"partition_cols", "storage_options"}
 
 
 @dataclass
@@ -61,8 +62,7 @@ class ParquetWriter(BaseWriter):
         if engine not in {None, "pyarrow"} or index is True:
             return False
 
-        arrow_write_options = set(signature(pq.write_table).parameters) - {"table", "where"}
-        if not set(write_kwargs).issubset(arrow_write_options):
+        if _PANDAS_ONLY_WRITE_OPTIONS.intersection(write_kwargs):
             return False
 
         pq.write_table(table, file_path, **write_kwargs)
--- a/nemo_curator/stages/text/io/writer/parquet.py
+++ b/nemo_curator/stages/text/io/writer/parquet.py
@@ -13,7 +13,11 @@
 # limitations under the License.
 
 from dataclasses import dataclass, field
+from inspect import signature
 from typing import Any
+
+import pyarrow as pa
+import pyarrow.parquet as pq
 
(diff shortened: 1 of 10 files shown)


nemo_curator/stages/text/io/writer/parquet.py
```diff
@@ -15,22 +15,30 @@
...
+    # Additional kwargs for the selected Parquet writer
     write_kwargs: dict[str, Any] = field(default_factory=dict)
     file_extension: str = "parquet"
     name: str = "parquet_writer"
 
     def write_data(self, task: DocumentBatch, file_path: str) -> None:
-        """Write data to Parquet file using pandas DataFrame.to_parquet."""
+        """Write Arrow batches directly and retain pandas compatibility."""
+        if isinstance(task.data, pa.Table) and self._write_arrow(task.data, file_path):
+            return
```
VibhuJawa, line 42: ~Nit:~

```suggestion
        if isinstance(task.data, pa.Table):
            return self._write_arrow(task.data, file_path):
            
```
Ignore nit, logic makes sense

VibhuJawa (approved):
Nit on code organization

nemo_curator/stages/text/io/writer/parquet.py
praateekmahajan: Under certain conditions `self._write_arrow` returns False, in which case we want to continue with pandas write.. Conditions are mainly
1. index=True
2. partition_cols=XYZ
3. engine="fastparquet"

VibhuJawa: yeah, makes sense, read understood logic leater

PR #2365, opened 2026-09-02
Add agent file for Dynamo

sarahyurick:


nemo_curator/core/serve/dynamo/AGENTS.md
```diff
@@ -0,0 +1,190 @@
...
+| `infra.py` | Actor naming, endpoint URLs, CLI-flag translation |
+| `constants.py` | Default ports, namespace, event/request plane names |
+
+## Base venv vs. actor venv
+
+Curator's own `pyproject.toml` pins `vllm[flashinfer,runai,otel]==0.22.0+cu129`
+directly (the `vllm` extra) and constrains `transformers>=4.56.0,<5.0`.
+`ai-dynamo>=1.3.1` is *also* a direct optional dependency (via the
+`inference_server` extra, pulled in by `sdg_cuda12`), so the driver/base venv
+can already have Dynamo and vLLM installed. `text_cuda12` pulls in the plain
```
praateekmahajan:
```suggestion
can already have Dynamo and vLLM installed, but never dynamo[vllm]. `text_cuda12` pulls in the plain
```

nemo_curator/core/serve/dynamo/AGENTS.md
```diff
@@ -0,0 +1,190 @@
...
+`vllm` extra but not `inference_server`, so it does not install `ai-dynamo`.
+
+That base install doesn't automatically reach a model's actor, though:
+**each Ray actor gets its own fresh, isolated venv that does not inherit the
```
praateekmahajan: I don't think Ray gets its own "fresh" venv, but rather a venv based on top of base venv.. IIUC how it works is copy over existing venv and then install on top of it. Which is why sometimes more issues show up. But please fact check me

sarahyurick: Yeah I have always thought about it as Curator venv + `runtime_env` installs on top of it. Sorry for the confusion, will update.

praateekmahajan (approved):
LGTM, maybe minor nit that the fact runtime_envs are merged for multi-model scenarios is mentioned pretty high (i.e. top level info) versus i might be wrong but that's a nuanced thing, and only relevant to folks doing multi-model inference (which I don't think any of the default pipelines are doing).. IMO when the time comes we should fix it such that frontend resolves more smartly than just merging... (i.e. i never built with multi-model in mind, but wanted to support it from day 1 so got it working)..

PR #2367, opened 2026-09-02
perf(semantic-dedup): bound parquet reads and memory

praateekmahajan:
## Summary

- Read Parquet footer metadata to group files by actual embedding element counts before loading them with cuDF.
- Preallocate KMeans and pairwise embedding arrays from exact row and element counts.
- When `fit_data_fraction=None`, select complete fit files within a budget derived from currently available GPU memory. Explicit fractions retain file-level sampling.
- Predict labels for every input row with `KMeansMG.predict`, including rows not selected for fitting.
- Remove `embedding_dim` from the semantic-deduplication API, benchmarks, tutorials, and documentation.
- Keep Parquet writes synchronous; asynchronous writes are scoped to #2361.

## Behavior and constraints

Read groups stay below cuDF's embedding-column element limit. Files are kept intact, so each individual Parquet file must itself fit below that limit.

## Benchmark results

EOS benchmark on 8× H100 GPUs. Baseline: main at `07920606`; PR measurement: `aa7529ea`. The current head only simplifies the JSONL fallback and benchmark reporting; the measured Parquet path is unchanged.

### 20% dataset, full fit — 65.8M rows

[Open the run comparison](http://rratzel-ws1:5050/compare?dir_a=login-eos-nightly%3A%2Flustre%2Ffsw%2Fcoreai_dlalgo_ci%2Fcurator_ci%2Fresults%2Fmain&dir_b=login-eos-nightly%3A%2Flustre%2Ffsw%2Fcoreai_dlalgo_ci%2Fcurator_ci%2Fresults%2Fmain-mirror&run_a=nightly-2026_09_02__06_08_52_UTC&run_b=pr-2367-2026_09_02__22_07_08_UTC-aa7529ea&entry=semdedup_identification_xenna_dataset_size_ratio)

| Metric | Main | PR | Change |
|---|---:|---:|---:|
| End-to-end workflow | 218.2 s | 199.5 s | **18.7 s faster (−8.6%)** |
| KMeans stage | 138.0 s | 125.5 s | **12.5 s faster (−9.1%)** |
| KMeans actor processing | 105.4 s | 94.3 s | **11.1 s faster (−10.5%)** |
| KMeans read | 33.9 s | 24.2 s | **9.7 s faster (−28.7%)** |
| KMeans fit + predict | 38.5 s | 35.3 s | **3.2 s faster (−8.2%)** |
| KMeans output | 33.0 s | 22.6 s | **10.4 s faster (−31.6%)** |
| Pairwise stage | 80.2 s | 74.0 s | **6.2 s faster (−7.8%)** |
| Run-average GPU utilization | 28.3% | 31.3% | **+3.0 pp** |
| Run-average GPU memory | 27.8% | 19.3% | **−8.5 pp (−30.6%)** |
| Peak GPU memory | 82.9% | 82.4% | Flat (−0.5 pp) |

The steady KMeans fit plateau drops from about 66% to 32% GPU memory, showing that exact preallocation removes the previous duplicate full embedding buffer. The similar absolute peak occurs later in distance calculation and partitioned Parquet output.

### Full dataset, 20% fit — 329.8M rows

[Open the run comparison](http://rratzel-ws1:5050/compare?dir_a=login-eos-nightly%3A%2Flustre%2Ffsw%2Fcoreai_dlalgo_ci%2Fcurator_ci%2Fresults%2Fmain&dir_b=login-eos-nightly%3A%2Flustre%2Ffsw%2Fcoreai_dlalgo_ci%2Fcurator_ci%2Fresults%2Fmain-mirror&run_a=nightly-2026_09_02__06_08_52_UTC&run_b=pr-2367-2026_09_02__22_07_08_UTC-aa7529ea&entry=semdedup_identification_xenna_fit_data_fraction)

| Metric | Main | PR | Change |
|---|---:|---:|---:|
| End-to-end workflow | 20m 54s | 18m 37s | **2m 17s faster (−11.0%)** |
| KMeans stage | 7m 43s | 6m 13s | **1m 30s faster (−19.4%)** |
| KMeans actor processing | 414.2 s | 316.7 s | **97.5 s faster (−23.5%)** |
| KMeans read | 201.3 s | 156.0 s | **45.3 s faster (−22.5%)** |
| KMeans fit | 40.2 s | 35.0 s | **5.2 s faster (−12.9%)** |
| KMeans predict + output | 172.6 s | 110.1 s | **62.5 s faster (−36.2%)** |
| Pairwise stage | 13m 11s | 12m 23s | **47.6 s faster (−6.0%)** |
| Run-average GPU utilization | 48.2% | 55.9% | **+7.8 pp** |
| Run-average GPU memory | 33.1% | 14.3% | **−18.8 pp (−56.8%)** |
| Peak GPU memory | 58.5% | 78.0% | **+19.5 pp** |

The partial-fit run uses substantially less memory on average and keeps GPUs busier, but its instantaneous peak is higher during output. Reducing that output-path peak is intentionally scoped to the asynchronous/bounded writer work in #2361.

## Validation

- Semantic KMeans, pairwise, and utility GPU tests: 55 passed, 1 skipped
- Pre-commit hooks


nemo_curator/stages/deduplication/semantic/utils.py
```diff
@@ -32,29 +43,102 @@ def get_array_from_df(df: "cudf.DataFrame", embedding_col: str) -> "cp.ndarray":
...
+) -> list[ParquetFileInfo]:
+    """Read exact row, embedding-element, and projected metadata sizes in input order."""
+    if not files:
+        return []
 
-    max_files_per_subgroup = int(cudf_max_num_elements / avg_num_rows)
-    max_files_per_subgroup = max(1, max_files_per_subgroup)  # Ensure at least 1 file per subgroup
+    retained = set(retained_columns or [])
+    try:
+        if storage_options:
```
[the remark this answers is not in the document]
praateekmahajan: Fixed in aa7529ea. Remote URIs now use the fsspec footer path even when storage_options is empty, and the regression test covers a remote-style URI without storage options.

nemo_curator/stages/deduplication/semantic/utils.py
```diff
@@ -12,18 +12,29 @@
...
 
 if TYPE_CHECKING:
     import cudf
     import cupy as cp
 
-from typing import Any
-
 import pyarrow.parquet as pq
-from fsspec.parquet import open_parquet_file
+from fsspec.parquet import open_parquet_files
```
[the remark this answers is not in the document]
praateekmahajan: It is valid

nemo_curator/stages/deduplication/semantic/kmeans.py
```diff
@@ -132,40 +136,208 @@ def process(self, task: FileGroupTask) -> EmptyTask:
...
+            return self._process_parquet(tasks, all_files)
+        if self.filetype != "jsonl":
             msg = f"Unsupported filetype: {self.filetype}. Only jsonl and parquet are supported."
             raise ValueError(msg)
 
         if self.fit_data_fraction is not None:
-            return self._process_batch_two_pass(tasks, groups)
-        return self._process_batch_single_pass(tasks, groups)
+            return self._process_batch_two_pass(tasks, all_files)
+        return self._process_batch_single_pass(tasks, all_files)
```
sarahyurick: Why we are keeping both of `_process_batch_two_pass` and `_process_batch_single_pass` if the new behavior defaults to "however much fits in 60% of GPU memory" (if I am understanding correctly)?

sarahyurick: Is it only because of Parquet vs JSONL?

nemo_curator/stages/deduplication/semantic/kmeans.py
```diff
@@ -132,40 +136,208 @@ def process(self, task: FileGroupTask) -> EmptyTask:
...
+            write_time += time.perf_counter() - write_start
+            predicted_rows += len(metadata)
+            output_index += 1
+
+        sampled_chunks.clear()
+        del fit_embeddings, fit_labels
+        gc.collect()
+        cp.get_default_memory_pool().free_all_blocks()
+
+        if remaining_info:
```
sarahyurick: Nit but to me `remaining_info` wasn't clear that it is "the remaining data that we didn't fit with", could we rename and/or add comments about that?

nemo_curator/stages/deduplication/semantic/kmeans.py
```diff
@@ -132,40 +136,208 @@ def process(self, task: FileGroupTask) -> EmptyTask:
...
+            raise ValueError(msg)
+
+        fit_frames = iter(self._iter_parquet_frames(fit_info, columns))
+        first_fit_frame = next(fit_frames)
+        embedding_width = get_array_from_df(first_fit_frame, self.embedding_field).shape[1]
+        fit_embeddings = cp.empty((fit_rows, embedding_width), dtype=cp.float32)
+        sampled_chunks: list[tuple[cudf.DataFrame, int, int]] = []
+        read_time = 0.0
+        offset = 0
+        read_start = time.perf_counter()
```
sarahyurick: I thought the read starts at `first_fit_frame = ...`? Shouldn't this timer be up there?

nemo_curator/stages/deduplication/semantic/kmeans.py
```diff
@@ -132,40 +136,208 @@ def process(self, task: FileGroupTask) -> EmptyTask:
...
+        for df in chain([first_fit_frame], fit_frames):
+            stop = offset + len(df)
+            embeddings = fit_embeddings[offset:stop]
+            embeddings[...] = get_array_from_df(df, self.embedding_field)
+            embeddings /= cp.linalg.norm(embeddings, axis=1, keepdims=True)
```
sarahyurick: Sorry why don't we want `normalize_embeddings_col_in_df` here?

nemo_curator/stages/deduplication/semantic/kmeans.py
```diff
@@ -132,40 +136,208 @@ def process(self, task: FileGroupTask) -> EmptyTask:
...
+                self._write_output_frame(
+                    f"{tasks[0].task_id}_{output_index}.parquet", df, embeddings, labels, centroids
+                )
+                write_time += time.perf_counter() - write_start
+                predicted_rows += len(df)
+                output_index += 1
+                read_start = time.perf_counter()
+            read_time += time.perf_counter() - read_start
+
+        del self.kmeans
```
sarahyurick: Why do we delete here? Should we do it for the JSONL path too?

nemo_curator/stages/deduplication/semantic/kmeans.py
```diff
@@ -544,8 +620,8 @@ def __post_init__(self):
         super().__init__()
         # Validate eagerly so bad values surface at construction, not later in
         # decompose() / on a worker.
-        if self.fit_data_fraction is not None and not 0.0 < self.fit_data_fraction < 1.0:
-            msg = f"fit_data_fraction must be in (0, 1), got {self.fit_data_fraction}; pass None to fit on the full dataset"
+        if self.fit_data_fraction is not None and not 0.0 < self.fit_data_fraction <= 1.0:
+            msg = f"fit_data_fraction must be in (0, 1], got {self.fit_data_fraction}; pass None for automatic sizing"
```
sarahyurick: General comment should we warn when automatic sizing fits on something other than the full dataset? Especially if it is a very small subset that ends up being fit.

nemo_curator/stages/deduplication/semantic/utils.py
```diff
@@ -32,29 +45,101 @@ def get_array_from_df(df: "cudf.DataFrame", embedding_col: str) -> "cp.ndarray":
...
 
-def break_parquet_partition_into_groups(
-    files: list[str], embedding_dim: int | None = None, storage_options: dict[str, Any] | None = None
-) -> list[list[str]]:
-    """Break parquet files into groups to avoid cudf 2bn row limit."""
-    if embedding_dim is None:
-        # Default aggressive assumption of 1024 dimensional embedding
-        embedding_dim = 1024
+def _root_column(path: str | list[str]) -> str:
+    return path[0] if isinstance(path, list) else path.split(".", maxsplit=1)[0]
```
sarahyurick: Couldn't this fail if a real column actually has a dot in it?

nemo_curator/stages/deduplication/semantic/utils.py
```diff
@@ -32,29 +45,101 @@ def get_array_from_df(df: "cudf.DataFrame", embedding_col: str) -> "cp.ndarray":
...
+) -> list[ParquetFileInfo]:
+    """Read exact row, embedding-element, and projected metadata sizes in input order."""
+    if not files:
+        return []
 
-    max_files_per_subgroup = int(cudf_max_num_elements / avg_num_rows)
-    max_files_per_subgroup = max(1, max_files_per_subgroup)  # Ensure at least 1 file per subgroup
+    retained = set(retained_columns or [])
+    try:
+        if storage_options or get_protocol(files[0]) != "file":
```
sarahyurick, line 72: Curious have you done an S3 benchmark?

nemo_curator/stages/deduplication/semantic/utils.py
```diff
@@ -32,29 +45,101 @@ def get_array_from_df(df: "cudf.DataFrame", embedding_col: str) -> "cp.ndarray":
...
+                embedding_elements = sum(
+                    column.meta_data.num_values
+                    for row_group in footer.row_groups
+                    for column in row_group.columns
+                    if _root_column(column.meta_data.path_in_schema) == embedding_column
+                )
+                result.append(ParquetFileInfo(path, footer.num_rows, metadata_bytes, embedding_elements))
+        return result  # noqa: TRY300
+    except Exception as error:
+        msg = f"Failed to read Parquet footer metadata for one of {len(files)} files"
```
sarahyurick: Should we try to say which file(s) failed?

nemo_curator/stages/deduplication/semantic/utils.py
```diff
@@ -32,29 +45,101 @@ def get_array_from_df(df: "cudf.DataFrame", embedding_col: str) -> "cp.ndarray":
...
+        result = []
+        for start in range(0, len(files), _FOOTER_BATCH_SIZE):
+            batch = files[start : start + _FOOTER_BATCH_SIZE]
+            footers = plc.io.parquet_metadata.read_parquet_footers(plc.io.SourceInfo(batch))
+            for path, footer in zip(batch, footers, strict=True):
+                metadata_bytes = sum(
+                    column.meta_data.total_uncompressed_size
+                    for row_group in footer.row_groups
+                    for column in row_group.columns
+                    if _root_column(column.meta_data.path_in_schema) in retained
```
sarahyurick: Nit but it does seem like we could reduce `read_parquet_file_info`'s complexity here? Like maybe I am reading it wrong but the `_root_column` logic here and in the previous code block are both running the same logic right (in terms of nested for loops and if blocks)?

nemo_curator/stages/deduplication/semantic/utils.py
```diff
@@ -32,29 +45,101 @@ def get_array_from_df(df: "cudf.DataFrame", embedding_col: str) -> "cp.ndarray":
...
+                )
+                embedding_elements = sum(
+                    column.meta_data.num_values
```
sarahyurick: Not too familiar with what is the difference between using `column.meta_data.num_values` here and `column.num_values` above?

nemo_curator/stages/deduplication/semantic/utils.py
sarahyurick, line 123: Should we try to say which file(s) failed?

nemo_curator/stages/deduplication/semantic/kmeans.py
praateekmahajan: Yes - these are now exclusively the JSONL paths. None preserves JSONL’s full single-pass behavior (because we can't infer the number of rows without reading the files), while an explicit fraction requires two passes. I’ll rename them to make that distinction obvious.

nemo_curator/stages/deduplication/semantic/kmeans.py
praateekmahajan: `prediction_only_info` sounds good?

nemo_curator/stages/deduplication/semantic/kmeans.py
praateekmahajan: Good catch! It misses that one read.

nemo_curator/stages/deduplication/semantic/kmeans.py
praateekmahajan: Actually no need to remove it, because the actor pool anyway dies after this. So can remove it

nemo_curator/stages/deduplication/semantic/kmeans.py
praateekmahajan: Yes we can add, however one thing after the other comment I realised that `None` behaves differently for jsonl/parquet so we should warn for that too

nemo_curator/stages/deduplication/semantic/utils.py
praateekmahajan: Yea, also TBH i don't find this code super clear, so will update it to be clearer

nemo_curator/stages/deduplication/semantic/utils.py
praateekmahajan: The batch might have 64 files which is too many to report, it might be an overkill to find which of the 64 failed. lmk what you think

nemo_curator/stages/deduplication/semantic/kmeans.py
praateekmahajan: That helper creates another full normalized Torch array and assigns it back into cuDF. Whereas the current proposed  path copies into the exactly allocated CuPy fit buffer and normalizes each slice in place, which is what removes the ~2× fit-memory plateau. 
I'll see if we can get rid of the `normalize_embeddings_col_in_df` or rather put this code there

nemo_curator/stages/deduplication/semantic/utils.py
praateekmahajan: Yea, sadly we use fsspec for remote and pylibcudf for local which both differ in the metadata's object that they return. So math might be the same however attribute handling still remains different

nemo_curator/stages/deduplication/semantic/utils.py
praateekmahajan: Same as above, different metadata objects, PyArrow returns `ColumnChunkMetaData` directly; pylibcudf returns a `ColumnChunk` whose metadata is under `.meta_data`, can rename them though

nemo_curator/stages/deduplication/semantic/utils.py
praateekmahajan: Nope sadly not

nemo_curator/stages/deduplication/semantic/kmeans.py
```diff
@@ -89,7 +94,7 @@ def __init__(  # noqa: PLR0913
...
-            fit_data_fraction (float | None): Fraction of the dataset (in (0, 1)) used to fit the KMeans model. Pass None to fit on the full dataset (single-pass mode). Sampling is done at the file level: each actor draws `round(fit_data_fraction x num_actor_files)` of its files (floor of 1) for the fit. When set, uses a two-pass approach: Pass 1 reads only the embedding column from the sampled files (so IO and GPU memory in Pass 1 scale with the fraction); Pass 2 then loads each (full) original group one at a time to predict labels and write results. This dramatically reduces peak GPU memory for large datasets. If None, all rows are loaded simultaneously.
+            fit_data_fraction (float | None): Fraction of whole files used to fit KMeans. None selects as many complete files as fit the live GPU-memory budget.
```
sarahyurick: Can we update the documentation and docstrings to be more precise about Parquet vs JSONL behavior?

And maybe the docs should be updated say that using Parquet is more optimal for semantic dedup because of xyz features.

nemo_curator/stages/deduplication/semantic/utils.py
sarahyurick: Yeah I guess not. I am just wondering about when this error is raised, what is a good way for the user to figure out how to fix their file(s) and/or file(s) to exclude in the next run?

tests/stages/deduplication/semantic/test_utils.py
```diff
@@ -45,121 +46,99 @@ def test_get_array_from_df() -> None:
...
+    file_info = read_parquet_file_info(files, retained_columns=["id"], embedding_column="embeddings")
+
+    # Grouping consumes this list positionally, so footer batching must not reorder the caller's files.
+    assert [info.path for info in file_info] == files
+    assert [info.num_rows for info in file_info] == [7, 7, 7]
+    assert [info.embedding_elements for info in file_info] == [14, 14, 14]
+
+
+@pytest.mark.gpu
+def test_parquet_file_info_uses_fsspec_for_remote_uri_without_storage_options() -> None:
```
sarahyurick, line 77: I was wondering if we can remove the GPU coupling here yet or not? Even though it is used for semantic dedup only, it is still IO.

praateekmahajan: Yea we have a todo below for `TODO : Remove this once we figure out how to import semantic on CPU` 


Maybe I should make it top-level.. 

The whole imports cause this issue, and would prefer not to resolve that in this PR

nemo_curator/stages/deduplication/semantic/utils.py
praateekmahajan: Today we have this issue even with parquet reads, a corrupt file doesn't share which file caused the failure. So I think we can let it be, added a TODO tho in case we run into it

nemo_curator/stages/deduplication/semantic/kmeans.py
```diff
@@ -109,11 +115,10 @@ def __init__(  # noqa: PLR0913
         self.n_init = n_init
         self.oversampling_factor = oversampling_factor
         self.max_samples_per_batch = max_samples_per_batch
-        if fit_data_fraction is not None and not 0.0 < fit_data_fraction < 1.0:
-            msg = f"fit_data_fraction must be in (0, 1), got {fit_data_fraction}; pass None to fit on the full dataset"
+        if fit_data_fraction is not None and not 0.0 < fit_data_fraction <= 1.0:
+            msg = f"fit_data_fraction must be in (0, 1], got {fit_data_fraction}; pass None for automatic sizing"
```
sarahyurick: Nit can you update this message to mention Parquet vs JSONL too?

nemo_curator/stages/deduplication/semantic/kmeans.py
```diff
@@ -544,9 +626,14 @@ def __post_init__(self):
         super().__init__()
         # Validate eagerly so bad values surface at construction, not later in
         # decompose() / on a worker.
-        if self.fit_data_fraction is not None and not 0.0 < self.fit_data_fraction < 1.0:
-            msg = f"fit_data_fraction must be in (0, 1), got {self.fit_data_fraction}; pass None to fit on the full dataset"
+        if self.fit_data_fraction is not None and not 0.0 < self.fit_data_fraction <= 1.0:
+            msg = f"fit_data_fraction must be in (0, 1], got {self.fit_data_fraction}; pass None for automatic sizing"
```
sarahyurick: Similar comment as above.

nemo_curator/stages/deduplication/semantic/workflow.py
```diff
@@ -211,8 +209,8 @@ def _validate_config(self) -> None:
             )
 
         # Validate fit_data_fraction
-        if self.fit_data_fraction is not None and not 0.0 < self.fit_data_fraction < 1.0:
-            msg = f"fit_data_fraction must be in (0, 1), got {self.fit_data_fraction}; pass None to fit on the full dataset"
+        if self.fit_data_fraction is not None and not 0.0 < self.fit_data_fraction <= 1.0:
+            msg = f"fit_data_fraction must be in (0, 1], got {self.fit_data_fraction}; pass None for automatic sizing"
```
sarahyurick: Similar comment as above.

nemo_curator/stages/text/deduplication/semantic.py
```diff
@@ -194,8 +194,8 @@ def __post_init__(self):
 
     def _validate_config(self) -> None:
         """Validate workflow configuration."""
-        if self.kmeans_fit_data_fraction is not None and not 0.0 < self.kmeans_fit_data_fraction < 1.0:
-            msg = f"kmeans_fit_data_fraction must be in (0, 1), got {self.kmeans_fit_data_fraction}; pass None to fit on the full dataset"
+        if self.kmeans_fit_data_fraction is not None and not 0.0 < self.kmeans_fit_data_fraction <= 1.0:
+            msg = f"kmeans_fit_data_fraction must be in (0, 1], got {self.kmeans_fit_data_fraction}; pass None for automatic sizing"
```
sarahyurick: Similar comment as above.

sarahyurick (approved):

PR #2370, opened 2026-09-04
perf(ray-actor-pool): balance weighted RAFT tasks

praateekmahajan:
## Summary

- let file partitioning opt in to emitting a task weight; semantic KMeans uses input bytes
- weight fuzzy connected-components inputs by edge count
- balance weighted RAFT tasks deterministically with largest-first assignment
- preserve existing equal-count splitting when weights are absent

Exact deduplication and fuzzy LSH remain unchanged because they use the shuffle path rather than RAFT batching.

## Motivation

Equal task counts can produce uneven per-GPU input sizes, leaving collective KMeans or connected-components workers waiting on the most heavily loaded actor.

## Alternatives considered

- deterministic randomization: smaller, but probabilistic
- blocksize grouping: simple, but requires tuning and does not handle oversized files

## Testing

- focused unit tests and pre-commit passed
- GPU-specific fuzzy unit tests were not run locally because cudf is unavailable

## EOS benchmark

[Run viewer](http://rratzel-ws1:5050/run-viewer?dir=login-eos-nightly%3A%2Flustre%2Ffsw%2Fcoreai_dlalgo_ci%2Fcurator_ci%2Fresults%2Fmain-mirror&run=pr-2370-2026_09_04__21_40_40_UTC-43948b27)  and [Comparision](http://rratzel-ws1:5050/compare?dir_a=login-eos-nightly%3A%2Flustre%2Ffsw%2Fcoreai_dlalgo_ci%2Fcurator_ci%2Fresults%2Fmain&dir_b=login-eos-nightly%3A%2Flustre%2Ffsw%2Fcoreai_dlalgo_ci%2Fcurator_ci%2Fresults%2Fmain-mirror&run_a=nightly-2026_09_04__06_23_39_UTC&run_b=pr-2370-2026_09_04__21_40_40_UTC-43948b27)
All three relevant 8xH100 entries passed. Compared with the same-day main nightly:

| Entry | Main | PR | Delta |
|---|---:|---:|---:|
| fuzzy dedup | 446.8s | 449.1s | +0.5% |
| semdedup 20% dataset | 214.9s | 219.1s | +2.0% |
| semdedup 20% fit | 1138.1s | 1141.1s | +0.3% |

**There is no material speedup or slowdown. The full semdedup run reduced per-actor input-row standard deviation from 51,585 to 25**, but the baseline imbalance was already only about 0.1% of the mean, so elapsed time remained unchanged. 
**This is directionally the intended behavior for skewed inputs while remaining neutral on already-balanced data.**

### DIFF
--- a/nemo_curator/backends/ray_actor_pool/executor.py
+++ b/nemo_curator/backends/ray_actor_pool/executor.py
@@ -325,7 +325,10 @@ class RayActorPoolExecutor(BaseExecutor):
                 for task, weight in sorted(
                     zip(tasks, task_weights, strict=True), key=lambda item: item[1], reverse=True
                 ):
-                    batch_index = min(range(num_batches), key=batch_weights.__getitem__)
+                    # Break equal-weight ties by task count so zero-weight tasks still reach every RAFT actor.
+                    batch_index = min(
+                        range(num_batches), key=lambda index: (batch_weights[index], len(batches[index]))
+                    )
                     batches[batch_index].append(task)
                     batch_weights[batch_index] += weight
                 return batches
--- a/nemo_curator/backends/ray_actor_pool/executor.py
+++ b/nemo_curator/backends/ray_actor_pool/executor.py
@@ -292,13 +292,18 @@ class RayActorPoolExecutor(BaseExecutor):
         return actors
 
     def _generate_task_batches(
-        self, tasks: list[Task], batch_size: int | None = None, num_output_tasks: int | None = None
+        self,
+        tasks: list[Task],
+        batch_size: int | None = None,
+        num_output_tasks: int | None = None,
+        task_weights: list[int] | None = None,
     ) -> list[list[Task]]:
         """Generate task batches from a list of tasks.
         Args:
             tasks: List of Task objects to process
             batch_size: The size of the batch
             num_output_tasks: The number of output tasks to generate.
+            task_weights: Optional weights used to balance tasks across output batches.
             Either batch_size or num_output_tasks must be provided but not both.
         Returns:
             List of task batches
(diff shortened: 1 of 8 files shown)


tests/stages/deduplication/fuzzy/test_buckets_to_edges_stage.py
```diff
@@ -107,6 +107,7 @@ def test_basic_edge_creation(self, input_task: FileGroupTask, tmp_path: Path) ->
         assert len(output_task.data) == 1
         assert output_task.dataset_name == "test_buckets_edges"
         assert "storage_options" in output_task._metadata
```
ayushdg, line 109: Is it possible to add some more tests either in kmeans and fuzzy or standalone with skewed incoming weights to see that after assignment the skew is balanced out compared to not using any task weights

nemo_curator/backends/ray_actor_pool/executor.py
```diff
@@ -341,7 +363,12 @@ def _process_stage_with_pool(
                     f"Stage {_stage.name} is a RAFT stage but has a batch size of {stage_batch_size}. Ignoring batch size."
                 )
             num_actors = len(actor_pool._idle_actors)
-            task_batches = self._generate_task_batches(tasks, num_output_tasks=num_actors)
+            task_weights = [(task._metadata or {}).get("task_weight") for task in tasks]
```
ayushdg: Is there a easy way to disable this if needed?

ayushdg (changes requested):
Minor questions/comments

tests/stages/deduplication/fuzzy/test_buckets_to_edges_stage.py
praateekmahajan: We have a test in `ray_actor_pool/test_executor.py` that is exactly that right, in the base class? It shows when task_weight is provided it breaks into two batches of `16, 17` as opposed to `27, 6` for the same weights. And then any `RaftAdapter` should process those batches accordingly.

ayushdg (approved):

PR #1911, opened 2026-05-05
ci: Major refactor of release-workflows

ko3n1g:
## Why

See the design discussion in NVIDIA-NeMo/FW-CI-templates#466.

## What

- **Delete** `.github/workflows/build-test-publish-wheel.yml`.
- **Rewrite** `.github/workflows/release.yml` as the single caller for both `push` and `workflow_dispatch`.

## Test plan

- [x] PR validate-only (push (mirror), sha fdbac7fc1c, 2026-05-07T12:01:00Z, success): https://github.com/NVIDIA-NeMo/Curator/actions/runs/25494488916
- [x] `workflow_dispatch dry-run=true` (sha fdbac7fc1c, 2026-05-07T12:00:56Z, success): https://github.com/NVIDIA-NeMo/Curator/actions/runs/25494485717
- [ ] Real release via `workflow_dispatch dry-run=false` on the next planned RC.

## Rollout

1. Land FW-CI-templates#466.
2. Cut FW-CI-templates `v1.0.0`.
3. Bump the SHA pin in this PR → tag.

PR #2371, opened 2026-09-05
fix(metrics): warn when the packaged Xenna Grafana dashboard is missing (#2189)

### LINKED ISSUE #2189: xenna_grafana_dashboard.json missing from the wheel — Xenna Grafana dashboard is never provisioned
## Summary

`nemo_curator/metrics/xenna_grafana_dashboard.json` exists in the git tree but is not included in the published wheel, so `launch_grafana`'s dashboard copy silently skips and wheel users never get the Xenna dashboard in Grafana.

## Reproduction (nemo-curator 1.2.0)

```
$ pip install nemo-curator
$ ls .venv/lib/python3.12/site-packages/nemo_curator/metrics/
__init__.py  constants.py  start_prometheus_grafana.py  utils.py
```

The json (and the module README) are absent. In `nemo_curator/metrics/utils.py` the copy is guarded:

```python
xenna_dashboard_src = os.path.join(os.path.dirname(__file__), "xenna_grafana_dashboard.json")
...
if os.path.isfile(xenna_dashboard_src) and not os.path.isfile(xenna_dashboard_dst):
    shutil.copy(xenna_dashboard_src, xenna_dashboard_dst)
```

so the missing file skips silently: `start_prometheus_grafana` provisions the four auto-generated Ray dashboards but never the Xenna one.

## Expected

The Xenna dashboard json ships in the wheel (package data), so monitoring set up per the metrics README includes the Xenna dashboard.

## Suggested fix

Include `nemo_curator/metrics/*.json` as package data in the build configuration; optionally log a warning instead of skipping silently when the source json is missing.

chethanuk:
## Description

`write_grafana_configs()` folded the packaged-file check into the copy condition:

```python
if os.path.isfile(xenna_dashboard_src) and not os.path.isfile(xenna_dashboard_dst):
    shutil.copy(xenna_dashboard_src, xenna_dashboard_dst)
```

so "the packaged dashboard is missing" and "it is already provisioned" took the same silent
branch. When the JSON was absent from the install the function returned normally, having written
only the Ray dashboards, with no log at any level. The reporter had to list `site-packages` to
work out why the Xenna dashboard never appeared in Grafana.

This guards on the destination first and warns only when the packaged source is genuinely
absent, so an already-provisioned dashboard stays quiet and is never overwritten.

Packaging itself is untouched: the missing-from-the-wheel half of the report was already fixed
by ad587436 (#1911), which added `recursive-include nemo_curator ... *.json` to `MANIFEST.in`.
What was left is the silent skip, which is what the issue's "Suggested fix" asks for.

closes #2189

## Usage

No API change. When the packaged dashboard is missing, `write_grafana_configs()` now logs it
instead of returning quietly:

```python
from nemo_curator.metrics.utils import write_grafana_configs

write_grafana_configs(3000, 9090)
# WARNING | Xenna Grafana dashboard not found at /.../nemo_curator/metrics/xenna_grafana_dashboard.json;
#           the Xenna dashboard will not be provisioned. This usually means the packaged data file is
#           missing from the installed nemo_curator distribution.
```

## Tests

Added a parametrized test in `tests/metrics/test_metrics_utils.py` covering the four
(source present, destination present) states for `write_grafana_configs`; only the
"source missing, destination missing" case fails on `main` today, with the reported symptom
(no warning logged). All four pass with this fix, and the rest of `tests/metrics/` (28 tests)
is unaffected.

- `pytest tests/metrics/test_metrics_utils.py -k WriteGrafanaConfigs` — 4 passed
- `pytest tests/metrics/` — 28 passed
- `ruff check` / `ruff format --check` on both changed files — clean

## Checklist

### DIFF
--- a/nemo_curator/metrics/utils.py
+++ b/nemo_curator/metrics/utils.py
@@ -263,8 +263,15 @@ def write_grafana_configs(grafana_web_port: int, prometheus_web_port: int, metri
     xenna_dashboard_src = os.path.join(os.path.dirname(__file__), "xenna_grafana_dashboard.json")
     xenna_dashboard_src = os.path.abspath(xenna_dashboard_src)
     xenna_dashboard_dst = os.path.join(dashboards_path, "xenna_grafana_dashboard.json")
-    if os.path.isfile(xenna_dashboard_src) and not os.path.isfile(xenna_dashboard_dst):
-        shutil.copy(xenna_dashboard_src, xenna_dashboard_dst)
+    if not os.path.isfile(xenna_dashboard_dst):
+        if os.path.isfile(xenna_dashboard_src):
+            shutil.copy(xenna_dashboard_src, xenna_dashboard_dst)
+        else:
+            logger.warning(
+                f"Xenna Grafana dashboard not found at {xenna_dashboard_src}; "
+                "the Xenna dashboard will not be provisioned. This usually means the "
+                "packaged data file is missing from the installed nemo_curator distribution."
+            )
 
     # Generate Ray's default Grafana dashboards
     _write_ray_default_dashboards(dashboards_path)
--- a/tests/metrics/test_metrics_utils.py
+++ b/tests/metrics/test_metrics_utils.py
@@ -16,6 +16,7 @@ import os
 import pathlib
 from unittest.mock import patch
 
+import pytest
 import yaml
 
 from nemo_curator.metrics.constants import (
@@ -33,6 +34,7 @@ from nemo_curator.metrics.utils import (
     is_grafana_running,
     is_prometheus_running,
     remove_ray_prometheus_metrics_service_discovery,
+    write_grafana_configs,
 )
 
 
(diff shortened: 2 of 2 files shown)


nemo_curator/metrics/utils.py
```diff
@@ -263,8 +263,15 @@ def write_grafana_configs(grafana_web_port: int, prometheus_web_port: int, metri
...
-    if os.path.isfile(xenna_dashboard_src) and not os.path.isfile(xenna_dashboard_dst):
-        shutil.copy(xenna_dashboard_src, xenna_dashboard_dst)
+    if not os.path.isfile(xenna_dashboard_dst):
+        if os.path.isfile(xenna_dashboard_src):
+            shutil.copy(xenna_dashboard_src, xenna_dashboard_dst)
+        else:
+            logger.warning(
+                f"Xenna Grafana dashboard not found at {xenna_dashboard_src}; "
+                "the Xenna dashboard will not be provisioned. This usually means the "
+                "packaged data file is missing from the installed nemo_curator distribution."
```
sarahyurick, line 273: Hi @chethanuk thank you for opening! I was wondering if the warning could include an instruction for how to grab the JSON file, and if there are any additional instructions for setting a custom `xenna_dashboard_src` path as needed.

PR #2376, opened 2026-09-07
fix(modules): treat DocumentSplitter separator as a literal string

### LINKED ISSUE #2374: DocumentSplitter treats multi-character separators as regex (wrong splits, re.error crash on "++")
### Description

`DocumentSplitter` passes its separator to `Series.str.split()` without `regex=False`, so pandas treats multi-character separators as regular expressions:

- `DocumentSplitter(separator="..")` on text `a..b..c` produces segments `['', '', '', 'c']` (regex `..` matches any two chars) instead of `['a', 'b', 'c']`
- `separator="||"` splits on empty string alternation — wrong boundaries
- `separator="++"` crashes with `re.error: nothing to repeat at position 0`
- the documented round-trip with `DocumentJoiner` is broken, because the joiner always joins with the literal separator

### Steps to reproduce

```python
import pandas as pd
from nemo_curator.stages.text.modules import DocumentSplitter

# crash case
DocumentSplitter(separator="++")
# re.error: nothing to repeat at position 0  (raised per batch, at str.split)

# silent corruption case
pd.DataFrame({"text": ["a..b..c"]})["text"].str.split("..")
# [['', '', '', 'c']]
```

### Expected behavior

The separator is documented as a plain string ("The separator to split the documents on") and `DocumentJoiner` treats it as literal — splitting should too.

### Actual behavior

`df[self.text_field].str.split(self.separator)` — pandas only treats single-char string patterns as literal; multi-char patterns go through `re`.

### Environment

nemo-curator main (`93b48525`), Python 3.12

ManoharPaturi:
Fixes #2374.

One line: pass `regex=False` to the `str.split` call so the separator is treated as a literal string, matching the docstring and the `DocumentJoiner` round-trip (which always joins literally).

- `".."`, `"||"`, `"(sep)"` now split on the actual separator
- `"++"` no longer raises `re.error`
- round-trip split→join now reproduces the original text for metacharacter separators

3 new tests (metacharacter split, `++` crash, DocumentJoiner round-trip) — all fail on main, pass here. Rest of the splitter suite unchanged.

### DIFF
--- a/tests/stages/text/modules/test_splitter.py
+++ b/tests/stages/text/modules/test_splitter.py
@@ -14,6 +14,7 @@
 
 import pandas as pd
 
+from nemo_curator.stages.text.modules.joiner import DocumentJoiner
 from nemo_curator.stages.text.modules.splitter import DocumentSplitter
 from nemo_curator.tasks import DocumentBatch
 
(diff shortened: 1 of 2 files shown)


tests/stages/text/modules/test_splitter.py
```diff
@@ -260,3 +261,71 @@ def test_reconstruction_with_unique_ids(self):
         doc2_segments = result_df[result_df["doc_id"] == "doc2"].sort_values("segment_id")
         reconstructed_doc2 = "\n\n".join(doc2_segments["text"].tolist())
         assert reconstructed_doc2 == "Foo\n\nBar"
+
+    def test_separator_is_treated_as_literal_string(self):
```
sarahyurick: Hi @ManoharPaturi thanks for opening! I was wondering is there any way for you to create a single parametrized test instead of adding 3 separate tests?

ManoharPaturi: Thanks @sarahyurick! That makes total sense I've consolidated the cases into a single `@pytest.mark.parametrize` test (`test_literal_separator_and_joiner_roundtrip`) covering multi-character metacharacters (`..`), invalid regexes (`++`), and groups (`(sep)`) along with the `DocumentJoiner` round trip. Formatted with Ruff and pushed in 0c2ceb4.

sarahyurick (approved):
LGTM, thanks!

PR #2377, opened 2026-09-07
fix(utils): split_large_files RecursionError on single row larger than target size

### LINKED ISSUE #2375: split_large_files: RecursionError when a single parquet row exceeds the target size
### Description

`split_large_files.py`'s `_split_table` recurses forever (ends in `RecursionError`) when a parquet chunk contains a single row that is itself larger than the target size. There is no base case for `num_rows <= 1`, so the bisection keeps slicing the same 1-row table in half (1 // 2 = 0) forever.

This is realistic for interleaved-document parquet files with large embedded payloads (e.g. base64 images) — one 3 MB row with `--target-size-mb 1` kills the tool.

### Steps to reproduce

```python
import pyarrow as pa
from nemo_curator.utils.split_large_files import _split_table

t = pa.table({"text": ["x" * 3_000_000]})
list(_split_table(t, 1024 * 1024))
# RecursionError
```

### Expected behavior

A row that can't be split further becomes its own shard (same behavior the JSONL splitter documents for oversized single records — write it alone, ideally with a warning), and the tool finishes.

### Actual behavior

`_split_table` slices `[0 : n//2]` and `[n//2 : n]`; for n=1 both slices are 1-row/0-row tables that never get smaller → infinite recursion.

### Environment

nemo-curator main (`93b48525`), Python 3.12

ManoharPaturi:
Fixes #2375.

Adds the missing base case to `_split_table`: when a chunk has `num_rows <= 1` and is still over target, it is returned as its own shard (with a warning), instead of recursing on the same 1-row slice forever. Mirrors the JSONL splitter's documented behavior for oversized single records.

2 new tests (unit base-case + end-to-end round-trip with a single oversized row) — both fail on main with RecursionError, pass here.

### DIFF
--- a/nemo_curator/utils/split_large_files.py
+++ b/nemo_curator/utils/split_large_files.py
@@ -50,6 +50,18 @@ def _join_out_path(output_path: str, filename: str, storage_options: dict[str, A
 
 
 def _split_table(table: pa.Table, target_size: int) -> list[pa.Table]:
+    if table.num_rows <= 1:
+        # A single row cannot be split any further, so return it as its own
+        # chunk even if it is larger than the target size. Without this base
+        # case, bisecting a one-row table produces the same one-row table
+        # forever and the recursion never terminates (RecursionError).
+        if table.num_rows == 1 and table.nbytes > target_size:
+            logger.warning(
+                "Single row ({} bytes) exceeds target ({} bytes); writing it as its own shard",
+                table.nbytes,
+                target_size,
+            )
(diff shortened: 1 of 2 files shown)


sarahyurick:
Not sure why 1 test failed, let me rerun to see whether or not it was random.

sarahyurick (approved):
LGTM thanks @ManoharPaturi ! Will plan to get this merged soon.

PR #2386, opened 2026-09-09
perf(semantic-dedup): reduce KMeans fit memory overhead

praateekmahajan:
KMeans uses two passes for Parquet input: it fits shared centroids from a contiguous float32 embedding array, releases that array, then rereads every input file to predict cluster assignments and write centroid-partitioned output. Separating the passes keeps the fit dataset out of memory while prediction and Parquet writing allocate their working buffers.

image

The purple series shows this PR; the table below reports the latest nightly comparison.

### Fit selection and loading

- `fit_data_fraction=None` automatically selects complete Parquet files per GPU actor. Selection uses exact footer embedding-element counts and budgets 90% of currently free GPU memory for the persistent float32 fit array.
- An explicit fraction in `(0, 1]` selects `round(fraction × actor file count)` files, with a minimum of one. `1.0` fits all files. Sampling uses `random_state`; the realized row fraction depends on file sizes.
- Footer counts determine the fit array's shape. The fit pass reads only the embedding column through cuDF's `ChunkedParquetReader`, dividing remaining memory between output chunks and decompression passes.
- Each chunk is copied and normalized directly into the preallocated array, then released before reading the next chunk. KMeans fits cooperatively across GPU actors; actor zero saves the shared centroids when a cache path is configured.

### Prediction and output

After fitting, the stage releases the fit array and rereads all Parquet input, including when automatic selection or an explicit `1.0` fraction fits every file. Full and sampled fits use the same prediction/write path.

The second pass processes one group of complete files at a time. It reads IDs, embeddings, and configured metadata; normalizes embeddings; predicts the assigned centroid; computes L2 and cosine distances; and writes Parquet under `centroid=<id>` directories. All input rows receive assignments, including rows outside the fit sample. Output embeddings use the configured float16 or float32 representation.

Prediction groups use footer embedding-element counts to stay below cuDF's child-column limit. This limit is distinct from available GPU memory: a large group or metadata-heavy input can still exhaust memory during reading, prediction, or writing. Fit selection is also an estimate; an explicit fit fraction must leave room for reader and KMeans workspace allocations.

### JSONL input

With `fit_data_fraction=None`, JSONL reads, fits, predicts, and writes all input in one pass. With an explicit fraction, it fits sampled files, then rereads all files together for prediction and writing. Automatic memory sizing and chunked fit loading apply to Parquet input.

### Benchmark results

**Before:** the latest `login-eos-nightly` run, September 14 at 06:08 UTC. **After:** this PR's completed run, September 14 at 18:02 UTC. Both runs passed the same two workloads on eight GPUs with 1,000 clusters and FP16 embedding storage / pairwise compute.

Each cell shows **nightly → PR (% change relative to nightly)**; lower is better for all three metrics.

| Fit workload | Workflow wall time | KMeans wall time | Peak sampled GPU memory |
| --- | ---: | ---: | ---: |
| Full fit: 100% of 65.8M rows | 168.4 s → 176.1 s (+4.54%) | 123.6 s → 130.7 s (+5.77%) | 98.02% → 77.63% (−20.80%) |
| Partial fit: 20% of 329.8M rows | 447.9 s → 464.6 s (+3.72%) | 335.4 s → 354.4 s (+5.64%) | 98.03% → 77.81% (−20.63%) |

Full fit clusters all 65.8M input rows. Partial fit trains on approximately 66.0M rows selected from 329.8M input rows, then assigns and writes all 329.8M rows. Within each comparison, input rows and fit rows are identical between nightly and the PR.

The PR lowers peak sampled GPU-memory usage by about **20 percentage points**, with **3.7–4.5% longer workflow time**. Peak memory is the maximum reported usage across all eight GPUs and all recorded samples, expressed as a percentage of device capacity; the parenthesized change is relative percent, not percentage points.

[Latest nightly run viewer](http://rratzel-ws1:5050/run-viewer?dir=login-eos-nightly%3A%2Flustre%2Ffsw%2Fcoreai_dlalgo_ci%2Fcurator_ci%2Fresults%2Fmain&run=nightly-2026_09_14__06_08_49_UTC) · [PR run viewer](http://rratzel-ws1:5050/run-viewer?dir=login-eos-nightly%3A%2Flustre%2Ffsw%2Fcoreai_dlalgo_ci%2Fcurator_ci%2Fresults%2Fmain-mirror&run=praateekmahajan-pr-2386-2026_09_14__18_02_42_UTC-74131a2e)

These are single runs on different EOS nodes. The environments also differ in `quack-kernels` version (nightly: 0.4.1; PR: 0.4.0), so the comparison does not isolate code changes completely.

### Validation

All 31 KMeans tests pass, including distributed clustering, automatic/partial/full fits, JSONL paths, output rows and schema, embedding precision, file sampling, and centroid caching. Repository pre-commit hooks pass.

### DIFF
--- a/nemo_curator/stages/deduplication/semantic/kmeans.py
+++ b/nemo_curator/stages/deduplication/semantic/kmeans.py
@@ -49,8 +49,10 @@ from loguru import logger
 # Column names
 L2_DIST_TO_CENT_COL = "l2_dist_to_cent"
 COSINE_DIST_TO_CENT_COL = "cosine_dist_to_cent"
-_AUTO_FIT_MEMORY_FRACTION = 0.6
-_FIT_READ_GROUP_SIZE_BYTES = 1 << 30
+_AUTO_FIT_MEMORY_FRACTION = 0.9
+_FIT_READ_MEMORY_FRACTION = 0.8
+_FIT_WRITE_MEMORY_FRACTION = 0.8
+_PARQUET_WRITE_MEMORY_AMPLIFICATION = 6
 KMeansEmbeddingOutputDtype = Literal["float16", "float32"]
 
 
@@ -187,7 +189,13 @@ class KMeansReadFitWriteStage(ProcessingStage[FileGroupTask, EmptyTask], Dedupli
             msg = f"KMeans fit sample has {fit_rows} rows but requires at least {self.n_clusters}"
             raise ValueError(msg)
 
-        fit_frames = iter(self._iter_parquet_frames(fit_info, columns, max_embedding_bytes=_FIT_READ_GROUP_SIZE_BYTES))
+        fit_frames = iter(
+            self._iter_parquet_frames(
+                fit_info,
+                columns,
+                max_embedding_bytes=self._max_fit_read_group_bytes(fit_info),
+            )
+        )
         read_start = time.perf_counter()
         first_fit_frame = next(fit_frames)
         embedding_width = get_array_from_df(first_fit_frame, self.embedding_field).shape[1]
@@ -232,17 +240,22 @@ class KMeansReadFitWriteStage(ProcessingStage[FileGroupTask, EmptyTask], Dedupli
         predicted_rows = 0
         output_index = 0
         for metadata, start, stop in sampled_chunks:
-            write_start = time.perf_counter()
-            self._write_output_frame(
-                f"{tasks[0].task_id}_{output_index}.parquet",
-                metadata,
-                fit_embeddings[start:stop],
-                fit_labels[start:stop],
-                centroids,
-            )
-            write_time += time.perf_counter() - write_start
-            predicted_rows += len(metadata)
-            output_index += 1
+            rows_per_write = self._max_fit_write_rows(fit_embeddings.shape[1])
+            for chunk_start in range(start, stop, rows_per_write):
+                chunk_stop = min(chunk_start + rows_per_write, stop)
+                metadata_start = chunk_start - start
+                metadata_stop = chunk_stop - start
+                write_start = time.perf_counter()
+                self._write_output_frame(
+                    f"{tasks[0].task_id}_{output_index}.parquet",
+                    metadata.iloc[metadata_start:metadata_stop],
+                    fit_embeddings[chunk_start:chunk_stop],
+                    fit_labels[chunk_start:chunk_stop],
+                    centroids,
+                )
+                write_time += time.perf_counter() - write_start
+                predicted_rows += chunk_stop - chunk_start
+                output_index += 1
 
         sampled_chunks.clear()
         del fit_embeddings, fit_labels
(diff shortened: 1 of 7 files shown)


nemo_curator/stages/deduplication/semantic/kmeans.py
```diff
@@ -179,119 +177,110 @@ def _process_parquet(self, tasks: list[FileGroupTask], files: list[str]) -> list
...
         read_time = time.perf_counter() - read_start
         if offset != fit_rows:
             msg = f"Parquet footers reported {fit_rows} fit rows but the reader returned {offset}"
             raise RuntimeError(msg)
-        del df, embeddings, first_fit_frame, fit_frames
+        del embeddings
 
         fit_start = time.perf_counter()
         self.kmeans.fit(fit_embeddings, sample_weight=None)
-        fit_labels = cp.asarray(self.kmeans.labels_).astype(cp.int32, copy=False)
```
sarahyurick, line 212: Sorry what was this doing before? Why can we delete it now?

nemo_curator/stages/deduplication/semantic/utils.py
```diff
@@ -116,7 +116,14 @@ def read_parquet_file_info(  # noqa: C901
...
                 )
-                result.append(ParquetFileInfo(path, footer.num_rows, metadata_bytes, embedding_elements))
+                result.append(
+                    ParquetFileInfo(
+                        path,
+                        footer.num_rows,
+                        metadata_bytes,
+                        embedding_elements=embedding_elements,
+                    )
+                )
```
sarahyurick, line 126: Nit but this does not look like an actual change?

sarahyurick (commented):
Mostly LGTM, left minor comments.

nemo_curator/stages/deduplication/semantic/kmeans.py
praateekmahajan: Previously, we reused `self.kmeans.labels_` to write the fit rows without predicting them again. That required retaining the fit embeddings and their metadata until writing finished.
Now we release the fit array, reread every file, and call `predict()` for each group - including the fit rows.

nemo_curator/stages/deduplication/semantic/utils.py
praateekmahajan: Yea not a real change just kwarg

nemo_curator/stages/deduplication/semantic/kmeans.py
```diff
@@ -36,20 +43,10 @@
...
-import random
-import time
-
-from loguru import logger
-
 # Column names
 L2_DIST_TO_CENT_COL = "l2_dist_to_cent"
 COSINE_DIST_TO_CENT_COL = "cosine_dist_to_cent"
-_AUTO_FIT_MEMORY_FRACTION = 0.6
+_AUTO_FIT_MEMORY_FRACTION = 0.9
```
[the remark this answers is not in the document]
praateekmahajan: Partially true that large `n_clusters` we don't know fit memory requirement and whether 0.9 will work, however given the defaults we've tested that auto works and fits upto ~60% data.
If users run into OOM they should change the fit_data_fraction as it'll be harder to consider for all other variables that goes into KMeans

ayushdg (approved):

sarahyurick (approved):

PR #2402, opened 2026-09-16
docs: fix Nemotron-Parse writer examples

VibhuJawa:
## What

Fix the released Nemotron-Parse examples to use `InterleavedParquetWriterStage` with the supported `path` argument.

## Why

VDR-12: the copied examples imported a class that does not exist.

## Validation

- Import and constructor smoke check
- Pre-commit hooks

**Scope:** Documentation only.


sarahyurick (approved):
Fyi you can bypass the full CI/CD on fern-only PRs by adding the `docs-only` label.

PR #2404, opened 2026-09-16
fix: make file discovery failures actionable

VibhuJawa:
## What

Fail clearly when file discovery finds no matching inputs, and warn when extension filtering skips files.

## Why

VDR-14: invalid paths and mixed-format directories could complete successfully with incomplete or empty input.

## Validation

- 39 file-partitioning and file-utility tests
- Pre-commit hooks

**Scope:** Code and tests.


nemo_curator/stages/file_partitioning.py
```diff
@@ -112,8 +112,8 @@ def process(self, _: EmptyTask) -> list[FileGroupTask]:
 
         logger.info(f"Found {len(files)} files")
         if len(files) == 0:
-            logger.warning(f"No files found under {self.file_paths}")
-            return []
+            msg = f"No files matching {self.file_extensions} found under {self.file_paths}"
+            raise ValueError(msg)
```
VibhuJawa, line 117: This feels like it could blow things up downstream. @praateekmahajan .

sarahyurick: +1 I am not sure it should be a hard error.

PR #2405, opened 2026-09-16
fix: standardize Nemotron OCR API key

VibhuJawa:
## What

Use `NVIDIA_API_KEY` consistently for OCR verification while retaining `NVINFERENCE_API_KEY` as a compatibility fallback that warns when used.

## Why

VDR-18: OCR used a different key name from every other NVIDIA integration.

## Validation

- 8 focused Omni client tests
- End-to-end GPU tutorial run with one image and verified OCR output
- Pre-commit hooks

**Scope:** Code, tests, and documentation.


nemo_curator/models/omni/base.py
```diff
@@ -41,6 +42,8 @@ def setup(self) -> None:
         if getattr(self, "client", None) is not None:
             return
         api_key = os.environ.get(self.api_key_env_var, "").strip()
+        if not api_key and self.api_key_env_var == "NVIDIA_API_KEY":  # pragma: allowlist secret
+            api_key = os.environ.get("NVINFERENCE_API_KEY", "").strip()
```
sarahyurick, line 48: Nit should it log that it is grabbing from `NVINFERENCE_API_KEY`?

VibhuJawa: Added in 33670950. The client now warns only when NVIDIA_API_KEY is absent and the legacy NVINFERENCE_API_KEY actually supplies the credential; the canonical path remains silent.

VibhuJawa:
Additional validation: ran tutorials/synthetic/omni/ocr_pipeline.py end-to-end in the existing GPU Curator image on one local image. The three-stage Xenna/Ray pipeline completed in 77 seconds and produced the expected text NVIDIA OCR TEST 12345 with a valid bounding box.

sarahyurick (approved):
Thanks!

PR #2406, opened 2026-09-16
fix: bound Ray Serve startup

VibhuJawa:
## What

Submit Ray Serve without its unbounded readiness wait, then enforce one Curator-owned startup deadline with existing cleanup.

## Why

VDR-19: `health_check_timeout_s` started only after `serve.run` returned and could not bound startup.

## Validation

- 8 focused serve tests passed; 1 optional-dependency module skipped
- Pre-commit hooks

**Scope:** Code and tests.


tests/core/serve/ray_serve/test_backend.py
```diff
@@ -12,11 +12,14 @@
...
+from unittest import mock
+
 import pytest
 
-from nemo_curator.core.serve import RayServeModelConfig
+from nemo_curator.core.serve import InferenceServer, RayServeModelConfig
 from nemo_curator.core.serve.ray_serve.backend import RayServeBackend
 
+serve = pytest.importorskip("ray.serve", reason="ray[serve] not installed")
 LLMConfig = pytest.importorskip("ray.serve.llm", reason="ray[serve] not installed").LLMConfig
```
sarahyurick: Should we remove both of these?

VibhuJawa: AI SLOP: 

Removed both module aliases. A single bare `importorskip` remains because `ray[serve]` is optional for the base test environment. The mocks now patch Ray Serve by import path, and the existing field assertions make the `isinstance` check redundant. Local validation: 8 passed, 1 expected skip; pre-commit passed.

VibhuJawa: aghh, sorry for the long random  ai slop comment. Let me review this in a bit

PR #2407, opened 2026-09-16
[Do not Merge] build: add Nemotron OCR extra

VibhuJawa:
## What

Add an opt-in `nemotron_ocr` extra compatible with Curator's Torch pins and document package and source installation.

## Why

VDR-20: host installs had no supported dependency path for `pipeline_v2`.

## Validation

- `uv lock --check --offline`
- Frozen dependency graph and package contents verified
- Pre-commit hooks

**Scope:** Build metadata and documentation.


sarahyurick (approved):

VibhuJawa:
Local GPU validation found a blocker in the default install path. The frozen extra resolves and installs nemotron-ocr 2.0.2 with Curator's Torch 2.11.0+cu129 and Torchvision 0.26.0+cu129, but the published wheel bundles CUDA runtime 13.0.48. It imports and loads the model, then silently returns zero detector regions for an image that the CUDA-12.9 source-built extension recognizes. The Python package sources are identical; substituting or locally building the native extension against CUDA 12.9 restores OCR output. Forcing the official source distribution with no build isolation also works, but the currently documented uv pip install and uv sync commands select the incompatible wheel. Marking this draft pending either an upstream CUDA-12-compatible wheel or an explicit source-build-only installation design.

PR #2411, opened 2026-09-16
docs: update ProcessingStage input contract

lbliii:
## Description

Updates the existing `ProcessingStage` reference for `StageInputSpecs`, task-type-specific input requirements, and fanout inferred from list return annotations. Includes one small mapping example.

Linear: https://linear.app/nvidia/issue/NMCUR-411

## Validation

- Applied `doc-style-guide` across all 12 categories after the final edit. No remaining findings in changed text; pre-existing page issues are outside scope.
- Verified the declarations and annotation behavior against source.
- Changed-file pre-commit and PR diff whitespace checks passed.
- `make docs-check`: zero errors, one warning. The external `Full-Library-Reference` folder is missing, so the missing-redirects check is skipped.


lbliii:
@sarahyurick could you review the ProcessingStage contract update, given your review of the fanout implementation in #2157? The change is limited to multiple input types and annotation-based fanout behavior. Style and docs checks passed.

sarahyurick:
Hi @lbliii it looks like you linked the wrong PR that this is supposed to be documenting. Can you update with the correct PR and re-assign?

PR #2418, opened 2026-09-16
feat: allow outputs() to declare a spec per input task type

### LINKED ISSUE #2245: Support more than 1 task type for stage outputs
Is your feature request related to a problem? Please describe.
Currently many stages typically work with a single type of task as input/output for eg: DocumentBatch or VideoTask.
Even if a stage accepts different task types often they may have the same input/output requirements ["data"]["text"].
However with more cases of stages being applicable cross modality and cross task there might be cases where the input and output spec vary depending on the task type.

For eg: MinHashStage can be extended to operate on DocumentBatch in addition to FilegroupTask. However the input requirements for fgtask is (["data"], []) however for DocumentBatch it would be (["Data"], ["text","id"]). Which breaks validate_inputs in many cases or forces the stages to include the least amount of validation requirements to allow multiple types to work.

Describe the solution you'd like
One thing to decide here is it the output type depends on the input type or maybe depends on other factors instead

Describe alternatives you've considered
Alternative is for each stage with this requirement to override the validate_input/validate_output stage to handle the custom branching logic.

Additional context

chethanuk:
## Description

Closes #2245

`outputs()` can now return one `(attributes, columns)` spec per input task type, the same dict form `inputs()` got in #2239. The tuple form still works. On `main`, a stage whose `outputs()` returns a dict makes `Pipeline.describe()` print `Error getting stage info: can only join an iterable`.

The issue left open whether the output spec should depend on the input type or on something else. I keyed it by input task type to match `inputs()`. Happy to change it if you had something else in mind.

- `describe()` renders inputs and outputs with one helper. Tuple outputs print the same lines as before.
- Nothing reads outputs per task type at run time, so there is no output validation. A key that isn't one of the stage's input types isn't rejected; `describe()` prints it like any other key.
- The Fern pages still show `outputs()` as tuple-only, as they already do for `inputs()`. Only the docstring is updated here.

Testing: `pytest tests/pipelines tests/stages/common/test_base.py -m "not gpu"` gives 68 passed; the dict cases of the new `test_describe_renders_output_specs` fail on `main`. A `Pipeline` with one dict-output and one tuple-output stage goes through `build()`, `describe()` and `validate_input()` on a `DocumentBatch` and a `FileGroupTask` cleanly. The same script on `main` prints the error above for the dict stage.

## Usage

```python
class MyStage(ProcessingStage[FileGroupTask | DocumentBatch, DocumentBatch]):
    def inputs(self) -> StageInputSpecs:
        return {
            FileGroupTask: (["data"], []),
            DocumentBatch: (["data"], ["text"]),
        }

    def outputs(self) -> StageOutputSpecs:
        return {
            FileGroupTask: (["data"], ["text"]),
            DocumentBatch: (["data"], ["text", "score"]),
        }
```

## Checklist


sarahyurick:
Hi @chethanuk thanks for opening. Your PR mostly looks good to me, although I was wondering if there is any output validation that needs to be updated and/or added?

PR #2423, opened 2026-09-17
feat(dedup): support batch_size/process_batch in MinHashStage

### LINKED ISSUE #1465: Support batchsize/process_batch in the minHash stage

Lingikaushikreddy:
## Description

Implements `process_batch` for `MinHashStage`, closes #1465.

GPUs do best on large blocks (1-3GB) while CPU stages do best on small ones (~256MB), but a pipeline has to pick one read blocksize for both. Saturating the GPU for minhash therefore meant risking OOM in the CPU stages upstream.

`MinHashStage.process_batch` coalesces several input tasks into a single minhash computation. As described in the issue, inputs are still read one at a time — that read is no more saturated than it is today — but the concatenated frame is hashed and written as one block, which decouples the read blocksize from the size of the block the GPU actually sees.

Details:

- `batch_size` is a new constructor argument defaulting to `1`, so existing pipelines keep the current one-task-in/one-file-out behavior. It can also be set through the existing `with_(batch_size=...)`.
- Both `FileGroupTask` and `DocumentBatch` inputs are supported, reusing the same per-task read path as `process`.
- Inputs are read and appended one at a time before a single `cudf.concat`, so peak memory is one input frame plus the combined frame rather than every input frame at once.
- The combined output file is named from `get_deterministic_hash` over the batch's task ids, matching how `IdentifyDuplicatesStage` names its fan-in output.
- The fan-in keeps the first task's metadata plus every input's final `_stage_perf` record, also mirroring `IdentifyDuplicatesStage`, so upstream timings are not dropped.
- The shared read/compute/write steps are factored into `_read_task`, `_minhash_and_write` and `_build_output_task`, used by both `process` and `process_batch`. The single-task path is unchanged.

## Usage

```python
from nemo_curator.stages.deduplication.fuzzy.minhash import MinHashStage

# Read in small blocks (CPU-friendly), minhash in large ones (GPU-friendly).
stage = MinHashStage(
    output_path="/path/to/minhash/output",
    text_field="text",
    read_format="parquet",
    batch_size=8,   # coalesce 8 input tasks into one minhash computation + one output file
)

# Equivalent, via the existing override helper:
stage = MinHashStage(output_path="...", text_field="text", read_format="parquet").with_(batch_size=8)
```

`batch_size` defaults to `1`, which is the current behavior.

## Testing

Added to `tests/stages/deduplication/fuzzy/test_minhash_stage.py`:

- `test_process_batch_matches_per_task_processing` — the key one: batching changes how many files are written, never the signatures. Compares the batched output against the concatenation of per-task `process` outputs.
- `test_process_batch_carries_metadata_and_stage_perf` — the combined task keeps the first input's metadata.
- `test_process_batch_empty_returns_empty_list`, `test_process_batch_without_setup` — edge cases.
- `test_batch_size_defaults_to_one_and_is_configurable` — the default stays 1.

These are `@pytest.mark.gpu` (cuDF), so they need CI to run end to end. Locally I verified the new control flow with the GPU/infra dependencies stubbed: inputs read one by one in order, exactly one write per batch, deterministic output name, `_stage_perf` merged as intended, and `process` producing the same output path, metadata and perf record as before the refactor. `ruff check` is clean on both files.

## Checklist

- [x] The documentation is up to date with these changes. (`batch_size` is documented in the `MinHashStage` parameter docstring; `process_batch` has its own docstring.)

Commits are signed off (DCO). Happy to adjust the fan-in naming or the `_stage_perf` merge if you'd prefer different semantics here.


Lingikaushikreddy:
Thanks — three of the four were real, pushed in 9e68bdc.

**Fan-in breaks checkpoint resumability (P1)** — agreed, and this was the important one. `MinHashStage` inherited `is_resumable = True`, which is correct at `batch_size=1` but wrong once batching fans several inputs into one output. Now `self.is_resumable = batch_size == 1`, so the default stays 1:1 and resumable and existing checkpointed pipelines are unaffected, while a batched stage is rejected up front by the `Pipeline.run(checkpoint_path=...)` check rather than silently reprocessing its sources. Matches the reasoning already spelled out on `ConnectedComponentsStage` and LSH.

**Invalid batch sizes lose work (P1)** — agreed. The constructor now rejects a non-positive `batch_size`. A negative value silently producing no batches was the worse half of this.

**Performance merging remains untested (P2)** — agreed, the test gave every input an empty history so it could not have caught a regression. Inputs now get two distinct `StagePerfStats` records each, and the test asserts the merged order, which distinguishes "kept the whole history" (first task) from "kept only the final record" (the rest).

**Concatenation doubles memory pressure (P1)** — correct about the behaviour, but the problem is my comment rather than the code, so I fixed the comment and kept the single concat. `cudf.concat` needs the inputs and the combined copy resident together, so peak really is ~2x the combined input size; my comment claimed one input frame plus the combined frame, which was wrong. Concatenating incrementally would not lower that peak — the accumulator and its new copy still coexist at each step — and it recopies the accumulator on every step, so it is worse on time for no memory gain. `IdentifyDuplicatesStage` builds a list and concats once for the same reason. Since the ~2x factor is inherent to batching rather than a bug, I documented it on the `batch_size` parameter so the value can be sized against free GPU memory. Happy to revisit if you would rather cap this some other way.

Commits are signed and signed off.

Lingikaushikreddy:
@greptile-apps the summary says two previously reported blocking issues remain outstanding at `9e68bdc`, but the findings list does not mark which two. Could you name them?

For the record, where I think each of the four stands after `9e68bdc`:

1. **Concatenation doubles memory pressure** — comment corrected, code deliberately unchanged. `cudf.concat` needs the inputs and the combined copy resident together, so ~2x the combined input size is inherent to batching. Concatenating incrementally would not lower that peak and would recopy the accumulator at every step; `IdentifyDuplicatesStage` builds a list and concats once for the same reason. I documented the factor on `batch_size` so it can be sized against free GPU memory. If this is one of the two, I am glad to discuss alternatives, but I do not think an incremental concat is one.
2. **Invalid batch sizes lose work** — fixed; the constructor rejects a non-positive `batch_size`.
3. **Fan-in breaks checkpoint resumability** — fixed; `is_resumable = batch_size == 1`, so batched use is rejected by the `Pipeline.run(checkpoint_path=...)` check while the 1:1 default stays resumable.
4. **Performance merging remains untested** — fixed; inputs carry distinct `StagePerfStats` records and the test asserts the merged order.

If the two you mean include something I have mis-scoped above, point me at it and I will fix it.

PR #2430, opened 2026-09-19
feat: support create mode in DataDesignerStage

### LINKED ISSUE #2190: Investigate `preview` vs `create` for `DataDesignerStage`
For https://github.com/NVIDIA-NeMo/Curator/pull/1449, `preview` is being used to work with DataFrames pre-loaded by a Curator stage. We can look into switching over to `create` that generates durable artifacts on disk.

Context: `preview` is not meant for any meaningful larger generation. While it will let you specify a large number of rows to generate, it doesn't have the durability mechanisms built into `create` that lets you resume from checkpoints and scale across concurrent row groups. `preview` also exposes your workers vulnerable to OOMs, at the mercy of the size of each partition a Curator stage loads and the underlying Data Designer pipeline.

Thread: https://github.com/NVIDIA-NeMo/Curator/pull/1449#discussion_r2777992738

Abdullah49645:
## Description

Closes #2190.

Adds opt-in support for using `DataDesigner.create()` in `DataDesignerStage` while preserving the existing `preview()` behavior by default.

The stage now supports:
- `use_create=False` by default, preserving the existing `preview()` workflow.
- `use_create=True` to run the artifact-backed `DataDesigner.create()` workflow.
- An explicit `artifact_path` when create mode is enabled.
- `resume`, defaulting to `dd.ResumeMode.NEVER` when create mode is enabled.
- Validation to prevent `artifact_path` and `resume` from being used when `use_create=False`.

When create mode is enabled, `DataDesignerStage` passes the input batch's `dataset_name`, the configured `artifact_path`, and `resume` mode through to `DataDesigner.create()`.

This provides access to the persistent `create()` workflow without changing the existing behavior of `DataDesignerStage` for current users.

## Usage

```python
from nemo_curator.stages.synthetic.nemo_data_designer import DataDesignerStage

stage = DataDesignerStage(
    config_builder=config_builder,
    use_create=True,
    artifact_path="/path/to/artifacts",
)
```

By default, `use_create` remains `False`, so existing code continues to use `DataDesigner.preview()`.

## Checklist


nemo_curator/stages/synthetic/nemo_data_designer/data_designer.py
```diff
@@ -134,8 +148,27 @@ def process(self, batch: DocumentBatch) -> DocumentBatch:
 
         try:
             t1 = time.perf_counter()
-            results = self.data_designer.preview(self.config_builder, num_records=num_input_records)
-            df = results.dataset
+            if self.use_create:
+                results = self.data_designer.create(
+                    self.config_builder,
+                    num_records=num_input_records,
+                    dataset_name=batch.dataset_name,
```
sarahyurick: I think since `batch.dataset_name` is not guaranteed to be unique, can you please add a deterministic function to ensure that the same `DocumentBatch` gets the same unique name every time?

tests/stages/synthetic/nemo_data_designer/test_data_designer.py
```diff
@@ -91,6 +91,132 @@ def test_post_init_validation(self) -> None:
...
+            dataset_name="ds1",
+            artifact_path=str(tmp_path),
+            resume=dd.ResumeMode.NEVER,
+        )
+        stage.data_designer.preview.assert_not_called()
+        create_results.load_dataset.assert_called_once_with()
+        create_results.load_analysis.assert_called_once_with()
+        assert out_batch.data is output_df
+
+    def test_process_uses_preview_by_default(self) -> None:
```
sarahyurick: Some of the added tests seem unnecessary. For example `preview` is already being tested elsewhere in this file right?

sarahyurick (commented):
Thanks for opening @Abdullah49645 ! I left some minor requests.

tests/stages/synthetic/nemo_data_designer/test_data_designer.py
Abdullah49645: Thanks Sarah! Addressed both points.

* Create mode now derives the artifact dataset name deterministically from the batch task identity, rather than relying on dataset_name being unique.
* The NDD synthetic-stage wrapper now preserves the incoming task_id when creating its temporary batch, so distinct partitions retain distinct artifact identities.
* I also removed the redundant preview-specific test and kept focused coverage for deterministic naming and task-ID preservation.

The targeted tests for these changes pass.

nemo_curator/stages/synthetic/nemo_data_designer/data_designer.py
```diff
@@ -94,6 +109,9 @@ def _init_data_designer(self) -> None:
         else:
             self.data_designer = DataDesigner()
 
+    def _get_dataset_name(self, batch: DocumentBatch) -> str:
+        return f"{batch.dataset_name}-{get_deterministic_hash([batch.task_id])}"
```
sarahyurick: I think `task_id` is not always deterministic (e.g., after a fan-out) or it can be empty too.

Can we do something like this instead?

```suggestion
    def _get_dataset_name(self, batch: DocumentBatch) -> str:
        source_files = batch._metadata.get("source_files")
        if source_files:
            digest = get_deterministic_hash([str(p) for p in source_files])
        elif batch.task_id:
            digest = get_deterministic_hash([batch.task_id])
        else:
            digest = get_deterministic_hash([f"{batch.dataset_name}:{batch.num_items}"])

        safe_dataset_name = re.sub(r"[^A-Za-z0-9_.-]+", "-", batch.dataset_name).strip("-.") or "dataset"
        return f"{safe_dataset_name[:80]}-{digest}"
```

Let me know what you think.

Abdullah49645: Thanks! I incorporated this approach. I also included task_id alongside source_files in the hash so that distinct partition tasks that share the same source_files metadata cannot collide on the persisted artifact dataset name.

PR #2438, opened 2026-09-22
fix(interleaved): handle null WDS content types

VibhuJawa:
## Summary
- treat pandas null scalars as missing WDS content types
- normalize non-null content types to strings before extension lookup
- cover pd.NA with the existing fallback test

## Test plan
- pytest -q tests/stages/interleaved/test_multimodal_writer.py -k ext_from_content_type
- pre-commit run --files nemo_curator/stages/interleaved/io/writers/webdataset.py tests/stages/interleaved/test_multimodal_writer.py
- 64-document Cosmos WDS integration check: 453/453 member names and bytes matched the serial reference


sarahyurick (approved):
LGTM!

